PR #10, opened 2021-02-25
Update requirements.txt to include TimeSeriesCrossValidation and pyth…

danielorf:
…on-Levenshtein


danielorf:
This also includes the additional requirements pulled in by requirements.txt install for full snapshot

DidierRLopes (commented):
Thanks for the review.

Why are there so many new libraries added? Does this picks up all libraries and dependencies?

If so, how did you do it?

danielorf:
I installed everything already in requirements.txt, installed TimeSeriesCrossValidation + python-Levenshtein and then ran `pip freeze > requirements.txt` again.

It's kinda overkill, but usually either go for a very light of requirements with flexible operators (e.g. ">=" vs "==") or take a full snapshot of every dependency.  With the latter you get a complete picture of deps up front which can be good for security scans.

DidierRLopes:
Amazing, thank you!

I was curious because I used pipreqs library, and it didn't work as smoothly as I was expecting.

PR #23, opened 2021-02-25
changes pyplot to use interactive mode so .show() won't block

clothespin:
- Adds block=False to all instances where show is called on plot objects

It seems like it limits the utility of the terminal a lot to have it just hang there while plots are showing, especially when you want to have more than one plot on screen at once. Pyplot supports a block=False option that just returns back to the main loop after kicking off the plotter. I went ahead and added that param to all the uses of show(). If it would be better, I could gate the nonblocking behavior behind a config option? but I can't really think of a situation in which you'd want the blocking behavior. Feel free to correct me if I'm wrong however.

### DIFF
--- a/due_diligence/business_insider_api.py
+++ b/due_diligence/business_insider_api.py
@@ -51,6 +51,8 @@ def price_target_from_analysts(l_args, df_stock, s_ticker, s_start, s_interval):
         df_analyst_data['Date'] = pd.to_datetime(df_analyst_data['Date'])
         df_analyst_data = df_analyst_data.set_index('Date')
 
+        plt.figure()
+
         # Slice start of ratings
         if s_start:
             df_analyst_data = df_analyst_data[s_start:]
--- a/due_diligence/business_insider_api.py
+++ b/due_diligence/business_insider_api.py
@@ -77,7 +77,7 @@ def price_target_from_analysts(l_args, df_stock, s_ticker, s_start, s_interval):
         plt.grid(b=True, which='major', color='#666666', linestyle='-')
         plt.minorticks_on()
         plt.grid(b=True, which='minor', color='#999999', linestyle='-', alpha=0.2)
-        plt.show()
+        plt.show(block=False)
         print("")
 
         pd.set_option('display.max_colwidth', -1)
@@ -303,7 +303,7 @@ def insider_activity(l_args, df_stock, s_ticker, s_start, s_interval):
         plt.grid(b=True, which='major', color='#666666', linestyle='-')
         plt.minorticks_on()
         plt.grid(b=True, which='minor', color='#999999', linestyle='-', alpha=0.2)
-        plt.show()
+        plt.show(block=False)
 
         l_names = list()
         for s_name in text_soup_market_business_insider.findAll('a', {'onclick':"silentTrackPI()"}):
--- a/due_diligence/business_insider_api.py
+++ b/due_diligence/business_insider_api.py
@@ -77,7 +77,8 @@ def price_target_from_analysts(l_args, df_stock, s_ticker, s_start, s_interval):
         plt.grid(b=True, which='major', color='#666666', linestyle='-')
         plt.minorticks_on()
         plt.grid(b=True, which='minor', color='#999999', linestyle='-', alpha=0.2)
-        plt.show(block=False)
+        plt.ion()
+        plt.show()
         print("")
 
         pd.set_option('display.max_colwidth', -1)
@@ -303,7 +304,8 @@ def insider_activity(l_args, df_stock, s_ticker, s_start, s_interval):
         plt.grid(b=True, which='major', color='#666666', linestyle='-')
         plt.minorticks_on()
         plt.grid(b=True, which='minor', color='#999999', linestyle='-', alpha=0.2)
-        plt.show(block=False)
+        plt.ion()
+        plt.show()
 
         l_names = list()
         for s_name in text_soup_market_business_insider.findAll('a', {'onclick':"silentTrackPI()"}):
--- a/due_diligence/business_insider_api.py
+++ b/due_diligence/business_insider_api.py
@@ -51,11 +51,12 @@ def price_target_from_analysts(l_args, df_stock, s_ticker, s_start, s_interval):
         df_analyst_data['Date'] = pd.to_datetime(df_analyst_data['Date'])
         df_analyst_data = df_analyst_data.set_index('Date')
 
-        plt.figure()
 
         # Slice start of ratings
         if s_start:
             df_analyst_data = df_analyst_data[s_start:]
+
+        plt.figure()
 
         if s_interval == "1440min":
             plt.plot(df_stock.index, df_stock['5. adjusted close'].values, lw=3)
--- a/technical_analysis/momentum.py
+++ b/technical_analysis/momentum.py
@@ -7,11 +7,11 @@ import matplotlib.pyplot as plt
 
 # ----------------------------------------------------- CCI -----------------------------------------------------
 def cci(l_args, s_ticker, s_interval, df_stock):
-    parser = argparse.ArgumentParser(prog='cci', 
-                                     description=""" The CCI is designed to detect beginning and ending market trends. 
-                                     The range of 100 to -100 is the normal trading range. CCI values outside of this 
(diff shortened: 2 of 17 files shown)


DidierRLopes (commented):
This is a great idea!

I've tested it quickly, and I'm seeing the figures to freeze.

Which matplotlib version did you use to test this?

clothespin:
I'm on 3.1.1.

https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.show.html#matplotlib.pyplot.show says that in order for block=False to work correctly the event loop needs to be kept running, but I'm not super clear on how to do that yet.

clothespin:
So on further review of the matplotlib docs it looks like the simplest way to deal with the freezing issue is to use pyplot interactive mode, which automatically makes .show not block as well as do some kind of magic to keep the event loop running in the background. I went ahead and made the change and tested with a couple of simultaneous plots and I was able to interact with (zoom etc) both of them, although I noticed my computer's fan spinning up when I had two plots open at once and some lag. May or may not be an issue on a better PC. I went ahead and made and pushed the change. You shouldn't see plots freezing anymore.

DidierRLopes:
Good work @clothespin !

Currently with your fix, the figure doesn't block the terminal, which is the point of this PR. However, if you don't close the figure, the next plt.plot() will occur on top of it - mixing both figures content.

What do you think about instantiating a new figure everytime plt.figure(), just so that we only close the figures we don't want to have a look anymore?

clothespin:
In my testing (running on Debian Linux) issuing a load command, a view command to bring up a plot of what you just loaded, a second load command for a different ticker, and a second view command, I get two independent plot windows, each with the correct ticker, and each zoomable/adjustable. Are you seeing different behavior on your platform?

clothespin:
image
Here's a screenshot of the behavior I'm getting.

DidierRLopes:
You're right. That is because the function to the stock price is instantiated with: plt.subplots()
https://github.com/DidierRLopes/GamestonkTerminal/blob/586d6499042dba9490365e4b6172d354f65bd8c7/helper_funcs.py#L44

So when this function is called, a new figure is instantiated.

If you try to do load of 1 ticker, and then attempt to run a prediction algorithm on it you'll see what I mean. The prediction algorithm plots don't have a plt.subplots() or plt.figure() at the beginning, they start straight with plt.plot(). Meaning that by default: If there's no figure open, it instantiates a new one. If there's one, it will plot on that same one.

So in the prediction techniques menu, it will not work this smoothly. We need to add this plt.figure() above those figure generations :)

clothespin:
Ah, I see, that makes sense. Would that go on the line before plt.plot() or the line before show()? And we'd want this change for all of the files that I modified that weren't already calling subplots() or figure(), correct?

DidierRLopes:
The line before plt.plot() because the way it works is that this plt.plot() will plot the graph in whatever figure was last opened. If you do plt.figure() just before it, it will plot the graph on such new plot.

If you did
plt.figure()
plt.show()

you'd have an empty figure being generated, because you 1. instantiate a matplotlib object, 2. render it.

If you can fix it, that would be amazing! If you can't, it's still amazing because I'll do it myself :)

clothespin:
Okay, I think I caught most instances where a new figure() was needed by looking at files where plot() was being called but where neither figure() nor subplots() were. Tested with a couple of the disc and ta functions and they appear to work as expected.

technical_analysis/trend.py
```diff
@@ -118,6 +120,8 @@ def aroon(l_args, s_ticker, s_interval, df_stock):
                          scalar=ns_parser.n_scalar, offset=ns_parser.n_offset).dropna()
         
         plt.subplot(311)
+        plt.figure()
```
DidierRLopes, line 123: This plt.figure() should be above the plt.subplot(311)

technical_analysis/volatility.py
```diff
@@ -30,9 +30,12 @@ def bbands(l_args, s_ticker, s_interval, df_stock):
             print(f"The following args couldn't be interpreted: {l_unknown_args}\n")
             return
 
+        # create new figure
+        plt.figure()
```
DidierRLopes, line 34: Although it seems like "duplicated code" I would put this just before the first plt.plot() on both cases

technical_analysis/volume.py
```diff
@@ -28,6 +28,8 @@ def ad(l_args, s_ticker, s_interval, df_stock):
             print(f"The following args couldn't be interpreted: {l_unknown_args}\n")
             return
 
+        plt.figure()
```
DidierRLopes, line 31: Same as before if possible

technical_analysis/trend.py
```diff
@@ -24,6 +24,8 @@ def adx(l_args, s_ticker, s_interval, df_stock):
             print(f"The following args couldn't be interpreted: {l_unknown_args}\n")
             return
 
+        plt.figure()
```
DidierRLopes, line 27: I would put this straight above both plt.subplot(211), to be more contained in that area of code

technical_analysis/momentum.py
```diff
@@ -293,6 +299,8 @@ def stoch(l_args, s_ticker, s_interval, df_stock):
             print(f"The following args couldn't be interpreted: {l_unknown_args}\n")
             return
 
+        plt.figure()
```
DidierRLopes, line 302: Same here

technical_analysis/momentum.py
```diff
@@ -201,6 +205,8 @@ def rsi(l_args, s_ticker, s_interval, df_stock):
...
```
DidierRLopes, line 208: Same here

technical_analysis/momentum.py
```diff
@@ -119,6 +121,8 @@ def macd(l_args, s_ticker, s_interval, df_stock):
...
```
DidierRLopes, line 124: same here

technical_analysis/momentum.py
```diff
@@ -25,6 +25,8 @@ def cci(l_args, s_ticker, s_interval, df_stock):
...
```
DidierRLopes, line 28: same here

due_diligence/business_insider_api.py
```diff
@@ -51,6 +51,8 @@ def price_target_from_analysts(l_args, df_stock, s_ticker, s_start, s_interval):
         df_analyst_data['Date'] = pd.to_datetime(df_analyst_data['Date'])
         df_analyst_data = df_analyst_data.set_index('Date')
 
+        plt.figure()
```
DidierRLopes, line 54: I would put this one below the slicing of the rating's slicing. I.e. under
```
 # Slice start of ratings
 if s_start:
     df_analyst_data = df_analyst_data[s_start:]
 ```

DidierRLopes (commented):
Perfect. I added a few comments on some of the plt.figure() mainly because I'm picky. If you can move those around I'll accept this and merge the PR :)

technical_analysis/volatility.py
clothespin: I can go ahead and do this, but I'm not totally sure why, since placing the call before the conditional has the same effect as placing it on both sides. Wouldn't it be cleaner to just have it in one place?

clothespin: I won't spam you by hitting resolve on all the comments but I implemented this request for all the files you commented on. I'll go ahead and resolve only this one.

DidierRLopes:
Hey @clothespin ,

Sorry, we had a major code re-architecture and bugfix PR being merged.

And there's a lot of conflicts with this. If you can fix them I would appreciate it massively, if you don't want to I understand and I'll fix it myself and merge it.

Either way let me know, thank you!

DidierRLopes:
Don't worry with fixing this, I'm doing it to comply with the changes! Thanks!

clothespin:
OK, cool. If you'll take on fixing the conflicts for this PR I'll go ahead and do the same for the export command in #73

DidierRLopes:
> OK, cool. If you'll take on fixing the conflicts for this PR I'll go ahead and do the same for the export command in #73

That's perfect! Thank you! I've merged already the changes around TA. #109 should take care of prediction ones :)

DidierRLopes:
Non blocking plots have now been implemented and merged for both prediction and technical analysis menus. See #109 and #108, respectively. 

On that basis, I'll close this PR!

PR #36, opened 2021-02-26
replace fuzzywuzzy with rapidfuzz

derrickknox:
This PR replaces FuzzyWuzzy with RapidFuzz, which is faster and uses a MIT license.
Following https://news.ycombinator.com/item?id=26259362 the current library might cause license issues, while a MIT licensed one definitely does not.
Another advantage is that it comes with wheels, so the user does not need a C/C++ compiler.


DidierRLopes:
It's the small things! Amazing!! Thank you a lot!

PR #63, opened 2021-02-28
it can load btcusd, add crypto

sabujp:
It seems you can load -t btcusd -s 2017-10-01 so mention crypto in the readme. Also you don't need an alpha vantage key to get the daily OHLC values (for now at least) so also mention that below the key section

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## About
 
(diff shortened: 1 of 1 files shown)


DidierRLopes:
Excellent!

Thanks for this! Much appreciated :)

PR #68, opened 2021-03-01
Fixed undefined timeseries var

benkulbertis:
This needs to be defined again here, throws an error otherwise. Produced by loading a ticker with an intraday interval and then using 'pred' to get to the prediction menu.

```
> load -t SOXL -s 2020-06-01 -i 60
Loading Intraday 60min SOXL stock with starting period 2020-06-01 for analysis.

> pred
  File "gamestonk_terminal.py", line 328, in <module>
    main()
  File "gamestonk_terminal.py", line 313, in main
    df_stock_pred, _ = ts.get_daily_adjusted(symbol=s_ticker, outputsize='full')
NameError: name 'ts' is not defined
```


DidierRLopes (approved):
Thank you for this. Will merge it now!

PR #73, opened 2021-03-02
Adds export command to save LOADed data to a variety of formats

clothespin:
This pull implements extremely basic export functionality and is intended to contribute toward a resolution of #7.

I'm just exposing the [built in pandas IO formats](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html) on df_stock, nothing fancy. The only things the user can specify are the file format (I went ahead and implemented the ones I thought would be most desired but it would be easy to expose others) and the destination filename (the export goes to stdout if -f is unspecified). Can throw an exception in case you specify a format (e.g. excel) that doesn't support output to stdout, but don't also specify the filename. Otherwise works well and does what it says on the tin.


gamestonk_terminal.py
```diff
@@ -150,6 +149,37 @@ def view(l_args, s_ticker, s_start, s_interval, df_stock):
     plot_view_stock(df_stock.iloc[:, ln_col_idx], ns_parser.s_ticker)
 
 
+# ----------------------------------------------------- EXPORT -----------------------------------------------------
+def export(l_args, df_stock):
+    parser = argparse.ArgumentParser(prog='export', description='Exports the current dataframe to a file or stdout')
+    parser.add_argument('-f', "--filename", type=str, dest="s_filename", default=stdout, help="Where to save the export (stdout if unspecified)")
+    parser.add_argument('-F', "--format", dest="s_format", type=str, default="csv", help="Export format (csv json excel clipboard)")
```
DidierRLopes, line 156: I'd put: "Export historical data into following formats: csv, json, excel, clipboard"

gamestonk_terminal.py
```diff
@@ -150,6 +149,37 @@ def view(l_args, s_ticker, s_start, s_interval, df_stock):
     plot_view_stock(df_stock.iloc[:, ln_col_idx], ns_parser.s_ticker)
 
 
+# ----------------------------------------------------- EXPORT -----------------------------------------------------
+def export(l_args, df_stock):
+    parser = argparse.ArgumentParser(prog='export', description='Exports the current dataframe to a file or stdout')
+    parser.add_argument('-f', "--filename", type=str, dest="s_filename", default=stdout, help="Where to save the export (stdout if unspecified)")
```
DidierRLopes, line 155: As default, wouldn't be better to save for a file with the following default name:
datestart_dateend_stockname.csv ?

I can't see how it can be useful printing 100 days worth of data, if it makes sense.

gamestonk_terminal.py
```diff
@@ -161,6 +191,7 @@ def print_help(s_ticker, s_start, s_interval, b_is_market_open):
     print("   clear       clear a specific stock ticker from analysis")
     print("   load        load a specific stock ticker for analysis")
     print("   view        view and load a specific stock ticker for technical analysis")
+    print("   export      export the currently loaded dataframe to a file or stdout")
```
DidierRLopes, line 194: Can you add above: 
`if s_ticker:`

This is so that this command only appears once we have loaded a ticker, which I think is the right thing to do in this case

DidierRLopes (changes requested):
minor comments, but this is extremely helpful. Thank you for this.

Once you update this PR, I'll merge it

gamestonk_terminal.py
clothespin: So the idea here was to enable a non-interactive, batch mode use of the terminal in line with other REPL-driven software like Python itself or the Unix shell. For instance, you could have a script like the following:
`#!/usr/bin/env -S python gamestonk_terminal.py
load -t GME -s $(date -d "7 days ago" '+%y%m%d')
export`
which could then be used in a shell pipeline. You might, say, gzip the output without having to write it to a temporary file first. So there's definitely a use case for having output to stdout I would say.

However, I agree that it may not make as much sense for the default behavior - I was just going by the typical Unix convention of having output go to stdout if an output filename isn't specified on the command line. It might be better to have a -s for stdout option or something like that.

The thing about using datestart and dateend as default filenames is that it's totally possible to load tickers without specifying dateend or datestart; dateend by default can just be now but I'm not sure what I would do for datestart. Or is that already calculated based on the earliest data returned from the API request?

DidierRLopes: Yea, that makes sense.

We can go for the conventional way, and if in the future we think there's a better default way, we change it :)

DidierRLopes:
Hey @clothespin,

Same as #23, I'm extremely keen on having this.

If you can resolve the conflicts, that would be extremely appreciated. If you don't want to, I'll do it myself.

Let me know either way. Thank you in advance!

clothespin:
I went ahead and implemented this on a new branch in #119 so this PR can be closed

PR #82, opened 2021-03-03
Feature: Refactor CICD

shadycuz:
This PR adds support for macOS builds. It also added windows but pystan doesn't support windows. It also refactors @cclauss work into a single workflow. I did this so that we can only run tests if linting pass. I also set it up to run tests on Linux and python 3.8 first and only then run the tests on os:[mac,linux] pythons [3.6, 3.7, 3.8]. 

The full workflow is Linting --> Test on Linux 3.8 --> Test on [mac, linux] with python [3.6, 3.7, 3.8]

This is important to try and keep our number of minutes as low as possible.


.github/workflows/test.yml
```diff
@@ -1,12 +1,86 @@
...
-  tests:
+  linting:
+    name: Linting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Setup Python 3.8
+        uses: actions/setup-python@v1
```
cclauss, line 12:
```suggestion
        uses: actions/setup-python@v2
```
https://github.com/actions/setup-python/releases

.github/workflows/test.yml
```diff
@@ -1,12 +1,86 @@
...
+        with:
+          python-version: "3.8"
+          architecture: x64
+
```
cclauss, line 16:
```suggestion
```
It will probably be fastest to just run the linting tools on the default latest and greatest.

.github/workflows/test.yml
```diff
@@ -1,12 +1,86 @@
...
+  test:
+    name: ubuntu-latest - Python 3.8
+    needs: linting
     runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Setup Python 3.8
+        uses: actions/setup-python@v1
```
cclauss, line 41:
```suggestion
        uses: actions/setup-python@v2
```

.github/workflows/test.yml
```diff
@@ -1,12 +1,86 @@
...
+          pytest tests/
+  tests:
+    needs: test
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: true
       matrix:
-        python-version: ['3.6.13', '3.7', '3.8']
-    name: Python ${{ matrix.python-version }}
+        python-version: ['3.6', '3.7', '3.8']
```
cclauss, line 78:
```suggestion
        python-version: ['3.6', '3.7', '3.9']
```
My experience is that if you test on Py37 and Py39 then at least one of them will catch everything that you would find on Py38.

.github/workflows/test.yml
```diff
@@ -1,12 +1,86 @@
...
+        os: [ubuntu-latest, macos-latest]
+        exclude:
+          - os: ubuntu-latest
+            python-version: '3.8'
+    name: ${{ matrix.os}} - Python ${{ matrix.python-version }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
```
cclauss, line 86: Below change `uses: actions/setup-python@v1` --> `v2`.

.github/workflows/test.yml
```diff
@@ -1,12 +1,86 @@
...
+      - run: codespell --ignore-words-list="mape" --quiet-level=2  # --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --check-only --profile black . || true
+      - run: mypy --ignore-missing-imports . || true
+      - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
+      - run: safety check
+  test:
+    name: ubuntu-latest - Python 3.8
+    needs: linting
     runs-on: ubuntu-latest
```
cclauss, line 35: Do you want to `fast-fail: false` to see the results of all failing tests?

cclauss (changes requested):
Looks cool!  Nice work.

.github/workflows/test.yml
shadycuz: We don't support the latest and greatest. Tensor flow doesn't support python >= 3.8 

I feel like It's best to test wit the version you support even if I agree its very unlikely to make a difference.

.github/workflows/test.yml
shadycuz: We have had so many weird issues with the application dependencies. I would feel better if we tested with what we actually supported. I'm about 95% sure this will fail because of tensorflow. I will test locally and try.

.github/workflows/test.yml
shadycuz: I went back and forth on this. I want to try and limit the number of minutes as much as possible. I feel like failing fast might be the best way to do that. I feel like this project will attract a lot of new contributors who will push lots of failing builds.

shadycuz:
We need to implement some checks in the future so that we don't run things like tests when we update readme etc. 


On the todo list.

.github/workflows/test.yml
cclauss: Oh yes..... TF is still not compatible with Py39

PR #89, opened 2021-03-04
Update pyproject.toml metadata and CICD

shadycuz:
Getting the package metadata ready to publish to pypi. 

Also made some tweaks to the CICD to reduce builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,9 @@
 name: Tests
-on: [pull_request, push]
+on:
+  pull_request:
+    paths:
+      - '**.py'
+      - 'poetry.lock'
 jobs:
   linting:
     name: Linting
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,9 @@
 [tool.poetry]
 name = "GamestonkTerminal"
 version = "0.1.0"
-description = ""
+description = "The next best thing after Bloomberg Terminal"
 authors = ["Didier Rodrigues Lopes"]
+readme = "README.md"
 
 [tool.poetry.dependencies]
 python = "^3.6.8"


.github/workflows/test.yml
```diff
@@ -1,5 +1,9 @@
 name: Tests
-on: [pull_request, push]
```
shadycuz, line 2: @cclauss I removed push because it was causing duplicate builds for branches that didn't come from a fork, see #82

.github/workflows/test.yml
```diff
@@ -1,5 +1,9 @@
...
+on:
+  pull_request:
+    paths:
+      - '**.py'
+      - 'poetry.lock'
```
shadycuz, line 6: This should be all we need for now and will help reduce the number of builds. It seems to have worked for this PR.

pyproject.toml
```diff
@@ -1,8 +1,9 @@
 [tool.poetry]
 name = "GamestonkTerminal"
 version = "0.1.0"
-description = ""
+description = "The next best thing after Bloomberg Terminal"
 authors = ["Didier Rodrigues Lopes"]
+readme = "README.md"
```
shadycuz, line 6: @DidierRLopes Good start for now. I think a lot of other options are supported but it shouldn't make much of a difference.

.github/workflows/test.yml
cclauss: I believe that the way this works is...  If someone who is a repo maintainer (has write permissions to the repo) pushes a change to master then no test is run.  If that is true and that is an acceptable risk then go for it.

shadycuz: Yeah pretty much the plan is PR only unless its a docs thing. Could always change it in the future or add a "manual" button to run from the actions page.

cclauss:
There are no GitHub Action runs for this PR?  Do we really want to have PRs with no tests?
https://github.com/DidierRLopes/GamestonkTerminal/actions

shadycuz:
@cclauss 

Nothing ran because I didn't change a python file or a dependency. So that was the expected behavior.

Now that I think about it. Should probably update the list of files to include the github actions file.

PR #98, opened 2021-03-05
GitHub Actions: Add more flake8 tests

cclauss:
Instead of `selecting` a small list of key tests, let's `ignore` only those tests that our codebase currently fails.

Future PRs can clean up failing tests and remove them from the list.  High priority items to remove would be:
```
1     E722 do not use bare 'except'
1     E741 ambiguous variable name 'l'
20    F401 '.discovery' imported but unused
5     F841 local variable 'line' is assigned to but never used
```


aia (approved):

DidierRLopes:
Nice, thanks for contributing. Highly appreciated!

PR #53, opened 2021-02-27
General refactoring.

exislow:
Fixed unversioned requirements.
Removed unused import.
Reformatted code for better readibility.
Fixed all warnings.
Fixed errors except one.
Refactored to proper Python modules.

PR #29, opened 2021-02-25
Cleanups – 𝐔𝐋𝐓𝐑𝐀 𝐌𝐎𝐃𝐄

akx:
Following up on the heels of #26, this PR does more than just "mechanical" cleaning by way of deduplicating code by extracting it into methods and modules.

Lots of plots, for example, were simply copy-pasted with only one field name changed. (In fact, some were mis-copy-pasted, using incorrect columns.)

PR #101, opened 2021-03-05
Do not use the ambiguous variable name 'l'

cclauss:
`l` and `1` look too much alike to a casual reader so `l` should not be used as a variable name.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
       - run: bandit -r . || true
       - run: black --check .
       - run: codespell --ignore-words-list="mape" --quiet-level=2  # --skip=""
-      - run: flake8 . --count --ignore=E203,E302,E722,E741,F401,F841,W503 --max-line-length=140 --show-source --statistics
+      - run: flake8 . --count --ignore=E203,E302,E722,F401,F841,W503 --max-line-length=140 --show-source --statistics
       - run: mypy --ignore-missing-imports . || true
       - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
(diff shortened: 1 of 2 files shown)


DidierRLopes (approved):
Agreed.

Quick question @cclauss and @aia. Using l_, s_, n_ before a variable comes from my C coding style. I'm now rewriting the code in a better way following #29 and #53, should I drop this s_,l_,n_ convention? If yes, now is the time to do so :)

cclauss:
I believe that single-letter variable names make code look like it was written in the 1970s.

DidierRLopes:
> I believe that single-letter variable names makes code look like it was written in the 1970s.

Yea, I didn't mean that single "l". That came from a copy-paste from stack-overflow very-likely.

I meant my variables being: l_args, s_ticker, d_submission. Where l_ is list, s_ string, d_ submission, and so on

cclauss:
MUCH cooler to...

from [typing](https://docs.python.org/3/library/typing.html) import DIct, List
def my_func(args: List, ticker: str, submission: Dict) -> float:
    return 0.5

The big upside is that `mypy` will catch if you try to pass in a list instead of a dict or if you try to return a str.

aia:
Type hinting is a big thing.

DidierRLopes:
> MUCH cooler to...
> 
> from [typing](https://docs.python.org/3/library/typing.html) import DIct, List
> def my_func(args: List, ticker: str, submission: Dict) -> float:
> ` `return 0.5
> 
> The big upside is that `mypy` will catch if you try to pass in a list instead of a dict or if you try to return a str.

I'm on it!

PR #119, opened 2021-03-07
Adds export command to save LOADed data to a variety of formats: the sequel

clothespin:
Recreated this on a new branch since there had been so many changes in the meantime. From the original PR:

This pull implements extremely basic export functionality and is intended to contribute toward a resolution of #7.

I'm just exposing the built in pandas IO formats on df_stock, nothing fancy. The only things the user can specify are the file format (I went ahead and implemented the ones I thought would be most desired but it would be easy to expose others) and the destination filename (the export goes to stdout if -f is unspecified). Can throw an exception in case you specify a format (e.g. excel) that doesn't support output to stdout, but don't also specify the filename. Otherwise works well and does what it says on the tin.

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import argparse
+from sys import stdout
 import pandas as pd
 from alpha_vantage.timeseries import TimeSeries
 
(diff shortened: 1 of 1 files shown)


DidierRLopes (approved):
Sweet! Good work!

PR #130, opened 2021-03-07
Give sub-menu context

DidierRLopes:
This idea came from a reddit user, that mentioned that would be cool when going into the menus to know which menu they were in.

I think this implementation is just what's needed!


aia:
We think alike, but mine is 200% better:
https://github.com/DidierRLopes/GamestonkTerminal/pull/131

DidierRLopes:
Closing this because #131 has more 🚀🚀

PR #132, opened 2021-03-08
Add several exponential smoothing techniques to prediction menu

DidierRLopes:
Someone requested this prediction methods' family at some point. More information in here: https://otexts.com/fpp2/taxonomy.html

ets

Captura de ecrã 2021-03-08, às 02 08 45

PS: I don't allow the chance to personalize these weights, as I think it's not going to be used. Instead, based on the historical data, the .fit() method returns the best fit parameters to minimize some criteria (e.g. AIC)


aia:
very interesting!

aia (approved):

PR #133, opened 2021-03-08
Update Dockerfile, correct CMD

backhand:
Hi, just tried to run this on docker and it seems it's calling the wrong .py file in the CMD call. This is just a quick fix that fixed it for me :)


DidierRLopes (approved):
Amazing! Thank you for this!

PR #141, opened 2021-03-10
Add comparison analysis menu with historical and correlation heatmap comparisons

DidierRLopes:
This is the initial Comparison Analysis menu. Feel free to comment what you would change.

This is what the menu looks like so far:
Captura de ecrã 2021-03-10, às 00 33 12

 Here are 2 resulting examples from "historical" and "corr" commands, respectively
historical

corr

I'll update the README during weekend to reflect the progress that we've been making.


DidierRLopes:
I'm using adjusted close, but that's a good point.

I'll add the possibility for the user to choose to compare open close high low or adjusted close.

Do you think that by default should be the close?

PR #147, opened 2021-03-11
Update .gitgnore for pyenv files

gmerrall:
When developing on mac with homebrew and poetry, managing versions requires using pyenv. pyenv can pin in specific version  in a directory using a .python-version file.  This is also recommended and used by poetry.


aia:
You are missing an empty line at the end of the file

DidierRLopes (approved):

PR #152, opened 2021-03-11
Wrapping the reddit loop logic in try catch blocks

bfxavier:
I added some error handling inside the while loops to prevent the run from being aborted.

I noticed that, sometimes, the post would be missing the author, for example, and it would throw an exception and abort the command.

This should take care of the issue


DidierRLopes (approved):
Didn't see this happening during development, good catch!

Thanks for this.

PR #153, opened 2021-03-11
Fixing prophet run

bfxavier:
Using prophet as is is giving a 

"['date'] not in index"

Checking the object, the index is actually Date.

### DIFF
--- a/gamestonk_terminal/prediction_techniques/fbprophet.py
+++ b/gamestonk_terminal/prediction_techniques/fbprophet.py
@@ -44,8 +44,8 @@ def fbprophet(l_args, s_ticker, df_stock):
 
         df_stock = df_stock.sort_index(ascending=True)
         df_stock.reset_index(level=0, inplace=True)
-        df_stock = df_stock[["date", "5. adjusted close"]]
-        df_stock = df_stock.rename(columns={"date": "ds", "5. adjusted close": "y"})
+        df_stock = df_stock[["Date", "5. adjusted close"]]
+        df_stock = df_stock.rename(columns={"Date": "ds", "5. adjusted close": "y"})
         df_stock["ds"] = pd.to_datetime(df_stock["ds"])
 
         model = Prophet(yearly_seasonality=False, daily_seasonality=False)


DidierRLopes:
Hey,

You're right, there's an issue there, but this is not the correct fix.

The problem is that we were getting historical data from Alpha Vantage, which returns a dataframe with the following format: index='date', columns='1. open', '2. high' ,...

We changed for daily data (not intraday) to use yahoo finance instead because Alpha Vantage only allows 5 calls per minute and 500 per day, which can be quite limiting.  The dataframe from yahoo finance has the following format:
index='Date', columns='open', 'High' ,...

So, the correct fix is before the line, which is called when s_interval=="1440min" (meaning daily):
```
b_quit = pm.pred_menu(df_stock_pred, s_ticker, s_start, s_interval)
```
we need to convert the dataframe to match how the prediction algorithms are interpreting it. Which is like this:
df_stock_pred = df_stock.rename(
    columns={
        "Open": "1. open",
        "High": "2. high",
        "Low": "3. low",
        "Close": "4. close",
        "Adj Close": "5. adjusted close",
        "Volume": "6. volume",
    },
)
df_stock_pred.index.name = "date"

aia:
Oops, my bad

DidierRLopes:
No problem, I was too slow. Let me fix this.

I'm investigating another issue, which is the fact that when we imported prompt_toolkit, the matplotlib figures aren't showing up on TA and PRED menus. This is because of the line of code: plt.ion() on the figure generation.

I think that prompt_toolkit has its own event loop, and therefore, doesn't allow the matplotlib figures event to run. Which was allowing us to be able to not be blocked by a figure being generated.

aia:
I am seeing it too

DidierRLopes:
I already have the fix for this, and tested it and it works well. We were "lucky" in this one because all prediction algorithms and technical analysis only rely on data points. Prophet is the special kid, the only that actually relies on datetime instances :)

I'm looking online regarding the non blocking figures and keeping the prompt_toolkit. Or if there's a way to pause one in detriment of the other, but they are rather interconnected.

A easy fix would be not having prompt_toolkit to TA and PRED menus. But that's not ideal, and would rather find a better solution. Blocking figures is also not nice, at least in this context.

aia:
I think it is fine with blocking figures.

aia:
Threads are probably the solution if you want non-blocking figures:

https://stackoverflow.com/questions/19662906/plotting-with-matplotlib-in-threads

DidierRLopes:
Sounds about right. 

I'll fix the figures not showing for now by removing the plt.ion() line. I'll then work on this to make figures non-blocking.

DidierRLopes:
See this: https://matplotlib.org/stable/users/interactive_guide.html

"Command Prompt Integration" section, sounds promising :)

PR #156, opened 2021-03-11
Add Bugs stats

aia:


README.md
```diff
@@ -9,9 +9,12 @@
 [![Stargazers][stars-shield]][stars-url]
 [![Forks][forks-shield]][forks-url]
 [![Contributors][contributors-shield]][contributors-url]
-[![Issues][issues-shield]][issues-url]
 [![MIT License][license-shield]][license-url]
 
+[![Issues][issues-shield]][issues-url]
+[![Bugs Open][bugs-open-shield]][bugs-open-url]
+[![Bugs Open][bugs-closed-shield]][bugs-closed-url]
```
DidierRLopes, line 16: Since the URL is Issues specific and doesn't take into account bug labels, shouldn't this be: issues-open and issues-closed.

Also, I think that both line 15 and 16 have [Bugs Open] and in line 16 you mean [Bugs Closed]

DidierRLopes (approved):

README.md
aia: Fixed Bugs Closed.

I am still thinking what to do for issues.

I definitely think we should add Features Implemented counter, but that would require a bit of relabeling and adding.

Also how about Features Accepted?

Maybe remove Issues all together?

DidierRLopes: That's a great idea!

We could get rid of issues all together and keep:
- Bugs open
- Bugs closed
- Enhancement counter
- Features implemented counter

Not sure how we can track these enhancement/features implemented though. Because for instance you have a lot of ideas for portfolio analysis. And its seems "unfair" to only increase based on user's request, when that portfolio analysis tab will be a major work.

aia: I think we should start labeling PRs and publish metrics based on PR labels.

DidierRLopes: Sounds good to me. Feel free to lead the way here, I'll follow your guidelines :)

aia: We can also T-shirt size features (very common practice in the industry) and have a separate label per feature size.

E.g. XS-feature, S-feature, M-feature, L-feature, XL-feature.

PR #108, opened 2021-03-06
Improve TA plotting, dont block plots

DidierRLopes:
This PR contains:
- Improve plotting of TA menu, i.e. a plot is non-blocking. When the user inserts another command, the plot is closed automatically
- Refactor overall TA code to avoid duplications

PR #158, opened 2021-03-12
Add Feature Flags and disable prompt_toolkit by default due to conflict with ion

aia:


clothespin:
This update cause view (and other actions that cause plt.show() to be called) to block the input loop, so the user can't have more than one plot open at a time. Why do this?

aia:
Because plots are broken are hanging at the moment.
Having a plot displaying graphs is better than having a hung plot.

clothespin:
Hmm. On checking out the latest main, plots are working for me (Debian Linux Python 3.7.3) but the prompt hangs while one is displaying, like it did before #108 was merged. I can interact with the plot (zoom etc) as normal, it's not hung. So that's a bit of a regression. I can take a look this weekend at what might have broken that.

Either way, I'm not sure how removing .ion() fixes the problem; that would cause show to block again (the prompt would freeze but the plot gui would be responsive) but it sounds like your plot is freezing. I saw that behavior when I passed block=False to show rather than setting .ion(), which is why I went that route. I wonder if block=False would cause the desired behavior on your platform. What is your platform, incidentally?

DidierRLopes:
I tried this yesterday @aia and the reason why I didn't do it is because when closing the plots manually, they were freezing. But if you didn't see that then it's fine.

Another solution instead of plt.show() which is non blocking is to use
Plt.draw()
plt.pause(0.1)

The issue with this one is that I wasn't able to close the plot after

clothespin:
plt.show() is only non-blocking in interactive mode (after .ion() is called), if you remove the .ion() it would block. Can you post a script (series of commands) that causes the issue for you all so I can try to repro it on my end?

aia:
Interesting. On windows I get this:

With ion:
with-ion

Without ion:
without-ion

clothespin:
Huh. I was able to repro the plot hanging when I called sma on current main. So it looks like some plots are hanging but not others? Regular view worked fine, but it also blocked. Ultimately, in the short term I think we should do whatever unblocks the most users, so if that's this then okay, but I also think we should figure out the root cause so we can restore the nonblocking behavior. Do we have an issue for this currently?

DidierRLopes:
I got similar to @aia on Macbook.

The difference is that with ion my figure doesn't even pop up, and without it the plot is frozen.

@clothespin we should look for ways to restore the non blocking figures as I really liked that feature. The reason why you didn't see it in view cmd is that commands outside pred and ta menu have never been non blocking.

clothespin:
It is unfortunately late my time so I'll have to take a look tomorrow and over the weekend but for background:
matplotlib has an [event loop](https://matplotlib.org/3.3.4/users/interactive_guide.html) which processes all the gui activity for a given plot that has to run on the main thread. If interactive mode is off, this is accomplished for plt.show() by having plt.show() block the main thread so the event loop can take it over entirely and handle drawing updates and so on. With interactive mode on, show() returns immediately and the matplotlib gui event loop relies on hooks to run the gui while the main loop is waiting for input(). I think that broke somehow when we moved to prompt_toolkit; however, it may just be a matter of registering the correct input hooks somewhere like it explains in that linked doc under "Eventloop integration mechanism".

aia:
Another option is to put prompt_toolkit behind a feature flag.

I have a feeling we need feature flags...

https://featureflags.io/python-feature-flags/

clothespin:
That would give us the option to cut prompt_toolkit out for the batch workflow case, which solves my other primary gripe, so I'm all for feature flags.

DidierRLopes:
> is accomplished for plt.show() by having plt.show() block the main thread so the event loop can take it over entirely and handle drawing updates and so on. With interactive mode on, show() returns immediately and the matplotlib gui event loop relies on hooks to run the gui while the main loop is waiting for input(). I think that

That's the problem, is that prompt_toolkit also has an event loop, which takes priority over matplotlib's one.

I worked on making the matplotlib figure being generated inside a thread, althought it worked, there was a warning popping-up every time due to the fact that matplotlib is not thread-safe, see https://matplotlib.org/3.1.0/faq/howto_faq.html#working-with-threads

DidierRLopes:
I'm up for feature flags too :)

DidierRLopes (approved):

clothespin:
We can make the two [work together](https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/input_hooks.html) by providing draw_idle from the canvas to set_eventloop_with_inputhook.

aia:
I added poor-man's feature flags and turned off Prompt Toolkit.

DidierRLopes:
Great work!

PR #80, opened 2021-03-03
Import readline from stdlib for better ergonomics

comradesmith:
This allows users to press up in GamestonkTerminal to recall and replay their previous commands.

PR #77, opened 2021-03-02
fixed short interest and low float commands list error

CosminIvanov:
Also made input support arrow keys

Maybe we can do something smarter later but this is a #quickfix for it to work at this moment.

Issue #174, opened 2021-03-15
NOT RESPONDING


Quickpipsfx:
I didn't have this issue in the old version of stonk, I tried to build a fresh environment as well and still the figures are not responding, Help me if anyone faced the same issue.

p.s. the figures in discover function are showing but in TA and PRED it's not

![image](https://user-images.githubusercontent.com/80470045/111112545-59954a00-8586-11eb-9483-6b06a7b99aad.png)

aia:
see https://github.com/DidierRLopes/GamestonkTerminal/pull/158

Quickpipsfx:
@aia  am still unable to resolve it :/

DidierRLopes:
Hey @Quickpipsfx the issue is that you must have both USE_ION and USE_PROMPT_TOOLKIT set to true. This is not yet compatible, unfortunately.

Set one of them to true, and other to false, in here:
https://github.com/DidierRLopes/GamestonkTerminal/blob/8508b166f41d93e39ce00a3b91140d7fdc5dbcd9/gamestonk_terminal/feature_flags.py#L5

and it should work! Let me know if it doesn't!

Quickpipsfx:
> Hey @Quickpipsfx the issue is that you must have both USE_ION and USE_PROMPT_TOOLKIT set to true. This is not yet compatible, unfortunately.
> 
> Set one of them to true, and other to false, in here:
> https://github.com/DidierRLopes/GamestonkTerminal/blob/8508b166f41d93e39ce00a3b91140d7fdc5dbcd9/gamestonk_terminal/feature_flags.py#L5
> 
> and it should work! Let me know if it doesn't!

Yes I got it now, It's Fixed Thanks !!

PR #142, opened 2021-03-10
Better terminal with autocompletion + history

ricleal:
Hello all,

I just spent a few minutes prototyping a terminal using the `prompt-toolkit` (https://python-prompt-toolkit.readthedocs.io/).
We can do a proper autocompletion with nested options...

This was just 15 mins of my time. Let me know if you want to go ahead with this.

Thanks

PR #176, opened 2021-03-15
Non-breaking import readline

comradesmith:
Importing readline will improve user experience by letting users do a few things:
- easily access their command history from the current session using up/down arrows
- reverse history search with crtl+r
- use emacs keybindings (ctrl+a, etc)

I'll add this to the PR description, thanks!

Attempted in #77 and #80. However there were issues due to the lack of availability of readline in windows environments.

If merged this will add readline support _where available_, improving the UX on those platforms. There is also room to later import a solution available on windows in the `except` block.

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -2,6 +2,11 @@
 
 import argparse
 
+try:
+    import readline  # noqa: F401
+except ImportError:
+    pass
+
 from datetime import datetime, timedelta
 import pandas as pd
 from alpha_vantage.timeseries import TimeSeries


comradesmith:
gotta run linter brb

comradesmith:
> gotta run linter brb

done :)

aia:
What is that going to accomplish?

comradesmith:
> What is that going to accomplish?

Importing readline will improve user experience by letting users do a few things:
- easily access their command history from the current session using up/down arrows
- reverse history search with crtl+r
- use emacs keybindings (ctrl+a, etc)

I'll add this to the PR description, thanks!

aia:
How does it compare to Prompt-Toolkit that has been implemented here https://github.com/DidierRLopes/GamestonkTerminal/pull/142?

comradesmith:
> How does it compare to Prompt-Toolkit that has been implemented here #142?

Actually yeah that is better but its behind a feature flag so I wasn't aware. I'll close this and open a PR marking that FF as on so we can start to discuss making that the default behaviour as it is really nice

comradesmith:
> How does it compare to Prompt-Toolkit that has been implemented here #142?

Hi @aia @DidierRLopes I have re-opened this because it has many of the benefits of #142, but it is compatible with the prediction tools which had a bug with prompt toolkit (tracked in #174). So I think this is a good compromise to get some (but not all) UX improvements in front of users without impacting the prediction tools

aia (changes requested):
I'm not sure implementing multiple unfinished approaches to the same problem is the way to go.

This PR is about a UI feature and need to have a feature flag.

Also this PR adds an optional dependency and does not update poetry/requirements.

DidierRLopes:
I think #190 has addressed this in a nice way, I'll close this PR on that basis.

PR #177, opened 2021-03-15
Enable prompt toolkit by default

comradesmith:
Enable prompt toolkit by default. This feature was added in #142 and greatly improves the UX of the terminal.

If there are any downsides to enabling this lets discuss them here

### DIFF
--- a/gamestonk_terminal/feature_flags.py
+++ b/gamestonk_terminal/feature_flags.py
@@ -3,7 +3,7 @@ import os
 USE_COLOR = os.getenv("GTFF_USE_COLOR") or True
 USE_FLAIR = os.getenv("GTFF_USE_FLAIR") or "stars"
 USE_ION = os.getenv("GTFF_USE_ION") or True
-USE_PROMPT_TOOLKIT = os.getenv("GTFF_USE_PROMPT_TOOLKIT") or False
+USE_PROMPT_TOOLKIT = os.getenv("GTFF_USE_PROMPT_TOOLKIT") or True
 
 # Enable Prediction features
 ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT") or False


DidierRLopes:
> Enable prompt toolkit by default. This feature was added in #142 and greatly improves the UX of the terminal.
> 
> If there are any downsides to enabling this lets discuss them here

The only downside of this feature as far as I could tell was the fact that I wasn't able to make it work cleanly with the plots in pred and ta menus. Hence why @aia added both USE_PROMPT_TOOLKIT and USE_ION as feature flags, so that it could be up to the user to select the features he wanted.

If we could make these work together we could enable both by default. @clothespin mentioned that that is possible, I believe he's attempting this. See https://github.com/DidierRLopes/GamestonkTerminal/pull/158#issuecomment-797368914

DidierRLopes:
See #174 that's what happens when both ION and PROMPT_TOOLKIT are set to True.

comradesmith:
> See #174 that's what happens when both ION and PROMPT_TOOLKIT are set to True.

Oh I see, sorry I missed that. In that case I will try to replicate this with my simpler PR #176, as that might be a good enough stopgap measure to get some of those UX improvements

DidierRLopes:
> > See #174 that's what happens when both ION and PROMPT_TOOLKIT are set to True.
> 
> Oh I see, sorry I missed that. In that case I will try to replicate this with my simpler PR #176, as that might be a good enough stopgap measure to get some of those UX improvements

Sounds good. On that basis should we close this PR?

aia (changes requested):
Before enabling Prompt Toolkit by default, the conflicts with matplotlib Ion need to be resolved.

Non-blocking graphs are far more important than terminal completion.

DidierRLopes:
#190 addresses this. I'll close this on that basis.

PR #179, opened 2021-03-15
Checking for ticker before calling view and candle

buzzCraft:
Added a check for ticker when calling candle and view.
Calling candle without the check crashes the program, calling view just opens up a empty window.

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -119,12 +119,19 @@ def main():
             main_cmd = True
 
         elif ns_known_args.opt == "candle":
-            candle(
-                s_ticker, (datetime.now() - timedelta(days=180)).strftime("%Y-%m-%d")
-            )
+            if s_ticker:
+                candle(
+                    s_ticker, (datetime.now() - timedelta(days=180)).strftime("%Y-%m-%d")
+                )
+            else: 
+                print("No ticker selected. Use 'load ticker' to load the ticker you want to look at.")
+                main_cmd = True
 
         elif ns_known_args.opt == "view":
-            view(l_args, s_ticker, s_start, s_interval, df_stock)
+            if s_ticker:
+                view(l_args, s_ticker, s_start, s_interval, df_stock)
+            else: 
+                print("No ticker selected. Use 'load ticker' to load the ticker you want to look at.")
             main_cmd = True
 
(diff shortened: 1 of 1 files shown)


buzzCraft:
First pull request iv done, and clearly I did something wrong. Just added some checks.

aia:
You want to fix the linter.
Right not the linter is asking to apply python-black formatting.

buzzCraft:
Thanks, black is a new tool to me, but it looks great, and i will be using it in the future.

terminal.py
```diff
@@ -119,12 +119,28 @@ def main():
...
+                candle(
+                    s_ticker,
+                    (datetime.now() - timedelta(days=180)).strftime("%Y-%m-%d"),
+                )
+
+            else:
+                print(
+                    "No ticker selected. Use 'load ticker' to load the ticker you want to look at."
+                )
+                main_cmd = True
```
DidierRLopes, line 133: I think this 
`main_cmd = True`
should be outside the else condition.

DidierRLopes (changes requested):
Just a minor detail. Once that is done, I'll approve and merge this :)

Thank you!

terminal.py
buzzCraft: I agree, and fixed. I was just unsure since it wasn't there at all. Didn't want to mess up something i didn't understand :)

DidierRLopes: Yea, I saw it wasn't there. Well spotted! Thank you!

If you scroll a bit down you can see that when you call one of the main menus, they return this b_quit flag which in case is true it exists the program, in case it's false it just means the user wants to go one level up. Since the user will be coming to this main level again, after being in a sub-level, we want to print the main menu again. Hence why if b_quit flag is false we print the help menu.

That main_cmd exists for the case we use one of those main functions, to say that we don't want to show the help menu in those cases, unless the user selects "help".

Hope this made sense! 
Once the checks are done I'll merge this, cheers :)

buzzCraft: I understood the function, but didn't understand why it wasn't present for candle already, make sense it was just a typo. 

I like this project and ill try to dive into the code and see if i can help in some way :) Keep up the good work

PR #190, opened 2021-03-16
Registers pyplot.draw as inputhook for prompt_toolkit

clothespin:
This patch resolves the issue that was causing interactive plots to hang when prompt_toolkit is used by registering [matplotlib.pyplot.draw()](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.draw.html) as the input hook for the prompt session, which means the graphs will get updated whenever the prompt is blocked waiting for input. It will still be necessary to use the feature flag to disable prompt toolkit for batch scripts (since it expects a tty and doesn't work if stdin is just a file) but for the interactive use case this should enable both prompt_toolkit and plt.ion to coexist peacefully so it can work as the default. I checked with basic (view) plots and it appears to work well.

Potentially unblocks #177.

### DIFF
--- a/gamestonk_terminal/feature_flags.py
+++ b/gamestonk_terminal/feature_flags.py
@@ -1,23 +1,13 @@
+from distutils.util import strtobool
 import os
 
-# env vars are strings not booleans
-# check for "false" or "0"
-# any other value is true (including empty string)
-
-USE_COLOR = os.getenv("GTFF_USE_COLOR", "true").lower() not in ["false", "0"]
-USE_FLAIR = os.getenv("GTFF_USE_FLAIR", "stars")
-USE_ION = os.getenv("GTFF_USE_ION", "true").lower() not in ["false", "0"]
-USE_PROMPT_TOOLKIT = os.getenv("GTFF_USE_PROMPT_TOOLKIT", "false").lower() not in [
-    "false",
-    "0",
-]
-
+USE_COLOR = strtobool(os.getenv("GTFF_USE_COLOR", "True"))
+USE_FLAIR = os.getenv("GTFF_USE_FLAIR") or "stars"
+USE_ION = strtobool(os.getenv("GTFF_USE_ION", "True"))
+USE_PROMPT_TOOLKIT = strtobool(os.getenv("GTFF_USE_PROMPT_TOOLKIT", "False"))
 
 # Enable Prediction features
-ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT", "false").lower() not in ["false", "0"]
+ENABLE_PREDICT = strtobool(os.getenv("GTFF_ENABLE_PREDICT", "False"))
 
 # Enable plot autoscaling
-USE_PLOT_AUTOSCALING = os.getenv("GTFF_USE_PLOT_AUTOSCALING", "false").lower() not in [
-    "false",
-    "0",
-]
+USE_PLOT_AUTOSCALING = strtobool(os.getenv("GTFF_USE_PLOT_AUTOSCALING", "False"))
--- a/gamestonk_terminal/feature_flags.py
+++ b/gamestonk_terminal/feature_flags.py
@@ -1,12 +1,12 @@
 import os
 
-USE_COLOR = os.getenv("GTFF_USE_COLOR") or True
+USE_COLOR = os.getenv("GTFF_USE_COLOR").lower() not in ['false', '0'] or True
 USE_FLAIR = os.getenv("GTFF_USE_FLAIR") or "stars"
-USE_ION = os.getenv("GTFF_USE_ION") or True
-USE_PROMPT_TOOLKIT = os.getenv("GTFF_USE_PROMPT_TOOLKIT") or False
+USE_ION = os.getenv("GTFF_USE_ION").lower() not in ['false', '0'] or True
+USE_PROMPT_TOOLKIT = os.getenv("GTFF_USE_PROMPT_TOOLKIT").lower() not in ['false', '0'] or False
 
 # Enable Prediction features
-ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT") or False
+ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT").lower() not in ['false', '0'] or False
 
 # Enable plot autoscaling
-USE_PLOT_AUTOSCALING = os.getenv("GTFF_USE_PLOT_AUTOSCALING") or False
+USE_PLOT_AUTOSCALING = os.getenv("GTFF_USE_PLOT_AUTOSCALING").lower() not in ['false', '0'] or False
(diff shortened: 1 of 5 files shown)


DidierRLopes:
What OS are you using?

I've tested on my macbook and when `inputhook=pyplot.draw` and `USE_PROMPT_TOOLKIT = True`, I don't seem to get any prompt toolkit feature.

If I remove the `inputhook=pyplot.draw` prompt toolkit works again.

Can you check this?

PS: The main view wasn't using view in the first place, you need to go to TA or PRED menu to assess if they can coexist peacefully :)

clothespin:
OK, I think I worked out what was going on here. It turns out the examples in the prompt_toolkit docs lag behind the actual code, where prompt_toolkit removed the inputhook keyword arg in 3.0. Since we silently catch and ignore any exception related to prompt creation, I was just falling back to the default input without realizing it. I've got another approach that should work, I'll go ahead and try that.

DidierRLopes (commented):
Just tested this. NICE!!! It's working well, I think we can set USE_ION and USE_PROMPT_TOOLKIT both to True per default.

Can you fix the lint complaints before I merge this? Thanks!

clothespin:
The linter is fixed (see most recent commit) but the test task is failing because it uses files as stdin, which triggers the issue with stdin being a file and prompt toolkit. Those tests should probably run with USE_PROMPT_TOOLKIT as well as USE_ION set to off, since there is no interactive terminal (stdin is just a file) and plots wouldn't work anyway because there's no display available.

clothespin:
So, the trouble I'm running into now is that I can't pass the test suite because the test runs with prompt_toolkit enabled, despite what I set GTFF_USE_PROMPT_TOOLKIT to in workflow.yml. It looks like the feature flags code needs to be reworked slightly so that a string value of "false" in that environment variable will turn off the feature.

clothespin:
Yeah, not sure how to proceed here. No matter what I set the feature flag env var to, the test tries to run prompt_toolkit.

aia:
Tests are not importing feature flags and are not picking up environment variables.

We need to change that.

gamestonk_terminal/menu.py
```diff
@@ -1,11 +1,22 @@
...
+    return False
+
+
 history_file = os.path.join(os.path.expanduser("~"), ".gamestonk_terminal.his")
 
 try:
     session = PromptSession(history=FileHistory(history_file))
+    set_eventloop_with_inputhook(inputhook)
 except Exception as e:  # noqa: F841
-    session = None
```
aia, line 11: I think the removal of ```session = None``` breaks the tests.

clothespin: Yeah, this was what did it. I added a warning message here so that if this code path is taken (prompt_toolkit is configured but constructing the session fails) it'd be clear what was going on. Thanks for the tip.

aia (approved):
Very nice!

gamestonk_terminal/feature_flags.py
```diff
@@ -1,12 +1,23 @@
...
+USE_ION = os.getenv("GTFF_USE_ION", "true").lower() not in ["false", "0"]
+USE_PROMPT_TOOLKIT = os.getenv("GTFF_USE_PROMPT_TOOLKIT", "false").lower() not in [
+    "false",
+    "0",
+]
+
 
 # Enable Prediction features
-ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT") or False
+ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT", "false").lower() not in ["false", "0"]
```
DidierRLopes, line 17: Why did you add this?

DidierRLopes (commented):
I've tested this and it's working fine!!

Question though, the
```
plt.close('all')
```
in the prediction_techniques menu loop isn't closing the plot. Is this expected? Is there a way to do so?

gamestonk_terminal/feature_flags.py
DidierRLopes: Could it be that in the environment variable you were using false as a string, rather than the bool "False"? Asking because I'm curious :)

aia: I also don't think this is necessary and looks very confusing.

```ENABLE_PREDICT = os.getenv("GTFF_ENABLE_PREDICT") or False``` will be a non-empty string if an environment variable is set or False if an environment variable is not set. Python behavior for non-empty strings is the same behavior as for the True value in the conditional statement. That is all that we need.

clothespin: This is correct for options that default to False. However, I don't think the same pattern works for options that default to True, which would look like `ENABLE_FOO = os.getenv("GTFF_ENABLE_FOO") or True`. There's nothing you can set an environment variable to in that case that will cause ENABLE_FOO to evaluate to False. However, at the time I made those changes I wasn't aware that, per your comment, the tests aren't importing the feature flags at all. I totally agree that it's ugly and dumb, but one of the PRs that's blocked behind this one sets prompt toolkit as enabled by default, and if we can't then turn it off with an environment variable we're right back to the original problem.

clothespin: Regarding the type of the environment variable: from the OS level, you can't set a type. os.getenv returns a string or whatever value is specified in the default kwarg (set to None if unspecified) if it can't find the key in the environment. Python evaluates "", 0, and None as falsey and pretty much everything else is truthy, and that's the conversion going on when we're doing comparisons against the result of os.getenv.

That said, I would absolutely welcome a cleaner and less ugly way of doing this, I'm just not sure what that looks like.

DidierRLopes: Let me have a got at this tomorrow, and I'll let you know :)

aia: > There's nothing you can set an environment variable to in that case that will cause ENABLE_FOO to evaluate to False.

That's a great point and something I missed.

clothespin: OK, there is a better way to do this, Python ships a [strtobool function in distutils.utils](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool) that does exactly what I'm trying to do with the awkward if not in [] construct. I'll use that instead.

aia: Check this out - https://github.com/DidierRLopes/GamestonkTerminal/pull/230

clothespin: > Check this out - #230

I'll leave it up to you and @DidierRLopes as maintainers to decide whether you prefer the version I already pushed or this one, but this one might have some unexpected behavior, particularly regarding integration with other (Unix) tools that output '1' and '0' or anything that converts strings to lowercase. `"true`, `'1'`, and `1` all evaluate as False with that patch but are handled as expected with `distutils.util.strtobool()`.

aia: I like your approach. Want to go with it?

aia: I'll go ahead with the merge.

Thank you for great work!

DidierRLopes: Great work on this @clothespin . Thank you!

PR #194, opened 2021-03-16
Made use of prawcore.exceptions.ResponseException

henrytdsimmons:
Changed the reddit search functions to catch specific praw response exceptions rather than general exceptions.

### DIFF
--- a/gamestonk_terminal/sentiment/reddit_api.py
+++ b/gamestonk_terminal/sentiment/reddit_api.py
@@ -141,7 +141,7 @@ def watchlist(l_args):
                         s_watchlist_tickers += f"{t_ticker[1]} {t_ticker[0]}, "
                     n_tickers += 1
                 except Exception as e:
-                    print(e)
+                    print(e, "\n")
                     # pass
             if n_tickers:
                 print(
@@ -151,7 +151,7 @@ def watchlist(l_args):
         print("")
 
     except Exception as e:
-        print(e)
+        print(e, "\n")
         print("")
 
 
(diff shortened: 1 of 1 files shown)


DidierRLopes (changes requested):
Captura de ecrã 2021-03-16, às 13 24 22

Can you add a return after each of the print("") ? It would prevent to keep iterating through the loop! Thanks

DidierRLopes (changes requested):
Break isn't enough because it will try to parse something, and if this condition happens that's related with API token, so you can just simply return directly.

Also, on the print(...) can you add a new line at the end, like print("...\n")

Just test with wrong reddit api keys and make sure that the user experience is good :) 

Thanks!

gamestonk_terminal/sentiment/reddit_api.py
```diff
@@ -356,23 +348,16 @@ def popular_tickers(l_args):
...
-            )
-            print("")
-
-    except Exception as e:
-        print(e)
-        print("")
+    except ResponseException:
+        print(
+            "Received a response from Reddit with an authorization error. check your token.\n"
+        )
```
DidierRLopes, line 360: missing a return here

gamestonk_terminal/sentiment/reddit_api.py
```diff
@@ -470,9 +449,11 @@ def spac_community(l_args):
...
                 else:
                     break
-            except Exception as e:
-                print(e)
-                print("")
+            except ResponseException:
+                print(
+                    "Received a response from Reddit with an authorization error. check your token.\n"
+                )
+                break
```
DidierRLopes, line 456: and here

DidierRLopes (changes requested):

henrytdsimmons:
Should be done now. Sorry for needing so many iterations!

DidierRLopes (approved):

DidierRLopes:
No problem! Thanks for the contribution :)

Issue #180, opened 2021-03-15
[IMPROVE] Automatically adjust the figure window size according to screen size 


avocardio:
**What's the feature that should be improved?**
I have to adjust the view or pred figures every time I run them, and they sometimes don't scale well. 

**Describe how you would like the feature improved**
I would like the figures to open a little bit bigger or at least according to some screen ratio. 

I have no idea how to do this, but I was going to try this myself but I have no idea where I would find this under the files? @DidierRLopes is there one single file that draws the figures or would I need to dig into every single file? Thanks for the help.

DidierRLopes:
Hey @avocardio ,

This contribution would be extremely welcome.

The lines you are looking for are:
`plt.figure()`

For instance in Arima case, that is:
https://github.com/DidierRLopes/GamestonkTerminal/blob/ce31ff6245489b535385105a3a3560e9b9fb7bac/gamestonk_terminal/prediction_techniques/arima.py#L171
and
https://github.com/DidierRLopes/GamestonkTerminal/blob/ce31ff6245489b535385105a3a3560e9b9fb7bac/gamestonk_terminal/prediction_techniques/arima.py#L244

There's always 2 plt.figures() per file, that's due to backtesting.

If you found a way that this scaling would work well for all OS, you could literally due a plt.figure() Replace all.

buzzCraft:
I see that https://pypi.org/project/screeninfo/ will provide screen resolution for Win/Mac/Unix. If you are open to have yet another package imported, I'll start looking into it.

DidierRLopes:
As long as it improves the user experience, I'm happy with more packages being imported! Thank you for this :)

PS: #189 will be merged soon. This touches the Neural Network figures code, so watch-out for conflicts when you create the PR.

buzzCraft:
I'm thinking about making config_plot.py where you can set the size of plots. After looking into autoscaling, I don't think this is the way to go. I'll keep it as an option and let you decide later.

DidierRLopes:
Sounds good to me!

If you think autoscaling may not be suitable for everyone,  but for some users it will. You can add it behind a feature flag, look into feature_flags.py.

E.g.
`
if gtff.USE_PLOT_AUTOSCALING:
   plt.figure(autoscaling nice feature)
else:
   plt.figure( get resolution from config_plot file)
`

DidierRLopes:
@buzzCraft has worked on this and his improvements are already merged.

[...]

PR #196, opened 2021-03-16
Autoscaling for candle and view

buzzCraft:
As requested in #180 I added a way to autoscale all plots. Implemented the changes in candle and view, and can easy be implemented to all plots. 

Added screeninfo to requirements.txt
Screeninfo should work cross platform, but i'v only tested it on Windows 10


buzzCraft:
Getting an error when importing screeninfo. Iv included it into:
requirements.txt
requirements-full.txt
requirements-win.txt
not sure where i went wrong (unit testing is a new field for me)

DidierRLopes:
I think you need to update the poetry lock file.

Can you do: "poetry add screeninfo" and then test with "poetry build". This should update both pyproject.toml and poetry.lock.

buzzCraft:
Ok, that seems to work, please check poetry.lock, as I don't understand all changes that happen there.
If this scaling feature is something you want to use, i'll update all the plots later.

gamestonk_terminal/config_plot.py
```diff
@@ -0,0 +1,9 @@
+SCREEN_HEIGHT = 0
+SCREEN_WIDTH = 0
```
DidierRLopes, line 2: I think those values are never used, as screeninfo library is getting it for us

gamestonk_terminal/config_plot.py
```diff
@@ -0,0 +1,9 @@
...
+
+PLOT_HEIGHT_PRECENTAGE = 100
+PLOT_WIDTH_PRECENTAGE = 50
+
+PLOT_HEIGHT = 800
+PLOT_WIDTH = 800
```
DidierRLopes, line 8: Also, the defaults look like this to me:

USE_PLOT_AUTOSCALING = True
Captura de ecrã 2021-03-16, às 19 22 03

USE_PLOT_AUTOSCALING = False
Captura de ecrã 2021-03-16, às 19 19 29

May I suggest these:
```
PLOT_HEIGHT_PRECENTAGE = 50
PLOT_WIDTH_PRECENTAGE = 70

PLOT_HEIGHT = 500
PLOT_WIDTH = 800
```

DidierRLopes: Also can you add a comment above these parameters that:
PLOT_HEIGHT_PRECENTAGE
PLOT_WIDTH_PRECENTAGE
are used when AUTO_PLOT is set to True,

And that the others are used when AUTO_PLOT is False.

DidierRLopes: Precentage is mispelled, it should be "Percentage"

gamestonk_terminal/helper_funcs.py
```diff
@@ -353,3 +355,25 @@ def get_flair() -> str:
         return flair[gtff.USE_FLAIR]
 
     return ""
+
+
+def get_screeninfo():
+    screens = get_monitors()  # Get all available monitors
+    main_screen = screens[0]  # Choose the first screen as main screen
```
DidierRLopes, line 362: Would be nice having the MAIN/SECONDARY screen as a configurable parameter!

DidierRLopes (changes requested):
Some minor tings because I'm picky, but overall I just tested this and it is great!! Feel free to apply it to the TA and PRED figures please. #welikeniceplots

gamestonk_terminal/config_plot.py
buzzCraft: Thanks for the reply. I'll get right to it:)
I really appreciate the feedback, as I'm new to coding in teams, and got no formal / educational coding background.

DidierRLopes: For someone with no coding background you're doing amazing!! I'll wait for your changes to merge this :) Thank you!

buzzCraft:
Okay, think I covered all of your request. I also added scaling to momentum, overlap, trend_indicators, volatility and volume. At of this moment I don't have my env ready for predictions, so I didn't want to start adding scaling there without the possibility to test it. 
I'll get predictions up and running as soon as possible (within a day or two).

DidierRLopes (approved):

DidierRLopes:
Again, this is amazing. Great contribution. Thank you for this :)

PR #197, opened 2021-03-16
removed rogue GME ticker in cell

lolrenx:
cleaned up DD template notebook


aia:
You need that cell because that cell indicates to Papermill what are the notebook parameters - https://github.com/nteract/papermill#parameterizing-a-notebook.

GME is the default value there because we like the stock.

lolrenx:
> You need that cell because that cell indicates to Papermill what are the notebook parameters - https://github.com/nteract/papermill#parameterizing-a-notebook.
> 
> GME is the default value there because we like the stock.

yes, the tag does survive without having GME in every report. the parameter tag is still there. 
if this is to be an Easter Egg, maybe can it be more subtle 🤷‍♂️

aia:
You have very interesting work going on here.
Are you going to open a PR to add options features?

Btw, I add datetime to know for sure when was the last time the full notebook run.
I oftentimes re-run parts of the notebook.

DidierRLopes:
Hey @aia,

I just had a long call with @lolrenx showing him how the code was structured and all, and our overall plans going forward. The idea is that he'll start working on the options menu since he has a lot of experience in this arena, and knows what users are looking for.

aia:
Awesome!

Issue #167, opened 2021-03-13
[FR] Add economic data from FRED


alexanderruchti:
**What is FRED?**
FRED is short for the Federal Reserve Bank of St.Louis. It is one of the best free providers on economic data.

**Why would adding FRED benefit GamestonkTerminal?**
FRED offers many data series for free that give investors a decent view on how the economy (and therefore potentially the market) is going. Some noteworthy series are the _interest rate yield curve_, the _weekly initial unemployment rate_ figures, and the _break even inflation rates_.

**How to start**
FRED already offers a Python [API](https://fred.stlouisfed.org/docs/api/fred/). 

PS: I am not to lazy to do it myself, I am just a noob who only installed Git for the first time today and still don't know enough to meaningfully contribute.

DidierRLopes:
Thanks for this, will have a look at this hopefully later this week! :)

jmaslek:
I am willing to take a look and add something for this feature.

DidierRLopes:
@jmaslek addressed this already in #199 ! Closing this due to that

Issue #87, opened 2021-03-03
[FR] Interest rate data


nanman357:
**What's the problem of not having this feature?**
Would it be possible to have interest rate data available through the terminal? 
Example: USD interest rate curve, official quotes from ON tenor to 1Y (example sourse: https://stooq.com/q/d/?s=ukousd1y )
Further tenors from 2Y can be from swap quotes like here : https://www.chathamfinancial.com/technology/us-market-rates

**Describe the solution you would like**
Have a history of IR tenors available. Without this, discounting is impossible.

**Describe alternatives you've considered**
ALternatively, finding a list of discount factors for future dates (ex. if today = 2020-12-31, find df's starting from 2021-01-04 to say 2051-01-02).

**Additional information**
Besides USD, it would be beneficial to have this solution for other currencies such as GBP, EUR, JPY and other nieche currencies as CND, AUD, NZD, PLN, SGD etc. (just an example list of currencies).

This would require introduce a different calendat for each location (because ex. there's no Marthin Luther King day in Eurozone).

DidierRLopes:
I believe #199 addresses this. Is this correct @jmaslek ?

jmaslek:
I think so.  Most of those should be on the fred site - which can be accessed with the custom function.

Did not do the discount factors.

DidierRLopes:
I'll close this as it has been addressed from #199 . Feel free to re-open it if you disagree

PR #199, opened 2021-03-16
Fred #167

jmaslek:
Hi. I tried to address issue #167 for a feature request.

I added a fred menu that has 2 options.  Either GDP or custom.  The custom just requires the user to know the SERIES ID for what data they want.  For custom data, it can be plotted/ displayed and GDP has a plot option as well

Hopefully this would allow for further dev if anyone wants to add specific yield curves/ unemployment etc.

Note that an API Key is required.


gamestonk_terminal/fred/README.md
```diff
@@ -0,0 +1,46 @@
+# fred
+
+The purpose of fred is to have the ability to look at economic data.
+
+* [GDP](#GDP)
```
DidierRLopes, line 5: Would be nice to have this command in lower case to match the remaining style of the program.

gamestonk_terminal/fred/README.md
```diff
@@ -0,0 +1,46 @@
+# fred
```
DidierRLopes, line 1: Capital letters preferably here

gamestonk_terminal/fred/README.md
```diff
@@ -0,0 +1,46 @@
...
+    * Get custom fred data  
+
+## GDP <a name="GDP"></a>
+```text
+usage: GDP [-n N_TO_GET] [-s START_DATE] [-p PLOT]]
+```
+Gets GDP data.  Data is released quarterly
+* -n : Number of points to display.  If unspecified, it will display all available.
+* -s : Start Date.  Format m/d/y. First point to get
+* -p : Option to plot.  Defaults to False
```
DidierRLopes, line 17: Is it enough to only look at the data? Genuinely asking, as I've not used this data before. 

If the answer to this is no, I'd make plot as default. Otherwise, this is good.

gamestonk_terminal/fred/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+To get the last 20 reported GDP values:
+````
+GDP -n 20
+````
+To plot all GDP data since 1/1/2010:
+````
+GDP -s 1/1/2010 -p True
```
DidierRLopes, line 25: To follow the remaining examples, I would just add a plot and a print of the result. But I can do this myself, no worries here!

gamestonk_terminal/fred/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+The fred database has many different datasets freely available.  They each have their own custom SERIES_ID which can be obtained through the website.
+
+````
+usage: cust [-id] [-s START_DATE] [-p PLOT] [-disp DISP]
+````
+* -id : Series ID for FRED data.  Required argument
+* -s : Start data to acquire data.  Defaults to 1/1/2020
+* -p : Option to plot. Defaults to False
+* -disp : Option to print data to console.  Defaults to True.
```
DidierRLopes, line 39: If the default of print data to console is set to True, shouldn't the flag be called "--nodisp"? 

Again, I would probably add plot and stdout print as default, but I'm not knowledgeable enough here to have a saying, so I'll take your input.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,122 @@
...
+
+    parser = argparse.ArgumentParser(
+        add_help=False,
+        prog="Custom",
+        description="""
+                    Custom Data
+                """,
+    )
+    parser.add_argument(
+        "-id", dest="series_id", required=True, type=str, help="FRED Series ID"
```
DidierRLopes, line 81: Due to the fact that this argument is required, if the user adds the help flag (-h) the program will exit, you need to catch that exception in here. See what I did here https://github.com/DidierRLopes/GamestonkTerminal/pull/182/files for the regression command

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,122 @@
...
+        type=int,
+        default=-1,
+        required=False,
+        help="Number of GDP Values to Grab",
+    )
+
+    parser.add_argument(
+        "-s",
+        dest="start_date",
+        type=str,
```
DidierRLopes, line 31: You can use the valid_date function in helper_funcs, since it checks for the validity of a date. 

I'd also use the same datetime structure the other functions use: YYYY-mm-dd. 

I think that as a default you should just plot/print all data available, rather than setting a default start time.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,122 @@
...
+    parser.add_argument(
+        "-id", dest="series_id", required=True, type=str, help="FRED Series ID"
+    )
+    parser.add_argument(
+        "-s",
+        dest="start_date",
+        type=str,
+        default="1/1/2020",
+        required=False,
+        help="Date to Start",
```
DidierRLopes, line 89: Same as previously.

On the help I would add the formatting of the date.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    parser.add_argument(
+        "-p", dest="plot_", type=bool, default=False, required=False, help="Plot data"
+    )
+    parser.add_argument(
+        "-disp", dest="disp_", type=bool, default=True, required=False, help="Printdata"
```
DidierRLopes, line 95: missing whitespace  on "Printdata"

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,122 @@
...
+        "-s",
+        dest="start_date",
+        type=str,
+        default="1/1/2020",
+        required=False,
+        help="Date to Start",
+    )
+
+    parser.add_argument(
+        "-p", dest="plot_", type=bool, default=False, required=False, help="Plot GDP"
```
DidierRLopes, line 38: I'd keep the convention that has being used, without the "_" at the end of the variable name

gamestonk_terminal/fred/fred_menu.py
```diff
@@ -0,0 +1,65 @@
...
+
+def print_fred():
+    """ Print help """
+
+    print("\nFred Economic Data:")  # https://github.com/JerBouma/FundamentalAnalysis
+    print("   help          show this menu again")
+    print("   q             quit this menu, and shows back to main menu")
+    print("   quit          quit to abandon program")
+    print(" ")
+    print("   GDP           get GDP")
```
DidierRLopes, line 18: I'd keep the GDP in lower case

gamestonk_terminal/fred/fred_menu.py
```diff
@@ -0,0 +1,65 @@
...
+from gamestonk_terminal.helper_funcs import get_flair
+from gamestonk_terminal.menu import session
+from prompt_toolkit.completion import NestedCompleter
+from gamestonk_terminal.fred import fred_api
+
+
+def print_fred():
+    """ Print help """
+
+    print("\nFred Economic Data:")  # https://github.com/JerBouma/FundamentalAnalysis
```
DidierRLopes, line 13: That comment is no longer necessary

gamestonk_terminal/fred/fred_menu.py
```diff
@@ -0,0 +1,65 @@
...
+
+        except SystemExit:
+            print("The command selected doesn't exist\n")
+            continue
+
+        if ns_known_args.cmd == "help":
+            should_print_help = True
+
+        elif ns_known_args.cmd == "q":
+            # Just leave the FA menu
```
DidierRLopes, line 54: the "FRED" menu

gamestonk_terminal/main_helper.py
```diff
@@ -52,6 +52,8 @@ def print_help(s_ticker, s_start, s_interval, b_is_market_open):
     print(
         "   sen         sentiment of the market, \t from: reddit, stocktwits, twitter"
     )
+    print("   fred        Data from Federal Reserve Bank of St.Louis")
```
DidierRLopes, line 55: I'd try to keep the structure of the main menu

Captura de ecrã 2021-03-16, às 20 13 53

DidierRLopes (changes requested):
Sorry for all the commenting, but without testing in place, it's good to have the code with a similar architecture/style.

Thank you for this feature, this is extremely helpful. I hope you don't mind fixing those comments I've made.

gamestonk_terminal/fred/README.md
jmaslek: So I can't say I've used it either.  My instinct would be that it depends on the data.  For example, since GDP is quarterly, doesn't make sense to plot if I am trying to YOLO based on GDP growth from last week.

If I was looking at unemployment data, probably better to plot.  Ill default to true for now.

gamestonk_terminal/fred/README.md
jmaslek: Ill set the default print to False (along with plotting true). 

This way the user can see the numbers if they want, but won't get their console just spewing numbers.

gamestonk_terminal/fred/fred_api.py
jmaslek: Changed the date time.

I changed he default time to Jan 1 2015.  I think some of the data goes back into the 90s and beyond, so unless someone specifically wants that much data, I think 5 years is plenty.

jmaslek:
So I made most of the changes.  Last one is the exception catching I will look at later.

I have a couple images for plots/ outputs for the readme.  To make that work, should I add them to the images folder?

gamestonk_terminal/fred/README.md
DidierRLopes: Sounds good to me!

gamestonk_terminal/fred/README.md
DidierRLopes: Yea, I guess it depends on the number of data.

E.g. viewing historical data, doesn't make sense to output this by default. Once using a prediction algorithm to predict next 5 market days, it makes sense to print those.

gamestonk_terminal/fred/fred_api.py
DidierRLopes: Good point. In that case you can even go as far as 3 years, would be good to have that thought written somewhere :)

DidierRLopes:
> So I made most of the changes. Last one is the exception catching I will look at later.
> 
> I have a couple images for plots/ outputs for the readme. To make that work, should I add them to the images folder?

The exception catching is a 2-liner, you just need to add this:
```
    except SystemExit:
         print("")
```
above the `except Exception as e` to prevent the code from exiting. In all the commands implemented that have a "required" parameter.

Regarding the images/output, the way can do it is the following:
- Create new issue
- Drag image to it and drop it
- Copy the images URL and paste it into readme

Here's a youtube video:
https://www.youtube.com/watch?v=qIaWozjDyPk

DidierRLopes:
Thank you for making the changes! :) Can you rebase your code to avoid the conflicts, and make sure it passes our lint testing, so that I can merge it after? Thank you!

jmaslek:
Let me know if I need to do anything else!  

Ill add the photos as well for the readme.

DidierRLopes:
It's late already in here. I'll try to review/test this during my lunch break tomorrow, and let you know by then.

Thank you for this, it's great to see contributors adding bugfixes, but new features it's even nicer!

gamestonk_terminal/fred/fred_menu.py
```diff
@@ -0,0 +1,65 @@
...
+def print_fred():
+    """ Print help """
+
+    print("\nFred Economic Data:")
+    print("   help          show this menu again")
+    print("   q             quit this menu, and shows back to main menu")
+    print("   quit          quit to abandon program")
+    print(" ")
+    print("   gdp           get GDP")
+    print("   cust          User Specified FRED Data")
```
DidierRLopes, line 19: Below this you can add an additional::
```
print("")
```

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+
+        for date, val in lastn.iteritems():
+            print(f'Date: {date.strftime("%m-%d-%Y")}, GDP: {val} ')
+
+        if ns_parser.plot:
+            plt.figure()
+            lastn.plot(style="-ok")
+            plt.xlabel("Time")
+            plt.ylabel("GDP")
+            plt.show()
```
DidierRLopes, line 70: Below this you can add an additional:
```
print("")
```

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+
```
DidierRLopes, line 71: Would be good to keep the same figure style, as in:
Captura de ecrã 2021-03-17, às 12 23 30
to give a better experience.

So from here you can add:
Captura de ecrã 2021-03-17, às 12 20 18

* Grid
* Title
* Add units on the ylabel

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+        if ns_parser.disp:
+            for date, val in data.iteritems():
+                print(f'Date: {date.strftime("%m-%d-%Y")}, DATA: {val} ')
+        try:
+            if ns_parser.plot:
+                plt.figure()
+                data.plot(style="-ok")
+                plt.xlabel("Time")
+                plt.ylabel("DATA")
+                plt.show()
```
DidierRLopes, line 133: Same as for figure above.

In this one I would replace the label "DATA" by the name of the series. This may seem meaningless, but if someone wants to save the figure directly, they'll have all the information on the image.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+        ns_parser = parse_known_args_and_warn(parser, l_args)
+
+        if not ns_parser:
+            return
+
+        data = fred.get_series(ns_parser.series_id, ns_parser.start_date)
+        if ns_parser.disp:
+            for date, val in data.iteritems():
+                print(f'Date: {date.strftime("%m-%d-%Y")}, DATA: {val} ')
+        try:
```
DidierRLopes, line 127: You shouldn't need a second try-catch block inside another try-catch one.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+            if ns_parser.plot:
+                plt.figure()
+                data.plot(style="-ok")
+                plt.xlabel("Time")
+                plt.ylabel("DATA")
+                plt.show()
+        except Exception as e:
```
DidierRLopes, line 134: This exception shouldn't be necessary, as you have a 2 layer of try-catch blocks.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+
+    parser = argparse.ArgumentParser(
+        add_help=False,
+        prog="Custom",
+        description="""
+                    Custom Data
+                """,
+    )
+    parser.add_argument(
+        "-id", dest="series_id", required=True, type=str, help="FRED Series ID"
```
DidierRLopes, line 89: This should be: `"-i", "--id"`

Following UNIX convention where you use 1 '-' when a single character, but '--' when more.

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+
+    parser.add_argument(
+        "-s",
+        dest="start_date",
+        type=valid_date,
+        default="2015-01-01",
+        required=False,
+        help="Date to Start",
```
DidierRLopes, line 98: You can add the formatting expected here, e.g.
```
help=The starting date (format YYYY-MM-DD) of the data
```

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+    try:
+        ns_parser = parse_known_args_and_warn(parser, l_args)
+
+        if not ns_parser:
+            return
+
+        data = fred.get_series(ns_parser.series_id, ns_parser.start_date)
+        if ns_parser.disp:
+            for date, val in data.iteritems():
+                print(f'Date: {date.strftime("%m-%d-%Y")}, DATA: {val} ')
```
DidierRLopes, line 126: Instead of doing this iteration, would be better to have a nice table format, where you had:
```
TIME                  UNRATE
2020-02-03     123123123
2020-02-03     123123123
2020-02-03     123123123
...
```

Same for gdp command

gamestonk_terminal/fred/fred_api.py
```diff
@@ -0,0 +1,143 @@
...
+
+    parser.add_argument(
+        "-s",
+        dest="start_date",
+        type=valid_date,
+        default="2015-01-01",
+        required=False,
+        help="Date to Start",
+    )
+    parser.add_argument(
```
DidierRLopes, line 100: Given that the data will never be much, I think the plot and disp arguments are unnecessary.

Also, you were parsing this in an inefficient way. Because instead of having to do:
`-p true` or `-p false` you can set a flag to have the value you want.

E.g. If the user doesn't add anything you display a plot by default, if in the command he does something like:
```
gdp --noplot
```
Then we recognize that `--noplot` is a flag and we set that to the value we want.

This is done using:
```
parser.add_argument(
        "-q",
        "--quarter",
        action="store_true",
        default=False,
        dest="b_quarter",
        help="Quarter fundamental data flag.",
    )
```
Notice the default to False, but if the user sets this flag ('store_true') that value will be True instead.

Anyway, for these functions I would just keep plots and prints a default

DidierRLopes (changes requested):
A few more comments to improve this even further. If you work on these, that should be enough for me to merge it.

Thanks in advance!

gamestonk_terminal/fred/fred_api.py
jmaslek: I had that just in case some of the fred data was non numeric (no idea if that exists).  Wanted to print the values but catch the plotting.  

Got rid of it and if it happens for some reason, I am sure someone will submit an issue

gamestonk_terminal/fred/fred_api.py
jmaslek: I had that just in case some of the fred data was non numeric (no idea if that exists).  Wanted to print the values but catch the plotting.  

Got rid of it and if it happens for some reason, I am sure someone will submit an issue

jmaslek:
Think I addressed those.  I also updated the plotting to use the auto-sizing from the other PR.  (Thanks for the patience going through all those)

I was also thinking that it doesn't quite make sense to have either GDP or Custom option, so I may edit that with some popular options and modularize the get_%% function since they are the same.  Not sure if you would rather me do that before merging first, but I can always open another.

jmaslek:
Actually I decided to change the menu to make it more useful if people want to look at more than just GDP.

This also addresses issue #87 somewhat.  There are options for Interest Rates as well as custom example for USD/EURO exchange.

DidierRLopes:
> Think I addressed those. I also updated the plotting to use the auto-sizing from the other PR. (Thanks for the patience going through all those)
> 
> I was also thinking that it doesn't quite make sense to have either GDP or Custom option, so I may edit that with some popular options and modularize the get_%% function since they are the same. Not sure if you would rather me do that before merging first, but I can always open another.

No need to thank. I thank you for everyone using the terminal all your efforts working into this! A lot of people will be grateful for that.

Let me give it a quick test again!

DidierRLopes:
This is AMAZING! Well done!

Just 2 minor things:
When i try to use: libor3m, I get:
```
Not Found.  The data for this series is not available via the API;                  the data can be found on the FRED website.
```
So probably not worth adding in the menu.

The second is if you try to run t01 t05 t30 t05 ... and more you'll see that the plots keep being created without being closed. I would add a 
```
plt.close('all')
```
on the main fred menu so that these get closed when you generate a new figure.

DidierRLopes:
Also, based on what you did here I would add these commands to the README, and the plots you're getting so far. So that people know what they would get if they were to run these. 

And also, in the menu, in front of gdp command, rather than having just "gdp" I would have "gross domestic product".

Good work!!

DidierRLopes:
Also, instead of t01 and t05, we could just have t1 and t5 for simplicity :)

jmaslek:
Removed the libor3m.  Thought I checked that one (was looking at #87 so I put that as an explicit example).  Added the Fed Rate and Cad/USD as examples.  (plus renames t1/t5)

Plots close now (possible future option of overlaying them?).

Added a couple images to readme.  Just 1 for the 'predefined' and 1 for the custom.

Not sure if the ReadME is clear?  Didn't want to create a section for each option, since the syntax is identical.

DidierRLopes (approved):
Great work! Thank you for such a big contribution!

DidierRLopes:
> Removed the libor3m. Thought I checked that one (was looking at #87 so I put that as an explicit example). Added the Fed Rate and Cad/USD as examples. (plus renames t1/t5)
> 
> Plots close now (possible future option of overlaying them?).
> 
> Added a couple images to readme. Just 1 for the 'predefined' and 1 for the custom.
> 
> Not sure if the ReadME is clear? Didn't want to create a section for each option, since the syntax is identical.

If needed I'll polish it a bit. No worries, this is amazing as is! Thanks! Feel free to work on other features :)

If you have some free time and were keen on this it would be good to:
* improve the code structure to follow #205. I'm doing so myself with the sentiment menu right now. 
* add unit tests to your commands.

Of course, these are only things that you can do to improve what you've already implemented. But leaving it like this is already great :) Thank you!

PR #205, opened 2021-03-17
Refactor Papermill menu and add an Econ Data report

aia:
check-that-volume


gamestonk_terminal/papermill/due_diligence_view.py
```diff
@@ -1,11 +1,22 @@
...
 import argparse
+from typing import List
 from datetime import datetime
 import os
 import webbrowser
 import papermill as pm
 
 
-def analysis(l_args):
+def due_diligence(other_args: List[str]):
```
DidierRLopes, line 12: Why the name "other_args"? Since these are the main args used in this function

gamestonk_terminal/papermill/econ_data_helper.py
```diff
@@ -0,0 +1,98 @@
...
+        Graph line type to pass to mplfinance, by default "candle"
+    draw_mas : tuple, optional
+        SMAs to draw, by default (20, 50)
+    draw_volume : bool, optional
+        Switch to draw trading volume by default True
+    high_trend : bool, optional
+        Switch to draw high trend line, by default True
+    low_trend : bool, optional
+        Switch to draw low trend line, by default True
+    """
```
DidierRLopes, line 44: Nice!!

gamestonk_terminal/papermill/papermill_controller.py
```diff
@@ -0,0 +1,93 @@
...
+            process_input = pc.switch(an_input)
+        except SystemExit:
+            print("The command selected doesn't exist\n")
+            continue
+
+        if process_input is False:
+            return False
+
+        if process_input is True:
+            return True
```
DidierRLopes, line 93: Thank you for this structure! This will help me when I start refactoring code :) 

I like this a lot!

DidierRLopes (approved):

gamestonk_terminal/papermill/papermill_controller.py
buzzCraft: Probably a dumb question, but why don't you use:

        if not process_input:
            return False
       else:
           return True

That way you don't need to do that second condition check. There is probably a good reason for why you chose not to do it, but since this is a learning experience for me I'm curious :)

DidierRLopes: Good point. And since we're at it, I beieve that variable isn't even needed, we could return straight from pc.switch()

gamestonk_terminal/papermill/due_diligence_view.py
aia: I think we go
input -> <menu item, e.g. dd> <more args - other_args>

It was l_args originally, but in a since we don't need l_ with type hints, and args sounds too generic plus we already chopped dd off the original input, I thought I'd call it other_args.

gamestonk_terminal/papermill/papermill_controller.py
aia: process_input can be True, False or None.

False - quit the program
True - quit the menu
None - continue the menu

gamestonk_terminal/papermill/due_diligence_view.py
DidierRLopes: Got you now. Yea, removing the "l_" only makes sense. Was just wondering about "other", sounds good to me. Will start following this convention

gamestonk_terminal/papermill/papermill_controller.py
buzzCraft: Ah, thanks.

PR #206, opened 2021-03-17
Added scaling to arima, ets, fbprophet, knn, nn, regression, sma and historical yahoo_finance_api

buzzCraft:
Added scaling to arima, ets, fbprophet, knn, nn, regression, sma and historical yahoo_finance_api.
Left corr at yahoo-finance_api untouched, as i think corr maps work best as squares.

### DIFF
--- a/gamestonk_terminal/comparison_analysis/yahoo_finance_api.py
+++ b/gamestonk_terminal/comparison_analysis/yahoo_finance_api.py
@@ -6,10 +6,10 @@ import matplotlib.pyplot as plt
 import yfinance as yf
 import seaborn as sns
 
-from gamestonk_terminal.helper_funcs import parse_known_args_and_warn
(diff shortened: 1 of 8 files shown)


buzzCraft:
Seems like I got wrong version of neural_networks.py.. Tried to fix it but probably just made a mess :P

DidierRLopes:
Yea, it seems there are still conflicts. The best way to go about this is to try to merge the main branch onto yours and make sure you resolved all conflicts.

If you do `git pull` on VSCode you'll see that it will fail to pull because it has conflicts. You need to fix these one by one (usually you can see the blue/green color on the right scrollbar), and commit that after. And that should be it.

Let me know if you need help with this :)

buzzCraft:
Ok, conflict is resolved.

DidierRLopes (approved):

DidierRLopes:
Amazing, thank you for this @buzzCraft !

PR #229, opened 2021-03-18
RH Portfolio

jmaslek:
I added the option to view current RH holdings and also to plot your historical portfolio.

This could easily be extended to other brokers besides RH, but that is what I have to test with.  The robin_stocks also offers a tda package, so the functions will likely be the same.

I did not add any buy/sell options as I figure that may be on the roadmap for dealing with the tickers.

### DIFF
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -1,20 +1,23 @@
-#PORTFOLIO
+# PORTFOLIO
 
-This was developed to interface with Robinhood.  
+This page shows the available brokers for loading in data.  If yours is not listed, please submit a new issue and we will look at adding it.
+
 
 Once this screen is accessed, the first command to be run is 
 ````
 login
 ````
-Which will use your robinhood login information, which should be stored as environment variables or in the config_terminal.py
+Which will use your  login information, which should be stored as environment variables in [config file](/gamestonk_terminal/config_terminal.py)
 
-Currently there are 2 options:
+[ROBINHOOD](#ROBINHOOD)
 * [hold](#hold)
     * Look at current (stock only) holdings
 * [rhhist](#rhhist)
-    * Get and plot rhhistorical portfolio
+    * Get and plot historical portfolio
 
-## hold <a name="hold"></a>
+## ROBINHOOD <a name="ROBINHOOD"></a>
+
+### hold <a name="hold"></a>
 
 Displays current holdings to the console:
 
@@ -25,8 +28,10 @@ There are no additional flags.  This just prints all current stonks, their last 
 the % change from previous close.
 
 Example:
+
 ![hold](https://user-images.githubusercontent.com/18151143/111685384-3c6ab080-87fe-11eb-80ce-9b256c396bf2.png)
-## rhhist <a name="rhhist"></a>
+
+### rhhist <a name="rhhist"></a>
 Display your RH portfolio based on provided interval and span.  Based on the API data availablility, plotted as a candlestick chart.
 ````
 usage: rhhist [-s --span] [-i --interval]
@@ -37,4 +42,5 @@ usage: rhhist [-s --span] [-i --interval]
     * [5minute, 10minute, hour, day, week]
     
 Example Default Output:
-![rhhist](https://user-images.githubusercontent.com/18151143/111685390-3d9bdd80-87fe-11eb-90f9-7ca8b0c1e7f8.png)
+
+![rhhist](https://user-images.githubusercontent.com/18151143/111718919-36da8e00-8831-11eb-99e1-957c8eccb583.png)
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -9,7 +9,7 @@ login
 ````
 Which will use your  login information, which should be stored as environment variables in [config file](/gamestonk_terminal/config_terminal.py)
 
-Robinhood has Two Factor Autentication, so you will likely be prompted to enter a code that is texted/emailed to you.
+Robinhood has Two Factor Authentication, so you will likely be prompted to enter a code that is texted/emailed to you.
 ###NOTE THAT LOGGING IN WILL SAVE A TOKEN TO `os.path.expanduser("~")/.tokens` WHICH CAN BE USED TO LOGIN EVEN IF USER CREDENTIALS ARE INCORRECT.
 
 [ROBINHOOD](#ROBINHOOD)
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -6,13 +6,13 @@ Once this screen is accessed, the first command to be run is
 ````
 login
 ````
-Which will use your robinhood login infomation, which should be stored as environment variables or in the config_terminal.py
+Which will use your robinhood login information, which should be stored as environment variables or in the config_terminal.py
 
 Currently there are 2 options:
 * [hold](#hold)
     * Look at current (stock only) holdings
-* [hist](#hist)
-    * Get and plot historical portfolio
+* [rhhist](#rhhist)
+    * Get and plot rhhistorical portfolio
 
 ## hold <a name="hold"></a>
 
@@ -26,10 +26,10 @@ the % change from previous close.
 
 Example:
 ![hold](https://user-images.githubusercontent.com/18151143/111685384-3c6ab080-87fe-11eb-80ce-9b256c396bf2.png)
-## hist <a name="hist"></a>
+## rhhist <a name="rhhist"></a>
 Display your RH portfolio based on provided interval and span.  Based on the API data availablility, plotted as a candlestick chart.
 ````
-usage: hist [-s --span] [-i --interval]
+usage: rhhist [-s --span] [-i --interval]
 ````
 * -s/--span : How long to look back.  Defaults to the day. Options:
     * [day, week, month, 3month, year, 5year, all]
@@ -37,4 +37,4 @@ usage: hist [-s --span] [-i --interval]
     * [5minute, 10minute, hour, day, week]
     
 Example Default Output:
-![hist](https://user-images.githubusercontent.com/18151143/111685390-3d9bdd80-87fe-11eb-90f9-7ca8b0c1e7f8.png)
+![rhhist](https://user-images.githubusercontent.com/18151143/111685390-3d9bdd80-87fe-11eb-90f9-7ca8b0c1e7f8.png)
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -1,0 +1,40 @@
+#PORTFOLIO
+
+This was developed to interface with Robinhood.  
+
+Once this screen is accessed, the first command to be run is 
+````
+login
+````
+Which will use your robinhood login infomation, which should be stored as environment variables or in the config_terminal.py
+
+Currently there are 2 options:
+* [hold](#hold)
+    * Look at current (stock only) holdings
+* [hist](#hist)
+    * Get and plot historical portfolio
+
+## hold <a name="hold"></a>
+
+Displays current holdings to the console:
+
+````
+usage: hold
+````
+There are no additional flags.  This just prints all current stonks, their last price, previous close, your equity and
+the % change from previous close.
+
+Example:
+![hold](https://user-images.githubusercontent.com/18151143/111685384-3c6ab080-87fe-11eb-80ce-9b256c396bf2.png)
+## hist <a name="hist"></a>
+Display your RH portfolio based on provided interval and span.  Based on the API data availablility, plotted as a candlestick chart.
+````
+usage: hist [-s --span] [-i --interval]
+````
+* -s/--span : How long to look back.  Defaults to the day. Options:
+    * [day, week, month, 3month, year, 5year, all]
+* -i/--interval: Data resolution. Defaults to 10minute intervals. Options:
+    * [5minute, 10minute, hour, day, week]
+    
+Example Default Output:
+![hist](https://user-images.githubusercontent.com/18151143/111685390-3d9bdd80-87fe-11eb-90f9-7ca8b0c1e7f8.png)
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -9,6 +9,9 @@ login
 ````
 Which will use your  login information, which should be stored as environment variables in [config file](/gamestonk_terminal/config_terminal.py)
 
+Robinhood has Two Factor Autentication, so you will likely be prompted to enter a code that is texted/emailed to you.
+###NOTE THAT LOGGING IN WILL SAVE A TOKEN TO `os.path.expanduser("~")/.tokens` WHICH CAN BE USED TO LOGIN EVEN IF USER CREDENTIALS ARE INCORRECT.
+
 [ROBINHOOD](#ROBINHOOD)
 * [hold](#hold)
     * Look at current (stock only) holdings
--- a/gamestonk_terminal/portfolio/port_menu.py
+++ b/gamestonk_terminal/portfolio/port_menu.py
@@ -6,6 +6,7 @@ from gamestonk_terminal.helper_funcs import get_flair
 from gamestonk_terminal.menu import session
 from gamestonk_terminal.portfolio import rh_api
 
+
 def print_port(show_login):
     """ Print help """
     print("\nPortfolio:")
--- a/gamestonk_terminal/portfolio/port_menu.py
+++ b/gamestonk_terminal/portfolio/port_menu.py
@@ -6,11 +6,8 @@ from gamestonk_terminal.helper_funcs import get_flair
 from gamestonk_terminal.menu import session
 from gamestonk_terminal.portfolio import rh_api
 
-
 def print_port(show_login):
-
     """ Print help """
-
     print("\nPortfolio:")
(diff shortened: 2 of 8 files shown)


DidierRLopes:
Amazing @jmaslek! I'm starting to review this now, could you send me an e-mail at dro.lopes@campus.fct.unl.pt?

gamestonk_terminal/portfolio/README.md
```diff
@@ -0,0 +1,40 @@
+#PORTFOLIO
```
DidierRLopes, line 1: I think this needs spacing between # and the letters in order to render nicely

gamestonk_terminal/portfolio/rh_api.py
```diff
@@ -0,0 +1,164 @@
...
+
+    holds = r.account.build_holdings()
+    stocks = []
+    equity = []
+    for stock, data in holds.items():
+        stocks.append(stock)
+        equity.append(round(float(data["equity"]), 2))
+
+    print("")
+    print("Stonk\t last price \t prev close \t equity \t % Change")
```
DidierRLopes, line 54: love that you called it stonk ahahah!

gamestonk_terminal/portfolio/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+This was developed to interface with Robinhood.  
+
```
DidierRLopes, line 4: Regarding the structure under I would divide it in sub-menus. Although currently we only support robinhood, it is likely that this will be extended in the future.

Having something like https://github.com/DidierRLopes/GamestonkTerminal/tree/main/gamestonk_terminal/behavioural_analysis
where we separate commands based on whether they belong from Twitter, Google, Stocktwits, Reddit, ... is probably nicer. #thisistheway.

gamestonk_terminal/portfolio/README.md
```diff
@@ -0,0 +1,40 @@
...
+Displays current holdings to the console:
+
+````
+usage: hold
+````
+There are no additional flags.  This just prints all current stonks, their last price, previous close, your equity and
+the % change from previous close.
+
+Example:
```
DidierRLopes, line 27: You need to add a spacing between images and the text, so that they don't appear in same line.

Also, I believe the other day I told you that you need to create a new issue for it, but I think you can literally drag and drop it to README and it works.

gamestonk_terminal/portfolio/README.md
```diff
@@ -0,0 +1,40 @@
...
+![hold](https://user-images.githubusercontent.com/18151143/111685384-3c6ab080-87fe-11eb-80ce-9b256c396bf2.png)
```
DidierRLopes, line 28: I usually change VSCode theme to light in order to have a white background. Not important though.

gamestonk_terminal/portfolio/README.md
```diff
@@ -0,0 +1,40 @@
...
+````
+usage: rhhist [-s --span] [-i --interval]
+````
+* -s/--span : How long to look back.  Defaults to the day. Options:
+    * [day, week, month, 3month, year, 5year, all]
+* -i/--interval: Data resolution. Defaults to 10minute intervals. Options:
+    * [5minute, 10minute, hour, day, week]
+    
+Example Default Output:
+![rhhist](https://user-images.githubusercontent.com/18151143/111685390-3d9bdd80-87fe-11eb-90f9-7ca8b0c1e7f8.png)
```
DidierRLopes, line 40: Instead of a screenshot of the plot, I would click on the cd icon in order to save the image. Usually I make it a as wide as my screen and don't change the height, I've noticed that they end up looking nice with a good ratio.

gamestonk_terminal/portfolio/port_menu.py
```diff
@@ -0,0 +1,90 @@
...
+from matplotlib import pyplot as plt
+from prompt_toolkit.completion import NestedCompleter
+from gamestonk_terminal import feature_flags as gtff
+from gamestonk_terminal.helper_funcs import get_flair
+from gamestonk_terminal.menu import session
+from gamestonk_terminal.portfolio import rh_api
+
+
+def print_port(show_login):
+
```
DidierRLopes, line 11: No need for spacing here

gamestonk_terminal/portfolio/README.md
```diff
@@ -0,0 +1,40 @@
+#PORTFOLIO
+
+This was developed to interface with Robinhood.  
+
+Once this screen is accessed, the first command to be run is 
+````
+login
+````
+Which will use your robinhood login information, which should be stored as environment variables or in the config_terminal.py
```
DidierRLopes, line 9: In this case you can add the config_terminal.py hyperlink, you can do so using:
```
[add here the name you want to show externally](hyperlink to whatever you want to show)
```

gamestonk_terminal/portfolio/port_menu.py
```diff
@@ -0,0 +1,90 @@
...
+                print(e)
+                print("")
+
+        elif ns_known_args.cmd == "rhhist":
+            rh_api.plot_historical(l_args)
+
+        elif ns_known_args.cmd == "login":
+            rh_api.login()
+            should_print_help = True
+            print_login = False
```
DidierRLopes, line 90: I think you missed a final "else" for "command not recognized"?

gamestonk_terminal/portfolio/port_menu.py
```diff
@@ -0,0 +1,90 @@
...
+    print("\nPortfolio:")
+    print("   help          show this menu again")
+    print("   q             quit this menu, and shows back to main menu")
+    print("   quit          quit to abandon program")
+    print(" ")
+    print("Robinhood:")
+    print("")
+    if show_login:
+        print("   login         login")
+    print("   hold          look at current holdings")
```
DidierRLopes, line 23: The way I see this in future would be, user goes into 'portfolio' and sees:
```
Portfolio
    help    
    q
    quit

    login

Current Broker(s): ?
    hold
    hist

Supported Brokers:
rh - Robinhood
tda -  TDAmeritrade
ig - IG
....
```

Then the user would do something along the lines:
```
(port)> login rh tda
```

which would show: `Current Broker(s): rh tda`

Then upon user doing 
```
(port)> hold
```
they would get the holdings of both of the stocks in both platforms, and the data would be merged.

Just a quick vision I had, let me know what you think

gamestonk_terminal/portfolio/rh_api.py
```diff
@@ -0,0 +1,164 @@
+import argparse
+import datetime
+from datetime import datetime as dt
+import pandas as pd
+from pandas.plotting import register_matplotlib_converters
+import matplotlib.pyplot as plt
+import mplfinance as mpf
+import numpy as np
+
+import robin_stocks.robinhood as r
```
DidierRLopes, line 10: I would do 
```
from robin_stocks import robinhood
```
and use `robinhood` instead of `r` for clarity purpose

gamestonk_terminal/portfolio/rh_api.py
```diff
@@ -0,0 +1,164 @@
...
+
+span_title_dict = {
+    "day": "Day",
+    "week": "Week",
+    "month": "Month",
+    "3month": "3 Months",
+    "year": "Year",
+    "5year": "5 Years",
+    "all": "All Time",
+}
```
DidierRLopes, line 33: I like what you did here, nice!

gamestonk_terminal/portfolio/rh_api.py
```diff
@@ -0,0 +1,164 @@
...
+            print(colored(to_print, "red"))
+
+    print("")
+
+
+def plot_historical(l_args):
+    parser = argparse.ArgumentParser(
+        add_help=False,
+        prog="Port",
+        description="""
```
DidierRLopes, line 75: The comment could be in the same line without breaking the black formatting

gamestonk_terminal/portfolio/rh_api.py
```diff
@@ -0,0 +1,164 @@
...
+        help="Interval to look at portfolio",
+    )
+
+    try:
+        ns_parser = parse_known_args_and_warn(parser, l_args)
+
+        if not ns_parser:
+            return
+
+        if ns_parser.interval not in valid_interval:
```
DidierRLopes, line 104: Instead of having it here, you can add this check to the `type`

E.g.
```
parser.add_argument(
        "-s",
        "--start",
        type=valid_date,
        dest="s_start_date",
        help="The starting date (format YYYY-MM-DD) of the stock",
    )
```
where:
```
def valid_date(s: str) -> datetime:
    try:
        return datetime.strptime(s, "%Y-%m-%d")
    except ValueError as value_error:
        raise argparse.ArgumentTypeError("Not a valid date: {s}") from value_error
``´

File: gamestonk_terminal/portfolio/port_menu.py [hunk 15]
```diff
@@ -0,0 +1,90 @@
...
+    print("   rhhist        look at historical portfolio")
+    print(" ")
+
+
+def port_menu():
```
DidierRLopes, line 28: Probably too late for this, but we are slowly changing the "architecture" of the code, to be more readable and modular.

See how https://github.com/DidierRLopes/GamestonkTerminal/blob/main/gamestonk_terminal/behavioural_analysis/ba_controller.py has been refactored.

gamestonk_terminal/portfolio/rh_api.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if last_price >= prev_close:
+            print(colored(to_print, "green"))
+        else:
+            print(colored(to_print, "red"))
+
+    print("")
+
+
+def plot_historical(l_args):
```
DidierRLopes, line 71: Same history about changing code to be more readable and modular. So, some things that would be advised in this area are:

- One function retrieves data, one function plots, and the other process the command
- Type hinting, e.g. `l_args` would be `other_args: List[str]`
- Docstring with numpy format

gamestonk_terminal/portfolio/README.md
jmaslek: Its a known fact that Dark Mode increases profits by 5% minimum :)

DidierRLopes: In order for my Portfolio to get green I better use Vantablack (https://www.businessinsider.com/the-surprising-origin-of-the-darkest-color-on-earth-2019-10?r=US&IR=T) :D

gamestonk_terminal/portfolio/README.md
jmaslek: I will look at that.  Im noticing that the ppl title is kinda squished up top as well.

I was also playing with trying plotly for the figure (since graph objects has candle stick built in) - but using those figures opened a browser instead of an image - so I didn't go into it further.

gamestonk_terminal/portfolio/README.md
jmaslek: When I drag in a photo into my pycharm, it doesn't do anything, I just found that the upload issue works very simply.

gamestonk_terminal/portfolio/README.md
jmaslek: I added GitHub  link to config_terminal.py.  Did you mean that or the local path?

gamestonk_terminal/portfolio/README.md
DidierRLopes: Yea, I usually don't do that bit locally. I do it after pushing my branch to remote, sometimes even before creating the PR.

E.g. once you push, and create a PR, at the top you'll see:
`jmaslek:robinhood`
if you create there it will take you to your branch on remote through web browser. Then you can access the readme document, edit it and drag+drop a figure there. This also allows you to preview what it would look like.

Then when you commit through web browser, this will be updated on the remote. In order for you to get that exact same commit locally, you'll want to do git pull.  If you don't do this last bit and commit something locally you'll get in this weird state where your local branch is 1 commit behind remote but also 1 commit ahead.

Hope this makes some sense

gamestonk_terminal/portfolio/README.md
DidierRLopes: Yea, ideally for now I'd strongly prefer us to focus on matplotlib figures. In the future we may follow the DASH road, where plotly would defo come in place.

gamestonk_terminal/portfolio/README.md
DidierRLopes: Yea, I mean doing this:

[config_terminal.py](/gamestonk_terminal/config_terminal.py)

gamestonk_terminal/portfolio/rh_api.py
jmaslek: So the example you have there would ensure that -s (stuff) results in stuff having the "%Y-%m-%d" format.

In my case, (I think you are mentioning line 104) I want to check that -i is in a list.  I tried this:

```
def input_in_list(input_val, check_list):
    if input_val in check_list:
        return input_val
    else:
        raise argparse.ArgumentTypeError("Input not in list of valid options")

```

But using type = input_in_list raises nm error (which doesn't surprise me as I am not passing in a list)

gamestonk_terminal/portfolio/README.md
jmaslek: Also I feel like a fraud for not having any GME.  paper hands here.

gamestonk_terminal/portfolio/rh_api.py
DidierRLopes: The issue is that when you use `type=input_in_list` you only use as an argument the input provided by the user, and in that function you have an additional argument called check_list. 

The way to go forward is what I did for the sources of information when loading a stock, i.e. hardcoding the list inside that function. There may be ways to have this type= to accept another argument, but since I knew how to do it with the hardcoded list inside, I went that way.
```
def check_sources(source: str) -> str:
    available_historical_price_sources = ["yf", "av"]
    if source in available_historical_price_sources:
        return source
    raise argparse.ArgumentTypeError(
        "This source for historical data is not available."
    )
```

jmaslek: So just curiously, what is the advantage of having that type (if the list is hard coded) vs just checking with the `if _ in _` ?

gamestonk_terminal/portfolio/port_menu.py
jmaslek: Yeah I think that makes sense.

As an aside.  I liked the `hist` function, but it didn't get through the code spell.  Any chance that could be edited on the check side?    Or I can go with `history`

jmaslek:
Tried to address most changes. (README still could probably be better)

I have a second account through alpaca that I can start adding merged portfolio features like you suggested (hence the poetry lock update), as well as menu structure.

As for the refactoring, I can play around with it.   Will need to dive into the structure, but a quick look didn't scare me too bad, so it should be doable on short term.

gamestonk_terminal/portfolio/rh_api.py
DidierRLopes: Modularity I would say.

That way all the checks are done at the time of parsing, rather than accepting the arguments, and raising the exception at a later stage.

This will make it easier to decouple functionalities at a later stage, which is something we are keen on looking forward

DidierRLopes:
> Tried to address most changes. (README still could probably be better)
> 
> I have a second account through alpaca that I can start adding merged portfolio features like you suggested (hence the poetry lock update), as well as menu structure.
> 
> As for the refactoring, I can play around with it. Will need to dive into the structure, but a quick look didn't scare me too bad, so it should be doable on short term.

Should we merge this as is for now? Or would you rather do that menu change now? I can walk you through it if you want :)

jmaslek:
I will leave that up to you for now.  I would maybe say merge since RH is the only one I have working.  

I will start working on adding more brokers then change the menu once I have functions for dealing with multiple.

DidierRLopes:
Few minor notes before merging this:

1. It exits program when wrong login details provided.

Captura de ecrã 2021-03-19, às 13 59 53

2. 'logoff' appears as a choice but command isn't recognized, neither appears in help menu

Captura de ecrã 2021-03-19, às 14 02 02

3. New line after 'command not recognized'

Captura de ecrã 2021-03-19, às 14 02 13


The usual pedantic stuff....

jmaslek:
1)  So I was just changing my env vars to get in and noticed that robin_stocks actually saves an authentication token that it uses if the login fails.  So should I add that fact to the readme (i.e don't use this feature in a public computer?)  The token avoids the TFA on RH's end?

For 2) I meant to get rid of that logoff - I automatically logoff with -q or -quit.

DidierRLopes:
I like it, thanks for fixing it.

Also, why the time.sleep(2)? Any particular reason?

DidierRLopes (approved):

jmaslek:
Since it prints the error message ("Login failed" - or whatever) then immediately goes back to the menu.  I figured a 2 s wait would let the user see the message before the menu came back.

DidierRLopes:
I think in that case there's no need to show menu again.

My thinking regarding this is the following. Once you start using the terminal, when something fails you don't want to look at the menu everytime. And if you want, you can always write "help".

At least to be coherent with remaining code structure, would be good to keep it this way I think. Unless you disagree :)

jmaslek:
Ah. I see how you did it in (example sen-wsb when not logged in).  Didn;t think to cross reference first.  Ill go about that way.

I was going to merge back to the main branch then update the readme.  I may also fix stuff like that (I also didn't refactor the argparse type).

DidierRLopes:
No worries, the code is functional. That's good for now, we'll need a general code structure refactoring anyway, so we can take it gradually :)

PR #230, opened 2021-03-19
Fixing Feature Flag setting from environment

aia:


aia:
Decided on another approach

PR #231, opened 2021-03-19
Fix docker visualization

kaitj:
Per #227, this PR partially resolves the problem of `matplotlib` figures not working with docker containers. Resolved by installing python3-tk and it's dependencies to the container. A display still needs to be connected and x11 socket mounted (for ubuntu). Not sure of the exact commands for Windows or WSL so would appreciate if someone can test it out on that.


aia (approved):
Very nice!

PR #232, opened 2021-03-19
Adds snapshot command to save the current graph to an image file

clothespin:
This patch implements the `snapshot` command to save the current figure to an image file. Supports .svg, .png, and .pdf formats. This may seem redundant in an interactive setup with a display where the plot window itself provides save functionality, but for users who want to run in a headless environment that doesn't have access to a GUI display this provides a way to get plots out. Particularly helpful in the context of a container environment where a host X display may not be available or wanted, although the full headless batch dream will also require some logic around `plt.show()` in headless environments.

### DIFF
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -490,7 +490,7 @@ def snapshot(l_args, s_ticker):
         print(f"Supported image formats: {supported_image_formats}")
         return
     if not s_ticker:
-        print(f"You must LOAD a ticker symbol and plot a graph first")
+        print("You must LOAD a ticker symbol and plot a graph first")
         return
     plt.savefig(
         f"{ns_parser.fname}.{ns_parser.image_format}", format=ns_parser.image_format
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -492,5 +492,7 @@ def snapshot(l_args, s_ticker):
     if not s_ticker:
         print(f"You must LOAD a ticker symbol and plot a graph first")
         return
-    plt.savefig(f"{ns_parser.fname}.{ns_parser.image_format}", format=ns_parser.image_format)
+    plt.savefig(
+        f"{ns_parser.fname}.{ns_parser.image_format}", format=ns_parser.image_format
+    )
     print(f"Saved {ns_parser.fname}.{ns_parser.image_format}")
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -1,5 +1,6 @@
 import argparse
 from sys import stdout
+from datetime import datetime
 import matplotlib.pyplot as plt
 import pandas as pd
 from alpha_vantage.timeseries import TimeSeries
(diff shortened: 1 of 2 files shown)


aia (approved):

DidierRLopes:
May I suggest something? 

Instead of having a snapshot command, what about doing the following:
1. Adding a line in config_terminal.py to select the save format default (png probably would be nicer)
2. To all commands that output a plot add the arguments:
   * -s/--save: Save plots using format specified + using ticker_cmd_datetime
   *  --saveas <name>:  Save plots using format specified + using the name provided

This way we achieve a more universal saving feature. What do you think? Because I was trying out the branch and thinking that the only use case for this would be when using view/candles. 

And if we just had this global functionality would be much powerful, and we could still do `candle -s`or `view -s`. Let me know what you think about this.

DidierRLopes:
Since we are at it, I also thought the same for the export command. Rather than only exporting the historical data. We could have a default saving format and upon user doing:
* --export they would export all data output from the terminal to their preferred file choices.

That way we minimize noise on the main screen info, while allowing for the same capabilities across the entire terminal.

clothespin:
That's not a bad idea as long as you preserve the ability to write to stdout as well when needed, but the -f switch I'm already using accomplishes that. We'd have to be thoughtful about how not to introduce code duplication or complexity because we'd be taking functionality in one place and creating specialized versions of it in each place in the code that produces a plot (for snapshot) or dataframe (for export).

Part of the difficulty is we don't yet have a clear conceptual model of how commands are supposed to interact with the state of the terminal. For instance, some commands in the feature submenus create new figures and some modify existing ones, and there's no clear way to specify "I want this info shown to me in a new graph" vs "I want this info shown to me as a new item in an existing graph" and different commands will do different things. Ideally I'd like to be able to do something like, say, have two graphs open, overlay a third measurement on a specific one of those two windows and snapshot, all with some combination of terminal flags and commands. All we have now is a fairly vague notion of the "current dataframe" and "current figure", with all tracking and coordination of different plots being handled by matplotlib.

That said, that's a whole big complicated discussion, and we may not be at the point as a project yet where we're actually stable enough that it makes sense to figure out what the "model" is. I'll defer to your judgment on whether this is useful as-is or it'd be better to look at other options.

DidierRLopes:
Yea I see what you mean.

Although it may look like it, we always open a new figure for every new plot. It's just that when you select another command, in prediction menu let's say, we close the previous and open a new one. Giving you the false impression of using the same matplotlib figure.

I think the idea would be as simple as adding:
```
DEFAULT_FIGURE_FORMAT_TO_SAVE = 'png'
DEFAULT_STDOUT_FORMAT_TO_SAVE = 'csv'
```

and then, where relevant:
```
parser.add_argument(
        "--fig",
        action="store_true",
        dest="fig",
        help="Flag for saving figure",
    )
parser.add_argument(
        "--out",
        action="store_true",
        dest="txt",
        help="Flag for saving output",
    )
```
then at the bottom of the code we would just have to:

```
if ns_parser.fig:
       plt.saveas('blablabla.' + DEFAULT_FIGURE_FORMAT_TO_SAVE)
       
if ns_parser.out:
     if DEFAULT_FIGURE_FORMAT_TO_SAVE == 'csv':
         df_whatever.to_csv()
     elif DEFAULT_FIGURE_FORMAT_TO_SAVE == 'json':
         df_whatever.to_csv()
```
obviously this chunk of code would be in helper_funcs, and could be just "coupled" to relevant commands.

clothespin:
I think pretty much any of the discussed alternatives would be better than this proposal, so I'll go ahead and close this PR. I've been watching some [YouTube](https://www.youtube.com/watch?v=2ee-x6IXWK8) [tutorials](https://www.youtube.com/watch?v=Ia158DGKnhU) to get some inspiration regarding what a "mature" GamestonkTerminal might look like. Since I don't have $25k lying around to actually experiment with one myself, it seemed like the next best option. If we were able to decouple how we collected data from how it was processed and displayed it seems like the existing project could provide a very powerful backend to a GUI that looked more like the actual Bloomberg terminal.

PR #243, opened 2021-03-20
Multiple Updates

pchaganti:
Multiple Updates

- Added FRED API env variable to README
- Added fredapi to requirements-win
- Added materials sector/industry to econ data summary for papermill. This tracks the industry that deals with basic materials required in an economy such as plastic, paper, concrete, metals, etc.

### DIFF
--- a/requirements-win.txt
+++ b/requirements-win.txt
@@ -28,6 +28,7 @@ filelock==3.0.12; python_full_version >= "3.6.0" and python_version >= "3.6"
 finviz==1.3.4
 flair==0.7; python_version >= "3.6"
 flatbuffers==1.12
+fredapi==0.4.3
 ftfy==5.9; python_version >= "3.6"
 fundamentalanalysis==0.2.6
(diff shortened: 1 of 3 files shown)


gvikei (approved):

requirements-win.txt
```diff
@@ -28,6 +28,7 @@ filelock==3.0.12; python_full_version >= "3.6.0" and python_version >= "3.6"
 finviz==1.3.4
 flair==0.7; python_version >= "3.6"
 flatbuffers==1.12
+fredapi==0.4.3
```
gvikei, line 31: Thanks for making this change. Do we not need this in `requirements-full.txt`, `requirements-old.txt` as well?

pchaganti: It is already in requirements-full.txt. I made an earlier PR that was merged with that change. Not sure about the requirements-old.txt. Is it used?

aia (approved):
Great work! Thank you.

PR #244, opened 2021-03-21
refactor portfolio menu

jmaslek:
This PR replaces the port_menu with the port_controller as per the previous pr conversation.  also updated requirements.txt with robin-stocks and alpaca trade api (next project)

@DidierRLopes on my end, the linting failed due to the use of eval - so I am not sure what to replace with (also a 'pointless string' which is in the menu?)

edit -- apparently the listing passed!


gamestonk_terminal/portfolio/port_controller.py
```diff
@@ -0,0 +1,161 @@
...
+    ]
+
+    def __init__(self):
+        self.port_parser = argparse.ArgumentParser(add_help=False, prog="port")
+        self.port_parser.add_argument("cmd", choices=self.CHOICES)
+        self.brokers_list = set()
+
+    @staticmethod
+    def print_help(broker_list):
+        if not broker_list:
```
DidierRLopes, line 34: You don't need these 7 lines, look at my next comment

gamestonk_terminal/portfolio/port_controller.py
```diff
@@ -0,0 +1,161 @@
...
+
+        """ Print help """
+        print("\nPortfolio:")
+        print("   help          show this menu again")
+        print(
+            "   q             quit this menu, and shows back to main menu, logs out of brokers"
+        )
+        print("   quit          quit to abandon program, logs out of brokers")
+        print("   login         login to your brokers")
+        print("\nCurrent Brokers :" + print_broke)
```
DidierRLopes, line 49: You can do something like this in one line:
```
print(f"\nCurrent Brokers : {('', ', '.join(broker_list))[bool(broker_list)]}")
```

So, `', '.join(['a', 'b', 'c']) ` would resulting in the string `a, b, c`. 

Then you use the ternary operation
```
('A', 'B')[False] = 'A'
('A', 'B')[True] = 'B'
```

gamestonk_terminal/portfolio/port_controller.py
```diff
@@ -0,0 +1,161 @@
...
+            rh_api.show_holdings()
+        except Exception as e:
+            print(e)
+            print("")
+
+
+def menu():
+    """Portfolio Analysis Menu"""
+
+    port_controller = PortfolioController()
```
DidierRLopes, line 137: To follow the other menus:
```
fundamental analysis
technical analysis
comparison analysis
behavioural analysis
..
```

should this be `pa` for `portfolio analysis` ?

DidierRLopes (commented):
Just commented on a code improvement.

All the rest looks good to go! Good work!!

gamestonk_terminal/portfolio/port_controller.py
jmaslek: I assume you mean starting at `if not broker_list` ?  It should be a static method, right?

gamestonk_terminal/portfolio/port_controller.py
jmaslek: Love the python golf!

gamestonk_terminal/portfolio/port_controller.py
DidierRLopes: Yea, I mean that the line of code that I mentioned below can do this all in 1 go :p

gamestonk_terminal/portfolio/port_controller.py
DidierRLopes: You live and you learn :D

jmaslek:
made those changes.  Still added the if not brokers_list to print None when not logged in

DidierRLopes:
> made those changes. Still added the if not brokers_list to print None when not logged in

Nicee!! I'll merge this once checks are done

You could also have added the **None** in that line, here:
```
print(f"\nCurrent Brokers : {('None', ', '.join(broker_list))[bool(broker_list)]}")
```

PR #246, opened 2021-03-21
Add a unit test for Thought of the Day

aia:
Will improve make more advanced but basic example is this.


DidierRLopes:
Nice, thank you!

@jmaslek @lolrenx here's a unit test example, at some point we need to add these

lolrenx:
Good stuff ! Definitely a must have, agreed !

PR #247, opened 2021-03-21
Add US Dollar strength against the Euro and Yen to the econ data

pchaganti:
The strength of the US Dollar is another good indicator of the general economy. This PR adds US dollar vs Euro and Yen to the econ data in papermill.


DidierRLopes:
Thank you! :)

Issue #240, opened 2021-03-20
[FR] Enable averaging around multiple predictions


Felixkruemel:
**What's the problem of not having this feature?**
Predictive Models like LSTM will differ a lot between several runs. A command to let it run e.g. 10times and average the results would probably provide better results overall.
**Describe the solution you would like**
Just implement a way of specifying to run multiple runs of a single prediction model.

DidierRLopes:
This is a great idea! This way we can even provide the standard deviation around the average prediction, something along these lines:

<img width="548" alt="Captura de ecrã 2021-03-20, às 14 55 04" src="https://user-images.githubusercontent.com/25267873/111874263-47ebe200-898c-11eb-83fb-a2ddf56a339a.png">

Felixkruemel:
> 
> 
> This is a great idea! This way we can even provide the standard deviation around the average prediction, something along these lines:
Exactly! Even for short term predictment that would work way better I guess. And since LSTM is pretty fast that should also be doable in a good timely manner.

And additionally it should not require too much new code as it should not be more than a simple loop and an average function.

DidierRLopes:
You're right, should have thought about it. 

I'm a bit busy this weekend, but I'll try to get to it during the week and let you know in here when done. :)

Felixkruemel:
@DidierRLopes Whiele I'm thinking about it, could we somehow combine the sentiment analysis with an prediction algorithm? Maybe there is even one available somewhere.

Because better data quality is way more important than just better model

DidierRLopes:
That's a good idea, although more of a long term goal, I'd say.

Can you try to dig something that is worth giving a go, and I can have a look at it?

Felixkruemel:
@DidierRLopes 
Have a look here:
https://www.thetechplatform.com/post/sentiment-analysis-for-stock-price-prediction-in-python
The chart on the bottom pretty much sums up that a positive sentiment will go to a positive stock continuation.

Can we feed the data of both into lstm? So like not only feed it with historical chart data but additionally with historical sentiment data. You can have a look here: https://github.com/EmielStoelinga/CCMLWI
I searched whole Github but couldn't find anything else, may be something which hasn't been done in OpenSource before. So let's do it ;)

I probably should open a new issue for that or?

DidierRLopes:
Nice!! 

[...]

PR #252, opened 2021-03-22
Enable median/quantiles around loop of Neural Networks predictions

DidierRLopes:
This aims to implement #240 .

As default I chose to plot the median prediction values. And filled the area between quantile 10 and 90%

Examples shown below:

loops

Captura de ecrã 2021-03-22, às 02 14 38

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -96,7 +96,7 @@
 
 - [x] Add several exponential smoothing techniques (@didier) - [PR #132](https://github.com/DidierRLopes/GamestonkTerminal/pull/132)
 - [x] Allow backtesting (@didier) - [PR #169](https://github.com/DidierRLopes/GamestonkTerminal/pull/169)
-- [ ] Add averaging around multiple predictions - [IS #240](https://github.com/DidierRLopes/GamestonkTerminal/issues/240)
+- [x] Add averaging around multiple predictions (@didier) - [PR #252](https://github.com/DidierRLopes/GamestonkTerminal/pull/252)
 
 ### Long term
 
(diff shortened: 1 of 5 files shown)


Felixkruemel:
@DidierRLopes 
Does that work with backtesting? That would be great!

jmaslek (commented):
Looks good!   noticing that the --loops aren't in the README usage.

Maybe not related to this PR, but if I enter a wrong date (under -e flag : assumed -e was epochs without looking at help )) then it quits the terminal, not just backs out to menu.
Screen Shot 2021-03-22 at 10 03 21 AM

DidierRLopes:
> @DidierRLopes
> Does that work with backtesting? That would be great!

It does yes! :)

DidierRLopes:
> Looks good! noticing that the --loops aren't in the README usage.
> 
> Maybe not related to this PR, but if I enter a wrong date (under -e flag : assumed -e was epochs without looking at help )) then it quits the terminal, not just backs out to menu.
> Screen Shot 2021-03-22 at 10 03 21 AM

Good finding!! I'll fix both of these now, cheers!

DidierRLopes:
> > @DidierRLopes
> > Does that work with backtesting? That would be great!
> 
> It does yes! :)

Example:

ex1

ex2

jmaslek (approved):
Hell yeahhhh.  Looks great

DidierRLopes:
> Hell yeahhhh. Looks great

I'll work today on adding a README on how to tune Neural Networks.

Otherwise this will happen https://www.reddit.com/r/GME/comments/makyk1/predicting_gme_price_for_the_week_using_ai_36279/
and we don't want to be responsible by bad YOLOs :D

jmaslek:
Haha sounds good.  

Just curiously, is there a reason for not doing a train/test split or validation data on the NNs?

DidierRLopes:
> Haha sounds good.
> 
> Just curiously, is there a reason for not doing a train/test split or validation data on the NNs?

Now it kind of does it. :)

The problem with just using train/test split is that this is time-series, and we rely on order. I'll write down about this on a new doc, and then you can comment against it if you agree or not

PR #259, opened 2021-03-22
Add Alpaca Broker

jmaslek:
This PR adds alpaca broker with the same hold and historical portfolio functions.  

Note I haven't invested in this account enough to show the alphist.  I just bought some penny stocks to test out, so I can update the readme in like a week with all the loss porn.

### DIFF
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -78,7 +78,7 @@ Alpaca has a nicely maintained python API that is used here.  To login, the best
 * "APCA_API_KEY_ID"
 * "APCA_API_SECRET_KEY"
 
-If these are defined, then the login command does not explictly need to be run, but the login command verifies all those keys are defined. 
+If these are defined, then the login command does not explicitly need to be run, but the login command verifies all those keys are defined. 
 
 Note that alpaca does support paper trading, so if your base_url is "https://paper-api.alpaca.markets", then this will be importing
 your paper trading account.  
--- a/gamestonk_terminal/portfolio/README.md
+++ b/gamestonk_terminal/portfolio/README.md
@@ -4,6 +4,7 @@ This page shows the available brokers for loading in data.  If yours is not list
 
 Current brokers:
 * Robinhood
+* Alpaca
 
 
 Once this screen is accessed, the first command to be run is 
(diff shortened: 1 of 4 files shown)


DidierRLopes:
So, this may be a bit futuristic. But the way I see this is the following:
* Having only 2 commands: hold, hist

PS: You can change hist to historical, as you suggested the other day. It made sense to me.

When you do login to rh, you get:
Current brokers: rh

This means that if you do hold or hist, it's pointing at "robinhood".

BUT, imagine that you have:
Current brokers: rh alp

Then, if you did hold or hist, it would be pointing at both and you would show the combined holdings or historical.

jmaslek:
So I was thinking something along that line as well.  I was going to play with merging them later this week (these are the only two that I actually have).  I think looking individually is good as well (I can bury the menu unless specified).

For the future, my thought process was going to be something like 

- `login brokers`
- `merge`
- `hold` or `hist`

So the command to merge all would be stated specifically and then allow for the options of hold/hist (still allowing individual hold/hist).

For the future as well, I am noticing that the api's return different types of data (alpaca has a nice historical portfolio get, but robin-stocks just returns high low of a period, not equity at a given time), so that is why I submitted a PR that just had this broker.

gamestonk_terminal/portfolio/README.md
```diff
@@ -60,4 +70,39 @@ usage: rhhist [-s --span] [-i --interval]
     
 Example Default Output:
 
-![rhhist](https://user-images.githubusercontent.com/18151143/111718919-36da8e00-8831-11eb-99e1-957c8eccb583.png)
\ No newline at end of file
+![rhhist](https://user-images.githubusercontent.com/18151143/111718919-36da8e00-8831-11eb-99e1-957c8eccb583.png)
+## ALPACA <a name="ALPACA"></a>
+
+Alpaca has a nicely maintained python API that is used here.  To login, the best approach is to just save the following environment variables:
+* "APCA_API_BASE_URL"
```
DidierRLopes, line 77: To follow others API Keys, probably use something like: GT_ALP_BASE/KEY/SECRET

DidierRLopes (approved):
Agreed with what you mentioned, we're on the right path here!

Just a minor comment. All the rest looks good to me! Feel free to merge it

gamestonk_terminal/portfolio/README.md
jmaslek: So I was going to keep it like that, but the alpaca api actually goes in and looks for those specific env variables.  Anyone using Alpaca probably has those stored as well.  

I can change it to keep the same convention if you would like.

Also - yes reqs are updated (from yesterday)

DidierRLopes: On that basis, I agree with you to leave it as is! Feel free to merge it

Issue #261, opened 2021-03-22
[Bug] - Twitter sentiment spits out 'text' when no API keys exist


Felixkruemel:
**Describe the bug**
If you have no API keys inserted the result will be 'text'

**To Reproduce**
Just do a clean install without changing anything, load a stock, go into the ba menu and do sentiment analysis

**Screenshots**
 
![image](https://user-images.githubusercontent.com/12109804/112042567-f2f4cb00-8b47-11eb-8810-1b536c76bf90.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Python version: 3.6.8

**Additional context**
Additionally, when setting a day limit of e.g. 3 days it will do 4 days as the array starts at 0.

DidierRLopes:
Thanks for this! I'll have a look at it.

PS: Regarding the additional days when looking in the past, this was a design choice. And the reason why is that if you select 3 days in the past, you gather data from 3 days in the past + the current day so far :)

jmaslek:
Closing on #262 262

PR #262, opened 2021-03-22
Twitter commands sentiment/infer to fail smoothly when wrong API Key

DidierRLopes:
This addresses #261


jmaslek (approved):
Screen Shot 2021-03-22 at 4 28 37 PM

jmaslek:
What causes the tests to Queue for so long?

DidierRLopes:
> What causes the tests to Queue for so long?

No clue. But it's recent. It used to be much faster

PR #263, opened 2021-03-22
Change news source from Finviz to News API

DidierRLopes:
This change was required due to #255 because:
1. Finviz only gives news titles + link, not sorted
2. Finviz is limited in the sources by focusing on Yahoo Finance mainly

News API seems to be the right way to go with this


jmaslek (approved):
Looks nice.  No issues on my end when api key not provided and when it is the output looks real good :)

hinxx:
Tested. Looks very neat! :+1:

PR #266, opened 2021-03-23
Add validate NN model guidelines

DidierRLopes:
This is not yet complete, it's just a quick draft of my main thoughts on the topic.

Since I've learned this on my spare time, would be great to have the contribution of someone that is skilled in this area. And has learned through experience.


DidierRLopes:
@arcutright can you have a look at this?

jmaslek:
This is a nice walkthrough!  I personally think some splitting/validation should be done (at least in the NNs where the time series is kind of irrelevant since the inputs are just arrays of N inputs -- the ARIMA and sma/regressions are different where the splitting should be done either adaptively or in order )

I think that the ideas of having the mlp/rnn/lstm are fun but they have pretty much been shown extensively that they do not work (and if they did then it would probably already be subject to alpha decay).  I think the fun will be fun to play with different sentiments and news sources.  (Also as a side note: Quantconnect has a backtested LSTM model [here](https://www.quantconnect.com/forum/discussion/6931/from-research-to-production-long-short-term-memory) )

DidierRLopes:
> This is a nice walkthrough! I personally think some splitting/validation should be done (at least in the NNs where the time series is kind of irrelevant since the inputs are just arrays of N inputs -- the ARIMA and sma/regressions are different where the splitting should be done either adaptively or in order )
> 
> I think that the ideas of having the mlp/rnn/lstm are fun but they have pretty much been shown extensively that they do not work (and if they did then it would probably already be subject to alpha decay). I think the fun will be fun to play with different sentiments and news sources. (Also as a side note: Quantconnect has a backtested LSTM model [here](https://www.quantconnect.com/forum/discussion/6931/from-research-to-production-long-short-term-memory) )

Can you give an example? Do you want to apply Forward Chaining, K-Fold or Group K-Fold?

What do you mean by splitting adaptively or in order?

I think the LSTM has a lot of potential personally. Only using the historical price, with the market being volatile as it is, makes LSTM look bad. However, if at some point we manage to use as input not only historical price, but the sentiment of the market, earnings, ... I think it can work quite well if validated properly.

We can use the Quantconnected backtested LSTM model as default, what you think?

jmaslek:
So for keras models (the NN's), you can just set the `validation_split=##` kwarg. This also allows to add the early stopping callback (which can actually be set on "loss" as well - possible update).  There can also be a hold out (maybe only train up until 45 days ago so that you can input the first 40 days to spit out 5 and compare the last 5 days with those -- which seems to be the backtesting?  see below)

I apologize for not diving deeper, but for the backtesting (which seems to be the equivalent of testing), but do you guess data that has been fed to the model (in that case, you are definitely overfitting)?

KNN could just be a simple [train_test_split](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html).  As for the adaptive/in order, I mostly mean that train data = data[:N] and test data = data[N:].  And the adaptive, I thought I have seen something where you train on (for example) year 1 and predict year 2, then train on year 1 + 2 then predict year 3 and so on...  Probably the exact same as using data[:N] and data[N:]

I don't think there is any need to change the default model.  The quant connect one uses all 5 OHLVC datapoints, (which I personally feels totally overfits the data).  Also if you try their backtester with longer times - it totally times out (at least for the free tier).

Another thing to consider for the NN (and my personal preference) - is instead of predicting price, is to predict up/down.  Basically you could just feed the last output into some kind of softmax layer), and then add another train output of a boolean being 1 if the price increased in the time frame or 0 if it decreased.

DidierRLopes:
I didn't know about validation_split, but after reading a bit about it I don't find this as an ideal method of cross-validation for timeseries, see https://towardsdatascience.com/addressing-the-difference-between-keras-validation-split-and-sklearn-s-train-test-split-a3fb803b733. I don't like the fact that it holds-out on the more recent data, which in my point of view should be used to tweak hyperparameters as the more recent data, the more important.

If we want to do cross-validation I think is best to do either Forward Chainging, K-Fold, or Group K-Fold. See https://github.com/DidierRLopes/TimeSeriesCrossValidation/blob/master/README.md or https://hub.packtpub.com/cross-validation-strategies-for-time-series-forecasting-tutorial/. Again, I'm no expert, I jus think it makes more sense but I may be wrong.

We can definitely take this offline as this is an interesting discussion for myself to learn more.

I know what you mean by the adaptively, that's the Forward Chaining example. It still keeps the order though. 

Hummm, yea, so probably leave it as it is for now. Would be good to add a dropout layer for LSTM though. I only added one to RNN by default I believe.

Personally I prefer to have a linear output layer because that gives you the value that the NN is predicting. And from that you can extrapolate whether price is going up or down (the coloring also helps here), while vice-versa is not possible. But if the user only wants a 1 for price going up, and -1 for price going down, they can define the last layer to have activation function tanh rather than linear, I believe.

jmaslek:
> I didn't know about validation_split, but after reading a bit about it I don't find this as an ideal method of cross-validation for timeseries, see https://towardsdatascience.com/addressing-the-difference-between-keras-validation-split-and-sklearn-s-train-test-split-a3fb803b733. I don't like the fact that it holds-out on the more recent data, which in my point of view should be used to tweak hyperparameters as the more recent data, the more important.

I can't say I have ever dug into the validation_split (that article is interesting), so I guess using train_test_split and then specifying `validation_data` would also be.

> If we want to do cross-validation I think is best to do either Forward Chainging, K-Fold, or Group K-Fold. See https://github.com/DidierRLopes/TimeSeriesCrossValidation/blob/master/README.md or https://hub.packtpub.com/cross-validation-strategies-for-time-series-forecasting-tutorial/. Again, I'm no expert, I jus think it makes more sense but I may be wrong.

The forward chaining you mentioned is what I was imagining for the other models.  Did not know what it was called :)

> Personally I prefer to have a linear output layer because that gives you the value that the NN is predicting. And from that you can extrapolate whether price is going up or down (the coloring also helps here), while vice-versa is not possible. But if the user only wants a 1 for price going up, and -1 for price going down, they can define the last layer to have activation function tanh rather than linear, I believe.

I just like it since you can try to associate a probability of going up to the input series, so you could (as an example), optimize a group of stocks using $\alpha$ = lstm output of probability to increase.

arcutright:
This walkthrough is good start for usage! We may want to add some links to keras documentation and/or papers, blog posts, etc to point to more background knowledge. As well as a warning that modeling like this is just for fun and probably less accurate than the user's guess.

> We can use the Quantconnected backtested LSTM model as default, what you think?

This seems fine as a default, but I don't have strong feelings either way. I definitely like the idea of adding validation testing as a default. Otherwise it presents itself as an overfitted magic black box.


> I think the LSTM has a lot of potential personally. Only using the historical price, with the market being volatile as it is, makes LSTM look bad. However, if at some point we manage to use as input not only historical price, but the sentiment of the market, earnings, ... I think it can work quite well if validated properly.

..

> I think that the ideas of having the mlp/rnn/lstm are fun but they have pretty much been shown extensively that they do not work (and if they did then it would probably already be subject to alpha decay). I think the fun will be fun to play with different sentiments and news sources. (Also as a side note: Quantconnect has a backtested LSTM model [here](https://www.quantconnect.com/forum/discussion/6931/from-research-to-production-long-short-term-memory) )

Of course all of this is mostly for fun. Generally to get accurate results with LSTM modeling you have to get into hyperparameter tuning, model augmentations, or manually massaging your datasets to account for external factors and it rarely generalizes. I suspect there may be a way to combine enough data sources to come up with an edge automatically (algo trading exists), but that's obviously nontrivial. Something that may be interesting to add if you want to add a single-output model (probability of price increase/decrease) would be to look into hybrid models. [Here is a paper](https://ieeexplore.ieee.org/document/8141873) I was reading recently of an augmented LSTM model for improved time-series forecasting.

In some local experiments I have had minor improvements from using intra-day and extended hours pricing, volume, options volume and premium prices, etc. I will probably push this up once I find a stable way to do this. Things gets complicated if you go beyond daily data, code is proof of concept at best. The worst part is that most data providers for stocks do not give good intraday/extended history data for free so the inputs are often spotty or inaccurate.

I also noticed there's no caching for `load` data. Would you like me to add this feature? I already wrote it for myself but I'm not sure how many non-devs would find it useful since I doubt users would fetch the same data repeatedly. It would end up being a large refactor since the various APIs are using different data layouts and you have hardcoded "5. adjusted close", etc all over the place. Kind of afraid of merge conflicts when it ends up behind master.

By the way, if you want to spend time tuning LSTM/RNN/MLP models specifically, you can set `batch_size` to `floor(log2(input_layer_size))` for much faster training if you use a large amount of input samples (assuming you have the RAM for it). It shouldn't have much effect on the results.

DidierRLopes:
This is amazing! Thanks for the feedback @arcutright !!

I'll incorporate some of this info in the next commit.

Regarding the caching of the data, out of curiosity what are you using it for? Yea, it's hardcoded, but you can simply use the df.rename({'wtv': '5. adjusted close'})

arcutright:
> Regarding the caching of the data, out of curiosity what are you using it for?

Making minor changes to model layers and trying to deal with intraday data turns into a loop of make changes -> load data -> train model -> evaluate -> make changes... Since intraday historical data is a pain to fetch from most sources due to rate limits and accuracy concerns, I just cache all the history data to a file and then update the cache with smaller calls to fetch past few days - past few weeks worth of data whenever it's needed.

> Yea, it's hardcoded, but you can simply use the df.rename({'wtv': '5. adjusted close'})

This is true, I just don't like how dirty it is to attach numbers to column names when order doesn't matter. It's copying a dirty part of AlphaVantage's API and pretending like it's standard. And then for intraday data it becomes philosophical whether to count the daily adjusted close or the end-of-day close at 4:00 since you would need to adjust for dividends and stock splits & offerings somehow. For large events you could end up with significant differences between 3:55 and 4:00 if you mix close with adjusted close, however going with raw close prices and ignoring events as a factor is a poor choice since they tend to affect liquidity and market sentiment, and [nasdaq argues that it affects their fundamentals](https://www.nasdaq.com/articles/splitting-stocks-changes-them-fundamentally-2020-09-24).

jmaslek (approved):
So I was thinking that we could have a default for the -e flag of the validation (something like a datetime.timedelta(days=5)).  I know that you haven't played with the code, so that could be later on.

Also if we feed in the validation data into the .fit() method, we can incorporate an early stopping to make it faster (although for a short validation window, it may locally converge too quickly).

One other super nitpick is that in the LSTM plot bottom, technically you only have 1 curve for the prediction error (the black line is always 0).
Screen Shot 2021-03-25 at 9 48 36 PM

Also I have no idea if this is a MPL thing or a my setup thing, but with the plots after calling lstm, I can't exit the plot figures with the top left x,  I have to "q" in the terminal.

I think this can be merged as is, with any updates with the next 'version'

PR #269, opened 2021-03-24
Options reworked 

lolrenx:
merged volume + oi in single graph 
added default expiry date = next available date 
added menu allowing user to select expiry date 
added class op_controller (based on ca_controller) 
added main menu helper


lolrenx:
So yeah, that linting test fails but I can't find out why, it doesn't point to a specific line of code

jmaslek:
I have found that running 'pyupgrade --py36-plus gamestonk_terminal/options/volume.py' auto fixes that error check. (That seems to be the file that runs error).

lolrenx:
> I have found that running 'pyupgrade --py36-plus gamestonk_terminal/options/volume.py' auto fixes that error check. (That seems to be the file that runs error).

good stuff ! thank you

jmaslek:
These plots look really nice :)

Is it possible to get the chains?  Something that after you select your expiry, you could call chain and it would show the usual premium/vol/maybe delta?  Not sure if that is something for down the road.

lolrenx:
> These plots look really nice :)
> 
> You should update the readme.
> 
> Is it possible to get the chains? Something that after you select your expiry, you could call chain and it would show the usual premium/vol/maybe delta? Not sure if that is something for down the road.

thanks ! will update the readme tomorrow morning yeah. 
printing option chain should be easy enough too 👍

pchaganti:
Printing the option chain would be really useful!

:)

DidierRLopes:
Nice work!

Issue #274, opened 2021-03-25
[Bug] Error out when backtesting


WyattLiu:
**Describe the bug**
In Pred, once specifies backtesting date such as lstm -e 2021-03-23, it has KeyError:1980

**To Reproduce**
load AMD
pred AMD
lstm -e 2021-03-23

**Screenshots**
```
(✨) (pred)> help

Daily Stock: AMD

Prediction Techniques:
   help        show this prediction techniques menu again
   q           quit this menu, and shows back to main menu
   quit        quit to abandon program

   sma         simple moving average
   ets         exponential smoothing (e.g. Holt-Winters)
   knn         k-Nearest Neighbors
   linear      linear regression (polynomial 1)
   quadratic   quadratic regression (polynomial 2)
   cubic       cubic regression (polynomial 3)
   regression  regression (other polynomial)
   arima       autoregressive integrated moving average
   prophet     Facebook's prophet prediction
   mlp         MultiLayer Perceptron
   rnn         Recurrent Neural Network
   lstm        Long-Short Term Memory

(✨) (pred)> lstm -e 2021-03-23
1980
Traceback (most recent call last):
  File "/home/wyatt/GamestonkTerminal/gamestonk_terminal/prediction_techniques/neural_networks.py", line 872, in lstm
    df_stock, df_future = _setup_backtesting(df_stock, ns_parser)
  File "/home/wyatt/GamestonkTerminal/gamestonk_terminal/prediction_techniques/neural_networks.py", line 290, in _setup_backtesting
    if ns_parser.s_end_date < get_next_stock_market_days(
  File "/home/wyatt/GamestonkTerminal/gamestonk_terminal/helper_funcs.py", line 211, in get_next_stock_market_days
    holidays = holidays + us_market_holidays(year)
  File "/home/wyatt/GamestonkTerminal/gamestonk_terminal/helper_funcs.py", line 119, in us_market_holidays
    validHolidays.append(datetime.strptime(goodFridays[year], "%Y-%m-%d").date())
KeyError: 1980
```
**Desktop (please complete the following information):**
 - OS:Ubuntu 18.04.5 LTS
 - Python 3.8.8

**Additional context**
N/A

jmaslek:
If you load the stock with a more recent start date, maybe load -t AMD -s 2015-01-01 does the pred work?

Looks like the key error is from a us holiday that year...

jmaslek:
[...]

PR #276, opened 2021-03-25
Added a default load data for stocks and edited good friday list

jmaslek:
This would fix the error associated with #274 (unless a start date is specified past 2010).

I added a default to the load function.  If not specified, it will load from 2015 on.  Also fixed the future good friday dates


DidierRLopes:
Good work, well spotted!

PR #277, opened 2021-03-25
Fix Market Watch functions on DD menu

DidierRLopes:


gamestonk_terminal/due_diligence/market_watch_api.py
```diff
@@ -15,6 +15,7 @@
 
 def sec_fillings(l_args, s_ticker):
```
jmaslek, line 16: shouldn't it be sec_filings (1 l)

jmaslek (approved):

gamestonk_terminal/due_diligence/market_watch_api.py
DidierRLopes: It should, but I eat crayons for breakfast 🙃 ahah 

I'm not on my laptop now until later today, I'll merge it for now, and then fix this

jmaslek: 🤣  doesn't really matter lol

Issue #275, opened 2021-03-25
[BUG] Some commands issue a line of text 


saintsupremacy:
Reddit stock popular command seems to be finding zero stocks for everything but finding wsb posts is no problem. As mentioned though I get this line of text when I go into DD then type either "warnings" or "sec" "argument -h/--help: conflicting option strings: -h, --help" .Thats all the mishaps ive found for now. Maybe its on my end but just something that hasnt happen in previous versions. 

I no longer have trouble with poetry lock though. You can ignore that part in the title.

DidierRLopes:
Thanks for this, regarding the last point the PR #277 should fix both of these. It was a silly mistake.

Let me have a further look to the popular command

DidierRLopes:
Should work now. Let me know @saintsupremacy  :)

saintsupremacy:
> Should work now. Let me know @saintsupremacy 
sorry for the late reply :p, it works! thank you.

PR #279, opened 2021-03-25
Remove after param from psaw, because it's breaking reddit api

DidierRLopes:
This attempts to solve the issue mentioned in #275 

The commands that used the function `psaw_api.search_submissions` and the parameter `after` are not working as before. Once removing this parameter it works fine. I just commented it out, we can revisit this at a later stage


jmaslek (approved):
I don't have the reddit credentials to test this, but code looks good

aia (approved):

PR #280, opened 2021-03-26
remove "fuck these guys" from Robinhood

sabujp:
I don't like Robinhood either, but not everyone who is going to use/see this in the future is going to have the same context. Let's keep the code clean and professional.


DidierRLopes:
Good point! Thanks :)

jmaslek:
Dang I was wondering how long that would last

DidierRLopes:
> Dang I was wondering how long that would last

I was about to comment on that to remove once merged but find it funny and forgot ahah

lolrenx:
could have changed to  "the best brokerage out there /s"

PR #281, opened 2021-03-26
Adding a Yield Curve model

aia:


jmaslek:
Do you want this to be callable from the terminal somewhere?

aia:
> 
> 
> Do you want this to be callable from the terminal somewhere?

Yes, I am still working on those pieces. Treasury Yields have many applications.

You will need treasury yields to calculate Sharpe ratio for portfolios.

jmaslek (approved):

PR #309, opened 2021-04-02
Like this?

alokan:


DidierRLopes:
Exactly like this, thanks!

Well, regarding @jmaslek suggestion, the way the code works we would have to add extra logic to terminal.py to accomodate for this. Which I think is unnecessary.

The best way to handle this would be to have this check before entering the options menu, so that you only enter the menu provided you have a valid ticker. However, I think what you have here is already better than not having anything.

Let me know your opinion :)

alokan:
I think it's unnecessary as well. Moving the check to terminal.py still printed the menu without printing the error in op_controller.py so I think this is fine.

DidierRLopes:
> I think it's unnecessary as well. Moving the check to terminal.py still printed the menu without printing the error in op_controller.py so I think this is fine.

Yea, for it to work you would have to add
```
main_cmd = True
```

to the except  condition.

Regardless I think is unnecessary :) Thanks!

DidierRLopes (approved):

PR #312, opened 2021-04-03
Exception handling for wsb popular function

rmassoth:
I discovered that the program would crash if your Reddit password is incorrect and you try to run the popular function. I was going to add a specific exception for OAuthException but all other Reddit functions just had a catch-all for this. The error message from prawcore is a bit ambiguous but at least it's consistent for all functions.


aia (approved):

DidierRLopes (approved):
Nice, thank you!

PR #317, opened 2021-04-06
Add view command to technical analysis with trendlines

DidierRLopes:
While looking into finviz API I found this graph interesting, which is the same as `candle` but with some more info around. 

aapl

I thought it would be good to also have a view command with the support/resistance trendlines in the technical analysis menu, and thought that we could add this so that we have another source of data, and not the same repeated command.

Let me know if you disagree, regarding:
- Having a view command on technical analysis
- Not using the same source as `candle` with a really similar outcome

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -68,6 +68,7 @@
 * [x] Add initial implementation of trendline analysis (@aia) - [PR #173](https://github.com/DidierRLopes/GamestonkTerminal/pull/173)
 * [x] Add technical summary report provided by FinBrain (@didier) - [PR #294](https://github.com/DidierRLopes/GamestonkTerminal/pull/294)
 * [x] Add recommendation based on technical indicators from Tradingview (@didier) - [PR #301](https://github.com/DidierRLopes/GamestonkTerminal/pull/301)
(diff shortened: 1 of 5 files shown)


jmaslek (approved):

aia:
Interesting find.

In candle I am trying to reconstruct the details of the making of that chart as we will probably want to give users the ability to draw sophisticated charts with pre-populated indicators in the future.

DidierRLopes:
> Interesting find.
> 
> In candle I am trying to reconstruct the details of the making of that chart as we will probably want to give users the ability to draw sophisticated charts with pre-populated indicators in the future.

I like that idea. The pandas-ta library (https://github.com/twopirllc/pandas-ta#types-of-strategies) actually provides the chance to get this custom technical analysis indicators.

Captura de ecrã 2021-04-07, às 21 38 06

The main difficulty here will be to process the different types of technical indicators based on what they represent so that the plot makes sense.

aia:
There is this too
https://github.com/polakowo/vectorbt/tree/master/apps/candlestick-patterns

https://candlestick-patterns.herokuapp.com/

PR #318, opened 2021-04-06
Add docstring to crypto + add test

jmaslek:
I did a small edit on crypto functions and attempted my first unit test.

@aia let me know if this is not done correctly :/


gamestonk_terminal/cryptocurrency/coin_api.py
```diff
@@ -81,8 +92,19 @@ def load(l_args):
         return [None, pd.DataFrame()]
 
 
-def view(coin, prices, l_args):
+def view(coin, prices, l_args: List[str]):
```
aia, line 95: You want to have
```
def view(coin: str, prices: DataFrame, l_args: List[str]):
```

I would also rename l_args to other_args. We use Python built-in type hinting, so we don't need to prefix list variables with l_

aia (approved):

gamestonk_terminal/cryptocurrency/coin_api.py
jmaslek: Got it.  Was the unit test the right idea?  I have never seen the mock/patch before/ so I always thought that the testing would be to just call (in my example) `load` and assert that the load output is equal to something (granted there is no date parameter there so I can't compare to an expected output, but I could compare shapes.

aia: The test is a start, but your mock doesn't do anything because you are not assigning anything to it.
```
class TestCoinGeckoAPI(TestCase):
    @mock.patch(
        "gamestonk_terminal.cryptocurrency.coin_api.CoinGeckoAPI.get_coin_market_chart_by_id"
    )
    def test_coin_api_load(self, mock_load):
        coin, _ = load(["-c", "bitcoin"])
        self.assertEqual(coin, "bitcoin")
        self.assertTrue(mock_load.called)
```
Broadly, the tests are divided into two categories - Unit tests and Integration tests.
Unit tests only execute code that is tested and pretty much no other code in the program.
Integration tests fully execute code and all of the code dependencies.
Assertions are verification of assumptions whichever assumptions you make.

Anatomy of a typical unit test is something like the following:
- Test scenario inputs (e.g. https://github.com/DidierRLopes/GamestonkTerminal/tree/main/tests/data)
- Mock hooks to remove calls to code out of the unit test scope
- Configuration of state before the execution of the unit test (sets up inputs, configures mocks)
- Execution of the code block that is the subject of the unit test
- Verification of the test scenario assumptions about the state after the code block execution

You have the correct idea that for a coin_api.load test, you want to mock (simulate) the execution of CoinGeckoAPI.get_coin_market_chart_by_id call because CoinGeckoAPI.get_coin_market_chart_by_id is an external call.

To execute a mock, however, you need to add a simulation part of the mock where you replace get_coin_market_chart_by_id  with a static result.

Looks like get_coin_market_chart_by_id returns a json.loads - https://github.com/man-c/pycoingecko/blob/master/pycoingecko/api.py#L26. So in a unit test you can replace get_coin_market_chart_by_id  with a load of your static json. Also you can see how pycoingecko is not a particularly great library because we have to dig into pycoingecko code to understand what it returns. A better library would tell us all of that in the pydoc.

pycoingecko

aia: Compare
ark_model

jmaslek: So the proper thing to do is run the `get_coin_market_chart_by_id` somewhere and save that so I can import that static dict (or json.loads).  Then i would run `load` and do some assertions that the load output = static dict (or something similar since the function pulls most recent with no option to set dates)? 

 I think my confusion comes in the mocking then.  So in your example there, you mock yf.download.  This means that when you call `add_order_total`, the `prices` df in that function will return nothing (since it is just a mock call)?  Skimming over the function, it looks like `ret` would just be `df_orders`.  

So in order to test this, you would have to feed in an expected_orders (not sure where that comes in).  Also is there a reason you do the `TestCase(__init__)` instead of just inheriting?

aia: > So the proper thing to do is run the `get_coin_market_chart_by_id` somewhere and save that so I can import that static dict (or json.loads). Then i would run `load` and do some assertions that the load output = static dict (or something similar since the function pulls most recent with no option to set dates)?

Correct.

> I think my confusion comes in the mocking then. So in your example there, you mock yf.download. This means that when you call `add_order_total`, the `prices` df in that function will return nothing (since it is just a mock call)? Skimming over the function, it looks like `ret` would just be `df_orders`.

yf.download will return this:

```
        mock_yf_download.return_value = pd.read_csv(
            StringIO(mock_yf_result), header=[0, 1], index_col=0, parse_dates=True
        )
```

I.e. a mock_yf_result formatted as a DataFrame.

Ret will return the result of merging df_orders with mock_yf_result.


> So in order to test this, you would have to feed in an expected_orders (not sure where that comes in). Also is there a reason you do the `TestCase(__init__)` instead of just inheriting?

expected_orders comes from https://github.com/DidierRLopes/GamestonkTerminal/blob/main/tests/data/discovery_ark_model.yaml

Yes. Inheriting creates a static set of test case. In many cases, I actually generate tests based on data. @parameterize_from_file and pytest_generate_tests do the generation. TestDiscoveryArkModel is not a collection of tests, but rather a template for tests generation. Tests are actually driven by the YAML config.

You can see generated tests like so:
```
pytest --collectonly tests/test_discovery_ark_model.py -s -vvv
```

You can debug tests too.

Test generation is one of the most advanced topics you encounter in testing as it involves meta programming. You can first try simpler techniques to initialize your inputs.

aia (approved):
Looking great!

PR #319, opened 2021-04-06
Add view to screener data

DidierRLopes:
Add `historical` command to screener menu, to allow to visualise stocks. Limit to 10, otherwise too noisy.

Had to add a workaround because YahooFinance API rate limit fails to download several stocks.

Looks like this:

screener_view

top_gainers


jmaslek:
So probably hard to test, but if there is a top gainer that is around $1 and another around $300, will the graph look weird?

jmaslek (approved):

PR #326, opened 2021-04-11
update main readme

jmaslek:
Noticed a couple issues involving pmdarima -- which requires pip install not conda install.  Added that under the ModuleNotFoundError to avoid that.  Also added that it can be updated through pit pull which could close out an open issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@
       <ul>
         <li><a href="#install">Install</a></li>
         <li><a href="#advanced-user-install---machine-learning">Advanced User Install - Machine Learning</a></li>
+        <li><a href="#update">Updating</a></li>
         <li><a href="#api-keys">API Keys</a></li>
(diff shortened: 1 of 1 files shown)


DidierRLopes:
You were faster than me https://github.com/DidierRLopes/GamestonkTerminal/tree/address-bugs :)

DidierRLopes:
I've also added a comment for Indian tickers working with ".NS" at end, which closes another ticket.

I've fixed the "view" cmd for indian tickers, because it was re-loading tickers from Alpha Vantage.

I'm now in the process of refactoring loading intraday data from yfinance because I found out that it's possible. Will also add a flag for before/after hours :)

jmaslek:
Haha.  Your description is better.  Want to just add the message about the pmdarima and I will close/delete branch

DidierRLopes:
Sweet! I've added your comments!

I'll do a PR soon

PR #329, opened 2021-04-11
Portfolio optimization

jmaslek:
Portfolio optimization draft.

Curious what your feedback is.  Just added an equal weighting right now (super simple).  This can also be called from the ca menu.

(Not sure why all the commits as well,  thought I merged the recent PR then `checkout -b this thing`)


gamestonk_terminal/comparison_analysis/ca_controller.py
```diff
@@ -15,6 +15,7 @@
 from gamestonk_terminal.comparison_analysis import market_watch_api as mw_api
 from gamestonk_terminal.comparison_analysis import finbrain_api as f_api
 from gamestonk_terminal.comparison_analysis import finviz_compare_view
+from gamestonk_terminal.portfolio_optimization import port_optimization_controller
```
DidierRLopes, line 18: `po_controller`, probably

gamestonk_terminal/comparison_analysis/ca_controller.py
```diff
@@ -107,6 +109,8 @@ def print_help(self):
         print("   performance   brief performance comparison")
         print("   technical     brief technical comparison")
         print("")
+        print("   po            portfolio optimzation from selected tickers")
```
DidierRLopes, line 112: To keep consistency, I would do:
```
print("  > po            portfolio optimzation from selected tickers")
```
where `>` means that it's a menu.

gamestonk_terminal/comparison_analysis/ca_controller.py
```diff
@@ -312,6 +316,10 @@ def call_technical(self, other_args: List[str]):
         """Process technical command"""
         finviz_compare_view.screener(other_args, "technical", self.ticker, self.similar)
 
+    def call_po(self, _):
+        """Open Portfolio Optimization menu with ticker and similar"""
+        port_optimization_controller.menu([self.ticker] + self.similar)
```
DidierRLopes, line 321: I would do something more convolved here, like:

User does
Menu -> Comparison Analysis -> Portfolio Optimization

* when they `quit`, they quit the program.
* when the `q` the go to the `Menu` instead of back to the Comparison Analysis.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,160 @@
...
+from gamestonk_terminal.menu import session
+from gamestonk_terminal.portfolio_optimization import port_opt_api as po_api
+from prompt_toolkit.completion import NestedCompleter
+
+class PortfolioOptimization:
+
+    CHOICES  = ["help",
+                "q",
+                "quit",
+                "clear",
```
DidierRLopes, line 20: As you mentioned before, I wouldn't bother with the `clear` instead I would add `select` command.
So that user could do:
* select - To select list of tickers
* add - To add to the existing list of tickers

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,160 @@
...
+        parser.add_argument(
+            "-t",
+            "--tickers",
+            dest="add_tickers",
+            type=lambda s: [str(item).upper() for item in s.split(",")],
+            default=[],
+            help="add tickers to optimzation.",
+        )
+        try:
+            # For the case where a user uses: 'select NIO,XPEV,LI'
```
DidierRLopes, line 108: For the case where a user uses: 'add NIO,XPEV,LI'

DidierRLopes (commented):
Yup, on top of this would be good to have the option from screener.

I can take care of that one once you have something working :)

gamestonk_terminal/comparison_analysis/ca_controller.py
jmaslek: Yeah I am not sure exactly how to do that.  If I do either `q` or `quit` from menu-ca-po, it just goes back to ca.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: Oops.  Just copied the select function from your ca menu

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: so select would overwrite any added so far?

gamestonk_terminal/comparison_analysis/ca_controller.py
jmaslek: no idea how I never caught that trend

gamestonk_terminal/comparison_analysis/ca_controller.py
DidierRLopes: You just need to do:
```
return port_optimization_controller.menu([self.ticker] + self.similar)
```

Because `port_optimization_controller.menu` already returns q/quit

gamestonk_terminal/comparison_analysis/ca_controller.py
DidierRLopes: I think it's only on the fundamental analysis menu, to dive into > av and > fmp.

I think it's nice to have a slight distinction compared normal commands. Probably even the main menu should have them..

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
DidierRLopes: Exactly :)

jmaslek:
My other question would be -- should optimize have different options like other menus (ie main menu look like

Optimize:
    equal_weight          optimize by equal weight
    market cap             optimize by market cap
    ...
)
Or just have optimize have a bunch of different flags for different options?

DidierRLopes:
> My other question would be -- should optimize have different options like other menus (ie main menu look like
> 
> Optimize:
> equal_weight optimize by equal weight
> market cap optimize by market cap
> ...
> )
> Or just have optimize have a bunch of different flags for different options?

Personally, I approach this on a scenario basis based on the amount of options and what makes sense.

**Prediction menu**
Here we have each prediction technique individually because they are not that many options and it's useful to add an explanation across each command.

**Screener menu**
In case of screening through a signal, I started by doing a sub-menu, but then when realized the amount of signals that there are..
 (Top Gainers, Top Losers, New High, New Low, Most Volatile, Most Active, Unusual Volume, Overbought, Oversold, Downgrades, Upgrades, Earnings Before, Earnings After, Recent Insider Buying, Recent Insider Selling, Major News, Horizontal S/R, TL Resistance, TL Support, Wedge Up, Wedge Down, Triangle Ascending, Triangle Descending, Wedge, Channel Up, Channel Down, Channel, Double Top, Double Bottom, Multiple Top, Multiple Bottom, Head & Shoulders, Head & Shoulders Inverse) 
 it didn't make sense to have an option for each, so now the user can do something like `valuation -s new_high`

Regarding portfolio optimization, I would have 1 command for each optimization technique because:
* It provides a faster usage of each optimization technique
* You can add more information on the command regarding that technique through the `-h`flag
* There aren't that many options that would make the menu extremely extensive

jmaslek:
something more like that?

gamestonk_terminal/comparison_analysis/ca_controller.py
```diff
@@ -109,7 +109,7 @@ def print_help(self):
         print("   performance   brief performance comparison")
         print("   technical     brief technical comparison")
         print("")
-        print("   po            portfolio optimzation from selected tickers")
+        print("   >po           portfolio optimzation from selected tickers")
```
DidierRLopes, line 112: I'd say `> po` with that spacing in between

gamestonk_terminal/comparison_analysis/ca_controller.py
```diff
@@ -318,7 +318,7 @@ def call_technical(self, other_args: List[str]):
 
     def call_po(self, _):
         """Open Portfolio Optimization menu with ticker and similar"""
-        port_optimization_controller.menu([self.ticker] + self.similar)
+        return po_controller.menu([self.ticker] + self.similar)
```
DidierRLopes, line 321: you can write a comment above that it's to either leave PO terminal to go back to main menu, or to quit. It doesn't go back to CA menu

gamestonk_terminal/main_helper.py
```diff
@@ -60,7 +60,7 @@ def print_help(s_ticker, s_start, s_interval, b_is_market_open):
     )
     print("   pa          portfolio analysis, \t\t supports: robinhood, alpaca, ally ")
     print("   crypto      cryptocurrencies, \t\t uses coingecko api")
-    print("   po          portfolio analysis, \t\t UPDATE ME BEFORE PR")
+    print("   po          portfolio analysis, \t\t menu to explore portfolio optimizations")
```
DidierRLopes, line 63: `print("   po          portfolio optimization, \t\t e.g. MVO, HRP, efficient frontier, sharp ratio")`

gamestonk_terminal/portfolio_optimization/port_opt_api.py
```diff
@@ -18,11 +18,40 @@ def optimize(list_of_stocks:List, other_args:List[str]):
...
+        weights : dict
+            Dictionary of weights where keys are the tickers
+
+        """
+    weights = {}
+    n_stocks = len(list_of_stocks)
+    for stock in list_of_stocks:
+        weights[stock] = round(1 / n_stocks, 5)
+
+    return weights
```
DidierRLopes, line 56: You have 2 duplicated functions here.

I would probably call this file like "mean_variance_optimizer_view" and add the methods that correspond to this category inside.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -46,7 +47,11 @@ def print_help(tickers: Set[str]):
         )
         print("")
         print("   add          add ticker to optimize")
-        print("   optimize     run optimization")
+        print("   select       overwrite current tickers with new tickers")
+        print("")
+        print("Optimization:")
```
DidierRLopes, line 52: Is it possible to split these across their categories? Or you think there's no point?

gamestonk_terminal/comparison_analysis/ca_controller.py
jmaslek: Will do.  Ill also add an option to go back to the ca option

gamestonk_terminal/portfolio_optimization/port_opt_api.py
jmaslek: yeah I just added a second one.  This is still a draft so I need to clean up.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: not sure I follow.  what categories?

DidierRLopes: From what I read there are several common methods within "Mean Variance Optimization", but then there's also:
* Efficient Semivariance
* Efficient CVaR
* Black-Litterman Allocation
* Hierarchical Risk Parity
* Markowitz's critical line algorithm

Again, I haven't studied this in detail, so there may not be a "clean" cut where a nice structuring is possible. 

I'm thinking about taking the course that aia suggested to have a better idea of what we can/should do

jmaslek:
Let me know what your thoughts so far are.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: So I made it kinda like this.  Portflolio theory is intense and there are a lot of ways it can be done.  I think there are many ML approaches as well

I am not sure how intense we should be.  I haven't dug too much into the pyportfolioopt package, but it seems like there are ways you can add constraints (maybe max allocation or max factor exposure, etc), which seem like a good entry level for the average user.  

I was thinking of having the options for the user to do : max_sharpe , min_volatility and efficient frontier at a given risk level.  Then we can start trying to add on more complex methods (allowing for factors

DidierRLopes: Yup, I agree with that approach.

It's hard for me to try to suggest a "cleanear approach" since I don't yet fully understand every moving piece.

But this sounds a great starting point to me

gamestonk_terminal/portfolio_optimization/README.md
```diff
@@ -0,0 +1,126 @@
+# Portfolio Optimization
```
DidierRLopes, line 1: I think overall the README can be improved, I can take care of this!

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+class PortfolioOptimization:
+
+    CHOICES = [
+        "help",
+        "q",
+        "quit",
+        "ca",
+        "scr",
```
DidierRLopes, line 25: I don't think this should have the Ca,Scr options.

It makes sense going from Ca,Scr to Portfolio Analysis, but not the other way around.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+        print("   > ca          comparison analysis menu")
+        print("   > scr         screener menu")
+        print(f"\nCurrent Tickers: {('None', ', '.join(tickers))[bool(tickers)]}")
+        print("")
+        print("   add          add ticker to optimize")
+        print("   select       overwrite current tickers with new tickers")
+        print("")
+        print("Optimization:")
+        print("")
+        print("   Property weighted:")
```
DidierRLopes, line 67: You can remove this, and go with your first idea.

Let all the command options with only 3 spaces, e.g.
```
print("   mkt_cap        marketcap weighted portfolio")
print("   div_yield       dividend weighted portfolio")
...
```

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+    def call_quit(self, _):
+        """Process Quit command - quit the program"""
+        return True
+
+    def call_ca(self, _):
+
+        return ca_controller.menu(pd.DataFrame(), self.ca_ticker, "", "1440min")
+
+    def call_scr(self, _):
+        return screener_controller.menu()
```
DidierRLopes, line 113: No need for this, and the one above IMO

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def call_add(self, other_args: List[str]):
+        self.add_stocks(self, other_args)
+
+    def call_select(self, other_args: List[str]):
+        self.tickers = set([])
+        self.add_stocks(self, other_args)
```
DidierRLopes, line 120: Nicely done!

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def call_equal_weight(self, other_args: List[str]):
+        weights = po_api.equal_weight(self.tickers, other_args)
```
DidierRLopes, line 123: To keep the structure the same, could you have this using argparse?

This way the user can add the -h flag and we can add a nice description to what the optimization does. Also, in the future we may need to add arguments for these, so this way it would be more future proof.

Another thing is that it would be nice to see bars with percentage, and even have an option that allows for discrete allocation based on a pre-defined user allocation.

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def call_yolo(self, _):
+        # Easter egg :)
+        print("DFV YOLO")
+        print({"GME": 200})
+        print("")
+
+    @staticmethod
+    def add_stocks(self, other_args: List[str]):
+
```
DidierRLopes, line 176: extra line in here

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
```diff
@@ -0,0 +1,277 @@
...
+            print(e)
+
+        print("")
+
+    @classmethod
+    def from_ca_menu(cls, ticker: str, similar: List[str]):
+        return cls(set([ticker] + similar))
+
+
+def menu_from_ca(ticker: str, similar: List[str]):
```
DidierRLopes, line 218: I don't think you need to have a new controller menu here, you can just use the normal one. This makes the code a bit confusing imo, unless I'm missing something.

jmaslek: If you think I should remove the jump to ca, then I can remove it.  I just did it this way so I could keep track of the loaded ticker to go back to ca.  I can also get rid of the class method in this case.

DidierRLopes: That would be perfect, I think!

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: So have an argument that would either be like --marketcap or --dividend instead of individual functions (where i input the property type arg)?

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: so no property weighted or MVO?  Just list all options?

gamestonk_terminal/portfolio_optimization/README.md
jmaslek: Yeah my bad,  definitely rushed through

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
DidierRLopes: No. Having it as you have, but parsing it through an argparse, so that the user can do:
```
market_cap -h
```
and get a description of what the command does. 

And in the future if we add that "value component" that the user can add, it's gonna be easier as the user can do:
```
market_cap --value 1000
```
and instead of getting: FB 0.5, AMZN 0.5, he gets: FB 500, AMZN 500. If that makes sense.

Also, the current outputs are not the prettiest, as you can still see the brackets. I would put these on a dataframe, and output:
```
ticker    value
AMZN    50%
FB          50%
```

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
DidierRLopes: Exactly. As you mentioned the first time

gamestonk_terminal/portfolio_optimization/README.md
DidierRLopes: No worries, I can take care of it. I'm usually really picky so I don't mind doing it

gamestonk_terminal/portfolio_optimization/port_optimization_controller.py
jmaslek: oh yeah that should be straightforward.  I was thinking there could be something like --value or --shares that would return the amount.  I also just printed the dictionary like in the pyportfolioopt module.

In terms of plotting, I was thinking I could default to a pie chart or something (in the case of long only) but have options for others

jmaslek:
Also any idea how to get around the mypy set/list thing?  I use set to make sure there aren't duplicate stocks.  Should I just use list(set( stuff) ) ?

DidierRLopes:
> Also any idea how to get around the mypy set/list thing? I use set to make sure there aren't duplicate stocks. Should I just use list(set( stuff) ) ?

I'd say so yes.

Also, can the port_optimization_controller file be only po_controller? To keep consistency across remaining controller files names

jmaslek:
Sure.  Just figured I'd change so its less likely to be confused with pa

DidierRLopes:
> Sure. Just figured I'd change so its less likely to be confused with pa

with pa and po, we're only missing pe, pi and pu eheh

jmaslek:
And sometimes py

jmaslek:
Just have a couple more that need argparsing then showing/plotting!

jmaslek:
Should be close.  Just gotta figure out the pyupgrade

jmaslek:
I think i commit too much :p

DidierRLopes:
This is looking amazing bro!! Even my gf looked and said "that looks cool" ahah

But you know how pedantic I am, I'd do these:

Captura de ecrã 2021-04-13, às 20 30 43

DidierRLopes:
Captura de ecrã 2021-04-13, às 20 37 29

the help talks about the usage of `ef` and prints the `Maximum Sharp Weights: ` which is unwanted.

Also, I would print in the console the percentages instead of `0.5484` as it is makes it easier to understand

gamestonk_terminal/portfolio_optimization/port_opt_api.py
```diff
@@ -0,0 +1,360 @@
+""" Portfolio Optimization Functions"""
```
DidierRLopes, line 1: You can call this `optimizer_view` instead, and `optimizer_helper` to the other one to follow same naming convention

gamestonk_terminal/portfolio_optimization/port_opt_api.py
```diff
@@ -0,0 +1,360 @@
...
+
+        plt.show()
+        print("")
+
+    except Exception as e:
+        print(e)
+        print("")
+
+
+def ef_portfolio(list_of_stocks: List[str], port_type: str, other_args: List[str]):
```
DidierRLopes, line 214: You don't need to specify list_of_stocks, you can use just `stocks`. The typing properties will tell you the type of the variable

gamestonk_terminal/portfolio_optimization/port_opt_api.py
```diff
@@ -0,0 +1,360 @@
...
+        List of the stocks to be included in the weights
+    port_type: str
+        Method to be used on ef object (example: max_sharpe, min_volatility)
+    Returns
+    -------
+    weights: dict
+        Dictionary of weights where keys are the tickers.
+    """
+
+    parser = argparse.ArgumentParser(add_help=False, prog="ef")
```
DidierRLopes, line 229: the prog should be equal to port_type

gamestonk_terminal/portfolio_optimization/port_opt_api.py
```diff
@@ -0,0 +1,360 @@
...
+            print("")
+            ef.portfolio_performance(verbose=True)
+            print("")
+            return weights
+        except Exception as e:
+            print(e)
+            print("")
+            return {}
+
+    elif port_type == "min_volatility":
```
DidierRLopes, line 279: It seems that you can optimize the code a bit further. The initial parts are the same for all, the only difference is when you get ef_min_vol, or ef_sharpe, ...  So you can just have an if there.

DidierRLopes: You can even have a function that get this data for you based on the port_type. This would be the preferred approach

gamestonk_terminal/portfolio_optimization/port_opt_api.py
```diff
@@ -0,0 +1,360 @@
...
+    )
+    parser.add_argument(
+        "-n", default=300, dest="n_port", help="number of portfolios to simulate"
+    )
+
+    try:
+        ns_parser = parse_known_args_and_warn(parser, other_args)
+        if not ns_parser:
+            return {}
+        stock_prices = process_stocks(list_of_stocks, ns_parser.period)
```
DidierRLopes, line 176: I think to follow MVC pattern you need to have a function this `argparse` and then call a function, that is the one that generates the plot

DidierRLopes:
`show_ef` fails for only 1 ticker

gamestonk_terminal/portfolio_optimization/port_opt_helper.py
```diff
@@ -0,0 +1,112 @@
...
+        List of tickers to get historical data for
+    period: str
+        Period to get data from yfinance
+
+    Returns
+    -------
+    stock_closes: DataFrame
+        DataFrame containing daily (adjusted) close prices for each stock in list
+    """
+    stock_prices = yf.download(list_of_stocks, period=period, group_by="ticker")
```
DidierRLopes, line 32: Can you add `progress=False` to avoid `***** 100% *****`

DidierRLopes:
Captura de ecrã 2021-04-13, às 21 02 06

This should mention what is the risk that is used as default, and which ranges are available.

jmaslek:
> This should mention what is the risk that is used as default, and which ranges are available.

Ill put in the default, but the range will depend on the stocks being used and the time period selected (since that changes covariance and expected returns)

> Most of the commands fail if only 1 ticker is provided
If there is only 1 ticker, then there is only 1 portfolio.  I'll add a message if only 1 is provided

DidierRLopes:
> > This should mention what is the risk that is used as default, and which ranges are available.
> 
> Ill put in the default, but the range will depend on the stocks being used and the time period selected (since that changes covariance and expected returns)
> 
> > Most of the commands fail if only 1 ticker is provided
> 
> If there is only 1 ticker, then there is only 1 portfolio. I'll add a message if only 1 is provided

In case it fails, can we get that info and output it along with the `use bigger target volatility`.

Yup, a message is perfect.

jmaslek:
> In case it fails, can we get that info and output it along with the `use bigger target volatility`.

The error message from them is pretty straightforward.  It tells what you have to use.

gamestonk_terminal/portfolio_optimization/port_opt_api.py
jmaslek: not following....

DidierRLopes: Don't worry with this one, it doesn't matter!

gamestonk_terminal/portfolio_optimization/optimizer_view.py
```diff
@@ -226,7 +180,7 @@ def ef_portfolio(list_of_stocks: List[str], port_type: str, other_args: List[str
         Dictionary of weights where keys are the tickers.
     """
 
-    parser = argparse.ArgumentParser(add_help=False, prog="ef")
+    parser = argparse.ArgumentParser(add_help=False, prog="port_type")
```
DidierRLopes, line 183: I think here you mean port_type without the ""

DidierRLopes (commented):
Nice!!

Just fix the lints and feel free to merge it. Looking great!

jmaslek:
> Just fix the lints and feel free to merge it. Looking great!

Ill let you take one more look to make sure I didn't break anything haha.

DidierRLopes:
> > Just fix the lints and feel free to merge it. Looking great!
> 
> Ill let you take one more look to make sure I didn't break anything haha.

Sorry, it has been a busy day! Will give you feedback in 1 second :)

DidierRLopes (approved):
Looking good!

PR #339, opened 2021-04-13
Randomize similar companies before selecting 10 for analysis

piiq:
This is for the case when more than 10 similar companies
are available. This allows getting better representation compared to
always getting the first 10 similar companies from the list that's
ordered alphabetically


DidierRLopes:
Good idea! 

Can you fix black complain and push it?

You can install black and run
```
black <file_that_is_failing>
```

piiq:
@DidierRLopes done

jmaslek (approved):

PR #342, opened 2021-04-14
Refactor options menu

DidierRLopes:


jmaslek:
Damn dude.  Do you sleep?

DidierRLopes:
> Damn dude.  Do you sleep?

Sometimes eheh

jmaslek:
So I have no idea if this is a super specific example, but currently if I do `vcalls` or `vputs` for `aapl` , exp 0 : I get very wide x labels:
Screen Shot 2021-04-14 at 3 24 08 PM

DidierRLopes:
> So I have no idea if this is a super specific example, but currently if I do `vcalls` or `vputs` for `aapl` , exp 0 : I get very wide x labels:
> 
> What if there is an input (`--width` maybe?) for plotting (current price - width : current price + width). Could be either a number or a percent...
> 
> Screen Shot 2021-04-14 at 3 24 08 PM

I thought about this case last night when I turned off actually. I will use the 50% percentile as default as Laurent had done it!

gamestonk_terminal/options/volume_view.py
```diff
@@ -0,0 +1,463 @@
...
+        if ns_parser.min_sp == -1:
+            min_strike = np.percentile(df_calls["strike"], 25)
+        else:
+            min_strike = ns_parser.min_sp
+
+        if ns_parser.max_sp == -1:
+            max_strike = np.percentile(df_calls["strike"], 75)
+        else:
+            max_strike = ns_parser.max_sp
+
```
jmaslek, line 346: Maybe default to a percentage of the current stock price instead of middle 50%. The percentile can be pretty large still.

jmaslek (approved):
Looks good!  Only 1 small comment

gamestonk_terminal/options/volume_view.py
DidierRLopes: Yea, you're right. That's a good point. I used 0.75 and 1.25, let me know if that is good enough :)

PR #344, opened 2021-04-14
docker image: fix install

gregwebs:
The existing docker image immediately has errors on startup when I ran it. Installing requirements-full fixed that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# Build as: docker build . -t gamestonkterminal:dev
 FROM python:3.6.13-slim-buster
 
 RUN apt-get update && apt-get -y install --no-install-recommends \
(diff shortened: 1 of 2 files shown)


aia:
Your pull request is removing poetry (our selected dependency management framework) and installing full dependencies instead of the standard dependencies that exclude ML.

What problems are you having with docker?

gregwebs:
This is using a cached list of dependencies already existing in the repo. So I would presume that this is not "removing poetry" but instead using a cache that was generated by poetry?

Here is what happens with the existing docker image on master. Have you tried running it?

```
docker run --rm -it gamestonkterminal:dev
     code/GamestonkTerminal  main
  File "terminal.py", line 27, in <module>
    from gamestonk_terminal.exploratory_data_analysis import eda_controller
  File "/home/python/gamestonk_terminal/exploratory_data_analysis/eda_controller.py", line 9, in <module>
    from gamestonk_terminal.exploratory_data_analysis import eda_api
  File "/home/python/gamestonk_terminal/exploratory_data_analysis/eda_api.py", line 13, in <module>
    import statsmodels.api as sm
ModuleNotFoundError: No module named 'statsmodels'
```

aia:
That error suggests that we are missing statsmodels from https://github.com/DidierRLopes/GamestonkTerminal/blob/main/pyproject.toml

aia:
The issue should be addressed now. Please try running the docker.

gregwebs:
Thanks, that works!

PR #349, opened 2021-04-16
Big-ish PO refactoring

DidierRLopes:
See https://github.com/DidierRLopes/GamestonkTerminal/tree/picky-po/gamestonk_terminal/portfolio_optimization

min_vol

Doesn't change the program functionality (apart from the fact that there's a new optimization command + from screener it is possible to jump to PO carrying last screened stocks), it's mostly details in user usage/visualization.

### DIFF
--- a/gamestonk_terminal/README.md
+++ b/gamestonk_terminal/README.md
@@ -9,6 +9,7 @@
 * [Due Diligence](#Due-Diligence-)
 * [Prediction Techniques](#Prediction-Techniques-)
 * [Portfolio Analysis](#Portfolio-Analysis-)
+* [Portfolio Optimization](#Portfolio-Optimization-)
 * [Cryptocurrencies](#Cryptocurrencies-)
 * [Comparison Analysis](#Comparison-Analysis-)
 * [Exploratory Data Analysis](#Exploratory-Data-Analysis-)
@@ -242,16 +243,19 @@ Command|Description|Brokers
 
 &nbsp;
 ## Portfolio Optimization [»](portfolio_optimization/README.md)
-command|desciription
+
+Command|Description
 ------|------
-`equal_weight`|Returns equal weights for all selected stocks
-`mkt_cap`| Returns weights that are weighted by market cap
-`div_yield`|Returns weights that are weighted by dividend yield
-`max_sharpe`| Returns portfolio of Efficient Frontier with maximum sharpe ratio
-`min_vol`| Returns portfolio on Efficient Frontier with minimum volatility
-`eff_risk`| Returns portfolio on Efficient Frontier that maximizes returns at a given risk
-`eff_ret`| Returns portfolio on Efficient Frontier that minimizes risk at a given return level
-`show_eff`|Plots random portfolios and shows the Efficient Frontier
+`add`| add ticker to optimize
+`select`| overwrite current tickers with new tickers
+`equal`| equally weighted
+`property`| weight according to selected info property (e.g. marketCap)
+`maxsharpe`| optimizes for maximal Sharpe ratio (a.k.a the tangency portfolio)
+`minvol`| optimizes for minimum volatility
+`maxquadutil`| maximises the quadratic utility, given some risk aversion
+`effret`| maximises return for a given target risk
+`effrisk`| minimises risk for a given target return
+`ef`| show the efficient frontier
 
 &nbsp;
 
@@ -331,7 +335,9 @@ Command|Description
 Command|Description|Source
 ------ | --------|----
 `exp`           | see/set expiry date
-`volume`        | plot options trading volume / open interest |[Yahoo Finance](https://finance.yahoo.com/)
+`volume`        | volume + open interest options trading plot |[Yahoo Finance](https://finance.yahoo.com/)
+`vcalls`        | calls volume + open interest plot |[Yahoo Finance](https://finance.yahoo.com/)
+`vputs`         | puts volume + open interest plot |[Yahoo Finance](https://finance.yahoo.com/)
 
 &nbsp;
 
--- a/gamestonk_terminal/portfolio_optimization/README.md
+++ b/gamestonk_terminal/portfolio_optimization/README.md
@@ -8,8 +8,12 @@ This menu aims to optimize a portfolio of pre-loaded stocks, and the usage of th
   * overwrite current tickers with new tickers
 * [equal](#equal)
   * equally weighted
+* [mktcap](#mktcap)
+  * weighted according to market cap (property marketCap)
+* [dividend](#dividend)
+  * weighted according to dividend yield (property dividendYield)
 * [property](#property)
-  * weight according to selected info property (e.g. marketCap)
+  * weight according to selected info property
 
 [Mean Variance Optimization](#Mean_Variance_Optimization)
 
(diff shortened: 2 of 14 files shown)


gamestonk_terminal/README.md
```diff
@@ -242,16 +243,19 @@ Command|Description|Brokers
...
-`eff_risk`| Returns portfolio on Efficient Frontier that maximizes returns at a given risk
-`eff_ret`| Returns portfolio on Efficient Frontier that minimizes risk at a given return level
-`show_eff`|Plots random portfolios and shows the Efficient Frontier
+`add`| add ticker to optimize
+`select`| overwrite current tickers with new tickers
+`equal`| equally weighted
+`property`| weight according to selected info property (e.g. marketCap)
+`maxsharpe`| optimizes for maximal Sharpe ratio (a.k.a the tangency portfolio)
+`minvol`| optimizes for minimum volatility
+`maxquadutil`| maximises the quadratic utility, given some risk aversion
```
jmaslek, line 255: I had intentionally left that out since it is more complicated (does market neutral break the pie chart since there will be negative weights?). But definitely another option

gamestonk_terminal/portfolio_optimization/README.md
```diff
@@ -1,169 +1,192 @@
...
...lgorithm, dividendRate, exDividendDate, beta, circulatingSupply, regularMarketDayLow, priceHint, currency, trailingPE, regularMarketVolume, lastMarket, maxSupply, openInterest, marketCap, volumeAllCurrencies, strikePrice, averageVolume, priceToSalesTrailing12Months, dayLow, ask, ytdReturn, askSize, volume, fiftyTwoWeekHigh, forwardPE, fromCurrency, fiveYearAvgDividendYield, fiftyTwoWeekLow, bid, dividendYield, bidSize, dayHigh, annualHoldingsTurnover, enterpriseToRevenue, beta3Year, profitMargins, enterpriseToEbitda, 52WeekChange, morningStarRiskRating, forwardEps, revenueQuarterlyGrowth, sharesOutstanding, fundInceptionDate, annualReportExpenseRatio, bookValue, sharesShort, sharesPercentSharesOut, fundFamily, lastFiscalYearEnd, heldPercentInstitutions, netIncomeToCommon, trailingEps, lastDividendValue, SandP52WeekChange, priceToBook, heldPercentInsiders, shortRatio, sharesShortPreviousMonthDate, floatShares, enterpriseValue, threeYearAverageReturn, lastSplitFactor, legalType, lastDividendDate, morningStarOverallRating, earningsQuarterlyGrowth, pegRatio, lastCapGain, shortPercentOfFloat, sharesShortPriorMonth, impliedSharesOutstanding, fiveYearAverageReturn, and regularMarketPrice.
```
jmaslek, line 74: Damn you went all out here. A lot of these are kinda useless and better used as factors in the optimization.

gamestonk_terminal/portfolio_optimization/README.md
```diff
@@ -1,169 +1,192 @@
...
+* --pie : Display a pie chart for weights
+
+![marketCap](https://user-images.githubusercontent.com/25267873/114958294-980a7700-9e5a-11eb-9db4-d06e6a1e8981.png)
+
+
+## Mean Variance Optimization <a name="Mean_Variance_Optimization"></a>
+
+These approaches are based off of the efficient frontier approach, which is meant to solve the following optimization problem.
+
+Minimize: <img src="https://latex.codecogs.com/svg.image?w^TS&space;w" title="w^TS w" />
```
jmaslek, line 85: So GitHub now has dark mode and these weren't showing on my browser so it may be better to drop the equations :(

gamestonk_terminal/portfolio_optimization/README.md
```diff
@@ -1,169 +1,192 @@
...
-* -p/--period Amount of time to retrieve data from yfinance. Options are: 1d,5d,1mo,3mo,6mo,1y,2y,5y,10y,ytd,max and it defaults to 3mo.
-* -r/--risk Risk tolerance.  Default is 0.1 (10%)
-* -v/--value If provided, this represents an actual allocation amount for the portfolio.  Defaults to 1, which just returns the weights.
-* --pie Flag that displays a pie chart of the allocations.
-
-### eff_ret
-This portfolio minimizes the risk at a given return level
-The usage is:
+
+Maximise the Sharpe Ratio. The result is also referred to as the tangency portfolio, as it is the portfolio for which the capital market line is tangent to the efficient frontier. This is a convex optimization problem after making a certain variable substitution. See Cornuejols and Tutuncu (2006) <http://web.math.ku.dk/~rolf/CT_FinOpt.pdf> for more. The sharpe ratio is defined as (Mean Returns - Risk Free Rate)/(Standard Deviation of Returns).
```
jmaslek, line 104: Seems a little overkill?  I would drop the complex optimization sentence

gamestonk_terminal/portfolio_optimization/README.md
```diff
@@ -1,169 +1,192 @@
...
-time period, which changes the expected returns and historical volatility, which changes the optimization.  We could also specify a dollar
-amount that you wish to allocate using the `-v` flag.
+
+Maximise return for a target risk. The resulting portfolio will have a volatility less than the target (but not guaranteed to be equal).
+
+* -p : Amount of time to retrieve data from yfinance. Default: 3mo.
+* -v : If provided, this represents an actual allocation amount for the portfolio.  Defaults to 1, which just returns the weights.
+* --pie : Display a pie chart for weights
+* -t : The desired maximum volatility of the resultingportfolio
+* -n : Whether the portfolio should be market neutral (weights sum to zero). Defaults: False.
```
jmaslek, line 175: As I mentioned above, curious if this breaks the pie plots when market neutral

jmaslek (approved):
Niceeeee

gamestonk_terminal/README.md
DidierRLopes: Yeah, I was looking at the API class and just saw there was another function so just added it since it was there.

I tested that only 1/2, and saw negative values but nothing broke. Let me investigate this further later on

gamestonk_terminal/portfolio_optimization/README.md
DidierRLopes: I thought the same, only removed the non numeric ones. We can have the marketCap, and divYield as commands if you think those are the only that kind of make sense

Or do like we do regression on pred menu. So,

mktcap
dividend
property

gamestonk_terminal/portfolio_optimization/README.md
DidierRLopes: Good point. I'll fix this, I'll add it as an image

gamestonk_terminal/portfolio_optimization/README.md
DidierRLopes: I just copied the one that the pyportfopt had for this funcionality. But I can shorten it

gamestonk_terminal/portfolio_optimization/README.md
DidierRLopes: I'll do some more testing around this. Thanks for pointing it out!

If it does I'll just remove that flag

gamestonk_terminal/README.md
jmaslek: So I finally got around to looking at it and for market neutral, the pie just shows positive weights.  Not sure if its simple enough to have a (-n AND --pie) condition that tells you there are negative weights and suppress the plot

gamestonk_terminal/portfolio_optimization/README.md
jmaslek: The only reason I did those are since those are common portfolio ones.  As I mentioned, we can have these as factors for the next step of PO (we can take this discussion offline)

gamestonk_terminal/README.md
DidierRLopes: Done!

gamestonk_terminal/portfolio_optimization/README.md
DidierRLopes: Let me refactor for this one :)

DidierRLopes:
This should now take care of everything :)

Issue #272, opened 2021-03-24
[FR] Add in options chain to option menu


jmaslek:
Based on discussion in #269, it would be dope to have an option to look at the option chains after you selected your expiry.

jmaslek:
@lolrenx

lolrenx:
Will be in the next options PR :)

pchaganti:
This looks like it has been merged. Close this issue?

jmaslek:
Closed on #351

PR #351, opened 2021-04-16
add in option chains :)

jmaslek:
Addressing #272.

This PR adds in the ability to pull options chains from Tradier and it displays them in table form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,10 +257,14 @@ These are the ones where a key is necessary:
   * Financial Modeling Prep: https://financialmodelingprep.com/developer
   * FRED: https://fred.stlouisfed.org/docs/api/api_key.html
   * News API: https://newsapi.org
+  * Tradier: https://developer.tradier.com/getting_started
 
 When these are obtained, don't forget to update [config_terminal.py](/gamestonk_terminal/config_terminal.py).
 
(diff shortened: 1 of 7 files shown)


DidierRLopes:
Really nice!!

You missed a `print("")` after the command runs.

The only thing I would say to improve the plot an additional step is:
1. Add an empty column on the left and on the right, of strike. In order to show better distinction between calls and puts.
2. Perhaps add green color on calls and red color for puts?

DidierRLopes (approved):

DidierRLopes:
Once this is merged,
- I'll refactor the options readme
- I'll add this command to features.md
- I'll add this PR to the ROADMAP

jmaslek:
> 2. Perhaps add green color on calls and red color for puts?
I thought about this but honestly I am not sure how to do that with tabulate.  I also thought about printing Calls/Puts in colored above the table, but I am not sure if the table size will change based on the chain and the formatting gets messed up.

DidierRLopes:
> > 2. Perhaps add green color on calls and red color for puts?
> 
> I thought about this but honestly I am not sure how to do that with tabulate.  I also thought about printing Calls/Puts in colored above the table, but I am not sure if the table size will change based on the chain and the formatting gets messed up.

Yea tabulate limits you a bit there. It would be good if you could add that extra delimiter, but if not It's still amazing, and feel free to merge it!

aia (approved):

PR #352, opened 2021-04-16
Support for .env files

jperkins12:
Users can define API keys/secrets in .env file.

Uses python-dotenv (https://pypi.org/project/python-dotenv/).


DidierRLopes:
Nice, thanks for this! 2 things:
* Do we need to add .env to gitignore?
* Can you add a line on the main README where we talk about API keys, to say that we can also add it to this file?

jperkins12:
Done and done. Looks like `.env` was already included in the gitignore.

aia (approved):

PR #356, opened 2021-04-17
Forex support through Oanda

alokan:


aia:
Great work!

You want to add oandapyV20 to poetry and regenerate requirements with poetry.

Check out how requirements.txt are generated in CI.

DidierRLopes:
Wow, amazing work right here!!!!

I'm impressed!

PR #360, opened 2021-04-17
Forex support through Oanda

alokan:
I used poetry export for requirements and requirements-full. I'm not sure why so many things changed.


gamestonk_terminal/oanda/oanda_functions.py
```diff
@@ -0,0 +1,517 @@
+import argparse
```
DidierRLopes, line 1: This file should be called "fx_view.py" instead

DidierRLopes (commented):
Commented on minor things. Good work!!

You also need to update main_helper where there's the main terminal print.
So that people know that there's a new `fx` menu.

Regarding the poetry, should be fine as you've done! 

I'll try to give this a go between today and tomorrow, will need to create a Oanda account. Also you may want to try to add a try-catch block that checks if user account is valid, because the terminal crashed when I accessed the menu without a valid account login

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+        response = client.request(request)
+        print(response)
+    except V20Error as e:
+        print(e)
+
+
+def cancel_pending_order(accountID, other_args: List[str]):
+    parser = argparse.ArgumentParser(
+        add_help=False,
+        prog="cancelpendingorder",
```
DidierRLopes, line 178: here should be `cancel`

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+        description="Cancel Pending Order",
+    )
+    parser.add_argument(
+        "-i",
+        "--id",
+        dest="orderID",
+        action="store",
+        type=str,
+        required=True,
+    )
```
DidierRLopes, line 188: Since that's the only parameter, you can add this:
```
if l_args:
       if "-" not in l_args[0]:
                l_args.insert(0, "-i")
```

So that the user can do: `cancel x`

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+    ns_parser = parse_known_args_and_warn(parser, other_args)
+    if not ns_parser:
+        return
+    try:
+        request = orders.OrderCancel(accountID, ns_parser.orderID)
+        client.request(request)
```
DidierRLopes, line 194: Can you add a `print("")` at end.

Also, when it fails, it should fail just with a simple error message to the user.

Captura de ecrã 2021-04-18, às 21 12 38

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+    except V20Error as e:
+        print(e)
```
DidierRLopes, line 196: Instead: `print(e, "\n")`

gamestonk_terminal/forex/fx_controller.py
```diff
@@ -0,0 +1,177 @@
...
+    @staticmethod
+    def print_help():
+        """Print help"""
+
+        print("\nForex Mode:")
+        print("    help          Show this menu again")
+        print("    q             Quit this menu and goes back to main menu")
+        print("    quit          Quit to abandon program")
+        print("")
+        print("    cancel        Cancel a pending order by ID -i order ID")
```
DidierRLopes, line 54: here you don't need to follow alphabetical order.

You can re-structure this by categories so that is easier for the user to understand the organization. Up to you entirely :)

gamestonk_terminal/forex/fx_controller.py
```diff
@@ -0,0 +1,177 @@
...
+        print("    candles       Show candles")
```
DidierRLopes, line 55: I get this with `candles`:

```
(✨) (fx)> candles
{"errorMessage":"Invalid value specified for 'instrument'"}
```

and this is because I haven't loaded an instrument. So I think you should hide the commands that require an instrument. Similarly to how the main menu is working now, where it doesn't show Technical Analysis for instance because user hasn't load ticker.

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+        df.columns = ["Open", "High", "Low", "Close", "Volume"]
+        if gtff.USE_ION:
+            plt.ion()
+        mpf.plot(
+            df,
+            type="candle",
+            style="charles",
+            volume=True,
+            title=f"{instrument} {ns_parser.granularity}",
+        )
```
DidierRLopes, line 361: It should fail more gracefully than:
```
(✨) (fx)> candles
{"errorMessage":"Invalid value specified for 'instrument'"}
```

by printing the error message and a new line.

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+                else:
+                    print(f"Actual: {actual}")
+            if "previous" in response[i]:
+                previous = response[i]["previous"]
+                unit = response[i]["unit"]
+                if unit != "Index":
+                    print(f"Previous: {previous}{unit}")
+                else:
+                    print(f"Previous: {previous}")
+            print("-" * 30)
```
DidierRLopes, line 471: print("") at the end.

I'll stop commenting this, but ideally all commands end up with a new line :)

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+    )
+    parser.add_argument(
+        "-u",
+        "--units",
+        dest="units",
+        action="store",
+        type=str,
+        required=False,
+    )
+    ns_parser = parse_known_args_and_warn(parser, other_args)
```
DidierRLopes, line 299: Here again, I would allow the user to do:
`close 3` by adding that code snipped that I mentioned before.

I'll stop commenting this forward, I'll leave it up to your judgment to add it. This is just to improve user experience.

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+    if not ns_parser:
+        return
+    data = {
+        "units": ns_parser.units,
+    }
+    try:
+        request = trades.TradeClose(accountID, ns_parser.orderID, data)
+        response = client.request(request)
+        print(response)
```
DidierRLopes, line 308: Instead of just printing response, you should check if there's an error, there, if there is just print the error string. This applies to all commands

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+        )
+        print(df)
+    except V20Error as e:
+        print(e)
+
+
+def close_trade(accountID, other_args: List[str]):
+    parser = argparse.ArgumentParser(
+        add_help=False,
+        prog="close_trade",
```
DidierRLopes, line 280: It looks weird, but in all commands `prog=` should have the name of the terminal option rather than the function, so that it appears correctly in `usage: `

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+        print(f"Margin Available:  ${margin_available}")
+        print(f"Margin Used: ${margin_used}")
+        print(f"Margin Closeout {margin_closeout}")
+        print(f"Margin Closeout Percent: {margin_closeout_percent}")
+        print(f"Margin Closeout Position Value: {margin_closeout_position_value}")
+    except V20Error as e:
+        print(e)
+
+
+def list_orders(accountID, other_args: List[str]):
```
DidierRLopes, line 72: This one crashed badly for me:

```
  File "/Users/DidierRodriguesLopes/Documents/git/GamestonkTerminal/gamestonk_terminal/forex/fx_controller.py", line 83, in switch
    )(other_args)
  File "/Users/DidierRodriguesLopes/Documents/git/GamestonkTerminal/gamestonk_terminal/forex/fx_controller.py", line 118, in call_list
    fx_view.list_orders(account, other_args)
  File "/Users/DidierRodriguesLopes/Documents/git/GamestonkTerminal/gamestonk_terminal/forex/fx_view.py", line 96, in list_orders
    df = df[["id", "instrument", "units", "price", "state", "type"]]
  File "/Users/DidierRodriguesLopes/anaconda3/envs/gst/lib/python3.6/site-packages/pandas/core/frame.py", line 2912, in __getitem__
    indexer = self.loc._get_listlike_indexer(key, axis=1, raise_missing=True)[1]
  File "/Users/DidierRodriguesLopes/anaconda3/envs/gst/lib/python3.6/site-packages/pandas/core/indexing.py", line 1254, in _get_listlike_indexer
    self._validate_read_indexer(keyarr, indexer, axis, raise_missing=raise_missing)
  File "/Users/DidierRodriguesLopes/anaconda3/envs/gst/lib/python3.6/site-packages/pandas/core/indexing.py", line 1298, in _validate_read_indexer
    raise KeyError(f"None of [{key}] are in the [{axis_name}]")
KeyError: "None of [Index(['id', 'instrument', 'units', 'price', 'state', 'type'], dtype='object')] are in the [columns]"
```

gamestonk_terminal/forex/fx_controller.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @staticmethod
+    def print_help():
+        """Print help"""
+
+        print("\nForex Mode:")
+        print("    help          Show this menu again")
+        print("    q             Quit this menu and goes back to main menu")
+        print("    quit          Quit to abandon program")
+        print("")
```
DidierRLopes, line 53: I would re-structure a bit this with the loaded instrument, doing something like:

```
print("    quit          Quit to abandon program")
print("")
print("Instrument: EUR/USD")
print("")
print("    load   ...")
```

gamestonk_terminal/forex/fx_view.py
```diff
@@ -0,0 +1,544 @@
...
+            print(f"Total Short P/L: {short_pl}")
+            print(f"Short Unrealized P/L: {short_upl}")
+            print("-" * 30 + "\n")
+    except V20Error as e:
+        print(e)
+
+
+def get_open_trades(accountID):
+    try:
+        request = trades.OpenTrades(accountID)
```
DidierRLopes, line 250: This one also crashed badly for me. 

Even if there are no arguments, it would be good to still use argparse to add a definition to the command so that the user can do:
```
trades -h
```
and see "shows open trades" or something like that

DidierRLopes (changes requested):
Overall, great work! Sorry for the late review, only now got around signing up for an account to test this!

I've got some minor points that I would like for this to be addressed before merging:

* Check for wrong api keys what happens. We have to try to ensure a clean experience to every user
* argparse for all commands so that the users can do `command -h`
* the `prog=` should have the name of the command rather than the name of the function
* When there is an error or so, don't print the dictionary, but print the important error message in string format to the user
* Always add a new line after a command, regardless of an error or not. To keep consistency across.
* In some commands it makes sense to allow user to do `load EUR/USD` instead of `load -i EUR/USD`. The snippet for this goes attached to my review
* Refactor fx_controller print help so that it makes more sense from a user perspective. E.g. adding the loaded instrument, and only printing some of the commands that rely on an instrument being loaded previously

Sorry for extensive review, but this will help users to have a smoother experience across the terminal! Great work!!

alokan:
Okay I think I got everything. V20Error isn't subscriptable, so I had to slice the string to get the error message out. Some of the longer ones I didn't encounter may still need to be done. Let me know if you find any or if I missed anything.

DidierRLopes:
> Okay I think I got everything. V20Error isn't subscriptable, so I had to slice the string to get the error message out. Some of the longer ones I didn't encounter may still need to be done. Let me know if you find any or if I missed anything.

Nice, thank you!!

Do you mind if I do a PR to your branch with the latest changes? Probably is easier that way

alokan:
No, I don't mind.

DidierRLopes:
Nice, thank you!

See https://github.com/alokan/GamestonkTerminal/pull/22

aia (approved):
Looks ready

alokan:
Okay I added a README, and a function to change the seperator. It takes an argument to what you want it changed to because stocktwits bullbear doesn't return anything for EUR_USD but does for EUR/USD. Also most posts on reddit have a / as the currency separators. Those aren't in this PR but will be when I get time to test what other parts of the terminal can be integrated and which formats return the most information. I also changed the default count for list_orders to 20 so it doesn't overrun the screen with past orders, added a check for impact from a key error that was raised when calling the calendar to see past events.

README.md
```diff
@@ -258,13 +258,9 @@ These are the ones where a key is necessary:
   * FRED: https://fred.stlouisfed.org/docs/api/api_key.html
   * News API: https://newsapi.org
   * Tradier: https://developer.tradier.com/getting_started
+  * Oanda API: https://developer.oanda.com
 
-When these are obtained, don't forget to update [config_terminal.py](/gamestonk_terminal/config_terminal.py).
```
DidierRLopes, line 262: I think you can leave this comment.

gamestonk_terminal/forex/README.md
```diff
@@ -0,0 +1,146 @@
...
+ * Display the positionbook if Oanda provides one for the forex instrument
+
+## summary <a name ="summary"></a>
+ ```text
+ usage: summary
+ ```
+
+ Displays current balance, net asset value, unrealized P/L, total P/L, count of open trades, how much margin is available, how much margin you've used, the margin call value, the percent of margin until closeout, and the total position value.
+
+## calendar <a name="calenar"></a>
```
DidierRLopes, line 41: *calendar

The link is broken because there's a typo

gamestonk_terminal/forex/README.md
```diff
@@ -0,0 +1,146 @@
...
+* [candles](#candles)
+	* Get a candlestick chart for the forex instrument
+* [price](#price)
+	* Show the current price for the forex instrument
+* [order](#order)
+	* Place a limit order
+* [orderbook](#orderbook)
+	* Display the orderbook if Oanda provides one for the forex instrument
+* [positionbook](#positionbook)
+ * Display the positionbook if Oanda provides one for the forex instrument
```
DidierRLopes, line 32: indentation

gamestonk_terminal/config_terminal.py
```diff
@@ -36,5 +36,9 @@
 RH_USERNAME = os.getenv("GT_RH_USERNAME") or "REPLACE_ME"
 RH_PASSWORD = os.getenv("GT_RH_PASSWORD") or "REPLACE_ME"
 
+# Oanda
```
DidierRLopes, line 39: Instead, can you write: https://developer.oanda.com

DidierRLopes (commented):
Things are looking good! This is the last round of review.

Btw, would you be happy if someone asked for a feature regarding Forex for us to ping you so that you keep maintaining it?

alokan:
> Btw, would you be happy if someone asked for a feature regarding Forex for us to ping you so that you keep maintaining it?

Yeah that would be great. I still plan on adding a few things. I just wanted to get your opinions on what was there so far and what needed to be changed.

DidierRLopes:
Can you resolve the conflicts before merging?

DidierRLopes (approved):

DidierRLopes:
Amazing work!! A lot of people will appreciate your effort with this. Thank you!

PR #363, opened 2021-04-18
Add dark pools (ATS) vs Non-ATS data over time

DidierRLopes:
dp

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -79,6 +79,7 @@
 ---
 
 ## Due Diligence
+* [x] Add dark pools (ATS) vs Non-ATS data over time (@didier) - [PR #363](https://github.com/DidierRLopes/GamestonkTerminal/pull/363)
 
 **NEXT**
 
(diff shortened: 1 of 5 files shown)


jmaslek:
That looks really awesome!  Still think the whole idea seems shady lol.

Curious how hard it is to switch the loaded stock from the dd menu (probably outside the scope of this PR).  I was trying to just quickly test out different stocks in that menu, but need to q first before.

jmaslek (approved):

DidierRLopes:
> That looks really awesome!  Still think the whole idea seems shady lol.
> 
> 
> 
> Curious how hard it is to switch the loaded stock from the dd menu (probably outside the scope of this PR).  I was trying to just quickly test out different stocks in that menu, but need to q first before.

Yea, I was thinking into adding a similar command to discovery.

That extracted all this data for last week and outputted the top 10 stocks with most ATS dat being traded. The issue here is that the data is delayed, but may give insight that something is happening behind the scenes.

Need to explore further :)

PR #365, opened 2021-04-18
improvements and bugfixes

DidierRLopes:
This PR includes:
* fix bug for yahoo finance tickers from other markets (e.g. SBIN.NS) 
* Add discord server to readme
* Option README improved

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ Distributed under the MIT License. See [LICENSE](https://github.com/DidierRLopes
 
 [James Maslek](https://www.linkedin.com/in/james-maslek-b6810186/) - jmaslek11@gmail.com
 
-Feel free to share porn loss, memes or any questions at https://discord.gg/FmKjHKMn
+Feel free to share loss porn, memes or any questions at https://discord.gg/FmKjHKMn
 
 ## Acknowledgments
 
--- a/README.md
+++ b/README.md
@@ -363,6 +363,9 @@ Distributed under the MIT License. See [LICENSE](https://github.com/DidierRLopes
 [Artem Veremy](https://www.linkedin.com/in/veremey/) - artem@veremey.net
 
 [James Maslek](https://www.linkedin.com/in/james-maslek-b6810186/) - jmaslek11@gmail.com
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -363,6 +363,9 @@ Distributed under the MIT License. See [LICENSE](https://github.com/DidierRLopes
 [Artem Veremy](https://www.linkedin.com/in/veremey/) - artem@veremey.net
 
 [James Maslek](https://www.linkedin.com/in/james-maslek-b6810186/) - jmaslek11@gmail.com
+
+Feel free to share porn loss, memes or any questions at https://discord.gg/FmKjHKMn
```
jmaslek, line 367: *loss porn

terminal.py
```diff
@@ -196,26 +196,51 @@ def main():
...
+                s_interval,
+            )
 
         elif ns_known_args.opt == "ca":
             b_quit = ca_controller.menu(df_stock, s_ticker, s_start, s_interval)
 
         elif ns_known_args.opt == "fa":
-            b_quit = fam.fa_menu(s_ticker, s_start, s_interval)
+            b_quit = fam.fa_menu(
+                s_ticker.split(".")[0] if "." in s_ticker else s_ticker,
```
jmaslek, line 215: So this could cause an issue if stock.XYX has stock in ( for example ) both NYSE and .NS ?

jmaslek (approved):
Looks good. I'm curious if there is an issue like I mentioned where if a symbol is on many exchanges does that mess with FA/ba menu searches?

aia (approved):

README.md
DidierRLopes: Ups

terminal.py
DidierRLopes: This is not really because a ticker is an exchange other than NYSE, NASDAQ. But because the way you load them using yahoo finance is by adding that .NS. 

Then when you are in one of this menus, if you search for google mentions of a ticker, instead of searching for SBIN you were searching for SBIN.NS. So, there was nothing really being found.

PR #367, opened 2021-04-19
Add vega, gamma and theta to option chains display

pchaganti:
Nice to have vega and gamma aslo displayed in the options chain, as these greeks are quite important when making choices.

### DIFF
--- a/gamestonk_terminal/options/chains_view.py
+++ b/gamestonk_terminal/options/chains_view.py
@@ -129,7 +129,9 @@ def display_chains(symbol: str, expiry: str, other_args: List[str]):
 
     default_columns = [
         "mid_iv",
+        "vega",
         "delta",
+        "gamma",
         "volume",
         "open_interest",
         "bid",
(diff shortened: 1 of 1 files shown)


jmaslek:
At least in my terminal, that wraps the table into multiple lines, which is why I kept the columns as is.  Ill leave the decision to @DidierRLopes 

Better option would be a flag for which greeks you want.

pchaganti:
True. Would be really nice to have the info at hand, but I understand.

thx!

DidierRLopes:
I agree in having those by default.

My terminal still has a lot of space, so that depends on the terminal font the user is using.

I agree with @jmaslek that would be good to allow user to configure this, but as default I would add the greeks.

Captura de ecrã 2021-04-19, às 18 43 39

pchaganti:
vega_gamma

jmaslek:
I added the flag option in #370 .

If you add those, then why not just add theta as well?

pchaganti:
Should I add theta to the existing PR?

DidierRLopes:
Nice work! Thank you!

PR #370, opened 2021-04-19
add option to configure options that get shown

jmaslek:
This adds the flag in `chains` that allows user to specify what they want to view.


gamestonk_terminal/options/chains_view.py
```diff
@@ -124,25 +136,26 @@ def display_chains(symbol: str, expiry: str, other_args: List[str]):
         default=-1,
         help="maximum strike price to consider.",
     )
+    parser.add_argument(
+        "-d",
+        "--display",
+        dest="to_display",
+        default=default_columns,
```
DidierRLopes, line 143: Instead, I would add here this:
```
type= check_valid_option_chains_headers
```
and define above:
```
def check_valid_option_chains_headers(headers: str) -> str:
    
    columns = [str(item) for item in headers.split(",")]
    
    for header in columns:
        if header not in df_columns:
            raise argparse.ArgumentTypeError("Invalid option chains header selected!")
    
    return columns
```

gamestonk_terminal/options/chains_view.py
```diff
@@ -124,25 +136,26 @@ def display_chains(symbol: str, expiry: str, other_args: List[str]):
...
-    chains_df = chains_df[default_columns].rename(columns=column_map)
-
     try:
         ns_parser = parse_known_args_and_warn(parser, other_args)
         if not ns_parser:
             return
+        if isinstance(ns_parser.to_display, str):
+            columns = list(ns_parser.to_display.split(",")) + ["strike", "option_type"]
+        if isinstance(ns_parser.to_display, list):
+            columns = ns_parser.to_display + ["strike", "option_type"]
```
DidierRLopes, line 156: This wouldn't be needed, as the ns_parser.to_display would contain the `List[str]` regardless

gamestonk_terminal/options/chains_view.py
```diff
@@ -124,25 +136,26 @@ def display_chains(symbol: str, expiry: str, other_args: List[str]):
         default=-1,
         help="maximum strike price to consider.",
     )
+    parser.add_argument(
+        "-d",
+        "--display",
+        dest="to_display",
+        default=default_columns,
+        help="columns to look at",
```
DidierRLopes, line 144: In here I would add the columns that the user can select!

gamestonk_terminal/options/README.md
```diff
@@ -84,6 +84,17 @@ Display options chains. [Source: Tradier]
 * --puts : Flag to show puts only
 * -m : Minimum strike price to consider.
 * -M : Maximum strike price to consider.
+* -d       : Columns to display.  Should be comma separated.  Must be in `['bid', 'ask', 'strike', 'bidsize', 'asksize', 'volume', 'open_interest',
```
DidierRLopes, line 87: Instead, can we have:

```
* -d : Columns to display from the following selection: bid, ask, strike, bidsize, asksize, volume, open_interest, delta, gamma, theta, vega, ask_iv, bid_iv, mid_iv. Default: mid_iv,vega_theta_gamma,delta,volume,open_interest,bid,ask. Note that they are separated by comma.
```

gamestonk_terminal/options/README.md
```diff
@@ -84,6 +84,17 @@ Display options chains. [Source: Tradier]
...
+        'delta', 'gamma', 'theta', 'vega', 'ask_iv', 'bid_iv','mid_iv']`
 
 
 <img width="948" alt="chains" src="https://user-images.githubusercontent.com/25267873/115161876-f708ff80-a097-11eb-8073-195979862a45.png">
+
+To change what columns are displayed, one can use the `-d` flag.  To show the same as above, but only the ask, ask size
```
DidierRLopes, line 93: I don't think we need this information to follow remaining README.

I think the idea is good, but if we were to add all different options for all different commands, I think these would become quite big and the idea is that the user scrolls through and gets an idea of what he can from the terminal.

gamestonk_terminal/options/chains_view.py
jmaslek: This makes mypy mad since headers is a List ??

DidierRLopes: It should be like this:
```
def check_valid_option_chains_headers(headers: str) -> List[str]:
```

DidierRLopes (approved):
Perfect!! Thank you!

PR #371, opened 2021-04-19
Build updates

aia:


jmaslek (approved):

terminal.py
```diff
@@ -95,6 +95,10 @@ def main():
     menu_parser.add_argument("opt", choices=choices)
     completer = NestedCompleter.from_nested_dict({c: None for c in choices})
 
+    if os.name == "nt":
```
DidierRLopes, line 98: @aia a discord user had his windows failing here, see:

Captura de ecrã 2021-04-23, às 21 41 36

aia: That's probably an issue with the terminal that they are using.

I don't have any issues on Windows 10 and CI does not have any issues either.

Issue #368, opened 2021-04-19
[FR]Implied Volatility rank and percentile


pchaganti:
**What's the problem of not having this feature?**
One of the things that is really nice in the better options apps is the ability to get IV rank and percentile info for tickers. This is quite useful to see how current IV for options compares to historical volatility.

**Describe the solution you would like**
Maybe an additional menu option to display this info for a loaded ticker if we are in the options view?

**Describe alternatives you've considered**

**Additional information**
This info is not available from RH or Yahoo Finance. I am not sure about Tradier or other sources. 
One example of this info (SPY) - https://www.alphaquery.com/stock/SPY/volatility-option-statistics/30-day/iv-mean

thanks!

jmaslek:
How does #375 address this?

pchaganti:
Works great! Looks awesome!

thanks!

<img width="416" alt="ivrank" src="https://user-images.githubusercontent.com/6500/115484243-76293f80-a220-11eb-83dc-22995b197223.png">

jmaslek:
No problem!  Thanks for the suggestion!

As an aside was the selenium part straightforward?

pchaganti:
Quite simple.

- pip install selenium
- download chrome driver from google https://sites.google.com/a/chromium.org/chromedriver/downloads
- put it in the PATH and update config_terminal.py

PR #375, opened 2021-04-20
Add in scraping of options info (IV Rank/ Hist IV)

jmaslek:
Trying to address #368 , this PR scrapes Barnhart.com for the options info.

For some reason, requests does not work, so I added selenium to the requirements/poetry.

Selenium is a little more confusing, but I added the link where the driver can be downloaded for either chrome or firefox.

@aia I tried adding a unittest to make sure the issue with cvxpy doesn't happen, but that just messes with the workflows so I will eliminate it on the next commit.


gamestonk_terminal/options/op_scrape_view.py
```diff
@@ -0,0 +1,102 @@
...
+        default=None,
+    )
+
+    try:
+        ns_parser = parse_known_args_and_warn(parser, other_args)
+        if not ns_parser:
+            return
+        if ns_parser.browser:
+            browser = ns_parser.browser
+        else:
```
DidierRLopes, line 59: Instead of this, you can add web as default of ns_parsers browser argument.

DidierRLopes (approved):
This looks good to me.

My only concern is: Is it it worth the effort given the information provided?

Regardless, I think that we could merge it, and if we found a better source for this info, we could probably update to use such source. Right?

jmaslek:
I'll leave that to you.  Selenium is a good package but the fact that the outside driver is needed may make it complicated.

DidierRLopes:
> I'll leave that to you. Selenium is a good package but the fact that the outside driver is needed may make it complicated.

I'm happy with it :) Let's go for it! If we stumble across something better in the future we'll think about it, but this is good! Cheers :) 

Feel free to merge. Before, can you just update:
* gamestonk_terminal/README with this new feature
* ROADMAP.md with this PR feature

PR #387, opened 2021-04-23
Stocktwits, reddit, and eda integration and added ta options to charts.

alokan:


DidierRLopes (commented):
I've done several posts on the state of the terminal, so we are seeing a huge influx of users currently. Hence will probably only have a look at this Sunday or so.

But from my quick scroll this is impressive!!! What a great contribution!!

gamestonk_terminal/forex/fx_view.py
```diff
@@ -228,7 +229,7 @@ def create_order(accountID, instrument, other_args: List[str]):
         "--price",
         dest="price",
         action="store",
-        type=float,
+        type=str,
```
DidierRLopes, line 232: you probably should check that the string parsed can be converted to a valid price.

I would personally keep the float and even use the method check_non_negative, and then when you use it instead of using it with `ns_parser.price` you can do `str(ns_parser.price)` if you want the string.

That way you ensure the user is providing a valid number

gamestonk_terminal/forex/fx_controller.py
```diff
@@ -152,6 +185,43 @@ def call_calendar(self, other_args: List[str]):
...
+        """Call bullbear fom stocktwits"""
+        instrument = fx_view.format_instrument(self.instrument, " ")
+        stocktwits_view.bullbear(other_args, instrument)
+
+    def call_messages(self, other_args):
+        """Call messages from stocktwits"""
+        instrument = fx_view.format_instrument(self.instrument, " ")
+        stocktwits_view.messages(other_args, instrument)
+
+    def call_edasummary(self, other_args: List[str]):
```
DidierRLopes, line 204: Instead of having 1 command for each of the EDA commands.

What do you think about having a menu that would dive into EDA?

The sequence of actions would be:
* You add the `> eda` sub-menu
* Then when user does `eda` you can do the following:
```
def call_eda(self, _):
        """Call the EDA menu"""
        eda_controller.menu(df_stock, ticker, s_start, s_interval)
```
* The only thing you need to take care of is the variables:
    * df_stock
    * ticker
    * s_start
    * s_interval
But this way you'd get the full menu in 1 simple call!

This is just an idea! Let me know what you think!

gamestonk_terminal/forex/README.md
```diff
@@ -108,13 +128,21 @@ Specify the instrument to use in the format EUR_USD
...
 * -c : Specify the number of candles to retrieve. (default=180)
+* -a : Adds AD (Accumulation/Distribution Index) to the chart. AD is a cumulative indicator that uses volume and price to assess whether a stock is being accumulated or distributed. This provides insight into how strong a trend is. [Investopedia](#https://www.investopedia.com/terms/a/accumulationdistribution.asp)
+* -b : Adds Bollinger Bands to the chart. Bollinger bands are a set of trendlines plotted two standard deviations (positively and negatively) away from a simple moving average of a security's price. [Investopedia](#https://www.investopedia.com/terms/b/bollingerbands.asp)
+* -C : Adds CCI (Commodity Channel Index) to the chart. CCI is a momentum-based oscillator used to help determine when an investment vehicle is reaching a condition of being overbought or oversold. [Investopedia](#https://www.investopedia.com/terms/c/commoditychannelindex.asp)
+* -e : Adds EMA (Exponential Moving Average) to the chart. EMA is a moving average that places a greater weight and significance on the most recent data points. [Investopedia](#https://www.investopedia.com/terms/e/ema.asp)
```
DidierRLopes, line 141: Since this is a flag, you can add in here what is the default moving average window that is used. Same for the others technical indicators that use pre-defined parameters.

gamestonk_terminal/forex/fx_view.py
alokan: I changed it to a string when I encountered a price precision error with some floats due to them being stored with trailing nines. It changed 1.20849 to something like 1.208499999999. I checked a few more times after changing it and it didn't seem to happen with strings. Once it's sent to Oanda they check that it's a valid price and return a V20Error if it's invalid, so I think having an extra check before sending it is unnecessary since the feedback to the user would happen at roughly the same time.

gamestonk_terminal/forex/fx_controller.py
alokan: This is a great idea! I didn't think to do that.

gamestonk_terminal/forex/fx_view.py
DidierRLopes: You could potentially use 
```
str(round(ns_parser.price, 2))
```

to round with 2 decimal places.

The reason why it works with strings is because the user will not add a lot of decimal places. However, in this situation, the main problem is that the user can use `--price HELLO` and this will break because we didn't check that the value provided was a number, if it makes sense.

gamestonk_terminal/forex/fx_controller.py
```diff
@@ -191,44 +179,27 @@ def call_news(self, other_args: List[str]):
         """Call news [News API]"""
         news_view.news(other_args, self.instrument)
 
-    def call_bullbear(self, other_args):
+    def call_bullbear(self, other_args: List[str]):
```
DidierRLopes, line 182: Would it be possible to do the same as you did for "EDA" with "BA" ?

gamestonk_terminal/forex/fx_controller.py
```diff
@@ -94,16 +90,8 @@ def print_help(self):
...
             print(
-                "    edasummary    brief summary statistics using exploratory data analysis"
-            )
-            print("    edarolling    rolling mean and std deviation")
-            print(
-                "    edadecompose  decomposition in cyclic-trend, season, and residuals"
-            )
-            print(
-                "    edacusum      detects abrupt changes using cumulative sum algorithm"
+                "    eda           exploratory data analysis,	 e.g.: decompose, cusum, residuals analysis"
```
DidierRLopes, line 93: Could you add a `>`?

This is to signal the user that this is a menu command :)

gamestonk_terminal/forex/fx_view.py
alokan: It just printed the invalid price error without crashing with regular strings like hello, I went ahead and changed it to use floats and round though.

gamestonk_terminal/forex/fx_controller.py
alokan: I left news and reddit since they're from dd. Most of the dd menu doesn't work with forex, so it won't integrate the same way.

gamestonk_terminal/forex/fx_view.py
```diff
@@ -229,14 +229,19 @@ def create_order(accountID, instrument, other_args: List[str]):
         "--price",
         dest="price",
         action="store",
-        type=str,
+        type=float,
```
DidierRLopes: Instead, can you do:
```
type=check_non_negative
```

And after we can merge it!!

PR #388, opened 2021-04-23
Improve install

DidierRLopes:


jmaslek (approved):

terminal.py
```diff
@@ -92,9 +92,12 @@ def main():
     menu_parser.add_argument("opt", choices=choices)
     completer = NestedCompleter.from_nested_dict({c: None for c in choices})
 
-    if os.name == "nt":
-        sys.stdin.reconfigure(encoding="utf-8")
-        sys.stdout.reconfigure(encoding="utf-8")
+    try:
```
DidierRLopes, line 95: @aia is it ok to add a "try-catch" here ?

At least 6 people now had to comment these lines to have the terminal working. Not sure what else we can do

aia:
People who have UTF-8 reconfigure failing will run into other issues. Pretty sure they need to use a better terminal on Windows 10. E.g. https://eugeny.github.io/terminus/.

```
            sys.stdin.reconfigure(encoding="utf-8")
            sys.stdout.reconfigure(encoding="utf-8")
```

aia (approved):

terminal.py
aia: Try is fine here. Especially if a lot of people are complaining.

aia: Windows and UTF-8 behavior is... interesting.

DidierRLopes:
> People who have UTF-8 reconfigure failing will run into other issues. Pretty sure they need to use a better terminal on Windows 10. E.g. https://eugeny.github.io/terminus/.
> 
> ```
> sys.stdin.reconfigure(encoding="utf-8")
> sys.stdout.reconfigure(encoding="utf-8")
> ```

I'll let them know about using a better terminal. Thanks!

aia:
I use Windows 10 every day and CI is working fine with this too - https://github.com/DidierRLopes/GamestonkTerminal/runs/2423128794?check_suite_focus=true. I added these reconfigures because CI was failing.

DidierRLopes:
> I use Windows 10 every day and CI is working fine with this too - https://github.com/DidierRLopes/GamestonkTerminal/runs/2423128794?check_suite_focus=true. I added these reconfigures because CI was failing.

Yea, that's why I we were a bit unsure about just commenting the lines since if you added them it wasn't randomly.

The users I told to comment those 3 lines, haven't come back to me with more issues since I told them that. So it may be that somehow for them it just works

PR #396, opened 2021-04-25
Easy open for Gamestonk Terminal following install

RowanHarley:
Changed readme to include optional process to speed up the terminal opening in the future, by using a simple batch file

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,15 +156,7 @@ python terminal.py
 ```
 10. (Optional) Speeding up opening process in the future
 
-To speed up the opening process, open up a notepad, and input this code:
-```
-call <Anaconda Dir>/Scripts/activate.bat
-call conda activate gst
-cd "<Gamestonk Dir>\GamestonkTerminal"
-call python terminal.py
-pause
-```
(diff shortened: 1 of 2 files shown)


DidierRLopes:
Hey @RowanHarley , thanks for this.

What do you think about adding the batch file template as well? I think the majority of users aren't devs, and not sure they would be able to set this batch file easily.

RowanHarley:
Good idea @DidierRLopes . Should I add it as a seperate file inside this repo?

DidierRLopes:
> Good idea @DidierRLopes . Should I add it as a seperate file inside this repo?

Yes, I'd say so. Outside the gamestonk_terminal folder :)

README.md
```diff
@@ -154,6 +154,9 @@ This is a library for package management, and ensures a smoother experience than
 ```
 python terminal.py
 ```
+10. (Optional) Speeding up opening process in the future
```
DidierRLopes, line 157: May be worth pointing out that this is for Windows only. Correct?

RowanHarley: Yes, good point. I'll change the readme to reflect that, and if I get the chance, I'll try to create a similar program for Mac OS to speed up the opening process

DidierRLopes: Nice sounds good!!

DidierRLopes (approved):

aia (approved):

PR #397, opened 2021-04-25
Improve user experience

DidierRLopes:
As requested by users, this PR does 2 things:
- Adds a feature flag to clear command history
- Adds a config of the port for the jupyter notebooks

As @aia would say, these are examples of bikeshedding :)


aia:
These are definitely useful changes.

aia (approved):

Issue #402, opened 2021-04-26
[Bug] Options max pain


spartan737:
GME max pain for 30/04/2021 is $420, but from SwaggyStocks, its only $155. TSLA is $800 here, $720 in SwaggyStocks. NIO max pain is almost similar to SwaggyStocks ($37.5 vs $38), SPY very similar ($410 vs $415). Please check it out. Thanks

jmaslek:
Closed #407

PR #407, opened 2021-04-27
Fix max pain calculation

jmaslek:
As per #402, this addresses the incorrect calculation

I think I also addressed a bug in disc->Fidelity when the top orders table is missing entries (at least in the price change column).  I raised #408 with this.


jmaslek:
@aia @DidierRLopes this failed test also appears in the disc -> orders

aia (approved):

Issue #373, opened 2021-04-20
[IMPROVE] Add Binance API access to price feed


partoneoftwo:
[IMPROVE] As a user it would be valuable to access the data available through Binance API.

kierzenka:
What data specifically do you think would be useful that isn't there?

jmaslek:
Closed #428

PR #428, opened 2021-05-02
Add binance

jmaslek:
Addressing #373 I added Binance API for coins.

Updated crypto menu and added another gecko command.


README.md
```diff
@@ -227,10 +227,14 @@ These are the ones where a key is necessary:
   * Tradier: https://developer.tradier.com/getting_started
   * Oanda API: https://developer.oanda.com
   * CoinMarketCap API: https://coinmarketcap.com/api/
+  * Binance: https://binance.us
 
 When these are obtained, don't forget to update [config_terminal.py](/gamestonk_terminal/config_terminal.py).
 
-Alternatively, you can also set them to the following environment variables: GT_API_KEY_ALPHAVANTAGE, GT_API_KEY_FINANCIALMODELINGPREP, GT_API_KEY_QUANDL, GT_API_REDDIT_CLIENT_ID, GT_API_REDDIT_CLIENT_SECRET, GT_API_REDDIT_USERNAME, GT_API_REDDIT_USER_AGENT, GT_API_REDDIT_PASSWORD, GT_API_TWITTER_KEY, GT_API_TWITTER_SECRET_KEY, GT_API_TWITTER_BEARER_TOKEN, GT_API_POLYGON_KEY, GT_FRED_API_KEY, GT_API_NEWS_TOKEN, GT_OANDA_TOKEN, GT_OANDA_ACCOUNT, GT_TRADIER_TOKEN, GT_CMC_API_KEY.
+Alternatively, you can also set them to the following environment variables: GT_API_KEY_ALPHAVANTAGE, GT_API_KEY_FINANCIALMODELINGPREP,
```
DidierRLopes, line 234: I think the formatting is weird:

Captura de ecrã 2021-05-02, às 13 23 20

README.md
```diff
@@ -227,10 +227,14 @@ These are the ones where a key is necessary:
...
+  * Binance: https://binance.us
 
 When these are obtained, don't forget to update [config_terminal.py](/gamestonk_terminal/config_terminal.py).
 
-Alternatively, you can also set them to the following environment variables: GT_API_KEY_ALPHAVANTAGE, GT_API_KEY_FINANCIALMODELINGPREP, GT_API_KEY_QUANDL, GT_API_REDDIT_CLIENT_ID, GT_API_REDDIT_CLIENT_SECRET, GT_API_REDDIT_USERNAME, GT_API_REDDIT_USER_AGENT, GT_API_REDDIT_PASSWORD, GT_API_TWITTER_KEY, GT_API_TWITTER_SECRET_KEY, GT_API_TWITTER_BEARER_TOKEN, GT_API_POLYGON_KEY, GT_FRED_API_KEY, GT_API_NEWS_TOKEN, GT_OANDA_TOKEN, GT_OANDA_ACCOUNT, GT_TRADIER_TOKEN, GT_CMC_API_KEY.
+Alternatively, you can also set them to the following environment variables: GT_API_KEY_ALPHAVANTAGE, GT_API_KEY_FINANCIALMODELINGPREP, 
+GT_API_KEY_QUANDL, GT_API_REDDIT_CLIENT_ID, GT_API_REDDIT_CLIENT_SECRET, GT_API_REDDIT_USERNAME, GT_API_REDDIT_USER_AGENT, GT_API_REDDIT_PASSWORD, 
+GT_API_TWITTER_KEY, GT_API_TWITTER_SECRET_KEY, GT_API_TWITTER_BEARER_TOKEN, GT_API_POLYGON_KEY, GT_FRED_API_KEY, 
+GT_API_NEWS_TOKEN, GT_OANDA_TOKEN, GT_OANDA_ACCOUNT, GT_TRADIER_TOKEN, GT_CMC_API_KEY, GT_BINANCE_API_KEY, GT_BINANCE_SECRET_KEY.
```
DidierRLopes, line 237: Could we update all of these to contain "API" at start.
E.g.
GT_CMC_API_KEY -> GT_API_COINMARKETCAP_KEY
GT_BINANCE_API_KEY - > GT_API_BINANCE_API
GT_BINANCE_SECRET_KEY -> GT_API_BINANCE_SECRET
GT_TRADIER_TOKEN -> GT_API_TRADIER_TOKEN

To follow: GT_API + source + token/secret/username/password

Also, I've noticed that we have:
```
COINMARKETCAP_KEY = os.getenv("GT_CMC_API_KEY") 
``
ideally the names should be identical with the difference being on the "GT_" at the start (which tbh I'm not sure what it stands for ahah)

File: ROADMAP.md [hunk 3]
```diff
@@ -201,7 +201,7 @@ ___
 **NEXT**
 * [ ] Add [Coinpaprika](https://coinpaprika.com/api/)
 * [ ] Add crypto charts
-* [ ] Add binance 
+* [x] Add binance (@jmaslek)
```
DidierRLopes, line 204: You should move this to the part above (to the "done" area) and add your PR number

gamestonk_terminal/README.md
```diff
@@ -274,9 +274,14 @@ Command|Description
 
 Command|Description
 ------ | ------------
-`load`| load cryptocurrency data
-`view`| load and view cryptocurrency data
+`load`| load cryptocurrency data from coingecko
```
DidierRLopes, line 277: Probably a 3 column table with:
Command Description Source would be ideal in this situation!

gamestonk_terminal/README.md
```diff
@@ -274,9 +274,14 @@ Command|Description
...
+`view`| view loaded cryptocurrency data
+`trend`| view top 7 coins from coingecko
 `top` | view top coins from coinmarketcap
+`add` | add binance ticker
```
DidierRLopes, line 281: I don't really like the "add" keyword here since we are not adding to anything, but selecting it.

Instead, could we have:
```
`select` | select coin-currency pair
```

gamestonk_terminal/README.md
DidierRLopes: Or even have a new line to distinguish between provider.

Similar to how behavioural analysis commands have been done

gamestonk_terminal/cryptocurrency/README.md
```diff
@@ -6,9 +6,21 @@ This menu aims to explore crypto world, and the usage of the following commands
   * load a given coin vs a given currency [CoinGecko]
 * [view](#view)
   * plot the loaded crypto data
+* [trend](#trend)
+  * show the top 7 coins from coingecko
 * [top](#top)
   * view top coins from coinmarketcap [coinmarketcap.com] 
-
+* [add](#add)
```
DidierRLopes, line 13: Again, probably the keyword select is better here.

gamestonk_terminal/cryptocurrency/README.md
```diff
@@ -6,9 +6,21 @@ This menu aims to explore crypto world, and the usage of the following commands
   * load a given coin vs a given currency [CoinGecko]
 * [view](#view)
   * plot the loaded crypto data
+* [trend](#trend)
```
DidierRLopes, line 9: Could we here distinguish between data providers?

Similarly to Behaviour Analysis README

gamestonk_terminal/cryptocurrency/README.md
```diff
@@ -35,6 +47,12 @@ Plot the loaded crypto data.
 
 ![crypto_view](https://user-images.githubusercontent.com/25267873/115787452-20889a80-a3ba-11eb-9216-f7fd1ffc98cf.png)
 
+## trend  <a name="trend"></a>
+Print the top 7 trending coins from coingecko
```
DidierRLopes, line 51: The order is reverse.

The usage comes first, and then the description

gamestonk_terminal/cryptocurrency/README.md
```diff
@@ -50,3 +68,39 @@ This command displays the top n cryptocurrencies from coinmarketcap.com.
 
 <img width="990" alt="crypto" src="https://user-images.githubusercontent.com/25267873/115787544-4746d100-a3ba-11eb-9433-b7cb9142404a.png">
 
+## add  <a name="add"></a>
```
DidierRLopes, line 71: I will refactor this part, so don't worry with it

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,253 @@
...
+    try:
+        client.get_avg_price(symbol=symbol.upper())
+        return symbol.upper()
+    except BinanceAPIException as e:
+        raise argparse.ArgumentTypeError(
+            f"{symbol} is not a valid binance symbol"
+        ) from e
+
+
+# pylint: disable=inconsistent-return-statements
```
DidierRLopes, line 27: Shouldn't this line be immediately above the function definition header?

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def add_binance_coin(other_args: List[str]):
+    """
```
DidierRLopes, line 31: I think this should be:
```
"""Define coin-currency pair from binance

Parameters
```

Note the spacing and that the description starts on the first line.

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,253 @@
...
+    except BinanceAPIException as e:
+        raise argparse.ArgumentTypeError(
+            f"{symbol} is not a valid binance symbol"
+        ) from e
+
+
+# pylint: disable=inconsistent-return-statements
+
+
+def add_binance_coin(other_args: List[str]):
```
DidierRLopes, line 30: You're missing the typing to return a str

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,253 @@
...
+    other_args: List[str]
+        Argparse arguments
+
+    Returns
+    -------
+    coin: str
+        Coin that is defined on binance
+    """
+    parser = argparse.ArgumentParser(
+        prog="add_coin",
```
DidierRLopes, line 44: this prog should contain name of function you're calling, in this case should be "add". Ideally, I think should be "select"

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -67,23 +94,52 @@ def call_quit(self, _):
...
+            other_args
+        )
+        print(f"{self.current_coin} loaded vs {self.current_currency}")
+        print("")
+
 
 def menu():
     crypto_controller = CryptoController()
-    crypto_controller.print_help(crypto_controller.current_coin)
+    crypto_controller.print_help(crypto_controller)
```
DidierRLopes, line 142: This is redundant, when you do this you are already inside crypto_controller object, so you have access to all variables that have been declared there.

What you should do is just a `print_help(None)`

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -7,36 +8,62 @@
...
 
         self.crypto_parser = argparse.ArgumentParser(add_help=False, prog="crypto")
         self.crypto_parser.add_argument("cmd", choices=self.CHOICES)
         self.current_coin = None
+        self.current_currency = None
         self.current_df = pd.DataFrame()
 
     @staticmethod
-    def print_help(current_coin):
+    def print_help(cont):
```
DidierRLopes, line 45: Here you shouldn't expect any outside argument. Instead, you want to use arguments from the class you are in. So this should be "self"

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -7,36 +8,62 @@
...
         """Print help"""
         print("\nCryptocurrency:")
         print("   help          show this menu again")
         print("   q             quit this menu, and shows back to main menu")
         print("   quit          quit to abandon program")
-        print(f"\nCurrent Coin: {current_coin}")
+        print(f"\nCurrent Coin: {cont.current_coin}")
```
DidierRLopes, line 51: self instead of cont

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -7,36 +8,62 @@
...
         print("")
+        print("Coingecko:")
```
DidierRLopes, line 53: This division is perfect! Nice

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -7,36 +8,62 @@
...
         print("   load          load cryptocurrency data")
         print("   view          load and view cryptocurrency data")
+        print("   trend         show top 7 trending coins")
         print("")
+        print("CoinMarketCap:")
         print("   top           view top coins from coinmarketcap")
         print("")
+        print("Binance:")
+        print("   add           add coin as current")
```
DidierRLopes, line 62: One thing is that this description "add coin as current" needs to be the same in FEATURES markdown, and also in table of contents of crypo README.

Also, this command should be "select" or so, and description can be improved

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -67,23 +94,52 @@ def call_quit(self, _):
         return True
 
     def call_load(self, other_args):
+        """Load command - get df from gecko"""
```
DidierRLopes, line 97: Just a "process load command" is sufficient

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -67,23 +94,52 @@ def call_quit(self, _):
...
+        """View command - plot loaded coin"""
         if self.current_coin:
             pycoingecko_view.view(self.current_coin, self.current_df, other_args)
 
         else:
             print("No coin selected. Use 'load' to load the coin you want to look at.")
             print("")
 
+    def call_trend(self, _):
+        """Trend command - show top 7 gecko coins"""
```
DidierRLopes, line 110: Just a "process trend command" is ideal. To keep convention across.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -67,23 +94,52 @@ def call_quit(self, _):
...
+    def call_balance(self, _):
+        """Balance command - check current holdings of coin in binance"""
+        binance_model.balance(self.current_coin)
+
+    def call_add(self, other_args):
+        """Add command - set current coin for binance operations"""
+        self.current_coin, self.current_currency = binance_model.add_binance_coin(
+            other_args
+        )
+        print(f"{self.current_coin} loaded vs {self.current_currency}")
```
DidierRLopes, line 136: I think this should be done in the add_binance_coin and not controller. The controller should be just a middleman let's say.

Also, if it fails loading, it shouldn't print anything. Otherwise we get something like:
```
None loaded vs None
```

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,253 @@
...
+    """
+    client = Client(BINANCE_API_KEY, BINANCE_SECRET_KEY)
+    try:
+        current_balance = client.get_asset_balance(asset=coin)
+        if current_balance is None:
+            print("Check loaded coin")
+            return
+        print("")
+        print(
+            f"You currently have {current_balance['free']} free and {current_balance['locked']} locked"
```
DidierRLopes, line 246: Can we round this somehow? 
```
You currently have 0.00000000 free and 0.00000000 locked
```
In case of bitcoin for instance we would want more decimal places. We can do something smart like multiply the coins by the current price, and see how many decimal places we need so that the value is under 1. That is the number of decimal places we would round to. What you think?

Also, I recommend something like:
```
FREE       : 15  (75%)
LOCKED : 5    (25%)
```
It's more nice visually, and you can add that percentage

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,253 @@
...
+        )
+        print("")
+        return
+    except Exception as e:
+        print(e)
+        print("")
```
DidierRLopes, line 252: I've been using lately:
```
print(e, "\n")
```

I think it's prettier. But this is minor

DidierRLopes (commented):
I've added a ton of comments, but all minor stuff.

Overall really good features, and crypto fans will be stoked by this!

Love the trend and book features!!

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: So `self` raises a pylint error (not sure why here and not in others) (Something about self in static method).  Since that doesn't assign anything to the object, that is just accepting the object (which is a controller object).

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: see above

gamestonk_terminal/cryptocurrency/binance_model.py
jmaslek: that didn't type set well, but I could do a tabulate table (also honestly I don't know how much someone would have locked...should be small right?

gamestonk_terminal/cryptocurrency/crypto_controller.py
DidierRLopes: I think you need to add @staticmethod above or so, there are other controllers that use members of the class, you can check those. E.g. research controller.

Regarding your comment that's true, it's just a parameter, but "_self is a parameter in function and the user can use a different parameter name in place of it. Although it is advisable to use self because it increases the readability of code._" https://www.edureka.co/blog/self-in-python/

gamestonk_terminal/cryptocurrency/binance_model.py
DidierRLopes: I'm not sure what this lock even means to be fully honest. 

I stacked DOT the other day, not sure if that's considered "locked". But in my case I stacked all my coins, so if that's considered "locked" then I think it can be a big amount.

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: It is wrapped as a static method.  And I pretty much copied it over so I am unsure why self raises an error in this case.

DidierRLopes: Let me see if I can figure it out. 

Do you remember the warning you were getting?

jmaslek: The error is `W0211 Static method with 'self' as first argument (bad-staticmethod-argument)`

jmaslek: Can it just be a non static method?

DidierRLopes: Got it.

You're trying to do this:
```
crypto_controller.print_help(crypto_controller)
```
when you should do this:
```
crypto_controller.call_help(None)
```

gamestonk_terminal/cryptocurrency/binance_model.py
jmaslek: So "locked" may be staked (is that what you mean by stacked...)?

DidierRLopes: Yes, it is indeed!

jmaslek:
not done yet -- ide issues

gamestonk_terminal/cryptocurrency/binance_model.py
jmaslek: So I haven't changed anything for rounding.  As you say it depends on the coin price, but I think crypto is just inherently messy.  Honestly this is also a command that is probably run very rarely (since I am not doing prices -- which I could) so you probably want to know your entire holdings when you run it.  (Should I add price?)

jmaslek:
I addressed everything except the rounding of `balance` (comment above).  Let me know if I goofed anything up on the merge.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -56,7 +82,7 @@ def switch(self, an_input: str):
 
     def call_help(self, _):
         """Process Help command"""
-        self.print_help(self.current_coin)
+        self.print_help()
```
DidierRLopes, line 85: I think as an argument this should take `self`

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,258 @@
...
+            print("Check loaded coin")
+            return
+        print("")
+        amounts = [float(current_balance["free"]), float(current_balance["locked"])]
+        total = np.sum(amounts)
+        df = pd.DataFrame(amounts)
+        df.columns = ["Amount"]
+        df.index = ["Free", "Locked"]
+        df["Percent"] = df.div(df.sum(axis=0), axis=1).round(3)
+        print(f"You currently have {total} coins and the breakdown is:")
```
DidierRLopes, line 252: You can use this trick: `str(float( number ))`

Captura de ecrã 2021-05-03, às 10 08 33

gamestonk_terminal/cryptocurrency/binance_model.py
```diff
@@ -0,0 +1,258 @@
...
+        print(tabulate(df, headers=df.columns, showindex=True, tablefmt="fancy_grid"))
```
DidierRLopes, line 253: Since you use a dataframe, you can use `.apply(lambda x: str(float(x))`

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: Since I dropped `@staticmethod`, `print_help` is now just a class method.

DidierRLopes: Yea, I think you're right!

DidierRLopes:
Let's gooo!

PR #431, opened 2021-05-03
Add ratings over time to due diligence menu

DidierRLopes:
Ratings over time command added to Due Diligence.

rot


jmaslek (approved):
So shouldn't finnhub_view just have the plot, and a new finnhub_model have the data acquisition?

DidierRLopes:
> So shouldn't finnhub_view just have the plot, and a new finnhub_model have the data acquisition?

I think so yes. But the finnhub_view.py has a single function that takes care of the data acquisition. I think for now that's fine, at least that's how 90% of code is.

PR #446, opened 2021-05-16
Government menu to also have house and senate trading

DidierRLopes:
This PR does:
* Fix absolute path to access screeners. This way the user can run terminal from anywhere, as presets folder is assessed relatively.
* Add Senate and House trading, and not only Congress.
* Add possibility to saw last days of transactions for a particular ticker for senate, house, and both (congress).
* Fix view command for subsequent tickers

### DIFF
--- a/gamestonk_terminal/README.md
+++ b/gamestonk_terminal/README.md
@@ -455,8 +455,16 @@ Command|Website
 `last_congress`         |last congress trading
 `buy_congress`          |top buy congress tickers
 `sell_congress`         |top sell congress tickers
+`last_senate`           |last senate trading
+`buy_senate`            |top buy senate tickers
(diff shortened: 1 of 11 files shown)


aia (approved):

jmaslek:
This looks really good.  My only comment would be if you wanna change the axis on the buy/sell plot to maybe be either 100's of thousands of dollars or millions (depending on what the max is). Other than that, looks great to me!

jmaslek (approved):

DidierRLopes:
> This looks really good. My only comment would be if you wanna change the axis on the buy/sell plot to maybe be either 100's of thousands of dollars or millions (depending on what the max is). Other than that, looks great to me!

Good point, let me add this :)

PR #452, opened 2021-05-20
Add option to go to ta menu from crypto

jmaslek:
Add the ability to load crypto data frame and use it with the ta menu.

Note that Binance is better for this since it gives candles, so coin gecko data will only work on indicators that use close data only

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -101,9 +101,10 @@
 * [x] Add several exponential smoothing techniques (@didier) - [PR #132](https://github.com/DidierRLopes/GamestonkTerminal/pull/132)
 * [x] Allow backtesting (@didier) - [PR #169](https://github.com/DidierRLopes/GamestonkTerminal/pull/169)
 * [x] Add averaging around multiple predictions (@didier) - [PR #252](https://github.com/DidierRLopes/GamestonkTerminal/pull/252)
+* [x] CNN 1D (@jmaslek) - [PR #436](https://github.com/GamestonkTerminal/GamestonkTerminal/pull/436)
+
 
 **NEXT**
-* [ ] CNN 1D (@didier)
 * [ ] Combine Sentiment Analysis with Chart data - [IS #240](https://github.com/DidierRLopes/GamestonkTerminal/issues/240)
 * [ ] Cross-Validation methods (Forward Chaining, K-Fold, Group K-Fold) (@didier)
 
@@ -204,10 +205,11 @@ ___
 * [x] Add Coingecko (@jmaslek) - [PR #283](https://github.com/DidierRLopes/GamestonkTerminal/pull/283)
 * [x] view top coins from coinmarketcap (@jmaslek) - [PR #378](https://github.com/DidierRLopes/GamestonkTerminal/pull/378)
 * [x] Add binance (@jmaslek) [PR #428](#https://github.com/DidierRLopes/GamestonkTerminal/pull/428)
+* [x] Add crypto ta (@jmaslek)
+
 
 **NEXT**
 * [ ] Add [Coinpaprika](https://coinpaprika.com/api/)
-* [ ] Add crypto charts
 * [ ] Add top altcoin lists
 ---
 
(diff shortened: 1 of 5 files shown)


ROADMAP.md
```diff
@@ -204,10 +205,11 @@ ___
 * [x] Add Coingecko (@jmaslek) - [PR #283](https://github.com/DidierRLopes/GamestonkTerminal/pull/283)
 * [x] view top coins from coinmarketcap (@jmaslek) - [PR #378](https://github.com/DidierRLopes/GamestonkTerminal/pull/378)
 * [x] Add binance (@jmaslek) [PR #428](#https://github.com/DidierRLopes/GamestonkTerminal/pull/428)
+* [x] Add crypto ta (@jmaslek)
```
DidierRLopes, line 208: You can add the PR number already if you want

gamestonk_terminal/cryptocurrency/README.md
```diff
@@ -24,18 +27,22 @@ This menu aims to explore crypto world, and the usage of the following commands
 * [balance](#balance)
   * show coin balance
 
+[MENUS](#Menu)
+* [> ta](technical_analysis/README.md)
+  * **technical analysis  for selected tickers**
```
DidierRLopes, line 32: double space here

gamestonk_terminal/cryptocurrency/README.md
```diff
@@ -102,13 +110,11 @@ Gets and shows the order book for the given coin/currency.  Shows the cumulative
 
 ## candle  <a name="candle"></a>
 ````
-usage: candle [-i --interval] [-l --limit]
+usage: candle
```
DidierRLopes, line 113: Was the removal of the interval and limit intentional?

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -117,21 +123,64 @@ def call_book(self, other_args):
...
+    # pylint: disable=inconsistent-return-statements
+    def call_ta(self, _):
+        """Process ta command"""
+        if not self.current_coin:
+            print("Please load a coin through either load or select", "\n")
+        if self.current_df.empty:
+            print("Price dataframe is empty")
+        else:
+            # Need to make the columns in df be compatible.  Also since there are no splits or dividends, there is no
+            # adj close.  To tell ta to use Close, we can just set the interval to anything but 1440.
```
DidierRLopes, line 155: I don't understand this part.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -117,21 +123,64 @@ def call_book(self, other_args):
...
+                self.current_df = self.current_df[["Price"]].rename(
+                    columns={"Price": "4. close"}
+                )
+                self.current_df.index.name = "date"
+
+            return ta_controller.menu(
+                self.current_df,
+                self.current_coin,
+                self.current_df.index[0],
+                "Hi",
```
DidierRLopes, line 181: This may be due to your comment above, right?

gamestonk_terminal/cryptocurrency/README.md
jmaslek: Yeah.  `select` now does that and candle just plots it.

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: So the TA menu is meant to interface with the `load` command which uses the finance data frame.  If you look in terminal.py, there is this renaming of OHLCV columns to `1. open` etc.  I need to do that so that ta doesn't complain about `[5. adj close]`  not  being found

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: The `hi`?  That can be anything except "1440" so that ta knows to not use adjusted close

PR #463, opened 2021-05-24
fix a mess created by making timeseries-cv pip installable

DidierRLopes:
Should fix the mess that I created by making: https://pypi.org/project/timeseries-cv/


requirements-full.txt
```diff
@@ -84,6 +85,7 @@ jupyter-core==4.7.1; python_full_version >= "3.6.1" and python_version >= "3.6"
 jupyter==1.0.0
 jupyterlab-pygments==0.1.2; python_version >= "3.6"
 jupyterlab-widgets==1.0.0; python_version >= "3.6"
+keras-nightly==2.5.0.dev2021032900
```
jmaslek, line 88: Still not sure what this is.  2.5 matches TF?

jmaslek (approved):

aia (approved):

requirements-full.txt
DidierRLopes: I've just removed it, will merge once checks go through

PR #466, opened 2021-05-26
update incorrect fred key from config file

jmaslek:
Fix #465 where the incorrect variable was being imported

Issue #465, opened 2021-05-26
[Bug] Wrong API key name for FRED on README.md


mchow01:
**Describe the bug**
On `README.md` under the "API Keys" table, the variable name for the Federal Reserve Economic Data (FRED) is incorrect.  The README lists `GT_FRED_API_KEY` but it needs to be `GT_API_FRED_KEY` as seen in `gamestonk_terminal/config_terminal.py`

**To Reproduce**
1. Go to https://github.com/GamestonkTerminal/GamestonkTerminal/
2. Take a look at the table "API Keys" https://github.com/GamestonkTerminal/GamestonkTerminal/#api-keys

**Additional context**
I was wondering why I kept on getting "Bad Request.  The value for variable api_key is not a 32 character alpha-numeric lower-case string.  Read https://research.stlouisfed.org/docs/api/api_key.html for more information. " error when trying to get any data under `econ` menu.  I know my API key is correct and the format of my `.env` file.  I took a look at the `gamestonk_terminal/config_terminal.py` file and sure enough, the variable name for the FRED API did not match up.

jmaslek:
Thanks for the catch!  I am updating this now

Issue #348, opened 2021-04-15
[Bug] pyportfolioopt dependency cvxpy breaks numpy 1.19


sabirjana:
**Describe the bug**
I updated the terminal and it stopped working. 

**To Reproduce**
Steps(from the start) and commands to reproduce the behavior

**Screenshots**
```
'(gst-env)'D:\P4Finance\GamestonkTerminal>python terminal.py
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
Traceback (most recent call last):
  File "terminal.py", line 26, in <module>
    from gamestonk_terminal.comparison_analysis import ca_controller
  File "D:\P4Finance\GamestonkTerminal\gamestonk_terminal\comparison_analysis\ca_controller.py", line 19, in <module>
    from gamestonk_terminal.portfolio_optimization import po_controller
  File "D:\P4Finance\GamestonkTerminal\gamestonk_terminal\portfolio_optimization\po_controller.py", line 12, in <module>
    from gamestonk_terminal.portfolio_optimization import optimizer_view as po_api
  File "D:\P4Finance\GamestonkTerminal\gamestonk_terminal\portfolio_optimization\optimizer_view.py", line 9, in <module>
    from pypfopt import plotting
  File "D:\P4Finance\gst-env\lib\site-packages\pypfopt\__init__.py", line 1, in <module>
    from .black_litterman import (
  File "D:\P4Finance\gst-env\lib\site-packages\pypfopt\black_litterman.py", line 13, in <module>
    from . import base_optimizer
  File "D:\P4Finance\gst-env\lib\site-packages\pypfopt\base_optimizer.py", line 14, in <module>
    import cvxpy as cp
  File "D:\P4Finance\gst-env\lib\site-packages\cvxpy\__init__.py", line 18, in <module>
    from cvxpy.atoms import *
  File "D:\P4Finance\gst-env\lib\site-packages\cvxpy\atoms\__init__.py", line 20, in <module>
    from cvxpy.atoms.geo_mean import geo_mean
  File "D:\P4Finance\gst-env\lib\site-packages\cvxpy\atoms\geo_mean.py", line 20, in <module>
    from cvxpy.utilities.power_tools import (fracify, decompose, approx_error, lower_bound,
  File "D:\P4Finance\gst-env\lib\site-packages\cvxpy\utilities\power_tools.py", line 18, in <module>
    from cvxpy.atoms.affine.reshape import reshape
[...]

PR #467, opened 2021-05-26
Fix FRED API KEY name #465

DidierRLopes:
This should supersede #466

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,9 +87,9 @@ As a modern Python-based environment, GamestonkTerminal opens access to numerous
 
 ## Donation
 
-GST is a completely free open-source terminal. My personal goal has always been to keep this terminal available for everyone to use at zero cost. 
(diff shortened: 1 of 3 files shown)


DidierRLopes:
@aia do you know what's happening here? Me and @jmaslek are struggling to find out why the test fails

aia:
```
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
```

This is your problem. We have seen it before.

You want to figure out the correct version here:
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/.github/workflows/windows.yml#L55

aia:
https://github.com/GamestonkTerminal/GamestonkTerminal/issues/348#issuecomment-822872464

DidierRLopes:
> #348 (comment)

You're right. @jmaslek already solved this. Thanks both of you :)

PR #468, opened 2021-05-27
Minor improvements to ETF menu

jperkins12:
- `overview` command now works with lower case symbols
- `web` command will open to page for specific etf if symbol is provided. Otherwise it will default to https://stockanalysis.com/etf/.


jmaslek:
Thanks! We are working on fixing the build issue so once we get that updated you can merge with that to pass checks.

DidierRLopes:
Thank you @jperkins12 !!

PR #469, opened 2021-05-28
Add menu option 'quote'

jperkins12:
Adds a main menu option `quote` that shows a table of the current price for a stock.

image

### DIFF
--- a/gamestonk_terminal/README.md
+++ b/gamestonk_terminal/README.md
@@ -58,6 +58,12 @@ view -t S_TICKER [-s S_START_DATE] [-i {1,5,15,30,60}] [--type N_TYPE]
 ![GNUS](https://user-images.githubusercontent.com/25267873/108925137-f2920e80-7633-11eb-8274-6e3bb6a19592.png)
 
 ```
+quote -t S_TICKER
+```
+
+* Show the current price of a stock.
+
+```
 candle
 ```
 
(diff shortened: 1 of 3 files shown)


jmaslek:
So I am not sure of @DidierRLopes opinion, but I would do this through yfinance so that this can be accessed without any api keys.

Also the Open/High/Low/Close depends on time interval, so that could be an option in the parser.

jperkins12:
It could be changed to a yfinance call. I chose AV in this case because it had the `ts.get_quote_endpoint()` function that pulled the latest price data.

jperkins12:
Changed the API to finance

jmaslek:
Thanks. I will test this after market open.  

Can you also update the Readme inside /gamestonkterminal ?

jmaslek:
One other comment.  Can you round all the prices to 2 decimals.  I don't recall if finance returns the prices and whatnot as strings or floats, but if you have them all as floats, you can use `floatfmt=".2f"` in tabulate.

jperkins12:
I can absolutely make all the requested changes. As for the theme, that is standard iTerm + some oh-my-zsh goodness. I'm actually on vscode 😛.

jmaslek:
Sweet looks good to me.  

Make sure your pyupgrade is updraded first then to fix the linter, you can run `pyupgrade --py36-plus gamestonkterminal/main_helper.py`  (or whatever the path is)

jmaslek (approved):

DidierRLopes:
Really nice! Good work!!

PR #471, opened 2021-05-29
Make `gst` executable

Chavithra:
Hi, this let run the command `gst` from anywhere in the `virtualenv` instead of having to call `terminal.py`


aia:
Sounds good to me.

aia (approved):

DidierRLopes:
Nice, thanks!

PR #473, opened 2021-05-29
Add Python 3.9 remove Python 3.7

aia:
Speaking about Python versions.

Should we drop python 3.6? What about FBProphet 0.6?


jmaslek:
I still use 3.7 :(

Im also totally for dropping prophet

aia:
Python 3.10 is almost done

jmaslek:
> Python 3.10 is almost done

Definitely gonna change everything to switch cases

PR #474, opened 2021-05-29
Attempting to fix requirements as both full Mac and Windows builds are getting stuck

aia:


aia:
Seems like these modules are broken:

https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/portfolio/degiro_view.py
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/portfolio/degiro_model.py
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/portfolio/degiro_controller.py

Degiro Connector used by these modules is not a proper Python library. A proper python library cannot be exporting a module named trading.
https://github.com/Chavithra/degiro-connector

Probably makes sense to disable these modules for now.

DidierRLopes:
> Seems like these modules are broken:
> 
> https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/portfolio/degiro_view.py
> https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/portfolio/degiro_model.py
> https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/portfolio/degiro_controller.py
> 
> Degiro Connector used by these modules is not a proper Python library. A proper python library cannot be exporting a module named "trading".
> https://github.com/Chavithra/degiro-connector
> 
> Probably makes sense to disable these modules for now.
> 
> ```
> Traceback (most recent call last):
>   File "/home/runner/work/GamestonkTerminal/GamestonkTerminal/terminal.py", line 42, in <module>
>     from gamestonk_terminal.portfolio import port_controller
>   File "/home/runner/work/GamestonkTerminal/GamestonkTerminal/gamestonk_terminal/portfolio/port_controller.py", line 11, in <module>
>     from gamestonk_terminal.portfolio import (
>   File "/home/runner/work/GamestonkTerminal/GamestonkTerminal/gamestonk_terminal/portfolio/degiro_controller.py", line 16, in <module>
>     from gamestonk_terminal.portfolio.degiro_view import DegiroView
>   File "/home/runner/work/GamestonkTerminal/GamestonkTerminal/gamestonk_terminal/portfolio/degiro_view.py", line 4, in <module>
>     import trading.helpers.payload_handler as payload_handler
> ModuleNotFoundError: No module named 'trading.helpers'
> ```

Humm, should we just not import degiro sub-menu in the interim?

@Chavithra can you have a look at this?

aia:
@Chavithra you want to restructure your module so that it can be imported with something like

degiro_connector.trading.helpers.payload_handler

That way, we will not be running into a naming collision with the word "trading" and we will know where the imports are going to. You also want to use "_" rather than "-" in names as python's standards are based around the Snake Case - https://en.wikipedia.org/wiki/Snake_case

Chavithra:
> @Chavithra you want to restructure your module so that it can be imported with something like
> 
> degiro_connector.trading.helpers.payload_handler
> 
> That way, we will not be running into a naming collision with the word "trading" and we will know where the imports are going to. You also want to use "_" rather than "-" in names as python's standards are based around the Snake Case - https://en.wikipedia.org/wiki/Snake_case

@aia Will check what I can do about that naming

PR #491, opened 2021-06-07
Changing NLP from Flair to VADER

jmaslek:
In order to get rid of the torch package in the optional, I switched the NLP sentiment to be VADER instead of flair.  My research seemed to show that VADER actually outperforms some Deep Learning packages for Social Media stuff like tweets.

I added this to the full poetry install, not optional, since it doesn't require any heavy ML packages (tf or torch).

@DidierRLopes I left the READMe images for you :)


gamestonk_terminal/feature_flags.py
```diff
@@ -14,13 +14,10 @@
 USE_ION = strtobool(os.getenv("GTFF_USE_ION", "True"))
 
 # Enable Prompt Toolkit
-USE_PROMPT_TOOLKIT = strtobool(os.getenv("GTFF_USE_PROMPT_TOOLKIT", "False"))
+USE_PROMPT_TOOLKIT = strtobool(os.getenv("GTFF_USE_PROMPT_TOOLKIT", "True"))
```
aia, line 17: Prompt toolkit breaks graphs for me :(
With prompt toolkit graphs get stuck.
We thought this would be fixed with this https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/menu.py#L13, but latest dependencies are not working with this.

jmaslek: I did not intend to push that flag.  Also interesting that my graphs are fine.

jmaslek:
Seems like I forgot to regenerate the full reqs.

gamestonk_terminal/feature_flags.py
jmaslek: > Prompt toolkit breaks graphs for me :(
> With prompt toolkit graphs get stuck.
> We thought this would be fixed with this https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/menu.py#L13, but latest dependencies are not working with this.

So is the default being False good enough or do I need to do something else to fix plots?

aia (approved):

jmaslek:
@aia mypy upgraded to 0.9, so the PRs will fail as is.  Should I pin the mypy==0.812 in tests.yml ?

aia:
I think we should update mypy and fix or disable the new concerns mypy 0.9 discovers

jmaslek:
Sounds good.  From the last action run, an example of the the errors are 

```
gamestonk_terminal/behavioural_analysis/finbrain_view.py:6: error: Library stubs not installed for "requests" (or incompatible with Python 3.9)
62
gamestonk_terminal/behavioural_analysis/finbrain_view.py:6: note: Hint: "python3 -m pip install types-requests"
63
gamestonk_terminal/behavioural_analysis/finbrain_view.py:6: note: (or run "mypy --install-types" to install all missing stub packages)
```

I tried to add --install-types to tests.yml but it exited since it requires a [yN] input to install the types.  Whats the correct way to fix that ?

aia:
https://github.com/GamestonkTerminal/GamestonkTerminal/pull/494

aia:
Should we merge this?

jmaslek:
Will do.  @DidierRLopes feel free to ping me when you get back from the beach if you want any changes

gamestonk_terminal/behavioural_analysis/README.md
```diff
@@ -165,34 +165,51 @@ Print up to the last 30 messages of a user. [Source: Stocktwits]
...
 ```
 usage: sentiment [-n N_NUM] [-d N_DAYS_PAST]
 ```
-Plot in-depth sentiment extracted from tweets from last days that contain pre-defined ticker. This model splits the text into character-level tokens and uses the DistilBERT model to make predictions. DistilBERT is a distilled version of the powerful BERT transformer model. Note that a big num of tweets extracted per hour in conjunction with a high number of days in the past, will make the algorithm take a long period of time to estimate sentiment. Inspired by https://towardsdatascience.com/sentiment-analysis-for-stock-price-prediction-in-python-bed40c65d178. [Source: Twitter]
-  * -n : num of tweets to extract per hour. Default 100.
-  * -d : num of days in the past to extract tweets. Default 7.
-
-<img width="949" alt="Captura de ecrã 2021-02-22, às 00 15 20" src="https://user-images.githubusercontent.com/25267873/108643678-7ff22900-74a3-11eb-9bea-54f84ffe42e1.png">
-
-![nvda](https://user-images.githubusercontent.com/25267873/108643604-2ab61780-74a3-11eb-950b-ba0bbec8293e.png)
```
DidierRLopes, line 188: Why where these plots removed?

The change of the NLP model used shouldn't really influence the plot. Am I right? If so, would be nice to keep them.

gamestonk_terminal/feature_flags.py
```diff
@@ -19,9 +19,6 @@
 # Enable Prediction features
 ENABLE_PREDICT = strtobool(os.getenv("GTFF_ENABLE_PREDICT", "False"))
 
-# Enable FB Prophet
```
DidierRLopes, line 22: Is this because prophet was dropped all along?

gamestonk_terminal/behavioural_analysis/README.md
jmaslek: Well I redid the plots, so while they should be similar, there are some differences I figured you would want them updated.

gamestonk_terminal/helper_funcs.py
```diff
@@ -22,6 +22,23 @@
...
+
+    # Define the function with default arguments
+    def int_range_checker(num) -> int:
+        num = int(num)
+        if num < mini or num > maxi:
+            raise argparse.ArgumentTypeError(f"must be in range [{mini},{maxi}]")
+        return num
+
+    # Return function handle to checking function
+    return int_range_checker
```
DidierRLopes, line 39: seems a bit redundant, it could have been just:
```
num = int(num)
if num < mini or num > maxi:
   raise argparse.ArgumentTypeError(f"must be in range [{mini},{maxi}]")
return num
```

gamestonk_terminal/feature_flags.py
jmaslek: I think I dropped it here based off a conversation in a different thread.

gamestonk_terminal/behavioural_analysis/README.md
DidierRLopes: Got it! With the Neutral? I'll test this at some point!

Nicely done :)

gamestonk_terminal/feature_flags.py
DidierRLopes: Yup, I read that somewhere. Was happy about that, ngl :D

gamestonk_terminal/helper_funcs.py
jmaslek: Thats what I initially thought, but the type has to callable, and doing it the way you suggest would never pass in a "num", no?

DidierRLopes: Hmmm, I think I got it now. Never had seen code like this, but I think it makes sense. Really smart.

Basically you would do this only:
int_range_checker(num) -> int

but since you want to specify the min,max you declare that function a layer above so that you can define these thresholds on each different context.

jmaslek: yeah.  Took me a while on stack exchange to find it lol.

PR #492, opened 2021-06-07
Fix error with intraday plotting

### LINKED ISSUE #487: Cannot set intervals on plots and "view" (crashes)
**Describe the bug**
It crashes when trying to view graphs with time intervals. It seems to be some issue with datetime objects?

**To Reproduce**
when trying: 

`load -t SPY -s 2021-06-01 -i 30`

and then 

`view`

**Screenshots**

Terminal

 - OS: Mac Big Sur (M1)
 - Python version 3.6

jmaslek:
This should fix #487.  Things that were done

- Fix dataframe slicing (this is where the error was)
- Print correct intraday interval to the command line (yf intraday was not redefining s_interval)
- Update the width of `plt.bar` to be the resolution.  This way it doesn't get messy with too little (much) data.
- Added types and docstrings to most functions in main_helper and helper_funcs.

### DIFF
--- a/gamestonk_terminal/helper_funcs.py
+++ b/gamestonk_terminal/helper_funcs.py
@@ -1,18 +1,21 @@
+"""Helper functions"""
+__docformat__ = "numpy"
 import argparse
 from datetime import datetime, timedelta, time as Time
 import os
 import random
 import re
 import sys
+import pandas as pd
 from pytz import timezone
 import iso8601
 import matplotlib
 import matplotlib.pyplot as plt
 from holidays import US as holidaysUS
 from colorama import Fore, Style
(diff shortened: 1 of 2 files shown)


aia (approved):

DidierRLopes:
> This should fix #487. Things that were done
> 
> * Fix dataframe slicing (this is where the error was)
> * Print correct intraday interval to the command line (yf intraday was not redefining s_interval)
> * Update the width of `plt.bar` to be the resolution.  This way it doesn't get messy with too little (much) data.
> * Added types and docstrings to most functions in main_helper and helper_funcs.

Nice cleaning up, thanks!

PR #496, opened 2021-06-11
Add Stockgrid io source

jmaslek:
Added a couple of commands to `>disc` and one to dd using the above mentioned source.

Also updated mypy in poetry so the correct install is made locally on poetry install.


aia (approved):

DidierRLopes (approved):
Amazing!!! Great work!!

Can you add these new commands you've added to the roadmap? Just so that we have a history of features.

Also, I'd rename enhancement to feature size L :)

jmaslek:
Yes I can update the roadmap.  Also this is definitely not a L feature :)

DidierRLopes:
> Yes I can update the roadmap. Also this is definitely not a L feature :)

It has 4 new commands, I'd say that is at least an M feature!

PR #508, opened 2021-06-19
Removing last mentions of fbprophet and updating dependencies 

aia:


DidierRLopes (approved):
Nice!

Thought about this when I removed the mention of prophet from the pred menu line on the main menu, but then forgot about it.

PR #530, opened 2021-07-04
Added code to save exported formats to respective directories under G…

Pranav4399:
Added code to export the files to respective folders under a folder named exports in \GamestonkTerminal\gamestonk_terminal\exports. The structure of exports folder is all follows : 

\gamestonk_terminal\exports
                                   I
                                   I--->csv
                                   I--->json
                                   I--->excel.
Please create these folders if the code is about to be approved

### DIFF
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -633,13 +633,16 @@ def export(other_args: List[str], df_stock):
         return
 
     if ns_parser.s_format == "csv":
-        df_stock.to_csv(ns_parser.s_filename)
+        outPath = os.path.join("r'", os.path.dirname( __file__ ), 'exports', 'csv', ns_parser.s_filename)
+        df_stock.to_csv(outPath)
 
     elif ns_parser.s_format == "json":
-        df_stock.to_json(ns_parser.s_filename)
+        outPath = os.path.join("r'", os.path.dirname( __file__ ), 'exports', 'json', ns_parser.s_filename)
+        df_stock.to_json(outPath)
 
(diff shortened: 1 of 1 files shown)


DidierRLopes:
> Added code to export the files to respective folders under a folder named exports in \GamestonkTerminal\gamestonk_terminal\exports. The structure of exports folder is all follows : 
> 
> 
> 
> \gamestonk_terminal\exports
> 
>                                    I
> 
>                                    I--->csv
> 
>                                    I--->json
> 
>                                    I--->excel.
> 
> Please create these folders if the code is about to be approved
> 
> 
> 
> 

May I suggest a different approach?

Personally for me I don't think it makes sense to split files based on their extension, as that will be easily seen from their extension.

What would be good for us to do is have an "exports" folder on the topdir, and then follow the code architecture within that exports folder. Ie if you are within disc/dd and do export on command "darkshort", the saved file would be saved in:

exports/disc/dd/darkshort_datetime.ext


Also, ideally the export wouldn't be a command but an argument. So one could do "--export csv" or "export json".

What you think?

Pranav4399:
That approach sounds good although  I thought export option is only available for load tickers as of now. I'll see how export works in all the other menus and work on the same again.

DidierRLopes:
> That approach sounds good although I thought export option is only available for load tickers as of now. I'll see how export works in all the other menus and work on the same again.

Yea, most of them don't. I'm adding these slowly as users requests.

Let me know which commands would you like to be able to export and I'll try to work on them!

DidierRLopes:
Closing this based on #535 . Let me know what you think @Pranav4399

PR #533, opened 2021-07-05
Add supply and customer feature

DidierRLopes:
Add suppliers and customers features to DD menu


jmaslek (approved):
Should we have an option to list the company name instead of just ticker?

DidierRLopes:
> Should we have an option to list the company name instead of just ticker?  

I was thinking that we could have even the table displayed in CSI market, based on user selection. It was just a quick win to add this.

jmaslek:
> I was thinking that we could have even the table displayed in CSI market, based on user selection. It was just a quick win to add this.

That was a big table haha.  The CSI market page also gives different categories for stuff (fundamentals it seems), so I would just keep it to what you have (flag for name was just a suggestion).  Once the user has a ticker they can dd it

PR #534, opened 2021-07-06
Small bugs

jmaslek:
Small fixes

- Changed pyproject script so that I can be lazy and just run get (was looking for def main() but was replaced with def terminal() )
- Fixed bt html error.  Error was in bt.get, not sure why now, but I just did a custom download of yf data.


DidierRLopes (approved):
What do you mean by the "just running get"?

jmaslek:
> What do you mean by the "just running get"?

Autocorrected "gst". You can do gst instead of python terminal.py

DidierRLopes:
> > What do you mean by the "just running get"?
> 
> Autocorrected "gst". You can do gst instead of python terminal.py

Got it, yea!

PR #535, opened 2021-07-06
Data export new concept

DidierRLopes:
@jmaslek can you have a look at the concept of exporting. What do you think?

We could add this piece of code to the bottom of all commands, so that all data could be exported from an user perspective.

### DIFF
--- a/gamestonk_terminal/helper_funcs.py
+++ b/gamestonk_terminal/helper_funcs.py
@@ -610,8 +610,20 @@ def find_tickers(submission):
     return l_tickers_found
 
 
-def export_data(export_type, dir_path, func_name, df):
+def export_data(export_type: str, dir_path: str, func_name: str, df: pd.DataFrame):
+    """Export data to a file.
 
+    Parameters
+    ----------
+    export_type : str
+        Type of export between: csv,json,xlsx,xls
+    dir_path : str
+        Path of directory from where this function is called
+    func_name : str
+        Name of the command that invokes this function
+    df : pd.Dataframe
+        Datframe containing the data
+    """
     if export_type:
         export_dir = dir_path.replace("gamestonk_terminal", "exports")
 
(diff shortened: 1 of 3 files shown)


gamestonk_terminal/helper_funcs.py
```diff
@@ -608,3 +608,28 @@ def find_tickers(submission):
             l_tickers_found.append(s_ticker.strip())
 
     return l_tickers_found
+
+
+def export_data(export_type, dir_path, func_name, df):
+
```
jmaslek, line 614: I FINALLY GOT YOU.  WHERE ARE THE DOCSTRINGS

jmaslek (commented):
So I think its a good idea.  We can add as needed.

You should also add "xls" as a choice (I forget the specifics of the differences, but both should point to df.to_excel).  Could also add a .pkl option or an .hdf option, but I feel that those would never get used.

gamestonk_terminal/helper_funcs.py
DidierRLopes: > I FINALLY GOT YOU. WHERE ARE THE DOCSTRINGS

To be fair, this was a concept PR :D

DidierRLopes:
> So I think its a good idea. We can add as needed.
> 
> You should also add "xls" as a choice (I forget the specifics of the differences, but both should point to df.to_excel). Could also add a .pkl option or an .hdf option, but I feel that those would never get used.

Yea, not sure it would be used. I can do the xls yes!

DidierRLopes:
> > So I think its a good idea. We can add as needed.
> > You should also add "xls" as a choice (I forget the specifics of the differences, but both should point to df.to_excel). Could also add a .pkl option or an .hdf option, but I feel that those would never get used.
> 
> Yea, not sure it would be used. I can do the xls yes!

I've tried this and can't use the to_excel() without installing another module. Let's leave it like this for now, and if a user requests xls, we can look into it.

Also, csv can be open as excel file.

jmaslek (approved):

PR #536, opened 2021-07-09
Allow to load data from IEX

DidierRLopes:
I went over the credits of IEX. I don't think it has a great free tier.

But anyway, it has a lot of stuff that we can perhaps integrate at some point. So probably worth having this as a starting point. See https://github.com/iexcloud/pyEX/blob/main/examples/all.ipynb

This PR includes:
  * IEX daily and intradaily data
  * Fix view help
  * Fix load help
  * View function to allow to print raw data
  * Remove export function and refactor it into a flag in view function
  * Add exports/ folder to gitignore
  * Improve FTD command by allowing to select a start/end date + plotting share price

### DIFF
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -548,7 +548,7 @@ def view(other_args: List[str], s_ticker: str, s_start, s_interval, df_stock):
     parser = argparse.ArgumentParser(
         add_help=False,
         prog="view",
-        description="Visualize historical data of a stock. An alpha_vantage key is necessary.",
+        description="Visualize historical data of a stock.",
     )
     if s_ticker:
         parser.add_argument(
@@ -565,7 +565,7 @@ def view(other_args: List[str], s_ticker: str, s_start, s_interval, df_stock):
             "--ticker",
             action="store",
             dest="s_ticker",
-            required=True,
+            required="-h" not in other_args,
             help="Stock ticker",
         )
     parser.add_argument(
@@ -573,7 +573,7 @@ def view(other_args: List[str], s_ticker: str, s_start, s_interval, df_stock):
         "--start",
         type=valid_date,
         dest="s_start_date",
-        default=s_start,
+        default=s_start if s_start else "2019-01-01",
         help="The starting date (format YYYY-MM-DD) of the stock",
     )
     parser.add_argument(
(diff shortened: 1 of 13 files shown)


gamestonk_terminal/main_helper.py
```diff
@@ -194,6 +195,38 @@ def load(other_args: List[str], s_ticker, s_start, s_interval, df_stock):
...
+                        "volume": "Volume",
+                    }
+                )
+
+                df_stock_candidate.sort_index(ascending=True, inplace=True)
+
+                # Slice dataframe from the starting date YYYY-MM-DD selected
+                df_stock_candidate = df_stock_candidate[ns_parser.s_start_date :]
+
+                print(df_stock_candidate.to_string())
```
jmaslek, line 228: Do you want this to print?

jmaslek (approved):

jmaslek (commented):
Also intraday doesn't work  (note `load msft --source iex` works)

Screen Shot 2021-07-09 at 10 28 59 AM

jmaslek (changes requested):
So the intraday option is not available for iex as written.  Not sure if you want to have a flag that says that or add it in (I think the `client.chart` has some extra kwargs for intraday (https://github.com/iexcloud/pyEX/blob/d9c38826ebc5c35ffb80903f006834a5989d70c5/pyEX/stocks/prices.py#L87))

gamestonk_terminal/main_helper.py
DidierRLopes: Nahhh

DidierRLopes:
> So the intraday option is not available for iex as written. Not sure if you want to have a flag that says that or add it in (I think the `client.chart` has some extra kwargs for intraday (https://github.com/iexcloud/pyEX/blob/d9c38826ebc5c35ffb80903f006834a5989d70c5/pyEX/stocks/prices.py#L87))

Yea, the intraday only gives 1m for the day... I'm gonna try to play with that

DidierRLopes:
@jmaslek I've added a couple of things to this PR :) Should be in a good shape now

gamestonk_terminal/main_helper.py
jmaslek: Still does

jmaslek:
After I do `load AAPL -i 1 --source iex` , the `view` is weird

jmaslek (approved):

aia (approved):

PR #537, opened 2021-07-10
Feature/cryptosentiment

JakubPluta:
Hi, it's my proposal for Finbrain Sentiment Analysis for Cryptocurrencies.

I mostly re-used all of the components for Finbrain that are already under `gamestonk_terminal/behavioural_analysis/finbrain_view.py`. The only thing that I created is a new view for crypto and I placed it under `gamestonk_terminal/crypto/finbrain_crypto_view.py` is it the right place for it?  Or you suggest putting it also in `behavioural_analysis/finbrain_view.py`? 

Also one more thing. I manually scraped all possible crypto symbols (there are like ~105 symbols) from the Finbrain website and stored in JSON file. But would be great if   Finbrain can provide to us endpoint with list of all Symbols, then we don't need to read it from file (and probably also the are adding systematically new coins, so would it be nice to have it always updated.

Let me know what you think :)


DidierRLopes (approved):
Absolutely love what you've done here with regards to showing what are the coins available!

And the implementation is exactly how I would have done it. Great work once again!

JakubPluta:
Thanks for review :)

PR #538, opened 2021-07-10
Fix dd warnings debug output

pkeogh:
Due diligence `warnings --debug` output multiplies long-term debt by 5 rather than net income [as specified by Sean Seah](https://www.drwealth.com/gone-fishing-with-buffett-by-sean-seah/comment-page-1/).

The actual test used to generate warnings is correct, so this patch updates debug output to print the same computations.


jmaslek (approved):
Good catch!

DidierRLopes:
Nice one!

PR #547, opened 2021-07-14
fix for typos on the template.ini screener

liunicholas:
just a simple edit of the template.ini screener to remove typos, the poetry.lock file is changed from a "poetry update", is there a way to only merge the template.ini?


aia:
You want to checkout the previous version of poetry.lock or simply copy the current version from the project. After you have the old copy, git commit the old copy.

liunicholas:
I replaced the poetry.lock file so now the only changed file is the template.ini

aia:
Very nice! Thank you.

aia (approved):

PR #548, opened 2021-07-15
Add market overview - type data form WSJ

jmaslek:
Based on a discord request, which was the have some kind of global market menu based on https://www.tradingview.com/markets/ .  

Instead of a new menu, I am just adding functions to the econ menu. This PR scrapes and prints the different data available on wsj.com/market-data.

It also includes movers for ETFs/mutual funds which I included in the etf menu.

Given the architecture change that is going on, I just put all the wsj functions in a 'common' file that lives in /gamestonk_terminal/.    I can change the file structure to be more consistent with what has been done

### DIFF
--- a/gamestonk_terminal/economy/wsj_view.py
+++ b/gamestonk_terminal/economy/wsj_view.py
@@ -1,0 +1,48 @@
+"""WSJ view module"""
+__docformat__ = "numpy"
+
+from tabulate import tabulate
+
+from gamestonk_terminal.wsj_functions import (
+    market_overview,
+    us_bonds,
+    us_indices,
+    global_bonds,
+    global_currencies,
+    top_commodities,
+)
+
+
+def display_wsj(fn: str):
+    """
+    Display dataframe from wsj
+
+    Parameters
+    ----------
+    fn: str
+        What function to show
+    """
+
+    if fn == "market":
+        data = market_overview()
+    elif fn == "us_bonds":
+        data = us_bonds()
+    elif fn == "gl_bonds":
+        data = global_bonds()
+    elif fn == "indices":
+        data = us_indices()
+    elif fn == "currencies":
+        data = global_currencies()
+    elif fn == "commodities":
+        data = top_commodities()
+
+    print(
+        tabulate(
+            data,
+            showindex=False,
+            headers=data.columns,
+            floatfmt=".2f",
+            tablefmt="fancy_grid",
+        )
+    )
+    print("")
--- a/gamestonk_terminal/economy/wsj_view.py
+++ b/gamestonk_terminal/economy/wsj_view.py
@@ -1,9 +1,11 @@
-"""WSJ view module"""
+"""WSJ view """
 __docformat__ = "numpy"
 
+import argparse
+from typing import List
 from tabulate import tabulate
 
-from gamestonk_terminal.wsj_functions import (
+from gamestonk_terminal.economy.wsj_model import (
     market_overview,
     us_bonds,
     us_indices,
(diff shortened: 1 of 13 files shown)


aia (approved):

gamestonk_terminal/economy/wsj_view.py
```diff
@@ -0,0 +1,48 @@
...
+    market_overview,
+    us_bonds,
+    us_indices,
+    global_bonds,
+    global_currencies,
+    top_commodities,
+)
+
+
+def display_wsj(fn: str):
```
DidierRLopes, line 16: Can you change this function to handle each command separately, and add the usual helper stuff. This allows to keep the helper functionality, and have the code more "explicit". Otherwise the way I see this is:
```
market            --\                                 / market
us_bonds        --->     display_wsj  ---- us_bonds
global_bonds  -- /                               \ global_bonds
```

which looks like we're adding complexity. I would only have a display_wsj like that if we got all the data from 1 single function as well.

gamestonk_terminal/etf/wsj_view.py
```diff
@@ -0,0 +1,63 @@
...
+                data.iloc[: ns_parser.num],
+                showindex=False,
+                headers=data.columns,
+                floatfmt=".2f",
+                tablefmt="fancy_grid",
+            )
+        )
+        export_data(
+            ns_parser.export,
+            os.path.dirname(os.path.abspath(__file__)) + "/movers/",
```
DidierRLopes, line 56: why are you adding movers here? This shouldn't be needed

DidierRLopes:
Nitpicky as always! But great functions you added!

gamestonk_terminal/etf/screener_model.py
```diff
@@ -183,7 +183,7 @@ def etf_screener(other_args: List[str]):
 
         export_data(
             ns_parser.export,
-            os.path.dirname(os.path.abspath(__file__)),
+            os.path.dirname(os.path.abspath(__file__)) + "/screeners/",
```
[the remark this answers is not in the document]
jmaslek: Since `>etf` can export screeners or this new stuff. Figured I would break into submenus to keep exports clean

gamestonk_terminal/wsj_functions.py
```diff
@@ -0,0 +1,271 @@
+"""Functions to scrape wsj.com/market-data"""
```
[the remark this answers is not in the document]
jmaslek: I can.  Is the location fine or do you want me to put it into etf and econ menu with the respective functions

gamestonk_terminal/etf/wsj_view.py
jmaslek: same as above. trying to keep exports cleanly labeled

gamestonk_terminal/etf/screener_model.py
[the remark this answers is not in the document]
DidierRLopes: Humm, I would agree with something like that if we added the configuration parameters along with the screener data.

Since we don't provide any additional information, I would keep it.

It shows like this:
```
Saved file: /Users/DidierRodriguesLopes/Documents/git/GamestonkTerminal/exports/etf/screeners/etf_screener_20210715_222100.csv
```

gamestonk_terminal/etf/wsj_view.py
DidierRLopes: I don't think it's needed, but let me know what you think from my comment above.

Also, when doing this sort of operations, @piiq mentioned to me that the preferred way to do it is using os.path.join() which is OS agnostic and doesn't depend on slash inclination

gamestonk_terminal/wsj_functions.py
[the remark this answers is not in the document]
DidierRLopes: Probably yes. Just to keep the current structure. Will make it easier to port functions to ApeCore afterwards

gamestonk_terminal/economy/wsj_view.py
jmaslek: well what helper stuff is there?  All this display_wsj function does is print the tabulate function.  I would just do it in the controller, but that goes against the structure.  So just make 6 individual functions that just load different data?

DidierRLopes: The usual helper stuff is this 
```
parser = argparse.ArgumentParser(
        prog="darkpos",
        add_help=False,
        description="Shows Net Short Vol. vs Position. [Source: Stockgrid]",
    )
```

Yea, I'd say to keep it on wsj_view but have 6 functions for it.

gamestonk_terminal/etf/screener_model.py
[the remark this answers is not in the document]
jmaslek: So if you specify using the flags, all those get added, and your screener file will be a really long yucky thing.  Its just when you provide --config that the save file is not that long (yeah...lazy on my end there since I can just loop through the ini file without having to define anything else)

jmaslek:
Think I addressed everything except the file paths, which I think makes sense to keep different exports sorted.  If you _reaaaalllllyyyy_ feel strongly about it then I can revert it

gamestonk_terminal/etf/wsj_view.py
piiq: to be OS agnostic the path in the `export_data` can be set to

```python
os.path.join(os.path.dirname(os.path.abspath(__file__)), "movers")
```

That way the forward slash will not break windows

piiq (approved):

aia:
Looks like charset-normalizer released a new version.

You can run poetry update and generate new requirements.

That should fix windows build.

jmaslek:
I have no idea what those .idea files are/where they came from

aia:
> 
> 
> I have no idea what those .idea files are/where they came from

Many packages are using charset-normalizer

```
poetry show -t
```

jmaslek:
> > I have no idea what those .idea files are/where they came from
> 
> Many packages are using charset-normalizer
> 
> ```
> poetry show -t
> ```

Do i need to revert back to 205fffb or are those files fine being committed

DidierRLopes:
> I have no idea what those .idea files are/where they came from

Check this: https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder- 

You added a .idea folder. Can you remove that file, and add `.idea` to the `.gitignore` ? This way this type of files will be ignored next time

DidierRLopes (approved):

aia:
https://github.com/GamestonkTerminal/GamestonkTerminal/pull/551
This is the fix for Conda

DidierRLopes:
I think you can merge this and there will be no conflict. Right @aia ? Same for Crypto feature?

jmaslek:
i merged so lets see if it passes

aia:
> 
> 
> I think you can merge this and there will be no conflict. Right @aia ? Same for Crypto feature?

Correct.

PR #554, opened 2021-07-16
This should make so that the project is considered as python only

DidierRLopes:
This should avoid the project being considered as 20% HTML, without having to remove the example files. @aia 

Relied on this: https://stackoverflow.com/questions/34713765/github-changes-repository-to-wrong-language

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* linguist-vendored
+*.py linguist-vendored=false


aia:
Sweet!

aia (approved):

DidierRLopes:
> Sweet!

I had to touch one of the .html to trigger linguistic (https://github.com/GamestonkTerminal/GamestonkTerminal/commit/9287f495ecbd7425ddca5d32107f74a8f0061f18)

but now project shows as 100% python :)

PR #555, opened 2021-07-17
Tidy code

DidierRLopes:
This aims to improve the code to keep consistency across.

Also, it adds the default arguments values on the helper when the user uses `command -h`

### DIFF
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -28,7 +28,6 @@ from gamestonk_terminal.helper_funcs import (
     plot_view_stock,
     parse_known_args_and_warn,
     plot_autoscale,
-    export_data,
 )
 
 from gamestonk_terminal import config_terminal as cfg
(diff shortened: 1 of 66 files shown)


gamestonk_terminal/main_helper.py
```diff
@@ -525,36 +560,38 @@ def quote(other_args: List[str], s_ticker: str):
 
 
 def view(other_args: List[str], s_ticker: str, s_interval, df_stock):
-    """
-    Plot loaded ticker or load ticker and plot
+    """Plot loaded ticker or load ticker and plot
+
     Parameters
     ----------
```
jmaslek, line 566: I actually changed this recently to just be plotting so it should be """Plot loaded ticker

gamestonk_terminal/main_helper.py
```diff
@@ -564,56 +601,6 @@ def view(other_args: List[str], s_ticker: str, s_interval, df_stock):
...
-    parser = argparse.ArgumentParser(
-        add_help=False,
-        prog="export",
-        description="Exports the historical data from this ticker to a file or stdout.",
-    )
-    parser.add_argument(
-        "-f",
-        "--filename",
-        type=str,
-        dest="s_filename",
```
jmaslek, line 577: Are we eliminating the data export and adding it to individual functions?

jmaslek (approved):

gamestonk_terminal/main_helper.py
DidierRLopes: Yup!

PR #557, opened 2021-07-19
Add initial changelogs. This will be used on landing page

DidierRLopes:
@jose-donato there's still a lot of changelogs to be added, but here's a start :)

This should be enough to get started with that section on the landing page.


jose-donato (approved):
All good to go!

PR #558, opened 2021-07-20
Use requests with pd.read_html for gainers/losers discovery

devwolff:
Fix "No table found" - https://github.com/pandas-dev/pandas/issues/6393


DidierRLopes:
Thank you for this! The test is currently failing at:
"gamestonk_terminal/discovery/yahoo_finance_view.py 8: C0411 third party import "import requests" should be placed before "from gamestonk_terminal.helper_funcs import parse_known_args_and_warn" (wrong-import-order)"

Can you fix this so we can push this change?

aia (approved):

PR #560, opened 2021-07-22
Options menu updates

jmaslek:
In this PR I reworked the op menu into different submenus.  I only added 2 new functions.  One is a simple call/put profit calculator, meant to be more educational. The second is historical option prices from tradier.

Marking as draft


DidierRLopes:
Captura de ecrã 2021-07-23, às 22 28 53

To make this plot more complete, what about adding the text part to the plot, in the x and y axis?  Like using plt.xticks([11], ["Breakeven Price"])


Captura de ecrã 2021-07-23, às 22 28 03

Missing a new line here

DidierRLopes:
The first new line when printing `(✨) (op)> info` can be removed

DidierRLopes:
No need to have:
```
aapl loaded 
```
when doing 
```
(✨) (op)> load aapl
```
is kind of redundant

DidierRLopes:
image

I'd put the load straight after the screener

DidierRLopes:
Could we put some sort of color here?
image

DidierRLopes:
So, I wouldn't split Tradier vs Yahoo Finance. I'd do the following:

```
Options:
   cls           clear screen
   ?/help        show this menu again
   q             quit this menu, and shows back to main menu
   quit          quit to abandon program

   disp          display all preset screeners filters
   scr           output screener options

Current Ticker: PLTR
   load          load new ticker
   info          display option information (volatility, IV rank etc)

   calc          basic call/put PnL calculator

Current Expiration: 2021-08-13
   exp          see and set expiration date

   chains     display option chains with greeks
   oi             plot open interest
   vol           plot volume
   voi           plot volume and open interest
   hist          plot option history
```

We can use Tradier as default since it's probably better. And add a `--source` flag to oi, vol and voi, to allow to get data from Yahoo Finance.

What do you think?

jmaslek:
vol
hist
VOI
Screen Shot 2021-07-23 at 11 09 44 PM

DidierRLopes:
```
(✨) (op)> load aapl
(✨) (op)> help
```
after `load` a new line is missing.

Captura de ecrã 2021-07-24, às 23 50 31
what about adding horizontal/vertical lines with some colors and adding a legend to them?

I would also add the Max Loss to the plot. And then no need to print it in the screen.

```
(✨) (op)> exp 13
(✨) (op)> help
```
new line missing from `exp`. Also, could we add a message when getting the expiry date to see what date it corresponds to, or so?

Another nice thing we could do is allow the user to specify directly an expiry date like `2022-09-16`?

```
(✨) (op)> chains
Error in request. Check TRADIER_TOKEN

"None of [Index(['mid_iv', 'vega', 'delta', 'gamma', 'theta', 'volume', 'open_interest',\n       'bid', 'ask', 'strike', 'option_type'],\n      dtype='object')] are in the [columns]" 
```
error message can be improved here

Captura de ecrã 2021-07-24, às 23 54 52

This crashed the terminal badly. Although I agree with tradier being default, we need to revert automatically to yahoo finance if the key isn't provided or not valid.

 In the menu itself we should have:
```
   chains        display option chains with greeks [Tradier]
   oi            plot open interest [Tradier/YahooFinance]
   vol           plot volume [Tradier/YahooFinance]
   voi           plot volume and open interest [Tradier/YahooFinance]
   hist          plot option history [Tradier]
   gr_hist       plot option greek history [Tradier]
```

Captura de ecrã 2021-07-25, às 00 03 37
can the legend be in front of the lines?

Also, what do you think about a -g option "all" that shows stock price and all greeks.

jmaslek:
> can the legend be in front of the lines?
> Also, what do you think about a -g option "all" that shows stock price and all greeks.

Gonna work backwards.  I set the legend to be `loc=0` so it should put it in the best spot....  As for showing all of them, the plot would look too crowded, or very small individual subplots. I think this would be an interesting Jupiter notebook application though

DidierRLopes:
> > can the legend be in front of the lines?
> 
> > Also, what do you think about a -g option "all" that shows stock price and all greeks.
> 
> 
> 
> Gonna work backwards.  I set the legend to be `loc=0` so it should put it in the best spot....  As for showing all of them, the plot would look too crowded, or very small individual subplots. I think this would be an interesting Jupiter notebook application though
> 
> 

I was thinking a subplot 4lines 2 columns, wouldn't look too bad.

Or smth more fancy, like allow thr user to select multiple and the plot adapts to number options selected. Eg if i say iv,gamma. 2 plots side by side, if i say iv,gamma, premium. 3 plots: 2 side by side, and 1 below alone.

jmaslek:
> what about adding horizontal/vertical lines with some colors and adding a legend to them?
> I would also add the Max Loss to the plot. And then no need to print it in the screen.

So looking at this, I actually made some errors that I will fix and update.  I think the console printing is actually better than just a visual.  If the user wants to use actual numbers, they can get a better sense from the actual numbers.

jmaslek:
So I addressed everything except the all greeks.  I don't see exactly what one would gain with that many plots all showing at once, but can explore  it

DidierRLopes (approved):
Really nice!!

Before merging, could you add a newline to `gr_hist`?

PR #567, opened 2021-07-25
Test user internet connection

DidierRLopes:
This is just to avoid weird issues if the user is somehow without internet connection.

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -7,6 +7,7 @@ import os
 import sys
 from datetime import datetime
 from typing import List
+import socket
 import pandas as pd
 import yfinance as yf
 from alpha_vantage.timeseries import TimeSeries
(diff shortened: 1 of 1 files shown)


jmaslek (approved):

howard6606:
I think this broke my terminal. See screen cap 
Screen Shot 2021-07-26 at 9 11 07 AM

DidierRLopes:
> I think this broke my terminal.



> Screen Shot 2021-07-26 at 9 14 24 AM
> I think this broke my terminal.

arghhh. Sorry. I didn't have any issue, neither did @jmaslek . I also use this at work and we haven't run into problems.

I'll just do something like `requests.get('https://www.google.com/').status_code`. Let me fix this now!

Thanks for much for reporting this promptly!

PR #572, opened 2021-07-27
Add fibonacci level indicator. Improve TA menu in general.

jmaslek:
In this PR:

- Added a fibonocci level indicator (under 'custom')
- Redid VWAP to look at daily chart only
- Rewrote ta plots
- Added doctstring/type hinting to ta menus

### DIFF
--- a/gamestonk_terminal/main_helper.py
+++ b/gamestonk_terminal/main_helper.py
@@ -42,7 +42,13 @@ from gamestonk_terminal import thought_of_the_day as thought
 from gamestonk_terminal.technical_analysis import trendline_api as trend
 
 
-def clear(other_args: List[str], s_ticker, s_start, s_interval, df_stock):
+def clear(
+    other_args: List[str],
+    s_ticker: str,
+    s_start: str,
+    s_interval: str,
+    df_stock: pd.DataFrame,
+):
     """Clears loaded stock and returns empty variables
 
     Parameters
@@ -89,7 +95,13 @@ def clear(other_args: List[str], s_ticker, s_start, s_interval, df_stock):
         return s_ticker, s_start, s_interval, df_stock
 
 
-def load(other_args: List[str], s_ticker, s_start, s_interval, df_stock):
+def load(
+    other_args: List[str],
+    s_ticker: str,
+    s_start,
+    s_interval: str,
+    df_stock: pd.DataFrame,
+):
     """Load selected ticker
 
     Parameters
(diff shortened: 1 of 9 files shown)


aia (approved):

gamestonk_terminal/main_helper.py
```diff
@@ -136,7 +136,7 @@ def load(other_args: List[str], s_ticker, s_start, s_interval, df_stock):
         "-s",
         "--start",
         type=valid_date,
-        default="2019-01-01",
+        default="2020-01-01",
```
DidierRLopes, line 139: Should we do like you did on one of the last PRs? 2 years prior to today?'

gamestonk_terminal/technical_analysis/ta_controller.py
```diff
@@ -114,6 +116,8 @@ def print_help(self):
         print("volume:")
         print("   ad          chaikin accumulation/distribution line values")
         print("   obv         on balance volume")
+        print("custom")
+        print("   fib          fibinocci retracement")
```
DidierRLopes, line 120: `custom` missing the `:`

`fibonacci retracement` has an extra space at the start.

gamestonk_terminal/technical_analysis/custom_indicators_view.py
```diff
@@ -0,0 +1,151 @@
...
+        df["Price"] = levels
+        print(
+            tabulate(
+                df,
+                headers=["Fib Level", "Price"],
+                showindex=False,
+                tablefmt="fancy_grid",
+            )
+        )
+        print("")
```
DidierRLopes, line 148: image

Could we add a xlim from start date to end date?

And could we make the fibonacci levels in percentage on the side instead of decimal notation?

DidierRLopes (commented):
Nice! Good work!

DidierRLopes:
We need to start having good PR names so that we can do this after:

https://60f83ea326b8850007ddda1a--gamestonkterminal.netlify.app/changelog/W08_2021

gamestonk_terminal/main_helper.py
jmaslek: lets do a year.  More than that seems overkill for visualizations (even the 1/1/20 seems overkill)

DidierRLopes: Agreed! my suggestion was just to not have a default date, but a default "offset"

DidierRLopes (approved):

PR #576, opened 2021-07-29
Feature/fraud indicators

colin99d:
I am adding a fraud score based using Beneish's mscore. Since this is based on fundamental analysis I have put it under the fa section, and I have put it under the Alpha Vantage section since it utilizes that API. Users can use the -s flag to get a breakdown on the different components of the mscore. Additionally, users can use the -d flag to get details on what the value of the mscore means. Using -s and -d together results in a detailed explanation of the mscore and all components.

### DIFF
--- a/gamestonk_terminal/fundamental_analysis/av_view.py
+++ b/gamestonk_terminal/fundamental_analysis/av_view.py
@@ -518,24 +518,6 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
 
     return df_fa
 
-def clean_mscore_df(df_fa: pd.DataFrame) -> List[int]:
-    """Clean fundamentals dataframe
-
-    Parameters
-    ----------
-    df_fa : pd.DataFrame
-        Fundamentals dataframe
-
-    Returns
-    ----------
-    List[int]
-        Last two years cleaned
-    """
-    # pylint: disable=no-member
-    ints = [int(x) for x in df_fa.to_list()]
-
-    return ints
-
 
 def mscore(other_args: List[str], ticker: str):
     """Mscore for given ticker
@@ -592,20 +574,20 @@ def mscore(other_args: List[str], ticker: str):
         df_bs = df_bs.set_index("fiscalDateEnding").iloc[:2]
         df_is = df_is.set_index("fiscalDateEnding").iloc[:2]
 
-        ar = clean_mscore_df(df_bs["currentNetReceivables"])
-        sales = clean_mscore_df(df_is["totalRevenue"])
-        cogs = clean_mscore_df(df_is["costofGoodsAndServicesSold"])
-        ca = clean_mscore_df(df_bs["totalCurrentAssets"])
-        ppe = clean_mscore_df(df_bs["propertyPlantEquipment"])
-        cash = clean_mscore_df(df_bs["cashAndCashEquivalentsAtCarryingValue"])
-        cash_and_sec = clean_mscore_df(df_bs["cashAndShortTermInvestments"])
+        ar = [int(x) for x in df_bs["currentNetReceivables"].to_list()]
+        sales = [int(x) for x in df_is["totalRevenue"].to_list()]
+        cogs = [int(x) for x in df_is["costofGoodsAndServicesSold"].to_list()]
+        ca = [int(x) for x in df_bs["totalCurrentAssets"].to_list()]
+        ppe = [int(x) for x in df_bs["propertyPlantEquipment"].to_list()]
+        cash = [int(x) for x in df_bs["cashAndCashEquivalentsAtCarryingValue"].to_list()]
+        cash_and_sec = [int(x) for x in df_bs["cashAndShortTermInvestments"].to_list()]
         sec = [y - x for (x, y) in zip(cash, cash_and_sec)]
-        ta = clean_mscore_df(df_bs["totalAssets"])
-        dep = clean_mscore_df(df_bs["accumulatedDepreciationAmortizationPPE"])
-        sga = clean_mscore_df(df_is["sellingGeneralAndAdministrative"])
-        td = clean_mscore_df(df_bs["totalLiabilities"])
-        icfo = clean_mscore_df(df_is["netIncomeFromContinuingOperations"])
-        cfo = clean_mscore_df(df_cf["operatingCashflow"])
+        ta = [int(x) for x in df_bs["totalAssets"].to_list()]
+        dep = [int(x) for x in df_bs["accumulatedDepreciationAmortizationPPE"].to_list()]
+        sga = [int(x) for x in df_is["sellingGeneralAndAdministrative"].to_list()]
+        td = [int(x) for x in df_bs["totalLiabilities"].to_list()]
+        icfo = [int(x) for x in df_is["netIncomeFromContinuingOperations"].to_list()]
+        cfo = [int(x) for x in df_cf["operatingCashflow"].to_list()]
         ratios = {}
         ratios["DSRI"] = {"raw": (ar[0] / sales[0]) / (ar[1] / sales[1]),
                         "description": """Days Sales in Receivables Index gauges whether receivables and revenue are out of balance, a large number is expected to be associated with a higher likelihood that revenues and earnings are overstated."""
@@ -650,7 +632,5 @@ def mscore(other_args: List[str], ticker: str):
                 print(ratios["MSCORE"]["description"])
             print("MSCORE" + ": " + str(ratios["MSCORE"]["raw"]))
 
-
-    except Exception as e:
-        print(e, "\n")
-
+    except ValueError:
+        print("The ticker does not have complete information so a mscore analysis cannot be ran.")
--- a/gamestonk_terminal/fundamental_analysis/av_view.py
+++ b/gamestonk_terminal/fundamental_analysis/av_view.py
@@ -596,8 +596,8 @@ def mscore(other_args: List[str], ticker: str):
         ratios["DSRI"] = {
             "raw": (ar[0] / sales[0]) / (ar[1] / sales[1]),
             "description": (
-                "Days Sales in Receivables Index gauges whether receivables and"
-                "revenue are out of balance, a large number is expected to be associated with a"
+                "Days Sales in Receivables Index gauges whether receivables and "
+                "revenue are out of balance, a large number is expected to be associated with a "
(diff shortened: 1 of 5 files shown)


jmaslek:
Looking forward to reviewing this tomorrow. Make sure to update the readme files both in the FA menu and in the gamestonk_terminal folder

jmaslek:
So the linter is failing on black.  To prevent that, in your terminal you can run black ./ (which will check all files).  You can also install the pre-commit hooks (you can dm us on discord for questions on how to do that) so that you can Check all the linters instead of having to do a bunch of commits

gamestonk_terminal/fundamental_analysis/av_view.py
```diff
@@ -517,3 +517,120 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
...
+            for rkey, value in ratios.items():
+                if ns_parser.description:
+                    print(value["description"])
+                print(rkey + ": " + str(value["raw"]))
+        else:
+            if ns_parser.description:
+                print(ratios["MSCORE"]["description"])
+            print("MSCORE" + ": " + str(ratios["MSCORE"]["raw"]))
+
+    except ValueError:
```
jmaslek, line 635: Can you add a newline (either \n to the string or just `print("")` after everything you print?

jmaslek, line 635: Can you also add the generic: `except Exception as e: print(e,"\n")`  just in case value error isn't what is raised

gamestonk_terminal/fundamental_analysis/av_view.py
```diff
@@ -517,3 +517,120 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
...
+
+        if ns_parser.sub:
+            for rkey, value in ratios.items():
+                if ns_parser.description:
+                    print(value["description"])
+                print(rkey + ": " + str(value["raw"]))
+        else:
+            if ns_parser.description:
+                print(ratios["MSCORE"]["description"])
+            print("MSCORE" + ": " + str(ratios["MSCORE"]["raw"]))
```
jmaslek, line 633: Also can you show less decimals?  (`MSCORE: -2.9346317560474686`) I assume that just rounding to 2 decimals is good.

Edit: Do that for all of them when the `-s` flag is supplied

jmaslek (changes requested):
Good work and thanks for the feature :)

I just left a couple comments to try to keep everything consistent.  @DidierRLopes will probably go through even deeper

colin99d:
All requested changes were made and pre-commit ran perfect!

DidierRLopes:
This is amazing work! 

A few suggestions, I wouldn't have any flag. I would do something like this:

* Have the descriptions on when using the `-h` flag since this is additional information to the command, and should be on the help.

```
(✨) (fa)> fraud -h
usage: cash [-h]

DSRI: Days Sales in Receivables Index gauges whether receivables and revenue are out of balance, a large number is expected to be associated with a higher likelihood that revenues and earnings are overstated. 

GMI: Gross Margin Index shows if gross margins are deteriorating. Research suggests that firms with worsening gross margin are more likely to engage in earnings management, therefore there should be a positive correlation between GMI and probability of earnings management. 

AQI: Asset Quality Index measures the proportion of assets where potential benefit is less certain. A positive relation between AQI and earnings manipulation is expected. 

SGI: Sales Growth Index shows the amount of growth companies are having. Higher growth companies are more likely to commit fraud so there should be a positive relation between SGI and earnings management. 

DEPI: Depreciation Index is the ratio for the rate of depreciation. A DEPI greater than 1 shows that the depreciation rate has slowed and is positively correlated with earnings management. 

SGAI: Sales General and Administrative Expenses Index measures the change in SG&A over sales. There should be a positive relationship between SGAI and earnings management. 

LVGI: Leverage Index represents change in leverage. A LVGI greater than one indicates a lowe change of fraud. 

TATA: Total Accruals to Total Assets is a proxy for the extent that cash underlies earnigns. A higher number is associated with a higher likelihood of earnings manipulation. 

MSCORE: The Beneish model uses financial ratios to check if it is likely that earnings are manipulated. A score of -5 to -2.22 indicated a low chance of fraud, a score of -2.22 to -1.78 indicates a moderate change of fraud, and a score above -1.78 indicated a high chance of fraud. 
```

* Since this has a rather small output, I would output stats by default and separate it from MSCORE a bit. Like shown below,

```
(✨) (fa)> fraud
Stats:
   DSRI: 0.77 
   GMI: 0.98 
   AQI: 1.09 
   SGI: 1.06 
   DEPI: 0.95 
   SGAI: 1.03 
   LVGI: 1.09 
   TATA: -0.07 

MSCORE: -2.93 (low chance of fraud)
```

* Finally, if there's a definition like the one you described to MSCORE, i.e. score between -5 to -2.22 being low chance of fraud, -2.22 to -1.78 moderate change of fraud, and more than -1.78 high chance of fraud. I would add it in brackets after the numeric value, as this information adds a lot of value to the user!

gamestonk_terminal/fundamental_analysis/av_view.py
```diff
@@ -517,3 +517,176 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
...
+                "Sales General and Administrative Expenses Index measures the change "
+                "in SG&A over sales. There should be a positive relationship between SGAI and "
+                "earnings management."
+            ),
+        }
+        ratios["LVGI"] = {
+            "raw": (td[0] / ta[0]) / (td[1] / ta[1]),
+            "description": (
+                "Leverage Index represents change in leverage. A LVGI greater than "
+                "one indicates a lowe change of fraud."
```
DidierRLopes, line 651: lower*

gamestonk_terminal/fundamental_analysis/av_view.py
```diff
@@ -517,3 +517,176 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
...
+    ----------
+    other_args : List[str]
+        argparse other args
+    ticker : str
+        Fundamental analysis ticker symbol
+    """
+    parser = argparse.ArgumentParser(
+        add_help=False,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        prog="cash",
```
DidierRLopes, line 535: the `prog` value needs to represent the command used to call this. I.e. in this case should be `fraud`

jmaslek:
So can you do for the -s flag either 1 of two things:  either eliminate the space in between each metric (just an endlisnen after the last one) OR look into the tabulate we use (would need a dataframe first).  @DidierRLopes may have more of a preference.  
Screen Shot 2021-07-29 at 6 19 26 PM

gamestonk_terminal/fundamental_analysis/av_view.py
```diff
@@ -517,3 +517,176 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
...
+        # pylint: disable=unbalanced-tuple-unpacking
+        # pylint: disable=no-member
+        df_cf, _ = fd.get_cash_flow_annual(symbol=ticker)
+        df_bs, _ = fd.get_balance_sheet_annual(symbol=ticker)
+        df_is, _ = fd.get_income_statement_annual(symbol=ticker)
+        df_cf = df_cf.set_index("fiscalDateEnding").iloc[:2]
+        df_bs = df_bs.set_index("fiscalDateEnding").iloc[:2]
+        df_is = df_is.set_index("fiscalDateEnding").iloc[:2]
+
+        ar = [int(x) for x in df_bs["currentNetReceivables"].to_list()]
```
DidierRLopes, line 577: Instead of converting the array to a list, you could have gathered the array of values directly, like this: `df_bs["currentNetReceivables"].values`

gamestonk_terminal/fundamental_analysis/av_view.py
```diff
@@ -517,3 +517,176 @@ def clean_fundamentals_df(df_fa: pd.DataFrame, num: int) -> pd.DataFrame:
...
+                "fraud, a score of -2.22 to -1.78 indicates a moderate change of fraud, and a "
+                "score above -1.78 indicated a high chance of fraud."
+            ),
+        }
+
+        if ns_parser.sub:
+            for rkey, value in ratios.items():
+                if ns_parser.description:
+                    print(value["description"], "\n")
+                print(rkey + ": " + str(round(value["raw"], 2)), "\n")
```
DidierRLopes, line 685: This works fine, just a suggestion for the future. 

Python 3.x print notation is nicer, you can have something like this
```
print(f"{rkey} : {value["raw"]:.2f }", "\n")
```

DidierRLopes:
> So can you do for the -s flag either 1 of two things: either eliminate the space in between each metric (just an endlisnen after the last one) OR look into the tabulate we use (would need a dataframe first). @DidierRLopes may have more of a preference.
> Screen Shot 2021-07-29 at 6 19 26 PM

Check my suggestion, what you think?

DidierRLopes:
Amazing PR though! I'm usually really thorough on reviews and your code was really nice! Good work!

jmaslek:
> Check my suggestion, what you think?

Maybe color instead?

DidierRLopes:
> > Check my suggestion, what you think?
> 
> Maybe color instead?

Color is good. But having something like "more likelihood of fraud" is better. Since it gives a bit of explanation as well. E.g. if I show you this: "asdasd (means that this stock has been talked a lot)" you know what asdasd means.

Whereas if I show you "asdasd" in green. You associate with something good, yet you don't really know the meaning.

Just an idea :)

gamestonk_terminal/fundamental_analysis/av_view.py
colin99d: Alpha Vantage returns all values as strings and this was the simplest way I could think of to convert them to integers. Using the values method returns them as strings, but if there is an easier way to return as integers let me know.

jmaslek: You can do df[column] = df[column].apply(lambda x: float(x))

colin99d:
Thanks for all the comments so far guys! Its great to get the feedback to help write better code. I think I fixed everything so I just submitted a new commit.

DidierRLopes (approved):
Amazing, really happy about this. Good work. Looking forward for more PRs from you :)

PR #639, opened 2021-08-03
Add unusual options activity + option screener checks

jmaslek:
As the title says, this PR:

- Scrapes https://app.fdscanner.com/unusualvolume for unusual options
- Adds checks to all options variables to warn about bad entries

### DIFF
--- a/gamestonk_terminal/README.md
+++ b/gamestonk_terminal/README.md
@@ -427,6 +427,7 @@ Command|Description|Source
 `load`          | Load new ticker
 `info`          | Display option information | [Barchart](https://barchart.com/)
 `calc`          | Basic option PnL calculator
+`act`           | Scrapes unusual options activity|
 `exp`           | see/set expiry date
 `chains`        | displays option chains    |[Tradier](https://developer.tradier.com/)
 `oi`            | Plot open interest
--- a/gamestonk_terminal/options/README.md
+++ b/gamestonk_terminal/options/README.md
@@ -14,6 +14,8 @@ This menu aims to give insight on options. Options can cause significant share p
   * Display option information [Source: Barchart.com]
 * [calc](#calc)
   * basic call/put PnL calculator
+* [act](#act)
+  * Display unusual options activity
 * [exp](#exp)
   * See/set expiry date
 * [chains](#chains)
@@ -84,6 +86,34 @@ usage: calc [-s STRIKE] [-p PREMIUM] [--put] [--sell]  [-h]
 * --sell: Flag to calculate for selling option
 
 ![calc](https://user-images.githubusercontent.com/18151143/126883982-c54cb5ec-7eb2-41c1-812e-75e830e3703d.png)
+
+## act <a name="act"></a>
+Show unusual options activity from fdscreener.com
+
+```python
+usage: act [-n NUM] [--sortby] [-a] [--export] [-h]
+```
+* -n/--num : Number to get.  Defaults to 20.  Each page scrapes 20 results
+* --sortby : Column to sort data by.  Can be {Option,Vol/OI,Vol,OI,Bid,Ask}.  Defaults to Vol/OI
+* -a : Flag to sort in ascending order
+* --export : Flag to export.  Can be {csv, json, xlsx}
+```python
+(✨) (op)> act -n 5
+Last Updated: 2021-08-03 18:06:07
+╒══════════╤════════════╤══════════╤══════════╤═══════╤══════╤═══════╤═══════╕
+│ Ticker   │ Exp        │ Option   │   Vol/OI │   Vol │   OI │   Bid │   Ask │
+╞══════════╪════════════╪══════════╪══════════╪═══════╪══════╪═══════╪═══════╡
+│ AMD      │ 2021-08-06 │ 112.0 P  │    102.9 │ 15639 │  152 │  1.67 │  1.69 │
+├──────────┼────────────┼──────────┼──────────┼───────┼──────┼───────┼───────┤
+│ MA       │ 2021-11-19 │ 375.0 C  │     50.3 │ 10165 │  202 │ 18.35 │ 18.65 │
+├──────────┼────────────┼──────────┼──────────┼───────┼──────┼───────┼───────┤
+│ GOOG     │ 2021-08-06 │ 2630.0 P │     38.5 │  5314 │  138 │  1.2  │  1.65 │
+├──────────┼────────────┼──────────┼──────────┼───────┼──────┼───────┼───────┤
+│ AMD      │ 2021-08-06 │ 111.0 P  │     37.2 │ 15245 │  410 │  1.29 │  1.31 │
+├──────────┼────────────┼──────────┼──────────┼───────┼──────┼───────┼───────┤
+│ GOOG     │ 2021-08-06 │ 2635.0 P │     34.1 │  5966 │  175 │  1.4  │  1.8  │
+╘══════════╧════════════╧══════════╧══════════╧═══════╧══════╧═══════╧═══════╛
+```
 
 ## exp <a name="exp"></a>
 See/set expiry dates.
--- a/gamestonk_terminal/options/README.md
+++ b/gamestonk_terminal/options/README.md
@@ -14,7 +14,7 @@ This menu aims to give insight on options. Options can cause significant share p
   * Display option information [Source: Barchart.com]
 * [calc](#calc)
   * basic call/put PnL calculator
-* [act](#act)
+* [unu](#unu)
   * Display unusual options activity
 * [exp](#exp)
   * See/set expiry date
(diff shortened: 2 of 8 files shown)


gamestonk_terminal/README.md
```diff
@@ -427,6 +427,7 @@ Command|Description|Source
 `load`          | Load new ticker
 `info`          | Display option information | [Barchart](https://barchart.com/)
 `calc`          | Basic option PnL calculator
+`act`           | Scrapes unusual options activity|
```
DidierRLopes, line 430: Can we call it "unu" and change desc to "Unusual options activity".

My reasoning is that act would imply most active options. But since they call it unusual, I'm assuming they compare the activity with the past and if it's much different then it's considered unusual.

E.g. if Apple is traded a lot yesterday, and today no trades. That's unusual, and "activity" wouldn't reflect this so well. But this is pure speculation :)

gamestonk_terminal/options/README.md
```diff
@@ -14,6 +14,8 @@ This menu aims to give insight on options. Options can cause significant share p
   * Display option information [Source: Barchart.com]
 * [calc](#calc)
   * basic call/put PnL calculator
+* [act](#act)
```
DidierRLopes, line 17: *unu, perhaps

gamestonk_terminal/options/fdscanner_view.py
```diff
@@ -0,0 +1,43 @@
...
+    sort_columns: pd.Timestamp
+        Data column to sort on
+    export: str
+        File type to export
+    ascending: bool
+        Whether to sort in ascending order
+    """
+    data, last_update = fdscanner_model.unusual_options(num, sort_column, ascending)
+
+    if export:
```
DidierRLopes, line 29: I would put this at the end. So, after displaying info to the user.

gamestonk_terminal/options/fdscanner_view.py
```diff
@@ -0,0 +1,43 @@
...
+        export_data(
+            export,
+            os.path.dirname(os.path.abspath(__file__)),
+            f"UnusualOptions_updated_{last_update}",
```
DidierRLopes, line 33: Here should be the name of the command. 

The export_data func already appends a timestamp to the file

gamestonk_terminal/options/op_controller.py
```diff
@@ -87,6 +89,7 @@ def print_help(self):
         print("   info          display option information (volatility, IV rank etc)")
         print("")
         print("   calc          basic call/put PnL calculator")
+        print("   act           show unusual options activity")
```
DidierRLopes, line 92: Since this is kind of "discovery mode", can we add this right below screener option?

gamestonk_terminal/options/op_controller.py
```diff
@@ -148,6 +151,60 @@ def call_calc(self, other_args: List[str]):
...
+
+        try:
+            ns_parser = parse_known_args_and_warn(parser, other_args)
+            if not ns_parser:
+                return
+            num = ns_parser.num
+            sort_column = ns_parser.sortby
+            export = ns_parser.export
+            ascending = ns_parser.ascend
+            fdscanner_view.display_options(num, sort_column, export, ascending)
```
DidierRLopes, line 203: I think here I mislead you, I meant doing something like:

```
fdscanner_view.display_options(num=ns_parser.num, 
             sort_column=ns_parser.sortby, 
             export=ns_parser.export, 
            ascending=ns_parser.ascend)
```
and not declaring new variables

gamestonk_terminal/options/op_controller.py
```diff
@@ -148,6 +151,60 @@ def call_calc(self, other_args: List[str]):
...
+        """Process act command"""
+
+        parser = argparse.ArgumentParser(
+            prog="act",
+            add_help=False,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+            description="This command gets unusual options from fdscanner.com",
+        )
+        parser.add_argument(
+            "-n" "--num",
```
DidierRLopes, line 164: you missed a comma here.

gamestonk_terminal/options/syncretism_view.py
```diff
@@ -203,6 +211,129 @@ def screener_output(other_args: List[str]):
...
+        "max-cap",
+    ]
+    bool_list = ["active", "stock", "etf", "puts", "calls", "itm", "otm", "exclude"]
+    error = ""
+    for key, value in preset_dict.items():
+        if key in float_list:
+            try:
+                float(value)
+                if value.startswith("."):
+                    error += f"{key} : {value} needs to be formatted with leading 0\n"
```
DidierRLopes, line 291: We could have done this convention ourselves. I.e. if user provides a .5 we know this is 0.5, so we could parse it still.

gamestonk_terminal/options/syncretism_view.py
```diff
@@ -203,6 +211,129 @@ def screener_output(other_args: List[str]):
...
+            except Exception:
+                error += f"{key} : {value}, should be float\n"
+
+        elif key in bool_list:
+            if value not in ["true", "false"]:
```
DidierRLopes, line 296: I would not rely on quotes, and try to use .eval to evaluate a string to a type. If it fails, we know the user input something wrongly.

gamestonk_terminal/options/syncretism_view.py
```diff
@@ -203,6 +211,129 @@ def screener_output(other_args: List[str]):
...
+                error += f"{key} : {value},  Should be [true/false]\n"
+
+        elif key == "tickers":
+            if value[0] != '"':
+                error += f'{key} : {value}, ticker should have quotes "AAPL" \n'
```
DidierRLopes, line 301: Same here. I wouldn't necessarily ask for the "".

DidierRLopes (commented):
Take the option screeners comments as suggestions :)

Could you add 1 or 2 examples of an option screener preset, so people could have a look at a concrete example

gamestonk_terminal/options/fdscanner_view.py
jmaslek: The new time stamp actually tells you when the options were updated.  You could run the command 20 minutes apart and get the same data

gamestonk_terminal/options/op_controller.py
jmaslek: I see

gamestonk_terminal/options/fdscanner_view.py
DidierRLopes: I think that it would be better to keep this as the name of the command, since this is what we're going to do with all commands. Also having 2 timestamps is a bit ugly.

Captura de ecrã 2021-08-03, às 21 31 41

Most of the commands rely on a certain update time from the server side. So, the user knows about this when they export.

gamestonk_terminal/README.md
jmaslek: I can. I think this source refers to unusual as just high Vol/OI (which is a typical description)

gamestonk_terminal/options/syncretism_view.py
jmaslek: I brought this up to the author of the api and I think it might be updated on their end.

DidierRLopes: Even better!

DidierRLopes (approved):

gamestonk_terminal/options/op_controller.py
```diff
@@ -83,13 +83,13 @@ def print_help(self):
         print("")
         print("   disp          display all preset screeners filters")
         print("   scr           output screener options")
+        print("   act           show unusual options activity")
```
DidierRLopes, line 86: *unu

gamestonk_terminal/options/syncretism_view.py
```diff
@@ -297,15 +300,11 @@ def check_presets(preset_dict: dict):
                 error += f"{key} : {value},  Should be [true/false]\n"
 
         elif key == "tickers":
-            if value[0] != '"':
-                error += f'{key} : {value}, ticker should have quotes "AAPL" \n'
-            else:
+            for ticker in value.split(","):
```
DidierRLopes, line 303: If the value is a string, isn't ticker gonna be a single char?

like if we do for ticker in "HELLO", we get "H", "E", ...

gamestonk_terminal/options/syncretism_view.py
```diff
@@ -297,15 +300,11 @@ def check_presets(preset_dict: dict):
...
                 try:
-                    for ticker in value.split(","):
-                        if ticker[0] != '"':
-                            error += f'{key} : {ticker}, all tickers should have quotes "AAPL","GME" \n'
-                except Exception:
-                    error += f"Error splitting tickers: {value} \n"
+                    eval(ticker)
```
DidierRLopes, line 305: Can we check if the stock exists from yfinance? Or does the API returns this?

gamestonk_terminal/options/syncretism_view.py
jmaslek: So I checked that  "AAPL" .split(",") gives ['"AAPL"'] which can be eval-ed.  

I tried  cases where I use different combinations and this seems to catch most issues.

DidierRLopes: Captura de ecrã 2021-08-04, às 01 03 34

you're right. I thought that if split didn't find the character it would return the string as is. But it still puts it into a list :)

jmaslek: I actually did that on accident 🤣

gamestonk_terminal/options/syncretism_view.py
jmaslek: One could, but honestly this is likely going to be blank since you want to find different stocks with your criteria, so it seems like overkill to have to check.  If the ticker doesn't exist then it returns empty, not an error

DidierRLopes: So, my opinion is that if the user provides a ticker that doesn't exist, we want to warn them about that.

It isn't overkill because if the user doesn't provide any ticker (which will be the case almost always, as you mentioned) then we don't use yfinance to check on ticker

PR #641, opened 2021-08-04
Hotfix for jsondecode error in coinpaprika

jmaslek:
Address the failing tests by throwing the coinpaprika functions in a try block.

### DIFF
--- a/gamestonk_terminal/cryptocurrency/coinpaprika/coinpaprika_view.py
+++ b/gamestonk_terminal/cryptocurrency/coinpaprika/coinpaprika_view.py
@@ -3,12 +3,15 @@ __docformat__ = "numpy"
 
 import argparse
 from typing import List
+from json.decoder import JSONDecodeError
+
 import difflib
 from tabulate import tabulate
 from pandas.plotting import register_matplotlib_converters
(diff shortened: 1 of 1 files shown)


DidierRLopes (approved):
Looks good for now. Go ahead and merge.

An improvement is your suggestion of moving these code to inside the paprika_controller. So that it's only initialized when we go into that option.

jmaslek:
> An improvement is your suggestion of moving these code to inside the paprika_controller. So that it's only initialized when we go into that option.

I agree.  But it is used with a check function so every time you would check a coin it would get redownloaded, so maybe @JakubPluta can do something clever.  This is mostly to block the tests from failing

JakubPluta:
@jmaslek @DidierRLopes 

I am already in progress of restructuring whole crypto menu, so I will try to figure out something to load is just one. Maybe some cache or downloading data to a file (and simple check there if file exists load from file if no use api)

jmaslek:
I agree that calling it once is the way to go.  Maybe they will update their API to avoid crashing more often.  

We could also just have a small file cached (could have GitHub run it daily and commit) that is loaded to compare with.

PR #643, opened 2021-08-05
Fix bug in ca->hcorr 

jmaslek:
There was an issue in `hcorr` where 'dataframe objects are mutable'.  Issue was looping through dictionary.items when it should have just been the keys

### DIFF
--- a/gamestonk_terminal/comparison_analysis/yahoo_finance_view.py
+++ b/gamestonk_terminal/comparison_analysis/yahoo_finance_view.py
@@ -269,8 +269,8 @@ def correlation(other_args, df_stock, ticker, start, interval, similar):
                     d_candle_types[ns_parser.type_candle]
                 ].rename(ticker)
 
-                # pylint:disable=consider-iterating-dictionary
-                for symbol in d_stock.keys():
+                for symbols in d_stock.items():
+                    symbol = symbols[0]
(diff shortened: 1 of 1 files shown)


gamestonk_terminal/comparison_analysis/yahoo_finance_view.py
```diff
@@ -268,7 +268,9 @@ def correlation(other_args, df_stock, ticker, start, interval, similar):
                 df_stock = d_stock[ticker][
                     d_candle_types[ns_parser.type_candle]
                 ].rename(ticker)
-                for symbol in d_stock.items():
+
+                for symbols in d_stock.items():
```
DidierRLopes, line 272: could also be `d_stock.keys()`:)

DidierRLopes (approved):

gamestonk_terminal/comparison_analysis/yahoo_finance_view.py
jmaslek: I did that in the first commit and got a bunch of pylint errors so just did this way

DidierRLopes: I was thinking if you needed to wrap the dict.keys() with a list(). Prob, yes.

PR #649, opened 2021-08-07
Added __init__ files and updated files for pylint

colin99d:
Added __init__ files, updated .coveragerc, and made fx_view work with pylint.

### DIFF
--- a/gamestonk_terminal/forex/fx_view.py
+++ b/gamestonk_terminal/forex/fx_view.py
@@ -1,6 +1,15 @@
+from datetime import datetime
 import argparse
-from oandapyV20 import API
 from typing import List
+
+from oandapyV20 import API
+import pandas as pd
+import pandas_ta as ta
+import mplfinance as mpf
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mticker
+import seaborn as sns
+
 import oandapyV20.endpoints.pricing as pricing
 import oandapyV20.endpoints.accounts as accounts
 import oandapyV20.endpoints.orders as orders
@@ -17,13 +26,6 @@ from gamestonk_terminal.helper_funcs import (
     plot_autoscale,
     check_non_negative,
 )
-import pandas as pd
-import pandas_ta as ta
-import mplfinance as mpf
-import matplotlib.pyplot as plt
-import matplotlib.ticker as mticker
-import seaborn as sns
-from datetime import datetime
 
 
 client = API(access_token=cfg.OANDA_TOKEN, environment="live")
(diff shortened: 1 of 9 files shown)


gamestonk_terminal/forex/fx_view.py
```diff
@@ -10,13 +10,28 @@
...
+
+# pylint: disable=R0402
 import oandapyV20.endpoints.orders as orders
+
+# pylint: disable=R0402
 import oandapyV20.endpoints.instruments as instruments
+
+# pylint: disable=R0402
 import oandapyV20.endpoints.positions as positions
+
```
jmaslek, line 27: can this not just be `from oandapyV20.endpoints import (pricing,accounts,...)`

colin99d: I was just getting my pylint to work. I didn't add this file.

jmaslek:
Was fx_view not passing before?

colin99d:
> Was fx_view not passing before?

colin99d:
I am not sure. I did not edit the view except to make it work with my pylint.

jmaslek:
Im just confused.  If it wasn't working with pylint then every other commit would fail

DidierRLopes:
Could have been due to the init in the forex folder which may trigger they pylint to check all files in that dir since that becomes a module

gamestonk_terminal/forex/fx_view.py
DidierRLopes: Can you do @jmaslek suggestion, so on the top do

```
from oandapyV20.endpoints import (
   pricing,
   accounts,
   orders,
   ...
)
```
and then you don't need the `pylint: disable=R0402` anymore

colin99d:
Ok I switched to the new method. This way is much cleaner but I still needed an ignore because `Imports from package oandapyV20 are not grouped (ungrouped-imports)`

DidierRLopes:
> Ok I switched to the new method. This way is much cleaner but I still needed an ignore because `Imports from package oandapyV20 are not grouped (ungrouped-imports)`

So, the reason for that is because when you import oanda related stuff the import commands are separate by other imports.

image

You just need to put all these commands next to each other

DidierRLopes:
Ideally, we should use `pylint: disable` only when we can't do it in another way. Since pylint is there to enforce good coding development

colin99d:
Cool. I did not not it was that easy to fix. Thanks for that!

DidierRLopes:
> Cool. I did not not it was that easy to fix. Thanks for that!

Np, let's merge this once tests pass!! Thanks!!

DidierRLopes (approved):

PR #651, opened 2021-08-08
make po work again for stock screener menu

devwolff:
Because of recent changes, portfolio opt no longer works from stock screener menu


DidierRLopes:
Sorry, my bad. Forgot about this!

Can you also do:
```
ret = po_controller.menu([])
```
in `portfolio_controller.py`

This makes it so it doesn't crash when going into `po` directly

DidierRLopes:
The tests are failing, can you do `black gamestonk_terminal/portfolio/portfolio_optimization/po_controller.py"? That should solve your pylint issues, and then git add and commit again :)

DidierRLopes (approved):
Good work!!

PR #654, opened 2021-08-08
make oanda practice account work on forex menu

devwolff:


gamestonk_terminal/config_terminal.py
```diff
@@ -53,6 +53,9 @@
 DG_TOTP_SECRET = os.getenv("GT_DG_TOTP_SECRET") or None
 
 # https://developer.oanda.com
+OANDA_ACCOUNT_TYPE = (
+    os.getenv("GT_OANDA_ACCOUNT_TYPE") or "REPLACE_ME"
```
DidierRLopes, line 57: What do you think about replacing "REPLACE_ME" by "live". So that the default is live, and if someone is testing their practice account they can replace it. You can write as comment
```
# "practice" - for a practice account
```

jmaslek: You should default to practice in this case. That way if someone wants to trade and risk money they have to go through extra steps to confirm they want live.

DidierRLopes: Agreed! :)

DidierRLopes (approved):
Sweet, thank you! :)

PR #656, opened 2021-08-08
Feature/refactoring crypto

JakubPluta:
Hi @DidierRLopes,

here is my proposal for the new crypto menu.  As we discussed I split everything into 3 main modules discovery, overview, due diligence. It's an initial proposal, so maybe there should be some reshuffle of commands between menus (between overview and discovery) but it's a minor thing :) 

Also I added for load method `--source` parameter. (In the future we can think about his mapping table for all sources and coins)
By default it uses coingecko.

I have not updated in 100% READMEs and some methods docs, but I think would be better first to take a look on my work if I did it right. If you like my idea and proposal, then I will complete the documentation and make any necessary changes.

If will have some time, please take a look and let me know what you think about my proposal.


gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -114,32 +148,99 @@ def call_quit(self, _):
...
+            self.current_coin, self.source = load(
+                coin=self.current_coin, other_args=other_args
+            )
+        except TypeError:
+            print("Couldn't load data\n")
+
+    def call_chart(self, other_args):
+        """Process chart command"""
+        if self.current_coin:
+            getattr(DD_VIEWS_MAPPING[self.source], "chart")(
```
DidierRLopes, line 163: This chart could use the mplfinance lib as it looks nicer :)

Feel free to not take some of these actions on this PR, and work on these at a later stage if needed

gamestonk_terminal/cryptocurrency/discovery/coinpaprika_model.py
```diff
@@ -0,0 +1,43 @@
+import pandas as pd
+from gamestonk_terminal.cryptocurrency.coinpaprika_helpers import (
+    ENDPOINTS,
+    PaprikaSession,
+)
+
+
+session = PaprikaSession()
```
DidierRLopes, line 8: Probably we could have this inside the commands?

gamestonk_terminal/cryptocurrency/discovery/coinpaprika_model.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def search(q, c=None, modifier=None):
+    """Search CoinPaprika
+    Parameters
+    ----------
+    q:  phrase for search
```
DidierRLopes, line 15: I would use variables more descriptive. Just chars makes it harder to understand what it is.

gamestonk_terminal/cryptocurrency/discovery/coinpaprika_view.py
```diff
@@ -0,0 +1,284 @@
...
+        letter = ns_parser.letter
+        if letter and isinstance(letter, str):
+            df = df[
+                df[ns_parser.key].str.match(f"^({letter.lower()}|{letter.upper()})")
+            ]
+
+        try:
+            df = df[ns_parser.skip : ns_parser.skip + ns_parser.top]
+        except Exception:
+            df = get_list_of_coins()
```
DidierRLopes, line 187: If it fails we should print an error, and not re-load the same dataframe. Otherwise, what I would do is not load against that dataframe, but instead copy the dataframe above so that that df is left untouched, if that makes sense

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
```diff
@@ -0,0 +1,232 @@
+"""Cryptocurrency Overview Controller"""
```
DidierRLopes, line 1: Discovery*

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
```diff
@@ -0,0 +1,232 @@
...
+
+class Controller:
+
+    CHOICES = [
+        "?",
+        "cls",
+        "help",
+        "q",
+        "quit",
+        "cp_search",
```
DidierRLopes, line 26: So, one of the things we are going to write on the Coding Guidelines is commands to not have underscores, since that requires 2 fingers to type, and adds latency to execution speed.

Could we have these so that they have no underscore _ ?

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
```diff
@@ -0,0 +1,232 @@
...
+            "   cg_gainers         top gainers - coins which price gained the most in given period "
+        )
+        print(
+            "   cg_losers          top losers - coins which price dropped the most in given period "
+        )
+        print("   cg_yfarms          top yield farms")
+        print("   cg_top_defi        top defi protocols")
+        print("   cg_top_dex         top decentralized exchanges")
+        print("   cg_top_nft         top non fungible tokens")
+        print("")
```
DidierRLopes, line 78: Since these are becoming bigger, we can remove the print("") in between of the data sources

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def __init__(self):
+        """CONSTRUCTOR"""
+
+        self._discovery_parser = argparse.ArgumentParser(add_help=False, prog="disc")
+        self._discovery_parser.add_argument("cmd", choices=self.CHOICES)
+
+    def print_help(self):
+        """Print help"""
+        print("\nDiscovery:")
```
DidierRLopes, line 53: One of the new refactoring rules we'll implement is to build the whole string with """ and then print(help_text) at the end. Not sure if you want to do this now or later

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
```diff
@@ -0,0 +1,232 @@
...
+            self, "call_" + known_args.cmd, lambda: "Command not recognized!"
+        )(other_args)
+
+    def call_help(self, _):
+        """Process Help command"""
+        self.print_help()
+
+    def call_q(self, _):
+        """Process Q command - quit the menu."""
+        print("Moving back to (crypto) menu")
```
DidierRLopes, line 131: Probably this print is not needed as the terminal relies on this q to jump back 1 layer functionality

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
```diff
@@ -0,0 +1,232 @@
...
+        """Process search command"""
+        coinpaprika_view.search(other_args=other_args)
+
+    def call_bin_coins(self, other_args):
+        """Process bin_coins command"""
+        binance_view.coins(other_args=other_args)
+
+
+def menu():
+    controller = Controller()
```
DidierRLopes, line 208: I would call this disc_controller.

gamestonk_terminal/cryptocurrency/due_diligence/binance_model.py
```diff
@@ -0,0 +1,166 @@
...
+from gamestonk_terminal.config_plot import PLOT_DPI
+from gamestonk_terminal.feature_flags import USE_ION as ion
+
+
+def _get_trading_pairs() -> List[dict]:
+    """Helper method that return all trading pairs on binance.
+
+    Returns
+    -------
+        List[dict]
```
DidierRLopes, line 25: docstring missing a better description

gamestonk_terminal/cryptocurrency/discovery/pycoingecko_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+CATEGORIES = {
+    "trending": 0,
+    "most_voted": 1,
+    "positive_sentiment": 2,
+    "recently_added": 3,
+    "most_visited": 4,
+}
+
+COLUMNS = {
```
DidierRLopes, line 37: This COLUMNS is almost exactly the same, I would do something different. Just because the only different I can see is the last_added: added

gamestonk_terminal/cryptocurrency/discovery/pycoingecko_model.py
```diff
@@ -0,0 +1,451 @@
...
+def get_top_volume_coins() -> pd.DataFrame:
+    """Scrapes top coins by trading volume "https://www.coingecko.com/en/coins/high_volume"
+
+    Returns
+    -------
+    pandas.DataFrame
+        rank, name, symbol, price, change_1h, change_24h, change_7d, volume_24h, market_cap
+    """
+    columns = [
+        COLUMNS["rank"],
```
DidierRLopes, line 284: This doesn't add much value because rank is already rank, and this rank is pre-defined

gamestonk_terminal/cryptocurrency/due_diligence/coinpaprika_model.py
```diff
@@ -0,0 +1,333 @@
+from datetime import datetime, timedelta
```
DidierRLopes, line 1: that header line is missing here

gamestonk_terminal/cryptocurrency/due_diligence/coinpaprika_model.py
```diff
@@ -0,0 +1,333 @@
...
+        lambda x: "\n".join(textwrap.wrap(x, width=80)) if isinstance(x, str) else x
+    )
+    df["status"] = df["status"].apply(lambda x: x.replace("  ", ""))
+    df["date"] = df["date"].apply(lambda x: x.replace("T", "\n"))
+    df["date"] = df["date"].apply(lambda x: x.replace("Z", ""))
+    return df
+
+
+def get_coin_events_by_id(coin_id="eth-ethereum"):
+    """
```
DidierRLopes, line 61: Add a small description

gamestonk_terminal/cryptocurrency/due_diligence/coinpaprika_view.py
```diff
@@ -0,0 +1,815 @@
...
+        if coin_id:
+            return coin_id
+
+    except SystemExit:
+        print("")
+    except Exception as e:
+        print(e, "\n")
+
+
+def ta(coin_id: str, other_args: List[str]):
```
DidierRLopes, line 700: I would call this function load_ta_data or similar

DidierRLopes: Also, we should have mypy return type here --> Tuple[pd.DataFrame, str]

gamestonk_terminal/cryptocurrency/due_diligence/pycoingecko_view.py
```diff
@@ -0,0 +1,593 @@
...
+        plt.show()
+        print("")
+
+    except SystemExit:
+        print("")
+    except Exception as e:
+        print(e, "\n")
+
+
+def ta(coin: gecko.Coin, other_args: List[str]):
```
DidierRLopes, line 133: load_ta_data or so :p

gamestonk_terminal/cryptocurrency/overview/coinpaprika_model.py
```diff
@@ -0,0 +1,295 @@
...
+    global_markets["last_updated"] = datetime.fromtimestamp(
+        global_markets["last_updated"]
+    )
+
+    for key, date in global_markets.items():
+        if "date" in key:
+            try:
+                global_markets[key] = parser.parse(date).strftime("%Y-%m-%d %H:%M:%S")
+            except (KeyError, ValueError, TypeError):
+                ...
```
DidierRLopes, line 35: I have not seen this before `...`. What does it mean?

gamestonk_terminal/cryptocurrency/overview/coinpaprika_model.py
```diff
@@ -0,0 +1,295 @@
...
+        "pct_change_24h",
+        # "pct_change_7d",
+        # "pct_change_30d",
+        "ath_price",
+        "pct_from_ath",
+    ]
+    return _get_coins_info_helper(quotes=quotes)[cols].sort_values(by="rank")
+
+
+def get_list_of_exchanges(quotes="USD"):
```
DidierRLopes, line 182: type of parameter and return -> missing

gamestonk_terminal/cryptocurrency/overview/coinpaprika_model.py
```diff
@@ -0,0 +1,295 @@
...
+    """
+    List exchanges from CoinPaprika API
+    Parameters
+    ----------
+    quotes: Coma separated quotes to return e.g quotes=USD,BTC
```
DidierRLopes, line 187: type of variable missing here

gamestonk_terminal/cryptocurrency/overview/coinpaprika_view.py
```diff
@@ -0,0 +1,629 @@
...
+    "TWD",
+    "ZAR",
+    "VND",
+    "BOB",
+    "COP",
+    "PEN",
+    "ARS",
+    "ISK",
+]
+PLATFORMS = paprika.get_all_contract_platforms()["platform_id"].tolist()
```
DidierRLopes, line 64: We should avoid having these calls not be contained to a controller, or a command even. Same for ones below

gamestonk_terminal/cryptocurrency/overview/overview_controller.py
```diff
@@ -0,0 +1,259 @@
...
+        print("\nOverview:")
+        print("   cls             clear screen")
+        print("   ?/help          show this menu again")
+        print("   q               quit this menu, and shows back to main menu")
+        print("   quit            quit to abandon program")
+        print("")
+
+        print("")
+        print("CoinGecko:")
+        print("   cg_global          global crypto market info")
```
DidierRLopes, line 65: Same here for underscores_

DidierRLopes: and same regarding
print(
print(
print(

However, again, these are not changes necessary for this PR :)

tests/test_cryptocurrency/test_coinpaprika_view.py
```diff
@@ -2,102 +2,134 @@
...
+        twitter("eth-ethereum", [])
         sys.stdout = sys.__stdout__
         capt = capturedOutput.getvalue()
         self.assertIn("Couldn't find", capt)
 
     def test_events(self):
         capturedOutput = io.StringIO()
         sys.stdout = capturedOutput
-        coinpaprika_view.events("eth-ethereum", [])
+        events("eth-ethereum", [])
```
DidierRLopes, line 116: I would leave here the mention to the modules, it makes it easier to understand where it is.

Also, I had to remove 3 crypto tests that were causing tests to fail stochastically. This is due to performing api requests on a global level, hence why it's best to have these commnand contained as we have discussed  before :)

DidierRLopes (commented):
A lot of comments, but this PR is also huge (50 files changed, 8.5k additions and 7.5k deletions).

Some generic comments are:
- We also want to refactor argparse to be at controller level. If you want we can do it now, but given how big this PR is, I think is best to leave that for a next iteration.
- Docstrings can be left as they are but once we tackle next refactoring, these should be worked on (i.e. missing types, better descriptions, returns ->, ..). All that boring stuff :D
- Don't worry about README or even Tests. Let's fully focus on this refactoring structure, and then we can work on Tests + README.
- Commands shouldn't have "_" to allow for a faster typing
- The controller help should have a single print
- We should not do API calls at a global level but be more contained, i.e. either at command level or at controller level
- I would have variables names more descriptive. E.g. having just a character is not good because it's hard to know what it means by reading code. And reading code is really important.
- In terms of folder organization and display, you nailed it :) The only thing missing is that function to convert between multiple sources, BUT I think this doesn't need to be worked on now and is a future goal.

@jmaslek can you also have a look/comment at this PR, it would be good to have a second pair of eyes + comments.

DidierRLopes:
Captura de ecrã 2021-08-08, às 22 44 31

PS: In terms of additions you are already TOP 4! Great work!

gamestonk_terminal/cryptocurrency/__init__.py
```diff
@@ -0,0 +1,208 @@
+import argparse
```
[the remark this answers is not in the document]
JakubPluta: I didn't know where to put it as it's common for all sources. I think first ideas was to put it in helpers, but I had some kind of conflict with imports. But will take a look at the evening :D

gamestonk_terminal/cryptocurrency/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+from typing import List, Tuple, Any, Optional
+import difflib
+import pandas as pd
+from tabulate import tabulate
+from gamestonk_terminal.helper_funcs import parse_known_args_and_warn, check_positive
+from gamestonk_terminal.cryptocurrency.due_diligence import pycoingecko_model
```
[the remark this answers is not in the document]
JakubPluta: Understood, I will try to make some adjustments.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -114,32 +148,99 @@ def call_quit(self, _):
         """Process Quit command - quit the program"""
         return True
 
-    def call_cg(self, _):
-        if pycoingecko_controller.menu():
-            return True
-        print("")
+    def call_load(self, other_args):
+        """Process load command"""
+        try:
```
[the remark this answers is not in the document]
JakubPluta: Okey, I will take a look on you implementation, but I also agree to have it with next PR. In current PR there a lot of changes, so maybe don't add new ones :D I will do that as a next step

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
JakubPluta: Yes, for sure. I agree that it's not to user friendly to type _, but my question should I go with cpcoins, cgtrending and so on ? Or try to change all commands names to not have any prefixes like in previous version (there will be some overlap like with `coins` methods or `info` , `exchanges` and few others. Do you have anything on mind ?

gamestonk_terminal/cryptocurrency/__init__.py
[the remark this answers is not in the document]
JakubPluta: Yea so there is import config if I put it in `cryptocurrency_helper`, because my views `load` and `find` are using `pycoingecko_view`, `binance_view`, `coinpaprika_view` and in the same time those `pycoingecko_view`, `binance_view`, `coinpaprika_view`  have imports from `cryptocurrency_helper`. I ve never had this kind of clutch with imports, and I am thinking how to deal with that in a proper way.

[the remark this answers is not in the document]
aia: This definitely should not be in ```__init__```

aia (changes requested):

gamestonk_terminal/cryptocurrency/__init__.py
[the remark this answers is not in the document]
DidierRLopes: That solution sounds perfect to me! 

crypto_helpers should be used to hold the main commands associated with crypto_controller. Then having an addition dataframe_helpers seems the right approach to me :)

gamestonk_terminal/cryptocurrency/crypto_controller.py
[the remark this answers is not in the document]
DidierRLopes: Agreed! One step at a time eheh

gamestonk_terminal/cryptocurrency/__init__.py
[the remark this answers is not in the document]
DidierRLopes: If it's to get a list of data, using _model is fine. It's more of a suggestion rather than a rule, if it makes sense

gamestonk_terminal/cryptocurrency/discovery/discovery_controller.py
DidierRLopes: I think for now we could go with the cgcoins, cgtrending, ...

In the future, we can thinking on combining them. I think this is a futuristic step, for now having cg,cp at start without underscore is the right thing to do!

DidierRLopes:
Btw, @JakubPluta don't worry with the README. I'll take care of it at a later stage! Or we can discuss that on the next iteration.

gamestonk_terminal/cryptocurrency/discovery/coinpaprika_model.py
JakubPluta: @DidierRLopes 

I agree to not keep it initialized in model. I tried here few things, but without good results. 

One thing I triend was injection of PaprikaSession from Controller to `coinpaprika_model`:

```
class CryptoController:
    def __init__(self, coin=None, source=None):
        """CONSTRUCTOR"""

        self._dd_parser = argparse.ArgumentParser(add_help=False, prog="dd")

        self.current_coin = coin
        self.current_currency = None
        self.current_df = pd.DataFrame()
        self.source = source

        self.CHOICES.extend(self.SPECIFIC_CHOICES[self.source])

        self._dd_parser.add_argument("cmd", choices=self.CHOICES)

        if not coinpaprika_model.session:
            coinpaprika_model.session = PaprikaSession()
```

And in the same time looked  `coinpaprika_model` looked like here:
```
"""CoinPaprika model"""
__docformat__ = "numpy"

from typing import Optional, Any
import pandas as pd

session = None

def search(
    query: str, category: Optional[Any] = None, modifier: Optional[Any] = None
) -> pd.DataFrame:
    ...
    return pd.DataFrame(results)
```

So it was an idea to just inject session to given _model.py at the moment of initialization CryptoController, but this is not working. I think it's because some order of imports in other .py files. When I use any command related to CoinPaprika I get that this `session` variable is still `None`
But I am not sure. Maybe there is a way to set this variable as a `global` , but I am not familiar with that, so hard to say. Maybe you have more knowledge with that. 

Another Idea that I have, to just wrap all methods in coinpaprika_model.py into a class which will have in `__init__` parameter session which I will provide in Controller classes. But it's just an idea, don't know if it's right approach. 

Maybe you have any other idea for that ? Would be great, cause I feel a little stuck with it :D

gamestonk_terminal/cryptocurrency/overview/coinpaprika_model.py
JakubPluta: It's probably my bad habbit, It's very similar to just `pass` statement. Probably I missed that one, I will add there print statement to know what failed.

gamestonk_terminal/cryptocurrency/crypto_controller.py
JakubPluta: In coingecko, there are only "price" column, so we don't have there Open, High, Low, Close, Volume. Mplfinance is amazing, but one constraint is to have OHLCV data, so for coingecko it's not aplicable.

DidierRLopes: This makes sensee, thanks for the explanation!

What do you think about having another default dataframe data source? So that we can have mplfinance?

gamestonk_terminal/cryptocurrency/discovery/coinpaprika_model.py
DidierRLopes: Thanks for this explanation, I'm ok to leave this as is for now. And once we go into refactoring we can look at ways to tackle this. We can have a joined session with @jmaslek to brainstorm ideas to achieve the best architecture possible.

This will lead to have a strong foundation, and a more robust code going forward!

gamestonk_terminal/cryptocurrency/overview/coinpaprika_model.py
DidierRLopes: I had not seen `...` before! Interesting. But yes, ideally we would output what is the error. This way in case it crashes the user can tell us what is the message they saw, and we can know what happened.

You can do something like
```
except (KeyError, ValueError, TypeError) as e:
    print(e)
```

So that you still capture the error without the terminal crashing!

DidierRLopes:
@jmaslek I'm happy with this PR for a first refactoring of the crypto menu. Could you have a look and comment if there are things that can be improved?

Note: This only aims at changing the UX from having menus based on source, to have menus based on functionality. The next step is the code refactoring, but this PR is not meant to tackle that just yet.

jmaslek:
Not sure opinions, but I think unloaded coin/source should do None instead of ?  (print(f"Coin : {self.coin or None}))
```Coin: ?
Source: ?```

**Comment by jmaslek** [comment]
We have improved plotting a bunch.  

Screen Shot 2021-08-11 at 1 57 38 PM

**Comment by jmaslek** [comment]
`(crypto)>(disc)> cgcoins` shows too many.  Add a flag for how many to show, maybe default to 15 or 20

**Comment by jmaslek** [comment]
`cgvolume` shows inconsistent price floats
Screen Shot 2021-08-11 at 2 01 46 PM

**Comment by jmaslek** [comment]
`cpcoins           coins available on CoinPaprika`

I would remove this.  Not sure if @DidierRLopes would agree, but does this function add value?  Listing all coins just prints a bunch to the terminal.  The one after it is useful in that you can actually search.

**Comment by jmaslek** [comment]
Also in general, when you have an argparse with `required = True`, you should have it be 

`required = '-h' not in other_args`

This way using `command -h` still shows you defaults and descriptions

**Comment by jmaslek** [comment]
`(crypto)>(ov)> cgnews`  set the default to 15 or 20

**Comment by jmaslek** [comment]
```python
(✨) (crypto)>(dd)> ath -h
usage: ath [-h]

All time high data for loaded coin. You can find there most important metrics regarding ath of coin price like: current_price_btc, current_price_eth,
current_price_usd, ath_btc, ath_eth, ath_usd, ath_date_btc, ath_date_eth, ath_date_usd, ath_change_percentage_btc, ath_change_percentage_btc,
ath_change_percentage_eth, ath_change_percentage_usd
```

Giving variable names to metrics is not helpful.  I have no idea what these are.  I would also imagine that if there is an all time high and all time low data, you could combine into one command  grouping all times

jmaslek:
For the info commands, since there is just one column, maybe drop nan rows instead of showing nothing:
Screen Shot 2021-08-11 at 2 18 47 PM

DidierRLopes:
> `cpcoins coins available on CoinPaprika`
> 
> I would remove this. Not sure if @DidierRLopes would agree, but does this function add value? Listing all coins just prints a bunch to the terminal. The one after it is useful in that you can actually search.
> 
> EDIT:: same with `bincoins`

Good point. We can probably incorporate this data on the helper when the user does `search -h`?

jmaslek:
So the ta menu fails on most.  Mostly due to some of them requiring candle data.  @DidierRLopes this is probably something we need to do with refactoring TA where we make sure all columns are present or have a specific crypto flag?.

jmaslek:
bug.  Should add the -t (or -c) flag if its not present

```python
(✨) (crypto)> load bitcoin --source cp
The following args couldn't be interpreted: ['--source', 'cp']
Could not find coin with given id: bitcoin
```

Haven't followed the discussion, but is there some kind of conversion structure that knows one source needs bitcoin or another needs btc, etc?

jmaslek:
The coin paprika chart is nice.  Since you need candle data to use that function, you could always try resampling the other sources and make your own candles for using mplfinance.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -114,32 +157,84 @@ def call_quit(self, _):
...
+
+    def call_chart(self, other_args):
+        """Process chart command"""
+        if self.current_coin:
+            getattr(self.DD_VIEWS_MAPPING[self.source], "chart")(
+                self.current_coin, other_args
+            )
+        else:
+            print("No coin selected. Use 'load' to load the coin you want to look at.")
+            print("")
```
jmaslek, line 177: very small formatting point, but whenever you have a `print("")`, just try to add a `"\n"` to a previous print if there is one

DidierRLopes:
> So the ta menu fails on most. Mostly due to some of them requiring candle data. @DidierRLopes this is probably something we need to do with refactoring TA where we make sure all columns are present or have a specific crypto flag?.

So this is the reason why we have a common/momentum.py, common/overlap.py, yet the controller lives on his own. This is so that al the necessary processing can be done on the controller and call this common helper funcs

JakubPluta:
> `cgvolume` shows inconsistent price floats
> 
> edit: looks like most of the cg---- functions do
> Screen Shot 2021-08-11 at 2 01 46 PM

Yea, so the reason of that is that data directly comes from scraped coingecko sites. Almost of all the values are string types. So here are 2 ways to go: 

1. We leave it as it is (string format e.g $550.330,456)
2. I will try to convert all of those strings into numbers, and be consistent with format.

What do you think ?  If you propose 2nd option, should I work on that with current PR or do that in next one.

JakubPluta:
> The coin paprika chart is nice. Since you need candle data to use that function, you could always try resampling the other sources and make your own candles for using mplfinance.

I don't know if I get it. If in coingecko I have only one value "price" - > you mean by reasempling to create Open, High, Low, Close and put the same value there ? 

```
index           | price
2021-01-01 | 25.12
2021-01-02 | 25.44

index           | Open | High  | Low   | Close
2021-01-01 | 25.12 | 25.12 | 25.12 | 25.12
2021-01-02 | 25.44 | 25.44 | 25.44 | 25.44

```

JakubPluta:
> `cpcoins coins available on CoinPaprika`
> 
> I would remove this. Not sure if @DidierRLopes would agree, but does this function add value? Listing all coins just prints a bunch to the terminal. The one after it is useful in that you can actually search.
> 
> EDIT:: same with `bincoins`

If I will create also search/find view for Binance, it won't be needed. But this find view it's just a primitive way to iterate through dictionary and with some string comprehension looking for best match. In future should be improved for sure.

JakubPluta:
> ```python
> (✨) (crypto)>(dd)> ath -h
> usage: ath [-h]
> 
> All time high data for loaded coin. You can find there most important metrics regarding ath of coin price like: current_price_btc, current_price_eth,
> current_price_usd, ath_btc, ath_eth, ath_usd, ath_date_btc, ath_date_eth, ath_date_usd, ath_change_percentage_btc, ath_change_percentage_btc,
> ath_change_percentage_eth, ath_change_percentage_usd
> ```
> 
> Giving variable names to metrics is not helpful. I have no idea what these are. I would also imagine that if there is an all time high and all time low data, you could combine into one command grouping all times

By that you mean to have on command for ath and atl (with parametera like --type [ath, atl] ? Or you mean to combine dataframes into one ?

Regarding column names, I will change names to more like `All time high date [USD]`

JakubPluta:
> bug. Should add the -t (or -c) flag if its not present
> 
> ```python
> (✨) (crypto)> load bitcoin --source cp
> The following args couldn't be interpreted: ['--source', 'cp']
> Could not find coin with given id: bitcoin
> ```
> 
> Haven't followed the discussion, but is there some kind of conversion structure that knows one source needs bitcoin or another needs btc, etc?
> 
> EDIT:: the following also fails:
> 
> ```python
> (✨) (crypto)> load -c bitcoin --source cp
> The following args couldn't be interpreted: ['--source', 'cp']
> Could not find coin with given id: bitcoin
> ```

So this one is the biggest pain. CoinPaprika, CoinGecko, Binance - all of those API's has different number of coins, different coin names, coin ids, so it's very hard to gues which coin id you need to use for given source. It's not like with stocks or etf or other financial assets where you have with one ticker only one stock. 

E.g. coins with symbol UNI - > there are like 5 coins with this symbol, that's why user should use coins ids in most cases. But here for example in CoinGecko bitcoin has id `bitcoin` but in coinpaprika: `btc-bitcoin`. So that's why for now I added there `cpcoins`, `bincoins`, `cgcoins` and also `find` views to let user find id of coin in which he is interested in. 

I think we should do here brainstorm how to make this mapping matrix. I would say that CoinGecko should be our main source. Because it has the biggest number of coins. And somehow we should glue to it other coin'ids from different sources. But I would say to it will be not easy one. 

I propose to maybe leave it for now. And in next step build this translation table. But here would be great guys if you can help my with some ideas :)

JakubPluta:
@jmaslek Thank you very much for review. Tomorrow I will try to apply most changes. I replied to your comments. Please take a look.

jmaslek:
> Yea, so the reason of that is that data directly comes from scraped coingecko sites. Almost of all the values are string types. So here are 2 ways to go:
> 
> 1. We leave it as it is (string format e.g $550.330,456)
> 2. I will try to convert all of those strings into numbers, and be consistent with format.
> 
> What do you think ? If you propose 2nd option, should I work on that with current PR or do that in next one.

So I just meant the price column.  You can either do something like `df["Price"] = pd["Price"].apply(lambda x: float(x.strip("$"))` with the floatfmt specified in tabulate.  Or you can try something else that prints just 2 decimal places.

jmaslek:
> I don't know if I get it. If in coingecko I have only one value "price" - > you mean by reasempling to create Open, High, Low, Close and put the same value there ?
> 
> ```
> index           | price
> 2021-01-01 | 25.12
> 2021-01-02 | 25.44
> 
> index           | Open | High  | Low   | Close
> 2021-01-01 | 25.12 | 25.12 | 25.12 | 25.12
> 2021-01-02 | 25.44 | 25.44 | 25.44 | 25.44
> ```

I don't remember how often the data is supplied for (ie what is index[1]-index[0]), but you can use built in pandas resample (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.resample.html) to get OHLC stuff.  

Ultimately I am not opposed to leaving as is, but the mplfinance plots look nicer, so even if you just did a line plot from them it may look nicer?

jmaslek:
> By that you mean to have on command for ath and atl (with parametera like --type [ath, atl] ? Or you mean to combine dataframes into one ?
> Regarding column names, I will change names to more like All time high date [USD]
> ```

So I would leave the -h flag as just "All time high data for loaded coin".  Then give the displayed columns useful names (no need for all of that in the help description).

As for combining them, envision you load a stock quote, you usually get 52 wk high and 52 wk low, so if you had an `AllTime` command it would give you a data frame with ATH and ATL columns.  One could also just give it in a single currency instead of all 3 (so just ath_usd for example), but that's a minor point.

jmaslek:
> So this one is the biggest pain. CoinPaprika, CoinGecko, Binance - all of those API's has different number of coins, different coin names, coin ids, so it's very hard to gues which coin id you need to use for given source. It's not like with stocks or etf or other financial assets where you have with one ticker only one stock.
> 
> E.g. coins with symbol UNI - > there are like 5 coins with this symbol, that's why user should use coins ids in most cases. But here for example in CoinGecko bitcoin has id `bitcoin` but in coinpaprika: `btc-bitcoin`. So that's why for now I added there `cpcoins`, `bincoins`, `cgcoins` and also `find` views to let user find id of coin in which he is interested in.
> 
> I think we should do here brainstorm how to make this mapping matrix. I would say that CoinGecko should be our main source. Because it has the biggest number of coins. And somehow we should glue to it other coin'ids from different sources. But I would say to it will be not easy one.
> 
> I propose to maybe leave it for now. And in next step build this translation table. But here would be great guys if you can help my with some ideas :)

So I agree that this is a pain and something we can brainstorm on.  Having a "base" source like CG would make sense then you can have a conversion dictionary.

Basically my biggest thing was that when I did the load command with --source flag, I got `The following args couldn't be interpreted: ['--source', 'cp']`, so I wasn't sure why the error message was being displayed.

aia (approved):

DidierRLopes:
@jmaslek note for future: Not merging something like #674 which directly impacts a big refactoring :)

jmaslek:
My bad. Thought you approved. 

If it's a pain just install isort then run isort gamestonk_terminal/**/*.py then black ./

DidierRLopes:
> My bad. Thought you approved. 
> 
> If it's a pain just install isort then run isort gamestonk_terminal/**/*.py then black ./

Yes, I did. When I mention "note for future" I meant for both of us. I didn't remember that it would impact this PR.

gamestonk_terminal/cryptocurrency/overview/overview_controller.py
```diff
@@ -0,0 +1,252 @@
...
+        self._overview_parser.add_argument("cmd", choices=self.CHOICES)
+
+    def print_help(self):
+        """Print help"""
+        help_text = """
+Overview:
+    cls         clear screen
+    ?/help      show this menu again
+    q           quit this menu, and shows back to main menu
+    q
```
DidierRLopes, line 62: an extra q in here!

DidierRLopes (commented):
Looks good! Once tests pass, let's merge this!!! 🚀

DidierRLopes (approved):

PR #657, opened 2021-08-09
Refactor discovery menu

DidierRLopes:
This PR includes a refactoring of the discovery menu. Some comments:
- [x] remove simply_wallst, spachero, uwhales from discovery
- [x] cmd popular,emerging back to behavioural analysis (not refactored)
- [x] cmds darkpool,darkshort and shortvol not refactored since they will be moved out.

@jmaslek what if we had a "Dark Pools and Short Data" menu? Where we had not only darkpool,darkshort, and darkpos.. BUT also: shortinterest, lowfloat, shortvol, shortview, dp, ftd?

Seems a more complete menu, what you think?

Also, I want to add this on discovery https://www.pennystockflow.com. Will add in a following PR, I added a "TODO" for it


gamestonk_terminal/stocks/discovery/disc_controller.py
```diff
@@ -150,57 +146,516 @@ def call_quit(self, _):
...
+        )
+        parser.add_argument(
+            "-n",
+            "--num",
+            action="store",
+            dest="num",
+            type=int,
+            default=5,
+            choices=range(1, 25),
+            help="Number of the top gainers stocks to retrieve.",
```
jmaslek, line 253: Can we use that one type function I have that checks n is between two numbers?  Using the range as choices looks bad in the `-h` command?

gamestonk_terminal/stocks/discovery/disc_controller.py
```diff
@@ -150,57 +146,516 @@ def call_quit(self, _):
...
+        )
+        parser.add_argument(
+            "-n",
+            "--num",
+            action="store",
+            dest="num",
+            type=int,
+            default=5,
+            choices=range(1, 25),
+            help="Number of the top losers stocks to retrieve.",
```
jmaslek, line 292: same

jmaslek:
`gainers` : PE ratio shows nan, so maybe fill with " "?  Same with losers

jmaslek:
`arkord` : I don't feel super strongly, but maybe get rid of the finviz link?  One the user has a ticker, they can load it into the stocks menu and do with it what they want?

jmaslek:
`upcoming` : I know we discussed this, but we should think of a way to decrease the number shown.

jmaslek:
I like the idea of the shorts nand dark pools

gamestonk_terminal/stocks/behavioural_analysis/ba_controller.py
```diff
@@ -104,6 +106,10 @@ def print_help():
         print("   queries       top related queries with this stock")
         print("   rise          top rising related queries with stock")
         print("SentimentInvestor:")
+        print("   popularsi     show most popular stocks on social media right now")
```
jmaslek, line 109: `ba: error: argument cmd: invalid choice: 'popularsi' `  Not sure which needs to be changed

jmaslek: Probably in choices  (although would popular make more sense since you already specify the SI source?)

gamestonk_terminal/stocks/discovery/yahoofinance_model.py
```diff
@@ -0,0 +1,31 @@
+""" Yahoo Finance Model """
+__docformat__ = "numpy"
+
+import pandas as pd
+import requests
+
+
+def get_gainers():
```
jmaslek, line 8: `-> pd.DataFrame`

gamestonk_terminal/stocks/discovery/yahoofinance_model.py
```diff
@@ -0,0 +1,31 @@
...
+    -------
+    pd.DataFrame
+        Stock Gainers
+    """
+    url_gainers = "https://finance.yahoo.com/screener/predefined/day_gainers"
+
+    return pd.read_html(requests.get(url_gainers).text)[0]
+
+
+def get_losers():
```
jmaslek, line 21: `-> pd.DataFrame`

tests/test_cryptocurrency/test_coingecko_crypto.py
```diff
@@ -292,13 +291,14 @@ def test_coin_top_nft(self):
...
-        capt = capturedOutput.getvalue()
-        self.assertIn("Metric", capt)
+    # TODO: Fix this test
+    # def test_coin_nft_of_the_day(self):
+    #    capturedOutput = io.StringIO()
+    #    sys.stdout = capturedOutput
+    #    nft_of_the_day([])
+    #    sys.stdout = sys.__stdout__
+    #    capt = capturedOutput.getvalue()
+    #    self.assertIn("Metric", capt)
```
jmaslek, line 301: I did this

aia (approved):

DidierRLopes:
> `upcoming` : I know we discussed this, but we should think of a way to decrease the number shown.

Any suggestions?

jmaslek:
> Any suggestions?

Idk if you wanna randomly just show n of them (I do that with the options screener where you can just do df.sample(n).

DidierRLopes:
> > Any suggestions?
> 
> Idk if you wanna randomly just show n of them (I do that with the options screener where you can just do df.sample(n).

I set upcoming default to 1, so shouldn't be that bad

PR #661, opened 2021-08-12
DCF Generated in Excel

colin99d:
This command generates a highly customizable DCF in excel. It currently users basic linear regression the predict financial statements 10 years out, and then uses those numbers to predict future cash flows. I plan on adding a host of the features in the following PRs to make this tool better for users. The following issues are present in the PR:
-The statement of cash flows data does not always add up (this is an issue with stock analysis and I will address in greater detail in the next PR
-Negative Beta can lead to a negative risk of return (I am looking into methods for fixing this by the next PR
-Stock price can be negative (In the next PR I will make the excel just give the value of assets minus liabilities, not perfect but better than giving a negative estimate)


jmaslek:
Great work!  Couple quick comments:

- Should have a better command name (dcf?)
- All the stuff you are doing should go into a file called {}_model.py  .  So dcf_model or excel_model
- You should be saving the xlsx to a folder, not to `../`
- You updated poetry but not requirements
- You didn't docstring all of your functions
- Top of file should have a docstring + __docformat__ (look at every other file)

I am still not a fan that you don't get any plots or tables displayed to the console.  A major advantage of the terminal is that everything is in one spot, not having to go open up excel after using it (you could write the entire notebook in excel actually with just a variable spot for the ticker and all the cells pulling the data).

Also what is the point of the all_tickers.yaml?  I think @aia would be better suited for this, but just testing either a single ticker or a couple would be fine.

PR #665, opened 2021-08-13
Improved Testing Framework

colin99d:
This pull request adds several improvements to the testing framework:

1. I added VCRPY to allow us to quickly create YAML files to test APIs against. This will speed up testing and remove our reliance on an internet connection for tests
2. I created a decorator that turns any printed python into a variable and then checks that a certain phrase appears in the python to have better assertions in testing.
3. The decorator also ensures that only errors are printed during testing. This makes running tests less confusing for new users.


aia (approved):
Awesome!

PR #667, opened 2021-08-14
Excel Discounted Cash Flow

colin99d:
This command adds the ability to create a DCF in excel. It is a highly customized report that helps to determine the value of a stock.


gamestonk_terminal/stocks/fundamental_analysis/dcf_model.py
```diff
@@ -0,0 +1,841 @@
...
+        self.ws3.column_dimensions["A"].width = 3
+        self.create_header(self.ws1)
+        self.create_header(self.ws2)
+        self.create_header(self.ws3)
+        self.add_estimates()
+        self.create_dcf()
+        if self.audit:
+            self.run_audit()
+
+        trypath = f"../GamestonkTerminal/exports/excel/{self.ticker}-{self.now}.xlsx"
```
DidierRLopes, line 113: So that this is cross platform instead of using "/" we should use os.path.join.

E.g.
```
trypath =os.path.join("..",  "GamestonkTerminal", "exports", "excel", f"{self.ticker}-{self.now}.xlsx")
```

DidierRLopes: So that the files get sorted by ticker and also by date, can we have the name of the file to be:

```
AAPL 2021/08/14 10:20:30
```

So that the order is from big to small.

gamestonk_terminal/stocks/fundamental_analysis/dcf_model.py
```diff
@@ -0,0 +1,841 @@
...
+        if self.years == []:
+            self.years = [x.get_text().strip() for x in columns]
+            self.len_data = len(self.years) - 1
+
+        if self.rounding == 0:
+            phrase = soup.find(
+                "div", attrs={"class": "text-sm pb-1 text-gray-600"}
+            ).get_text()
+            if "thousand" in phrase:
+                self.rounding = 1000
```
DidierRLopes, line 152: you can do 1_000, 1_000_000 and 1_000_000_000 :)

gamestonk_terminal/stocks/fundamental_analysis/dcf_model.py
```diff
@@ -0,0 +1,841 @@
...
+
+    try:
+        ns_parser = parse_known_args_and_warn(parser, other_args)
+        if not ns_parser:
+            return
+
+        dcf_view = CreateExcelFA(ticker, ns_parser.audit)
+        trypath = dcf_view.create_workbook()
+        print(
+            f"Analysis successfully ran for {ticker}\nPlease look in {trypath} for the file."
```
DidierRLopes, line 62: Can we add a new line here at the end?

DidierRLopes (commented):
Really good, this looks really nice. Couple of comments architecture-wise.

- Can we have dcf_model to be dcf_view?
- Can we have a dcf_model that contains all the variables and helper files from excel folder? This way we don't need to add an extra folder. Ideally extra folders should be added when they represent a sub-menu inside.
- Once you have such dcf_model, any helper model that does regression or something like that should be in there. And dcf_view should pick up these and generate the excel.
- Also, a folder is not tracked by git. So when I tested this, it failed because there was no excel file. Can you add a file called ".gitkeep" onto the excels folder and then push that file? This is a cheeky way of pushing a folder to the repo.

colin99d:
> Really good, this looks really nice. Couple of comments architecture-wise.
> 
> * Can we have dcf_model to be dcf_view?
> * Can we have a dcf_model that contains all the variables and helper files from excel folder? This way we don't need to add an extra folder. Ideally extra folders should be added when they represent a sub-menu inside.
> * Once you have such dcf_model, any helper model that does regression or something like that should be in there. And dcf_view should pick up these and generate the excel.
> * Also, a folder is not tracked by git. So when I tested this, it failed because there was no excel file. Can you add a file called ".gitkeep" onto the excels folder and then push that file? This is a cheeky way of pushing a folder to the repo.

All the helper models that do regression also directly add to the excel. This is something that I was going to change for PR2. Is this something that needs changed before this PR? It will take a while.

DidierRLopes:
> > Really good, this looks really nice. Couple of comments architecture-wise.
> > 
> > * Can we have dcf_model to be dcf_view?
> > * Can we have a dcf_model that contains all the variables and helper files from excel folder? This way we don't need to add an extra folder. Ideally extra folders should be added when they represent a sub-menu inside.
> > * Once you have such dcf_model, any helper model that does regression or something like that should be in there. And dcf_view should pick up these and generate the excel.
> > * Also, a folder is not tracked by git. So when I tested this, it failed because there was no excel file. Can you add a file called ".gitkeep" onto the excels folder and then push that file? This is a cheeky way of pushing a folder to the repo.
> 
> All the helper models that do regression also directly add to the excel. This is something that I was going to change for PR2. Is this something that needs changed before this PR? It will take a while.

No, the moving of some elements of _view to _model, aren't a priority, so that's fine as is.

However, if you can rename dcf_model to dcf_view; And then put all the contents of the files on that excel folder into a dcf_model, that would be perfect for this PR and we could merge then :)

gamestonk_terminal/stocks/fundamental_analysis/dcf_model.py
```diff
@@ -1,841 +1,281 @@
-""" DCF Model """
+""" Excel Model """
```
DidierRLopes, line 1: *DCF Model

gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
```diff
@@ -0,0 +1,841 @@
+""" DCF Model """
```
DidierRLopes, line 1: *DCF View

gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
```diff
@@ -0,0 +1,841 @@
...
+from openpyxl.styles import Font
+from sklearn.linear_model import LinearRegression
+from bs4 import BeautifulSoup
+import yfinance as yf
+import pandas as pd
+import numpy as np
+import requests
+
+from gamestonk_terminal.helper_funcs import parse_known_args_and_warn
+import gamestonk_terminal.stocks.fundamental_analysis.dcf_model as md
```
DidierRLopes, line 21: can we keep this as "dcf_model" ? Instead of using a shorten version?

DidierRLopes (approved):
Amazing!

@jmaslek I'm happy with this!! If you are too, let's merge it!

PR #669, opened 2021-08-15
Implement news command on main stocks menu

DidierRLopes:
As per our discussion I added the news command at top of stocks. I was thinking about doing the same for crypto and options, the reasons I didn't are:
- Adding it to options would be redundant. Even more so because we can switch between these menus.
- Crypto has a big refactoring going on, adding it now would make the merging a pain.

@JakubPluta this news function on the common folder can be used on the main crypto menu


gamestonk_terminal/common/newsapi_view.py
```diff
@@ -0,0 +1,48 @@
...
+    s_from = (datetime.now() - timedelta(days=7)).strftime("%Y-%m-%d")
+
+    response = requests.get(
+        f"https://newsapi.org/v2/everything?q={term}&from={s_from}"
+        f"&sortBy=publishedAt&language=en&apiKey={cfg.API_NEWS_TOKEN}",
+    )
+
+    # Check that the API response was successful
+    if response.status_code != 200:
+        print("Invalid News API token\n")
```
jmaslek, line 30: should print out the error code, since there can be other errors than authentication (which is probably the most likely tbf) 

`f"Error in request {response.status_code}.  Check News API token"`

gamestonk_terminal/common/newsapi_view.py
```diff
@@ -0,0 +1,48 @@
...
+
+    else:
+        print(
+            f"{response.json()['totalResults']} news articles from {term} were found since {s_from}\n"
+        )
+
```
jmaslek, line 36: "news articles for {term}"

gamestonk_terminal/common/newsapi_view.py
```diff
@@ -0,0 +1,48 @@
...
+        for idx, article in enumerate(response.json()["articles"]):
+            print(
+                article["publishedAt"].replace("T", " ").replace("Z", ""),
+                " ",
+                article["title"],
+            )
+            # Unnecessary to use name of the source because contained in link article["source"]["name"]
+            print(article["url"])
+            print("")
+
```
jmaslek, line 46: `print(article["url"]. "\n)` ?

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -125,20 +132,22 @@ def print_help(self):
...
+>   report      generate automatic report,   \t e.g.: dark pool, due diligence"""
+
+        help_text += f"""{Style.DIM if not self.ticker else ''}
 >   fa          fundamental analysis,    \t e.g.: income, balance, cash, earnings
 >   res         research web page,       \t e.g.: macroaxis, yahoo finance, fool
+>   dd          in-depth due-diligence,  \t e.g.: news, analyst, shorts, insider, sec
 >   ca          comparison analysis,     \t e.g.: historical, correlation, financials
 >   bt          strategy backtester,      \t e.g.: simple ema, ema cross, rsi strategies
 >   ta          technical analysis,      \t e.g.: ema, macd, rsi, adx, bbands, obv
 >   ba          behavioural analysis,    \t from: reddit, stocktwits, twitter, google
```
jmaslek, line 146: can't everything be one string?

jmaslek (approved):
small critiques.  Looks good to me

Issue #460, opened 2021-05-24
Wyckoff and Elliot waves


markoreza:
Hi,

when I saw this
https://www.youtube.com/watch?v=Lhf_2gJJS1I&t=151s
I said to myself, Well, this could be a good addition to Gamestonk
Terminal. Then I forgot about it.

Today, I saw another Reddit post about Wyckoff, this one showing the
pattern before stock shoots up. It was in the morning, so I can't find it
any more, well, it's my smooth brain. But then, I said to myself:

If all this Wyckoff theories could be implemented into the Gamestonk
Terminal, so the program would check all stocks for all Wyckoff's
patterns, and then just tell the user which stock has [how many %]
potential to go up or down or whatever - well, this would prolly be an
useful thing...

You know what I mean - so the Terminal would tell its user which stock is
about to what - according to Wyckoff, of course.

More data on Wyckoff:

https://en.wikipedia.org/wiki/Richard_Wyckoff

https://archive.org/search.php?query=Wyckoff%2C%20Richard%20D.%20How%20I%20Trade%20and%20Invest%20in%20Stocks%20and%20Bonds&sin=TXT

http://libgen.lc/ads.php?md5=C45AEA2639529958D3E8A66A313DA9E7

https://school.stockcharts.com/doku.php?id=market_analysis:the_wyckoff_method

... and I think there's some more on the net...

The other thing which should be perhaps included ito the GameStonk
Terminal are Elliot waves. Believe me, I have no idea, but someone is
posting on Reddit quite frquently his price prediction, base on Elliot
Waves Theory. This is supposed to be a good thing, if I remember
correctly.

And, if you implement both into the gameStonk Terminal, then perhaps a
price prediction based on both this system (crossvalue) could be an
interesting think.

Thank you :-)

PR #670, opened 2021-08-15
update ta doc

devwolff:


DidierRLopes (approved):

DidierRLopes:
Nice, thank you!

Not sure if you are keen on, but #460 requests technical indicators: Wyckoff and Elliot Waves. If you want to have a go at that, that would be amazing :)

devwolff:
I don't really use them. From my understanding, they are more rules of thumbs, rather than any specific calculations right?  But I will take a proper crack at them when I have more time

DidierRLopes:
> I don't really use them. From my understanding, they are more rules of thumbs, rather than any specific calculations right? But I will take a proper crack at them when I have more time

Yes, I think you're right. I haven't looked into how to use them myself. 

Don't worry with those if you don't have an interest in them :) Someone that uses these will have a look at them at some point.

PR #677, opened 2021-08-17
add rolling metrics to eda

devwolff:
A small proposal here: Using pandas_ta to include rolling metrics in eda. Though it could be on ta menu, I think it is more suitable to place it in eda as these calculations are definitely not exclusive to ta discourse. Besides, it aligns better with the existing calculations in eda. 

Let me know what you guys think before I go further with this subsection

### DIFF
--- a/gamestonk_terminal/common/exploratory_data_analysis/rolling.py
+++ b/gamestonk_terminal/common/exploratory_data_analysis/rolling.py
@@ -143,20 +143,19 @@ def spread(
 def quantile(
     other_args: List[str], s_ticker: str, s_interval: str, df_stock: pd.DataFrame
 ):
-    """Median & Quantile
+    """Overlay Median & Quantile
 
     Parameters
     ----------
-    other_args:List[str]
+    other_args: List[str]
         Argparse arguments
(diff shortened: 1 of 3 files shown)


DidierRLopes:
I agree with you, and think is a great idea.

I'm away of the laptop today for 7/8h. But once I'm back I'll try these new commands! 😀

DidierRLopes:
In terms of code it looks goods to me!

jmaslek:
I too think this is a good idea.

Could you rename the rolling to rolling_view for now.  

If you wanna be my best friend you can take a look at the recent TA refactoring (I'll also push examples) and change the code so that the argparse is in _controller and there is a _view and _model you can, but I will go ahead and merge this once I confirm nothing strange on my end.  We will be requiring this shortly and you have been moving faster than me :)

jmaslek (approved):

devwolff:
@DidierRLopes @jmaslek I will help to refactor eda menu this weekend, before we add more materials

jmaslek:
@devwolff No worries. I am almost done with it :)

PR #684, opened 2021-08-20
add wma & hma to ta menu

devwolff:
The ta menu is getting a bit thick lol


DidierRLopes:
> The ta menu is getting a bit thick lol

That is fine. Usually we can have sub-menus, but I think for technical analysis it makes sense to have a huge menu with a lot of options. The more the user uses our terminal, the more familiar they get, and at some point they won't even need to look into the help menu but will know commands by heart - like Bloomberg Terminal

devwolff:
Definitely!

DidierRLopes:
#685 adds a fix for the test failing here, which is crypto related. If 685 passes we can merge this followed by 685 :)

DidierRLopes (approved):

PR #686, opened 2021-08-20
Refactor EDA -> Qa

jmaslek:
No new commands.  

Changed menu to quantitative analysis to indicate it is more numerical techniques than eda in a data science sense.


DidierRLopes:
From testing the command (without looking at code):
- Summary missing new line.
- I get this with `hist`: `stat` must be one of ['count', 'density', 'probability', 'frequency'], but proportion was passed.` 
- Something is wrong with rolling graph due to legends and what is displayed
- decompose gives: `/Users/DidierRodriguesLopes/Documents/git/GamestonkTerminal/gamestonk_terminal/common/quantitative_analysis/qa_view.py:376: UserWarning: This figure was using constrained_layout==True, but that is incompatible with subplots_adjust and or tight_layout: setting constrained_layout==False. 
  fig.tight_layout()`

In general I think the menu is a bit confusing because sometimes we perform statistics on the share price, and others on the daily returns.

Could we either work with 1 or the other. I.e. this could be set by the user, under Ticker:. Like,
```
Ticker: AAPL
Analyse: Price
```
Or we could have Analyse: Volume, or even Analyse: Returns.

jmaslek:
> * I get this with `hist`: `stat` must be one of ['count', 'density', 'probability', 'frequency'], but proportion was passed.`

Hmm.  I have no issue.  Are you on seaboard 0.11.2 ?

> * Something is wrong with rolling graph due to legends and what is displayed

So I just copy/pasted code.   I am keeping the dual legends.  honestly this command can probably go since there is spread and you can look at moving averaged in ta

> In general I think the menu is a bit confusing because sometimes we perform statistics on the share price, and others on the daily returns.
> 
> Could we either work with 1 or the other. I.e. this could be set by the user, under Ticker:. Like,
> 
> ```
> Ticker: AAPL
> Analyse: Price
> ```
> 
> Or we could have Analyse: Volume, or even Analyse: Returns.

So I didn't add returns to the last 3 commands (since I don't completely follow what they do from when you added them and if it would make sense to do those analysis on returns - also acf you do returns (what you called diff) and 'price acceleration' so those are default)

Since I didn't touch the new rolling features from last week (which only look at price) would it be better for you if I add a [default: Returns] to the commands that do those?

I was also thinking I could be a little better structured and have something like 
```python
daily Stock: AAPL (from 2020-08-19)
                 summary
Quantitative Analysis:
    cls         clear screen
    help        show this menu again
    q           quit this menu, and shows back to main menu
    quit        quit to abandon program
    
Plots
    hist        histogram with density plot
    cdf         cumulative distribution function
    bw          box and whisker plot
      acf         (partial) auto-correlation function differentials
Rolling Metrics:
    rolling     rolling mean and std deviation
    spread      variance and std deviation
    quantile    median and quantile
    skew        skewness of distribution
Some other title:    
    decompose   decomposition in cyclic-trend, season, and residuals
    cusum       detects abrupt changes using cumulative sum algorithm
  ```

DidierRLopes (approved):

PR #696, opened 2021-08-21
Adding a container running terminal and jupyter

aia:


docker/entry-point-poetry-jupyter.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# stty erase ^h
+poetry run jupyter notebook --ip=0.0.0.0 &
```
DidierRLopes, line 4: Shouldn't this be 8.8.8.8?

DidierRLopes (approved):

docker/entry-point-poetry-jupyter.sh
aia: No, 0.0.0.0 means listen on all addresses

DidierRLopes: Got it!

PR #697, opened 2021-08-21
Improved Discount Rate Calculation for DCF

colin99d:
- Allowed for selection of CAPM or Fama French for risk valuation
- Disallowed r from going below 0.5%
- Pulled T-Bill rates from Fed API
- If firm value is negative replace value with A-L
- Added warning that Cash Flow Statement might not tie-out


gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
```diff
@@ -51,19 +51,19 @@ def dcf(other_args: List[str], ticker: str):
...
-        ns_parser = parse_known_args_and_warn(parser, other_args)
-        if not ns_parser:
-            return
-
-        dcf_view = CreateExcelFA(ticker, ns_parser.audit)
-        trypath = dcf_view.create_workbook()
-        print(
-            f"Analysis successfully ran for {ticker}\nPlease look in {trypath} for the file.\n"
-        )
+    # try:
```
DidierRLopes, line 54: You should keep the try here, to avoid the terminal crashing in case there's an issue

DidierRLopes (commented):
Nice!

Can we change saved file to:
```
TICKER YEAR-MONTH-DAY HOUR:MINUTES:SECONDS
```
?

E.g.
```
../GamestonkTerminal/exports/excel/AAPL 2021-08-21 21:45:12.xlsx
```

Also, the print has the format `hh:mm:ss` but the saved file has `hh/mm/ss`

colin99d:
> Nice!
> 
> Can we change saved file to:
> 
> ```
> TICKER YEAR-MONTH-DAY HOUR:MINUTES:SECONDS
> ```
> 
> ?
> 
> E.g.
> 
> ```
> ../GamestonkTerminal/exports/excel/AAPL 2021-08-21 21:45:12.xlsx
> ```
> 
> Also, the print has the format `hh:mm:ss` but the saved file has `hh/mm/ss`

This is actually a weird file naming issue. The file is saved with hh:mm:ss but I guess that cannot be show, so the system replace it with hh/mm/ss. However, if you open up the file you will see it is saved as hh:mm:ss at the top of your excel spreadsheet.

DidierRLopes:
> > Nice!
> > Can we change saved file to:
> > ```
> > TICKER YEAR-MONTH-DAY HOUR:MINUTES:SECONDS
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> > E.g.
> > ```
> > ../GamestonkTerminal/exports/excel/AAPL 2021-08-21 21:45:12.xlsx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also, the print has the format `hh:mm:ss` but the saved file has `hh/mm/ss`
> 
> This is actually a weird file naming issue. The file is saved with hh:mm:ss but I guess that cannot be show, so the system replace it with hh/mm/ss. However, if you open up the file you will see it is saved as hh:mm:ss at the top of your excel spreadsheet.

Since the spreadsheet contains stock analysis data, which should be updated daily. Can we remove the time? And keep the date only?

colin99d:
> > > Nice!
> > > Can we change saved file to:
> > > ```
> > > TICKER YEAR-MONTH-DAY HOUR:MINUTES:SECONDS
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ?
> > > E.g.
> > > ```
> > > ../GamestonkTerminal/exports/excel/AAPL 2021-08-21 21:45:12.xlsx
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Also, the print has the format `hh:mm:ss` but the saved file has `hh/mm/ss`
> > 
> > 
> > This is actually a weird file naming issue. The file is saved with hh:mm:ss but I guess that cannot be show, so the system replace it with hh/mm/ss. However, if you open up the file you will see it is saved as hh:mm:ss at the top of your excel spreadsheet.
> 
> Since the spreadsheet contains stock analysis data, which should be updated daily. Can we remove the time? And keep the date only?

The only reason I have the time is so that a new file does not override an old one. How else would I get around this?

DidierRLopes:
> > > > Nice!
> > > > Can we change saved file to:
> > > > ```
> > > > TICKER YEAR-MONTH-DAY HOUR:MINUTES:SECONDS
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > ?
> > > > E.g.
> > > > ```
> > > > ../GamestonkTerminal/exports/excel/AAPL 2021-08-21 21:45:12.xlsx
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Also, the print has the format `hh:mm:ss` but the saved file has `hh/mm/ss`
> > > 
> > > 
> > > This is actually a weird file naming issue. The file is saved with hh:mm:ss but I guess that cannot be show, so the system replace it with hh/mm/ss. However, if you open up the file you will see it is saved as hh:mm:ss at the top of your excel spreadsheet.
> > 
> > 
> > Since the spreadsheet contains stock analysis data, which should be updated daily. Can we remove the time? And keep the date only?
> 
> The only reason I have the time is so that a new file does not override an old one. How else would I get around this?

You can check if the file exists a priori. If it does you can add a message "The file blablabla already exists. Hence, it won't be generated". Then the user will know that he has to deal with it.

DidierRLopes (approved):

PR #701, opened 2021-08-22
Improve contributing doc

DidierRLopes:


jmaslek (approved):

devwolff:
very well documented. Good work. 

What about functions in _view that could return a table and a plot. We'll just stick with plot_ then?

DidierRLopes:
> very well documented. Good work. 
> 
> 
> 
> What about functions in view files that could return multiple values like a table and a plot. We'll just stick with plot_ then?

Wasn't expecting that, really good question. 

What do you reckon? 

@jmaslek thoughts?

JakubPluta:
@DidierRLopes Very well documented. Good job 👍

jmaslek:
> 
> What do you reckon? 
> 
> @jmaslek thoughts?

So I thought we said all functions in view should be view_{}?

Anyways there are 2 cases.  First is when table data = plotted data.  In this case we have been doing --raw to show the table + --export.  If you want to show table by default we can decide on a universal flag.

In the second case when the data is different they should have 2 different view_ functions. 

Does this answer your question @devwolff or did you have a specific example?

DidierRLopes:
> > 
> > What do you reckon? 
> > 
> > @jmaslek thoughts?
> 
> So I thought we said all functions in view should be view_{}?
> 
> Anyways there are 2 cases.  First is when table data = plotted data.  In this case we have been doing --raw to show the table + --export.  If you want to show table by default we can decide on a universal flag.
> 
> In the second case when the data is different they should have 2 different view_ functions. 
> 
> Does this answer your question @devwolff or did you have a specific example?
> 
> 

Yea, prob too much confusion. Let's have display for all.

The reason beinf that '_view.view_' is rather ugly. So we can do _view.display_ instead

jmaslek (approved):

devwolff:
> > What do you reckon?
> > @jmaslek thoughts?
> 
> So I thought we said all functions in view should be view_{}?
> 
> Anyways there are 2 cases. First is when table data = plotted data. In this case we have been doing --raw to show the table + --export. If you want to show table by default we can decide on a universal flag.
> 
> In the second case when the data is different they should have 2 different view_ functions.
> 
> Does this answer your question @devwolff or did you have a specific example?

I totally agree

PR #702, opened 2021-08-22
Features website with Hugo

DidierRLopes:


DidierRLopes:
Merging this as a starting point, as I want to test the deployment of this Hugo server

PR #707, opened 2021-08-23
fix scale for better volume plotting

devwolff:


DidierRLopes (approved):

gamestonk_terminal/common/technical_analysis/volume_view.py
```diff
@@ -16,7 +17,7 @@
 
 def plot_ad(
     s_ticker: str, s_interval: str, df_stock: pd.DataFrame, use_open: bool, export: str
-):
+) -> pd.DataFrame:
```
DidierRLopes, line 20: This is not the case. this function is not returning any dataframe, but plotting it

gamestonk_terminal/common/technical_analysis/volume_view.py
```diff
@@ -206,7 +213,9 @@ def plot_adosc(
     )
 
 
-def plot_obv(s_ticker: str, s_interval: str, df_stock: pd.DataFrame, export: str):
+def plot_obv(
+    s_ticker: str, s_interval: str, df_stock: pd.DataFrame, export: str
+) -> pd.DataFrame:
```
DidierRLopes, line 218: Same here

jmaslek (approved):

PR #709, opened 2021-08-23
fix black on accidental push

jmaslek:
this should not have all the commits I had...


jmaslek:
merging since no effect on user.  still learning the subtleties of git.

PR #715, opened 2021-08-25
add offset feature to all filters

devwolff:
- add offset feature to all available filters
- changed window_length to length

### DIFF
--- a/gamestonk_terminal/common/technical_analysis/overlap_model.py
+++ b/gamestonk_terminal/common/technical_analysis/overlap_model.py
@@ -5,7 +5,9 @@ import pandas as pd
 import pandas_ta as ta
 
 
-def ema(s_interval: str, df_stock: pd.DataFrame, window_length: int) -> pd.DataFrame:
+def ema(
+    s_interval: str, df_stock: pd.DataFrame, length: int, offset: int
+) -> pd.DataFrame:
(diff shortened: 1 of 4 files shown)


jmaslek:
So I will leave this up to @DidierRLopes.  I removed all of these offsets intentionally as they are not useful.  

If you go into any trading software and add an any MA, you can select length only.

devwolff:
@jmaslek I think having offset is important in dealing with digital filters especially when it comes to perceiving their lag values. I hear you though

jmaslek:
I mean it costs nothing to have it there. Just as a technical indicator any source is gonna give you the offset 0.

As a digital filter, absolutely you care about lags and you can go into more depth with different arima etc models -- which is something that needs to be developed better in the terminal if that is of interest to you :)

DidierRLopes:
> So I will leave this up to @DidierRLopes. I removed all of these offsets intentionally as they are not useful.
> 
> If you go into any trading software and add an any MA, you can select length only.

Probably won't be our most used argument but I don't see reasons to to not have it over having it :)

DidierRLopes (changes requested):
Captura de ecrã 2021-08-25, às 17 34 09

That chunk of code should be changed to this:
Captura de ecrã 2021-08-25, às 17 38 54

PR #718, opened 2021-08-26
Introduce Jupyter Lab extensions

piiq:
The PR is created to frame the discussion and the work on the introduction of Jupyter Lab extensions that help work with the Terminal.

The initial idea is to have 3 extensions added to the Jupyter Lab launcher panel:
1. The terminal launcher
2. The settings editor
3. The documentation link

---


piiq:
@aia I've updated the mypy command that is executed in the hooks and in the test to exclude `setup.py`. The jupyterlab extensions come with such files for but the extension folder is not a valid python module - so excluding them from check is the only way to get past mypy.

After I pass the mypy check I get stuck on `pyupgrade` (and it also complains about the same files). Can you help me out figuring out what to do to dix this?
image

DidierRLopes:
> @aia I've updated the mypy command that is executed in the hooks and in the test to exclude `setup.py`. The jupyterlab extensions come with such files for but the extension folder is not a valid python module - so excluding them from check is the only way to get past mypy.
> 
> After I pass the mypy check I get stuck on `pyupgrade` (and it also complains about the same files). Can you help me out figuring out what to do to dix this?
> image

Can you try:
```
pyupgrade --py36-plus jupyterlab/gst-settings/setup.py
pyupgrade --py36-plus jupyterlab/gst/setup.py
```

aia:
@piiq pyupgrade will rewrite files for you.

Just run
```
pyupgrade --py36-plus **/*.py
```

Look at the diff
```
git diff
```

Commit if acceptable.

aia (approved):

PR #729, opened 2021-09-01
Bump tf to 2.6.0 to avoid security issue

jmaslek:
`pred` still works on my end after this update

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,10 +64,10 @@ oandapyV20 = "^0.6.3"
 valinvest = "^0.0.2"
 bt = "^0.2.9"
 python-binance = "^1.0.1"
-grpcio = "1.34.0"
+grpcio = "^1.34.0"
 pyupgrade = "2.18.2"
 cvxpy = "1.1.12"
-tensorflow = {version = "2.5.0", optional = true}
+tensorflow = {version = "2.6.0", optional = true}
 jupyter-client = "6.1.12"
 vaderSentiment = "3.3.2"
 degiro-connector = "1.0.0"
@@ -94,7 +94,7 @@ pyupgrade = "^2.11.0"
 Sphinx = "4.1.1"
 
 [tool.poetry.extras]
-prediction = ["tensorflow", "pmdarima", "transformers", "torch"]
+prediction = ["tensorflow", "pmdarima"]
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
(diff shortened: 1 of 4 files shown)


jmaslek:
need to bump pywin back to 300

pyproject.toml
```diff
@@ -94,7 +94,7 @@ pyupgrade = "^2.11.0"
 Sphinx = "4.1.1"
 
 [tool.poetry.extras]
-prediction = ["tensorflow", "pmdarima", "transformers", "torch"]
+prediction = ["tensorflow", "pmdarima"]
```
aia, line 97: Are we removing transformers and torch?

requirements-full.txt
```diff
@@ -162,15 +163,15 @@ python-dotenv==0.19.0; python_version >= "3.5"
 pytrends==4.7.3
 pytz==2021.1; python_full_version >= "3.6.8" and python_version >= "3.6" and python_full_version < "4.0.0"
 pyupgrade==2.18.2; python_full_version >= "3.6.1"
-pywin32==300; sys_platform == "win32" and python_version >= "3.6"
+pywin32==301; sys_platform == "win32" and python_version >= "3.6"
```
aia, line 166: you need to regenerate requirements-full.txt

requirements.txt
```diff
@@ -148,15 +148,15 @@ python-dotenv==0.19.0; python_version >= "3.5"
 pytrends==4.7.3
 pytz==2021.1; python_full_version >= "3.6.8" and python_version >= "3.6" and python_full_version < "4.0.0"
 pyupgrade==2.18.2; python_full_version >= "3.6.1"
-pywin32==300; sys_platform == "win32" and python_version >= "3.6"
+pywin32==301; sys_platform == "win32" and python_version >= "3.6"
```
aia, line 151: And requirements.txt

pyproject.toml
jmaslek: These have been long gone (whenever I added vader for twitter)

requirements-full.txt
jmaslek: I regenerated after running poetry update them, but now I can't get poetry to revert pywin32 `Unable to find installation candidates for pywin32 (300)`.  

pip install pywin32==300 also fails, so I may just close this for now.

aia: I can do a big dependency update PR.

PR #730, opened 2021-09-01
Updating dependencies and cleaning up

aia:
- Updated all relevant dependencies
- Cleaned up prediction extras
- Updated pylint
- Added jupyterlab
- pywin301 is no longer broken... Go figure.


jmaslek (approved):
thanks.  should  (or even can) we have a separate optional for jupyterlab?

aia:
> thanks. should (or even can) we have a separate optional for jupyterlab?

We can. I was thinking about it.

PR #733, opened 2021-09-03
reuse patch_pandas_text_adjustment in ca stocks menu

devwolff:
without tabulate, the text alignment is off


devwolff:
> Can't you just do this once before the print(df.to_string())?

It looks like it is caused by the coloring features. It works fine when USE_COLOR is set to False

DidierRLopes (approved):

PR #737, opened 2021-09-08
Add docker documentation

jmaslek:
This PR adds the documentation for pulling the latest tag from the ghcr.io and running.

The DOCKER_ADVANCED goes into the env_files and the X Server.  My description of setting up the x server may be kind of confusing, but I could not get it to work with a single DISPLAY env variable, so feel free to PR on this with alternative approaches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Open the docker desktop app in this case.
 docker pull ghcr.io/gamestonkterminal/gst-poetry:latest
 ```
 
-Upon running this the first time, you should see the various layers downloading (note the random series of letters numbers will vary).
+Upon running this the first time, you should see the various layers downloading (note the random series of letters numbers will vary).  The first time this is run, it will take a few minutes.  Subsequent updates will be much faster, as the changes will be in the MB instead of GB.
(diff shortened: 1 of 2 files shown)


aia (approved):

README.md
```diff
@@ -86,6 +87,41 @@ As a modern Python-based environment, GamestonkTerminal opens access to numerous
...
+```
+Open the docker desktop app in this case.
+
+1. Download the latest docker image.
+
+```bash
+docker pull ghcr.io/gamestonkterminal/gst-poetry:latest
+```
+
+Upon running this the first time, you should see the various layers downloading (note the random series of letters numbers will vary).
```
aia, line 106: I would also add that the first download will take time, but the subsequent downloads will be significantly faster.

README.md
```diff
@@ -86,6 +87,41 @@ As a modern Python-based environment, GamestonkTerminal opens access to numerous
...
+
+You are now ready to run the terminal.
+
+```docker run -it --rm ghcr.io/gamestonkterminal/gst-poetry:latest```
+
+This will open up the terminal in your command prompt or terminal.  Note that this has provided now environment file, so you will not be able to view plots or use keys at this stage.
+
+To read more on adding the environment keys and how to configure your X-server to show plots, hop over to the [Advanced Docker Setup](/DOCKER_ADVANCED.md).
+
 ### Install
```
aia, line 125: I would also update this header to something like - Install with local Python, Git and Python dependencies

PR #739, opened 2021-09-10
Allow economy menu to save figures

DidierRLopes:
I also used USE_TABULATE_DF wherever I could :)


jmaslek (commented):
What about the case where there is table data that we want to both export and save figure?

I think we should have a separate save figure command.

DidierRLopes:
> What about the case where there is table data that we want to both export and save figure?
> 
> I think we should have a separate save figure command.

No need for another command, this should do.

I tested with rtps by being able to output both csv and jpg with:
```
rtps --export csv,jpg
```

Note that I had to change the code to test for this, namely `if raw:` became `if 1:`

DidierRLopes:
Merging this since this will be useful for developing Economy Context on Discord Bot in https://github.com/GamestonkTerminal/DiscordBot

PR #741, opened 2021-09-11
Implement Jupyter Lab container

aia:


piiq:
@aia there's one thing I'd like to discuss.

The `lab-settings.json` is copied into the container to overwrite the settings schema. As far as I understand the `schema.json` holds the settings structure and the default values but is not the same file as the user settings.
The user settings for the extension are being saved to the `~/.jupyter/lab/user-settings/@gamestonk/settings/settings.jupyterlab-settings`. In our case by default this file contains the same info as the `schema`. If copying the `lab-settings.json` is done to keep user settings - it should be copied to a different location.
On the other hand if the user updates the settings the update will be save inside the container and will not be available on the local filesystem where it can be stored persistently between if the container re-builds. Do you think we should mount the `lab-settings.json` inside the container instead of copying it?

piiq:
Aaaaand another couple of things:

1. The default lab ACL requires a token. When I start the container with the `-d` flag the token can be obtained from `docker logs <containerID>` which is not straightforward. We have some options to get the token:
   - We can run the container interactively (`-it`) and see all the links produced by the console
   - We can set a default token or password or even disable access control altogether. The launch command would be based on this `jupyter lab --ServerApp.token="${DEFAULT_TOKEN}" --ServerApp.password="${DEFAULT_PASSWORD}"`
   - We can mount the file that launches the lab (`file:///home/python/.local/share/jupyter/runtime/jpserver-*-open.html`) to the host FS.
What do you think about this?

2. Viewing the charts requires X forwarding to the host machine. The default primary x-server display in a docker env is `host.docker.internal:0`. Should we add this the default setting to the `setenv`? In this case we would get rid of the `IP` variable setting that described in [DOCKER_ADVANCED](https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/DOCKER_ADVANCED.md) and the launch command sequence for macOS would look like this:
```bash
docker pull ghcr.io/gamestonkterminal/gst-poetry-jupyter:latest
xhost + 127.0.0.1
docker run -it -p 8888:8888 --env-file=docker/setenv ghcr.io/gamestonkterminal/gst-poetry-jupyter:latest
```

aia:
> 
> 
> @aia there's one thing I'd like to discuss.
> 
> The `lab-settings.json` is copied into the container to overwrite the settings schema. As far as I understand the `schema.json` holds the settings structure and the default values but is not the same file as the user settings.
> The user settings for the extension are being saved to the `~/.jupyter/lab/user-settings/@gamestonk/settings/settings.jupyterlab-settings`. In our case by default this file contains the same info as the `schema`. If copying the `lab-settings.json` is done to keep user settings - it should be copied to a different location.
> On the other hand if the user updates the settings the update will be save inside the container and will not be available on the local filesystem where it can be stored persistently between if the container re-builds. Do you think we should mount the `lab-settings.json` inside the container instead of copying it?

lab-settings.json here is used to only build a functioning image. In order for the lab launcher to work, I need to set PYTHON to "poetry run python" in the image.

We will have a different process for managing settings.json in the container.

aia:
> 
> 
> Aaaaand another couple of things:
> 
>     1. The default lab ACL requires a token. When I start the container with the `-d` flag the token can be obtained from `docker logs <containerID>` which is not straightforward. We have some options to get the token:
>        
>        * We can run the container interactively (`-it`) and see all the links produced by the console
>        * We can set a default token or password or even disable access control altogether. The launch command would be based on this `jupyter lab --ServerApp.token="${DEFAULT_TOKEN}" --ServerApp.password="${DEFAULT_PASSWORD}"`
>        * We can mount the file that launches the lab (`file:///home/python/.local/share/jupyter/runtime/jpserver-*-open.html`) to the host FS.
>          What do you think about this?

Yes. I was looking into that. I will try ServerApp.token. That looks very convenient.

Right now you can get the token by running docker run -d, docker logs <container id>.

> 
>     2. Viewing the charts requires X forwarding to the host machine. The default primary x-server display in a docker env is `host.docker.internal:0`. Should we add this the default setting to the `setenv`? In this case we would get rid of the `IP` variable setting that described in [DOCKER_ADVANCED](https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/DOCKER_ADVANCED.md) and the launch command sequence for macOS would look like this:
> 
> 
> ```shell
> docker pull ghcr.io/gamestonkterminal/gst-poetry-jupyter:latest
> xhost + 127.0.0.1
> docker run -it -p 8888:8888 --env-file=docker/setenv ghcr.io/gamestonkterminal/gst-poetry-jupyter:latest
> ```

I'll check this out.

Another thing is SSL. I will probably add a self-signed cert and start the lab with SSL.

DidierRLopes (approved):

piiq:
> lab-settings.json here is used to only build a functioning image. In order for the lab launcher to work, I need to set PYTHON_BINARY to "poetry run python" in the image.

@aia if you want you can override the defaults in the `jupyter/gst-settings/schema/settings.json` making the container settings default for the extension

piiq (approved):

PR #749, opened 2021-09-14
Added start date argument to newsapi

Flodur871:
I added an argument to specify date start for the news command, as was described in the ```# TODO```.

### DIFF
--- a/gamestonk_terminal/common/newsapi_view.py
+++ b/gamestonk_terminal/common/newsapi_view.py
@@ -35,7 +35,10 @@ def news(term: str, num: int, s_from: str):
     )
 
     # Check that the API response was successful
-    if response.status_code != 200:
+    if response.status_code == 426:
+        print(f"Error in request: {response.json()['message']}", "\n")
+
+    elif response.status_code != 200:
         print(f"Error in request {response.status_code}. Check News API token", "\n")
 
     else:
--- a/gamestonk_terminal/common/newsapi_view.py
+++ b/gamestonk_terminal/common/newsapi_view.py
@@ -1,6 +1,8 @@
 """ News View """
 __docformat__ = "numpy"
 
+from datetime import datetime, timedelta
+
 import requests
 
 from gamestonk_terminal import config_terminal as cfg
@@ -20,6 +22,13 @@ def news(term: str, num: int, s_from: str):
     """
     # TODO: Add argument to specify news source being used
 
+    if s_from:
+        show_newest = False
+
+    else:
+        show_newest = True
+        s_from = (datetime.now() - timedelta(days=7)).strftime("%Y-%m-%d")
+
     response = requests.get(
         f"https://newsapi.org/v2/everything?q={term}&from={s_from}"
         f"&sortBy=publishedAt&language=en&apiKey={cfg.API_NEWS_TOKEN}",
@@ -34,7 +43,13 @@ def news(term: str, num: int, s_from: str):
             f"{response.json()['totalResults']} news articles for {term} were found since {s_from}\n"
         )
 
-        for idx, article in enumerate(response.json()["articles"]):
+        if show_newest:
+            articles = response.json()["articles"]
+
+        else:
+            articles = response.json()["articles"][::-1]
+
+        for idx, article in enumerate(articles):
             print(
                 article["publishedAt"].replace("T", " ").replace("Z", ""),
                 " ",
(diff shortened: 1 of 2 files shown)


jmaslek:
So when I run `news` I get the following error:

```
datetime.datetime' object has no attribute stftime
 ```

I think that should be strftime. 

Other than that, good work and thanks !

File: gamestonk_terminal/stocks/stocks_controller.py [hunk 1]
```diff
@@ -233,14 +235,25 @@ def call_news(self, other_args: List[str]):
...
+            type=valid_date,
+            default=datetime.now() - timedelta(days=7),
+            help="The starting date (format YYYY-MM-DD) to search articles from",
+        )
         try:
             ns_parser = parse_known_args_and_warn(parser, other_args)
             if not ns_parser:
                 return
 
+            s_date = ns_parser.n_start_date.stftime("%Y-%m-%d")
```
jmaslek, line 252: *strftime

Flodur871: My bad

jmaslek (approved):

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -233,14 +235,25 @@ def call_news(self, other_args: List[str]):
...
         try:
             ns_parser = parse_known_args_and_warn(parser, other_args)
             if not ns_parser:
                 return
 
+            s_date = ns_parser.n_start_date.strftime("%Y-%m-%d")
             newsapi_view.news(
                 term=self.ticker,
                 num=ns_parser.n_num,
+                s_from=s_date,
```
DidierRLopes, line 256: You can add here the line that you have above since it's only used here, there's no need to create an s_date variable :)

DidierRLopes (commented):
A possible improvement is to allow to sort order of news shown. E.g. let's say I want to check the news that occurred on 2021-08-08. There's no easy way to do that, because you basically select start time, but then only get shown the "tail" of the news.

Also, just noticed that the free plan of NewsAPI only allows 1 month of news in the past. 
```
{"status":"error","code":"parameterInvalid","message":"You are trying to request results too far in the past. Your plan permits you to request articles as far back as 2021-08-13, but you have requested 2021-07-13. You may need to upgrade to a paid plan."}
```

aia (approved):

gamestonk_terminal/stocks/stocks_controller.py
Flodur871: Will do

DidierRLopes: @Flodur871 should we merge what you have here? Or do you want to push this on this same PR?

Flodur871: It's almost ready, so I think it would be better to push it here

gamestonk_terminal/common/newsapi_view.py
```diff
@@ -17,26 +17,42 @@ def news(term: str, num: int):
...
     # TODO: Add argument to specify news source being used
-    # TODO: Add argument to specify date start to search for articles
-    s_from = (datetime.now() - timedelta(days=7)).strftime("%Y-%m-%d")
+
+    if s_from:
+        show_newest = False
+
+    else:
+        show_newest = True
+        s_from = (datetime.now() - timedelta(days=7)).strftime("%Y-%m-%d")
```
jmaslek, line 30: So I think the best way to do this would be to have these as default args.  basically add to the function:

```
def news(term: str, num: int, s_from: str=(datetime.now() - timedelta(days=7)).strftime("%Y-%m-%d"), show_newest:bool = True):
```
This way you can just have a show_newest flag in addition to the start date  and they aren't coupled together.

Flodur871: Good point

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -233,6 +235,25 @@ def call_news(self, other_args: List[str]):
...
+        )
+        parser.add_argument(
+            "-s",
+            "--sort",
+            action="store",
+            choices=["newest", "oldest"],
+            dest="n_sort",
+            default="newest",
+            help="Sort by newest articles first or oldest first (default newest)",
+        )
```
jmaslek, line 255: You should have this be a true/false to indicate reverse instead of using ns_parser.n_sort=="newest".  So this would be default=True and action = "store_false"

Flodur871: Got you

DidierRLopes:
Good work!

Issue #746, opened 2021-09-13
[IMPROVE] Shorthand notation inputs for main menu entries


sug4ndh:
**What's the feature that should be improved?**
In the main menu, perhaps, the entries can have a shorthand notation for input? For e.g. instead of entering 'stocks', maybe just 's' should present the stocks menu.

**Describe how you would like the feature improved**
Right now, the users have to input the full word in the main menu like stocks, crypto, etc. Having a shorthand notation like s for stocks, c for crypto would be relatively faster.

jmaslek:
#751 closes this.  Thanks for the input!

PR #751, opened 2021-09-14
Allow for shorthand menu input

jmaslek:
This addresses #746 and allows for (example) s to be input instead of stocks.

Also commented out forex eda stuff so that the menu doesn't crash. Added TODO of incorporating common qa menu.

#750 was not on main branch so I am doing this instead.

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -164,9 +164,7 @@ What do you want to do?
 
     def call_s(self, _):
         """Process stocks command"""
-        from gamestonk_terminal.stocks import stocks_controller
-
-        return stocks_controller.menu()
+        return self.call_stocks(_)
 
     def call_crypto(self, _):
         """Process crypto command"""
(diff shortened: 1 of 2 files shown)


terminal.py
```diff
@@ -160,24 +162,48 @@ def call_stocks(self, _):
 
         return stocks_controller.menu()
 
+    def call_s(self, _):
+        """Process stocks command"""
+        from gamestonk_terminal.stocks import stocks_controller
```
DidierRLopes, line 167: Instead of repeating code, can we just do a relay?

E.g.
```
def call_s(self, _):
   call_stocks(self, _)
```

jmaslek: can't do it like that, but yes i relayed

aia (approved):

DidierRLopes (approved):

PR #753, opened 2021-09-19
Improve docker docs

DidierRLopes:
Note the command change to search for docker display.

In my case the address wasn't on en1 but on en0. Therefore I think is best to use grep to search for inet, and ignore the localhost 127.0.0.1. @jmaslek if you can confirm that this works in your machine, but I think it should


DidierRLopes:
Screenshot 2021-09-19 at 15 04 55

Now everything works :)

jmaslek (approved):

PR #754, opened 2021-09-20
Major brokers refactor into separate menus

jmaslek:
I have refactored the brokers menu into separate menus.

As is, I currently eliminated the `merge` command, but I have a TODO  in there to incorporate the coinbase menu.  

I would like to have a merged menu potentially, but not ready at this stage

Edit::: Builds were failing so I updated poetry packages and pinned the openpyxl

PR #756, opened 2021-09-23
Added sources argument to the news command

Flodur871:
I added an argument that allows filtering of news by source, as was mentioned in the ```#TODO```.

### DIFF
--- a/gamestonk_terminal/common/newsapi_view.py
+++ b/gamestonk_terminal/common/newsapi_view.py
@@ -11,6 +11,7 @@ def news(
     num: int,
     s_from: str,
     show_newest: bool,
+    sources: str,
 ):
     """Display news for a given title. [Source: NewsAPI]
 
@@ -24,13 +25,18 @@ def news(
         date to start searching articles from formatted YYYY-MM-DD
     show_newest: bool
         flag to show newest articles first
+    sources: str
+        sources to exclusively show news from
     """
-    # TODO: Add argument to specify news source being used
-
-    response = requests.get(
-        f"https://newsapi.org/v2/everything?q={term}&from={s_from}"
-        f"&sortBy=publishedAt&language=en&apiKey={cfg.API_NEWS_TOKEN}",
+    link = (
+        f"https://newsapi.org/v2/everything?q={term}&from={s_from}&sortBy=publishedAt&language=en"
+        f"&apiKey={cfg.API_NEWS_TOKEN}"
     )
+
+    if sources:
+        link += f"&domains={sources}"
+
+    response = requests.get(link)
 
     # Check that the API response was successful
     if response.status_code == 426:
--- a/gamestonk_terminal/stocks/stocks_controller.py
+++ b/gamestonk_terminal/stocks/stocks_controller.py
@@ -251,6 +251,14 @@ Market {('CLOSED', 'OPEN')[b_is_stock_market_open()]}
             default=True,
             help="Show oldest articles first",
         )
+        parser.add_argument(
+            "-s",
+            "--sources",
+            dest="n_sources",
+            default=[],
+            nargs="+",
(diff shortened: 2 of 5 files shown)


DidierRLopes:
I think this error will be fixed when #754 is merged

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -251,6 +251,14 @@ def call_news(self, other_args: List[str]):
             default=True,
             help="Show oldest articles first",
         )
+        parser.add_argument(
+            "-s",
+            "--sources",
+            dest="n_sources",
+            default=[],
```
jmaslek, line 258: So the "n_" in dest (if used) is used for number variables (int or floats).  So can we can either name this "s_sources", (s for string) or just "sources".

gamestonk_terminal/common/newsapi_view.py
```diff
@@ -24,14 +25,19 @@ def news(
...
-    response = requests.get(
-        f"https://newsapi.org/v2/everything?q={term}&from={s_from}"
-        f"&sortBy=publishedAt&language=en&apiKey={cfg.API_NEWS_TOKEN}",
+    link = (
+        f"https://newsapi.org/v2/everything?q={term}&from={s_from}&sortBy=publishedAt&language=en"
+        f"&apiKey={cfg.API_NEWS_TOKEN}"
     )
 
+    if sources:
+        link += f"&domains={sources}"
```
jmaslek, line 37: Could we check that each source ends in a .com, and if not add it, this way one could just do `news -s yahoo`

gamestonk_terminal/stocks/stocks_controller.py
Flodur871: I see

gamestonk_terminal/common/newsapi_view.py
Flodur871: Sure

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -254,23 +254,27 @@ def call_news(self, other_args: List[str]):
...
-            help="Show news only from the sources specified (e.g bbc.com yahoo.com)",
+            help="Show news only from the sources specified (e.g bbc yahoo.com)",
         )
 
         try:
             ns_parser = parse_known_args_and_warn(parser, other_args)
             if not ns_parser:
                 return
 
+            sources = ns_parser.sources
```
DidierRLopes, line 267: A more elegant solution would be:
```
sources = 
[src for src in ns_parser.sources if ".com" in src] + [f"{src}.com" for src in ns_parser.sources if ".com" not in src]
```

list comprehensions are quite powerful :)

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -254,23 +254,27 @@ def call_news(self, other_args: List[str]):
         parser.add_argument(
             "-s",
             "--sources",
-            dest="n_sources",
             default=[],
             nargs="+",
-            help="Show news only from the sources specified (e.g bbc.com yahoo.com)",
+            help="Show news only from the sources specified (e.g bbc yahoo.com)",
```
DidierRLopes, line 259: you're missing a comma here between bbc and yahoo

DidierRLopes (approved):
I commented 2 tiny things, but looks good to me :)

Great work!

gamestonk_terminal/stocks/stocks_controller.py
jmaslek: Doesn't "nargs=+" eliminate the need for commas?

DidierRLopes: You're right. I saw the `",".join()` and got confused, but we just process it like that after. So all good

PR #760, opened 2021-09-24
Wales alerts

JakubPluta:
Hello there!

Today very simple feature that make it possible to track crypto whales transactions who send huge amount of coins through blockchain. Using that you can sniff out if something bigger is going to happen on the market.

I added docs for hugo server, hope that I did it properly :)

Let me know if all good!


gamestonk_terminal/config_terminal.py
```diff
@@ -90,3 +90,7 @@
 API_COINBASE_KEY = os.getenv("GT_API_COINBASE_KEY") or "REPLACE_ME"
 API_COINBASE_SECRET = os.getenv("GT_API_COINBASE_SECRET") or "REPLACE_ME"
 API_COINBASE_PASS_PHRASE = os.getenv("GT_API_COINBASE_PASS_PHRASE") or "REPLACE_ME"
+
```
DidierRLopes, line 93: @jmaslek didn't you forget to add:

```
APCA_API_BASE_URL = os.getenv("GT_APCA_API_BASE_URL") or "REPLACE_ME"
APCA_API_KEY_ID = os.getenv("GT_APCA_API_KEY_ID") or "REPLACE_ME"
APCA_API_SECRET_KEY = os.getenv("GT_APCA_API_SECRET_KEY") or "REPLACE_ME"
```

in here?

gamestonk_terminal/cryptocurrency/onchain/whale_alert_model.py
```diff
@@ -0,0 +1,124 @@
...
+    limit: int
+        Limit of transactions. Max 100
+
+    Returns
+    -------
+    pd.DataFrame
+        Crypto wales transactions
+    """
+
+    min_value = 800000 if min_value > 800000 else min_value
```
DidierRLopes, line 66: A more elegant way to do this is using max, i.e.
```
min_value = max(min_value, 800000)
limit = max(limit, 100)
```

And if you want to limit a number between a and b, where a < b, you can do:
```
number =max(min(b, number), a)
```

DidierRLopes (approved):
I commented on a suggestion to have more pretty code, but apart from that all looks good. 

I also commented on something I believe James forgot, I can fix that one at a later stage

DidierRLopes:
Hugo server was done perfectly.

It was easy to add right?

gamestonk_terminal/config_terminal.py
jmaslek: So those actually just need to live in environment.  The api doesn't take in the password and stuff.

JakubPluta:
> Hugo server was done perfectly.
> 
> It was easy to add right?

Yes, indeed! It was matter of 2 minutes :) That was good step to use it for docs.

gamestonk_terminal/cryptocurrency/onchain/whale_alert_model.py
JakubPluta: Cool :) I knew this way for that, but just I had this kind of "bad habbit" of always doing things with `x if x else y` . But will try to have this elegant way in mind and implement it in the code!

Issue #767, opened 2021-09-28
[IMPROVE] Stock news 


MLNW:
**What's the feature that should be improved?**
The stock news option is not really functional.
The retrieved news is not relevant to the selected stock at all.

**Describe how you would like the feature improved**
The retrieved news should be more relevant to the selected stock.

**Additional information**
A simple example showcasing the issue. Note that none of the linked news articles have much to do with the selected stock ($COST - Costco in this case):
![image](https://user-images.githubusercontent.com/20930814/135156296-dd2ccfe8-a0d9-41c6-ad66-3e48aa779155.png)

DidierRLopes:
The reason for this was that we were looking for news that matched the ticker name, so this worked well for the examples I tested, e.g. AAPL, MSFT, ... but for tickers like COST it doesn't work well because it was searching for the word "cost". You can open those articles and you can confirm this.

I've changed the command to get the corporation name from the ticker, and then look for the words corresponding to the corporation name instead. Thus, instead of looking for MSFT, we'll look for Microsoft.

Here are the new results with COST:
<img width="947" alt="Screen Shot 2021-09-29 at 01 26 17" src="https://user-images.githubusercontent.com/25267873/135183559-d8d3b985-ff6c-4c80-837a-eb187df21eaf.png">

DidierRLopes:
Closing this since the improvement has already been merged :)

PR #769, opened 2021-09-29
Improve news to address #767

DidierRLopes:
Instead of searching for the term "COST", "AAPL", ... on the news, which may lead to incorrect articles popping. We get the name of the company from the ticker, and search for that on the news instead. This leads to a better search.

### DIFF
--- a/gamestonk_terminal/stocks/stocks_controller.py
+++ b/gamestonk_terminal/stocks/stocks_controller.py
@@ -3,6 +3,7 @@ import os
 from typing import List
 
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


jmaslek:
What about a flag for allowing to search just the ticker?

DidierRLopes:
> What about a flag for allowing to search just the ticker?

So, I thought about that. And did some tests, and it seems that the search was pretty horrible. And the corporation name was always more efficient.

I even tried giving the "$" sign, but the newsapi doesn't recognize it

jmaslek (approved):

PR #772, opened 2021-10-01
Refactor export functionality concept

DidierRLopes:
@jmaslek what do you think having it like this?

### DIFF
--- a/gamestonk_terminal/helper_funcs.py
+++ b/gamestonk_terminal/helper_funcs.py
@@ -1,7 +1,6 @@
 """Helper functions"""
 __docformat__ = "numpy"
 import argparse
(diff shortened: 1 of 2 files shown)


jmaslek (approved):
Clever. I like it.

DidierRLopes:
> Clever. I like it.

Which ones did you add it to, so I can fix at least those for now?

jmaslek:
Should just be whichever ba commands have export as a variable

DidierRLopes:
> Should just be whichever ba commands have export as a variable

Does it look good?

jmaslek:
Yes.  Think you got them.  I can add this to my current pr as well

PR #774, opened 2021-10-02
Discord Bot: Logging and error output capabilities

northern-64bit:
Adds logging for the server and displays errors with command corrections to the user. (Note: I don't know why git added 7 commits that already have been merged)


discordbot/economy/currencies.py
```diff
@@ -5,34 +5,61 @@
...
-            colour=cfg.COLOR,
-        )
-        embed.set_author(
-            name=cfg.AUTHOR_NAME,
-            icon_url=cfg.AUTHOR_ICON_URL,
-        )
+    try:
+        # Debug
+        if cfg.DEBUG:
+            print("-- STARTED COMMAND: !stocks.economy.currencies " + arg + " --")
```
DidierRLopes, line 11: Can I suggest using python3 print convention?

Like
```
print(f"-- STARTED COMMAND: !stocks.economy.currencies {arg} --")
```

discordbot/economy/currencies.py
```diff
@@ -5,34 +5,61 @@
...
+                description=help_txt,
+                colour=cfg.COLOR,
+            )
+            embed.set_author(
+                name=cfg.AUTHOR_NAME,
+                icon_url=cfg.AUTHOR_ICON_URL,
+            )
 
-        await ctx.send(embed=embed)
+            await ctx.send(embed=embed)
```
DidierRLopes, line 26: No need for this line here because you can use the one below if you remove 1 indentation

discordbot/economy/currencies.py
```diff
@@ -5,34 +5,61 @@
...
+                title="Economy: [WSJ] Currencies",
+                description=df_data_str,
+                colour=cfg.COLOR,
+            )
+            embed.set_author(
+                name=cfg.AUTHOR_NAME,
+                icon_url=cfg.AUTHOR_ICON_URL,
+            )
+
+            await ctx.send(embed=embed)
```
DidierRLopes, line 45: if you remove 1 indentation you can get this line to be used for both 2 things :)

discordbot/economy/currencies.py
```diff
@@ -5,34 +5,61 @@
...
             name=cfg.AUTHOR_NAME,
             icon_url=cfg.AUTHOR_ICON_URL,
         )
-
-        await ctx.send(embed=embed)
+        embed.set_description(
+            "Try updating the bot, make sure DEBUG is True in the config "
+            "and restart it.\nIf the error still occurs open a issue at: "
+            "https://github.com/GamestonkTerminal/GamestonkTerminal/issues"
+        )
```
DidierRLopes, line 58: If we want the user to see this we need a `await ctx.send(embed=embed)` I think

DidierRLopes: Since the error will be the same with the exception with what is caught by the exception, should we have this default embed in config in a function, something like:
```
def exception_embed(e: Exception):
        embed = discord.Embed(title="EXCEPTION", colour=cfg.COLOR)
        embed.set_author(
            name=cfg.AUTHOR_NAME,
            icon_url=cfg.AUTHOR_ICON_URL,
        )
        embed.set_description(e)
```

discordbot/economy/currencies.py
```diff
@@ -5,34 +5,61 @@
...
+    except Exception as e:
+        title = "INTERNAL ERROR"
+        embed = discord.Embed(title=title, colour=cfg.COLOR)
         embed.set_author(
             name=cfg.AUTHOR_NAME,
             icon_url=cfg.AUTHOR_ICON_URL,
         )
-
-        await ctx.send(embed=embed)
+        embed.set_description(
```
DidierRLopes, line 54: I think as a description we should pass the exception error `e`, what do you think? 

E.g. if the problem is on the server side, we want to tell the user that the problem is with the API and not with our terminal

discordbot/economy/currencies.py
DidierRLopes: Also, we can just have: 
```
print(f"!stocks.economy.currencies {arg}")
```

And then if there's a problem we know it will be associated with this

discordbot/economy/currencies.py
```diff
@@ -5,34 +5,61 @@
...
+            "Try updating the bot, make sure DEBUG is True in the config "
+            "and restart it.\nIf the error still occurs open a issue at: "
+            "https://github.com/GamestonkTerminal/GamestonkTerminal/issues"
+        )
+        if cfg.DEBUG:
+            print("-- ERROR at COMMAND: !stocks.economy.currencies " + arg + " --")
```
DidierRLopes, line 60: No need for all these prints, we can keep it more simples and just have a print(e).

The reason why it should be fine is that I foresee us having a debug thing like:
```
!economy
Timeout bla bla bla issue

!stocks.dps.sidtc AAAPL
AAAPL is not recognized...
```

DidierRLopes (commented):
Feel free to update just that command where I commented and we can keep discussing it until reaching a solution we are both happy. Let me know your thoughts

discordbot/economy/currencies.py
northern-64bit: Sounds good! I forgot the python 3 convention.

discordbot/economy/currencies.py
northern-64bit: I didn't think of this possibility.

discordbot/economy/currencies.py
northern-64bit: I tried to make it clear and easy to find the error, but you're right that this may just do it harder. I'll take a look at it.

discordbot/economy/currencies.py
northern-64bit: I forgot the ```await ctx.send(embed=embed)``` part.

discordbot/economy/currencies.py
northern-64bit: Fixed it!

DidierRLopes (commented):
It looks good to me, but can I suggest some things:
- When I tried to do !economy and then choosing option 1 it failed. This is because of 2 things, the commands require a ctx and an arg arguments. For the case we call the function from the menu, there is no argument provided, hence, when we declare the function we should have a default argument, e.g.
```
async def overview_command(ctx, arg=""):
```

- As a debug, we should have the name of the message that the user triggered on the discord. I.e. if the user does `!economy.overview` that's what should be printed. If the user does `!economy` and then selects `overview` we should print something smarter like:
```
!economy
Reaction selected: "1 - !economy.overview"
```
Does that make sense?

northern-64bit:
> 
> 
> It looks good to me, but can I suggest some things:
> 
>     * When I tried to do !economy and then choosing option 1 it failed. This is because of 2 things, the commands require a ctx and an arg arguments. For the case we call the function from the menu, there is no argument provided, hence, when we declare the function we should have a default argument, e.g.
> 
> 
> ```
> async def overview_command(ctx, arg=""):
> ```
> 
>     * As a debug, we should have the name of the message that the user triggered on the discord. I.e. if the user does `!economy.overview` that's what should be printed. If the user does `!economy` and then selects `overview` we should print something smarter like:
> 
> 
> ```
> !economy
> Reaction selected: "1 - !economy.overview"
> ```
> 
> Does that make sense?

Yes, it very much makes sense. I thought that the commands already had default argument, but I now I see the arg argument isn't passed to the function. The issues that you pointed out will be quickly fixed.

discordbot/economy/economy_menu.py
```diff
@@ -91,22 +94,40 @@ def check(reaction, user):
                 "reaction_add", timeout=cfg.MENU_TIMEOUT, check=check
             )
             if reaction.emoji == "0️⃣":
-                await overview_command(ctx)
+                if cfg.DEBUG:
+                    print("Reaction selected: 0")
+                await overview_command(ctx, "")
```
DidierRLopes, line 99: I think the argument should be handled from the other side.

I.e. we should have a default on overview_command, that is like this:
```
overview_command(ctx, arg="")
```

This is because imagine that we absolutely need one argument to have the functionality working. The controller shouldn't have to select this. It should be responsibility of the function implementation to have a saying over it.

DidierRLopes: Also, since we are doing the print in the controller, I think we should do the other commands prints in the controller as well. So inside each function, we basically just do a print if there's a try-catch error.

northern-64bit: Oh yeah you're right it makes the code way more robust. Will fix this

DidierRLopes (approved):
Perfect!

PR #778, opened 2021-10-03
Added autoinstallation

colin99d:
To expand our userbase I think it is important to make the startup process as easy as possible. I understand we have a docker image, but I am thinking about something much simpler than that for everyone who doesn't even know what Python or docker are and just want to do finance. This installation automatically created a virtual environment and installs all the dependencies. However, I cannot get it to open the same script inside a virtual environment once it has been clicked. @aia I was told you might be able to help me with this (see lines 394-397 of terminal.py).

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -5,21 +5,10 @@ __docformat__ = "numpy"
 import argparse
 import os
 import sys
-
-from prompt_toolkit.completion import NestedCompleter
-
-from gamestonk_terminal import config_terminal
-from gamestonk_terminal import feature_flags as gtff
-from gamestonk_terminal.helper_funcs import get_flair
-from gamestonk_terminal.menu import session
-from gamestonk_terminal.terminal_helper import (
-    about_us,
-    bootup,
-    check_api_keys,
-    print_goodbye,
-    reset,
-    update_terminal,
-)
+import platform
+import subprocess
+import json
+import pathlib
 
 # pylint: disable=too-many-public-methods,import-outside-toplevel
 
@@ -59,7 +48,9 @@ class TerminalController:
     CHOICES += CHOICES_SHORTHAND_MENUS
 
     def __init__(self):
-        """Constructor"""
+        from prompt_toolkit.completion import NestedCompleter
+
+        # Constructor
         self.update_succcess = False
         self.t_parser = argparse.ArgumentParser(add_help=False, prog="terminal")
         self.t_parser.add_argument(
@@ -138,6 +129,7 @@ What do you want to do?
         return True
 
     # COMMANDS
+
     def call_reset(self, _):
         """Process reset command"""
         return True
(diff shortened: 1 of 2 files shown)


terminal.py
```diff
@@ -5,21 +5,10 @@
...
-from gamestonk_terminal.helper_funcs import get_flair
-from gamestonk_terminal.menu import session
-from gamestonk_terminal.terminal_helper import (
-    about_us,
-    bootup,
-    check_api_keys,
-    print_goodbye,
-    reset,
-    update_terminal,
-)
```
aia, line 22: Moving imports from the top of the file to code internals decreases readability of the code.

terminal.py
```diff
@@ -239,6 +231,7 @@ def terminal():
     bootup()
     process_input = False
     t_controller = TerminalController()
+    from gamestonk_terminal import config_terminal
```
aia, line 234: The imports are now all over the code, making the code very confusing.

terminal.py
```diff
@@ -308,5 +305,93 @@ def terminal():
...
+        new = json_data["installed"]
+
+    else:
+        with open("data.txt", "w+") as f:
+            data = {"installed": False}
+            json.dump(data, f)
+        new = False
+
+    if not new:
+        if platform.system() == "Darwin":
```
aia, line 321: What about Windows and Linux systems?

terminal.py
```diff
@@ -308,5 +305,93 @@ def terminal():
...
+                            "-x86_64",
+                            "/usr/bin/python3",
+                            "-m",
+                            "pip",
+                            "install",
+                            "--user",
+                            "--upgrade",
+                            "virtualenv",
+                        ],
+                        check=True,
```
aia, line 345: Logic like that will be very difficult to implement, test and maintain.

Ignoring testing and maintenance for a moment, how are you handling failures coming from the subprocess.run?

aia:
Overall, here's what I see.
- The proposal is to add another method of terminal installation.
- The installation is implemented in raw python with the use of subprocess.
- The installation only covers a very basic case of OS X / native python 3 / virtualenv.

The problems are the following
- Adding another method of installation will increase user confusion. Our installation is already very confusing.
- Supporting another installation method will add significant overhead.
- The benefits of the new installation method are unclear.
- The new installation method does not implement Linux or Windows installation path.
- The new installation implementation will be challenging and voluminous if at all feasible.
- The testing methodology for the new method is unclear.

colin99d:
> Overall, here's what I see.
> 
> * The proposal is to add another method of terminal installation.
> * The installation is implemented in raw python with the use of subprocess.
> * The installation only covers a very basic case of OS X / native python 3 / virtualenv.
> 
> The problems are the following
> 
> * Adding another method of installation will increase user confusion. Our installation is already very confusing.
> * Supporting another installation method will add significant overhead.
> * The benefits of the new installation method are unclear.
> * The new installation method does not implement Linux or Windows installation path.
> * The new installation implementation will be challenging and voluminous if at all feasible.
> * The testing methodology for the new method is unclear.

I think I agree on all these points. One of those ideas that looked a lot better in my head than in the code. Should I delete the PR?

PR #781, opened 2021-10-04
Updated Portfolio Rf rate.

colin99d:
The portfolio maxsharpe command assumed a risk-free rate of 2%, but I updated it to be a risk-free rate determined by the scraping the current US T-Bill rate. 
Note: while doing this I found an existing error. Using any rate other than 0.02, whether through my new formula or specifying -r, throws an error:
 `/Users/colindelahunty/Desktop/GamestonkTerminal/env/lib/python3.8/site-packages/pypfopt/efficient_frontier/efficient_frontier.py:409: UserWarning: The risk_free_rate provided to portfolio_performance is different to the one used by max_sharpe. Using the previous value.` 
I could not find a solution to this error after 1.5 hours of searching. And left the print statements showing the values are the same so you guys could see what I was talking about. This issue also arrises in main, so it is not because of my changes.

### DIFF
--- a/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_helper.py
+++ b/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_helper.py
@@ -2,6 +2,7 @@
 __docformat__ = "numpy"
 
 import argparse
+import requests
 
 # These are all the possible yfinance properties
 l_valid_property_infos = [
@@ -105,3 +106,16 @@ def check_valid_property_type(aproperty: str) -> str:
         return aproperty
 
     raise argparse.ArgumentTypeError(f"{aproperty} is not a valid info")
+
+
+def get_rf():
+    """Uses the fiscaldata.gov API to get most recent T-Bill rate"""
+    try:
+        base = "https://api.fiscaldata.treasury.gov/services/api/fiscal_service"
+        end = "/v2/accounting/od/avg_interest_rates"
+        filters = "?filter=security_desc:eq:Treasury Bills&sort=-record_date"
+        response = requests.get(base + end + filters)
+        latest = response.json()["data"][0]
+        return round(float(latest["avg_interest_rate_amt"]) / 100, 8)
+    except Exception:
+        return 0.02
--- a/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_helper.py
+++ b/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_helper.py
@@ -108,7 +108,7 @@ def check_valid_property_type(aproperty: str) -> str:
     raise argparse.ArgumentTypeError(f"{aproperty} is not a valid info")
 
 
-def get_rf():
+def get_rf() -> float:
     """Uses the fiscaldata.gov API to get most recent T-Bill rate"""
     try:
         base = "https://api.fiscaldata.treasury.gov/services/api/fiscal_service"
--- a/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_model.py
+++ b/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_model.py
@@ -65,7 +65,7 @@ def get_property_weights(
 
 
 def get_maxsharpe_portfolio(
-    stocks: List[str], period: str = "3mo", rfrate: float = 0.02
+    stocks: List[str], period: str, rfrate: float
 ) -> Tuple[Dict, EfficientFrontier]:
     """Generate weights for max sharpe portfolio
 
@@ -87,6 +87,7 @@ def get_maxsharpe_portfolio(
     """
     stock_prices = yahoo_finance_model.process_stocks(stocks, period)
     ef = prepare_efficient_frontier(stock_prices)
+    print(rfrate)
     return dict(ef.max_sharpe(rfrate)), ef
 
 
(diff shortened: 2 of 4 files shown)


jmaslek:
Haven't reviewed yet, but dug into the warning and the issue is that the `rfrate` needs to be passed to `ef.portfolio_performance()` in optimizer_view. so print(ef.portfolio_performance(verbose=True, riskfreerate=rfrate)`

colin99d:
> Haven't reviewed yet, but dug into the warning and the issue is that the `rfrate` needs to be passed to `ef.portfolio_performance()` in optimizer_view. so `print(ef.portfolio_performance(verbose=True, riskfreerate=rfrate)`

I pushed the update with this included but the error is still there.

jmaslek:
> I pushed the update with this included but the error is still there.

Think I goofed up.  The kwarg is `risk_free_rate` for portfolio_performance

gamestonk_terminal/portfolio/portfolio_optimization/optimizer_helper.py
```diff
@@ -105,3 +106,16 @@ def check_valid_property_type(aproperty: str) -> str:
         return aproperty
 
     raise argparse.ArgumentTypeError(f"{aproperty} is not a valid info")
+
+
+def get_rf():
```
jmaslek, line 111: Type:
`def get_rf(): -> float`

gamestonk_terminal/portfolio/portfolio_optimization/optimizer_model.py
```diff
@@ -87,6 +87,7 @@ def get_maxsharpe_portfolio(
     """
     stock_prices = yahoo_finance_model.process_stocks(stocks, period)
     ef = prepare_efficient_frontier(stock_prices)
+    print(rfrate)
```
jmaslek, line 90: So we want to avoid printing in _model.  So if you want to print the rfrate, do it in view (which you actually have already)

colin99d: This is only to show that the rate matches so I can diagnose the error. Will be removed once the error is gone.

gamestonk_terminal/portfolio/portfolio_optimization/po_controller.py
```diff
@@ -385,7 +387,7 @@ def call_maxsharpe(self, other_args: List[str]):
             if len(self.tickers) < 2:
                 print("Please have at least 2 loaded tickers to calculate weights.\n")
                 return
-
+            print(ns_parser.risk_free_rate)
```
jmaslek, line 390: same about printing.  We can just print once in _view

jmaslek:
Small details about prints, but it works and is a great improvement using real data!

gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
```diff
@@ -160,7 +160,7 @@ def display_max_sharpe(
     else:
         print("\n", s_title)
         display_weights(weights)
-    ef.portfolio_performance(verbose=True)
+    ef.portfolio_performance(verbose=True, risk_free_rate=get_rf())
```
jmaslek, line 163: still want the user to be able to define a rfrate

colin99d: I was thinking about that myself. Where could we maintain the custom -r? I don't think it happened before my pr either.

jmaslek: I think the best solution would be similar to what you have implemented.  Have get_rf() be the default rfrate in the parser, so if it not supplied then we use the T Bill.  Then only use the rfrate in your functions as was, not redefining them in _view.

colin99d: I see!!! Thank you changes made.

jmaslek:
So now when I do this, I get:
```
(✨) (portfolio)>(po)> maxsharpe -r .05
('Please check your objectives/constraints or use a different solver.', 'Solver status: infeasible') 

(✨) (portfolio)>(po)> maxsharpe -r .06
('Please check your objectives/constraints or use a different solver.', 'Solver status: infeasible') 
```
Even though the rf_rate is .051 ish.  Usually I see the solver fail when rf is too high (since the convex problem doesn't have a solution).  Is there some multiplication going on somewhere?

colin99d:
> So now when I do this, I get:
> 
> ```
> (✨) (portfolio)>(po)> maxsharpe -r .05
> ('Please check your objectives/constraints or use a different solver.', 'Solver status: infeasible') 
> 
> (✨) (portfolio)>(po)> maxsharpe -r .06
> ('Please check your objectives/constraints or use a different solver.', 'Solver status: infeasible') 
> ```
> 
> Even though the rf_rate is .051 ish. Usually I see the solver fail when rf is too high (since the convex problem doesn't have a solution). Is there some multiplication going on somewhere?

.05 is 5%, the current risk free rate is .0005 or .05%.

jmaslek:
> .05 is 5%, the current risk free rate is .0005 or .05%.
Ew.  better off in a high yield savings at 1% lol

jmaslek (approved):

PR #782, opened 2021-10-04
Fix options scr bug

jmaslek:
There was a bug in options screener when a ticker was provided.

The issue was in how I tested in the ticker was okay.  When reading in key, value fro the config parser, the value was actually `'"AMC"'` not just `"AMC"`, so I passed eval(Ticker) into the finance to check its valid.


jmaslek:
Abusing admin power and merging small fix.

PR #785, opened 2021-10-04
Risk Free Line

colin99d:
Adds a line showing optimal portfolios combined with the risk-free asset.


aia:
Looks like you want to fix Black formatting.

gamestonk_terminal/portfolio/portfolio_optimization/po_controller.py
```diff
@@ -711,24 +711,33 @@ def call_ef(self, other_args):
...
+            dest="risk_free",
+            default=False,
+            help="Adds the optimal line with the risk-free asset",
+        )
 
-        try:
-            ns_parser = parse_known_args_and_warn(parser, other_args)
-            if not ns_parser:
-                return
+        # try:
```
jmaslek, line 723: Need to put back

gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
```diff
@@ -345,6 +352,17 @@ def display_ef(stocks: List[str], period: str = "3mo", n_portfolios: int = 300):
...
+    # Add risk free line
+    if risk_free:
+        y = max(rets)
+        b = get_rf()
+        m = (ret_sharpe - b) / std_sharpe
+        x = (y - b) / m
+        ax.set_xlim(xmin=0)
+        x = [0, x]
+        y = [b, y]
+        line = Line2D(x, y, color="#FF0000")
```
jmaslek, line 364: could condense to just line = Line2D([0,x],[b,y], color)

gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
```diff
@@ -345,6 +352,17 @@ def display_ef(stocks: List[str], period: str = "3mo", n_portfolios: int = 300):
         verbose=True, risk_free_rate=rfrate
     )
     ax.scatter(std_sharpe, ret_sharpe, marker="*", s=100, c="r", label="Max Sharpe")
+    # Add risk free line
+    if risk_free:
+        y = max(rets)
+        b = get_rf()
+        m = (ret_sharpe - b) / std_sharpe
+        x = (y - b) / m
+        ax.set_xlim(xmin=0)
```
jmaslek, line 361: Personal opinion, but I don't think it looks as good with the xlim going to 0.  I would keep it to either like .95 of min(max_std) or just subtract a few percent from the min.

colin99d: The rf line starts at x=0. Shouldn't we show the entire line?

jmaslek: I guess.  I just did an example and the volatility axis I saw was from .2 to .3, so extending to 0 looked weird

colin99d: It should always be at 0 since the risk-free asset has no volatility. Was this including -r? And which ticker?

jmaslek: yeah.  My comment was that the ef (with no -r) went from like .2 to .28 and then the plot looked "squished" for the ef (fig here).  This was aapl,sbux , `ef -p 1y -r`.  Also you should add a label for the legend
Figure_1
.

colin99d: Ooh yeah thats not good. Should I just remove the xmin altogether?

jmaslek: Well without it, it will still go all the way down to zero, so as I mentioned earlier, just set it to be a little less than the min volatility value

colin99d: .95 looked weird when there was a deep negative stock like GME. So I switched it to .8 which was a lot better. Graph looks a lot better overall thanks for the input!

jmaslek (approved):

aia (approved):

PR #792, opened 2021-10-06
adds a GitHub action to sync basic repo info (using `askgit`) to Postgres

patrickdevivo:
will require the `POSTGRES_CONNECTION` to be set in the repo secrets

### DIFF
--- a/.github/workflows/askgit-data-sync.yml
+++ b/.github/workflows/askgit-data-sync.yml
@@ -1,0 +1,34 @@
+on:
+  schedule:
+    - cron:  '0 0 * * *' # at midnight (utc) every day
+jobs:
+  build:
+    name: Run AskGit
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


DidierRLopes:
> will require the `POSTGRES_CONNECTION` to be set in the repo secrets

And GITHUB_TOKEN as well. I thought we had it, but I don't think we do.

Thanks for this!!

DidierRLopes:
@patrickdevivo when creating the GITHUB_TOKEN secret, I get a 
```
Failed to add secret. Secret names must not start with GITHUB_.
```

Am I doing something wrong?

patrickdevivo:
@DidierRLopes so you should actually be okay - https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret the `GITHUB_TOKEN` secret is already set by default in the action I believe

PR #800, opened 2021-10-08
Option Payoff Diagrams

colin99d:
Creates a payoff diagram for an underlying asset (stock) and an unlimited number of derivatives added to it. Can choose any strike price available, choose between a put and call and between buying or selling the derivative.

**Note:** I have NOT added this to the website yet because I am assuming it will get torn up first (:

### DIFF
--- a/gamestonk_terminal/options/payoff_controller.py
+++ b/gamestonk_terminal/options/payoff_controller.py
@@ -70,13 +70,13 @@ class Payoff:
 
 Underlying Asset: {text}
 
-    pick          long, short, or choose not to hold the underlying asset
+    list          list available strike prices for calls and puts
 
+    pick          long, short, or none (default) underlying asset
     add           add option to the list of the options to be plotted
     rmv           remove option from the list of the options to be plotted
 
     plot          show the option payoff diagram
-    list          list available strike prices for calls and puts
         """
         print(help_text)
 
@@ -217,7 +217,7 @@ Underlying Asset: {text}
         """Add an option to the diagram"""
         parser = argparse.ArgumentParser(
             add_help=False,
-            prog="add/select",
+            prog="add",
             description="""Add options to the diagram.""",
         )
         parser.add_argument(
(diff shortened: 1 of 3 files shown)


jmaslek:
If you enter `payoff` without having ticker and exp, the terminal crashes, so you should add a check first.

jmaslek:
So if I do 

```(✨) (options)>(payoff)> add -k 150
list index out of range 
```
(AAPL - exp 2, 150 is in list)

colin99d:
> So if I do 
> 
> 
> 
> ```
> 
> (✨) (options)>(payoff)> add -k 150
> 
> list index out of range 
> 
> ```
> 
> (AAPL - exp 2, 150 is in list)

So when you type list you are supposed to use the numbers on the left after -k, not the actual strike price.

jmaslek:
I see.  I was able to do `add -k 10` so there should be some distinguishing between selecting a strike price and a strike index

colin99d:
> So @DidierRLopes can chime in here.  The way this currently is really goes away from the MVC we have been aiming for.  We should have specific files for payoff_view/payoff_model that do the usual stuff, that way we can import the functions into a Jupyter notebook for example.  
> 
> 
> 
> 

Should I move the data getting functions to a model and the matplotlib stuff to a view?

jmaslek:
Yeah. That was we have consistency across repo and it helps with the "importability"

DidierRLopes:
> > So @DidierRLopes can chime in here.  The way this currently is really goes away from the MVC we have been aiming for.  We should have specific files for payoff_view/payoff_model that do the usual stuff, that way we can import the functions into a Jupyter notebook for example.
> 
> Should I move the data getting functions to a model and the matplotlib stuff to a view?

Exactly. You can read more about the code structure here: https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/CONTRIBUTING.md.

colin99d:
> So @DidierRLopes can chime in here. The way this currently is really goes away from the MVC we have been aiming for. We should have specific files for payoff_view/payoff_model that do the usual stuff, that way we can import the functions into a Jupyter notebook for example.

Awesome! Got this updated.

colin99d:
> Some comments. Overall looks great!
> 
> I think you forgot to add the new line after each command as well.
> 
> And I couldn't figure out the add/rmv commands, but I'm smooth brain ahah

So when you type list it gives the index on the left and then the corresponding call/put on the right. You use the index on the left to add/rmv

colin99d:
> I think you still forgot the "." in ">. payoff"
Screen Shot 2021-10-10 at 4 51 23 PM
I am seeing it.

colin99d:
> > > I think you still forgot the "." in ">. payoff"
> > 
> > 
> > Screen Shot 2021-10-10 at 4 51 23 PM
> > I am seeing it.
> 
> Yes, I meant the opposite :p We usually don't have the "."

Ohhhhh. Lol

Issue #757, opened 2021-09-23
[IMPROVE] Get similar companies using Polygon api also returns companies listed on Canadian stock exchange


sug4ndh:
**What's the feature that should be improved?**
The command, 'getpoly', which gives similar companies using Polygon api also returns companies listed on Canadian stock exchange when users might be interested only in companies listed either on nyse, nasdaq or nyse arca.

**Describe how you would like the feature improved**
The response object from polygon api for a ticker contains an "exchangeSymbol" key as well as an "exchange" key. Their values could be used to filer out contents if the user is also given the option to enter the market. 

**Possibly describe the ideal way to improve this**
The `get_similar_companies(ticker: str) -> Tuple[List[str], str]` function can take another argument, market, maybe?

**Additional information**
For RWLK, I got this result in the terminal:

DidierRLopes:
We are using this API command here: https://polygon.io/docs/get_v1_meta_symbols__stocksTicker__company_anchor. 

I'm not sure we want to go down the road of modifying polygon API here. @jmaslek, what you think?

jmaslek:
> I'm not sure we want to go down the road of modifying polygon API here. @jmaslek, what you think?

So we could do something that loops through all the similar tickers and checks what their exchange is......but polygon is limited to 5 api calls/minute for free, so you would only be able to do this for 4 tickers before we get limited.

Could do the same with something like yahoo.

Would be open to a PR if this is something you wanna do OP

sug4ndh:
@DidierRLopes, I was thinking what @jmaslek pointed out but I did not consider the 5 calls/minute limitation. Maybe, once we get similar tickers, we can check them using marketwatch by validating if the
sent request object is the same as the prepared object? For e.g. 
Prepared objects would be -> "https://www.marketwatch.com/investing/stock/$ticker"
We send requests using `requests.get(prepared objects)`, if the ticker is in the us stock exchange then the sent object would be the same as the prepared object, if it is not then the stock exchange would be auto appended to the sent object.
So, for RWLK
```
>>> p = "https://www.marketwatch.com/investing/stock/rwlk"
>>> r = requests.get(p)
>>> if p == r.request.url:
...     print("us stock")
... else:
...     print("not us stock")

```
the above would output "us stock" but for symbol BABY it would be "not us stock" since the sent url would be not the same as prepared object and would instead become
 ```
>>> p="https://www.marketwatch.com/investing/stock/BABY"
[...]

PR #801, opened 2021-10-08
Fix getpoly to return only US stocks

sug4ndh:
* Fix addresses #757
* Getpoly will only return stocks listed on US stock exchanges.
* Index/mutual funds will also not be returned by getpoly.


jmaslek:
So I think this should be a flag in the getpoly, so adding a --us_only or something flag to the parser so it will return us only if specified.  Then you would just have in your addition the screening under `if us_only:` or something to that effect.

DidierRLopes:
@sug4ndh can you have a look at this?

PR #802, opened 2021-10-09
Refactoring pred

jmaslek:
@DidierRLopes can you provide feedback on what your thoughts are on the `ets` code

### DIFF
--- a/gamestonk_terminal/common/prediction_techniques/pred_helper.py
+++ b/gamestonk_terminal/common/prediction_techniques/pred_helper.py
@@ -481,7 +481,7 @@ def plot_data_predictions(
         plt.plot(
             forecast_data.index,
             forecast_data.values,
-            "-ok",
+            "-og",
             ms=3,
             label="Forecast",
         )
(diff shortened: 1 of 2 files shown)


jmaslek:
> What do you mean by ets code? The refactoring? Looks good to me, and I really like the target variable option!! Feel free to clean this up and merge these 2 to start with, we can tackle each model individually in order to avoid big PRs.

Yeah.  Mostly curious if you were good with how I broke up the functions into model/view.

Only issue is that I changed the helper functions to be usable by this "target" option so the nn option will fail for right now :P

DidierRLopes:
> > What do you mean by ets code? The refactoring? Looks good to me, and I really like the target variable option!! Feel free to clean this up and merge these 2 to start with, we can tackle each model individually in order to avoid big PRs.
> 
> Yeah. Mostly curious if you were good with how I broke up the functions into model/view.
> 
> Only issue is that I changed the helper functions to be usable by this "target" option so the nn option will fail for right now :P

Yup, it looks good to me!

That's unfortunate! Oh well, at least we'll have 100% of refactoring done which is a great step already!

DidierRLopes:
Can we have prediction line in green? It's kind of hidden in black

PR #811, opened 2021-10-15
Made smile generic

colin99d:
I changed the option to show volatility smiles with a function that allows users to choose their own X and Y value. The user can also select from presets that are preprogrammed.


DidierRLopes:
Also needs conflicts resolved

PR #812, opened 2021-10-16
Parity Dividends

colin99d:
Forgot to include present value of dividends in the previous PR. This assumes dividend stays the same as average of previous 4 dividends paid quarterly.


DidierRLopes:
Looks good to me! Just needs the conflicts fixed, this is because you branched off of your branch locally and not the main one with the parity merged :)

Issue #816, opened 2021-10-17
[Bug] Stocks/Gov - could not convert string to float


deeleeramone:
Stocks/Gov Top Sells returning:
could not convert string to float: '>5000000
![image](https://user-images.githubusercontent.com/85772166/137613699-6170adee-d847-47b7-9a87-10f69a23ad34.png)

PR #819, opened 2021-10-17
Added handling for >$5,000,000 - solves #816

colin99d:
Fixes bug related to issue 816.


DidierRLopes:
Btw, you should use # when mentioning an issue, because git automatically connects these. And if you add it on the title, it's even better because when the PR is merged and closed, the issue gets closed automatically.

Issue #817, opened 2021-10-17
Comprehensive historical daily volume and price statistics [FR]


deeleeramone:
Volume is probably the single most important metric along with daily price open/close/high/low. There currently is no method for retrieving the historical daily data.

jmaslek:
So is there anything specific you would like to see?

If you go to `qa` and do `pick Volume` you can do some basic quantitative analysis on volume.

deeleeramone:
Raw data. Not a chart not quantitative analysis. Basic raw data. Displayed tables with daily historical values for Volume and open/close/high/low/change since previous day price data. Something like this:

https://chartexchange.com/symbol/nasdaq-hut/historical/ <https://chartexchange.com/symbol/nasdaq-hut/historical/>

There should be a way to filter by date range and sort volume and change in price columns ascending or descending.

deeleeramone:
The same can be applied to individual options like this:

https://chartexchange.com/symbol/opra-hut20211015c2.5/historical/ <https://chartexchange.com/symbol/opra-hut20211015c2.5/historical/>

colin99d:
I can work on this tonight

colin99d:
That feature is not currently available.

deeleeramone:
This is why I am requesting it.

DidierRLopes:
![Screenshot 2022-01-20 at 13 46 40](https://user-images.githubusercontent.com/25267873/150350649-fb168847-5491-4018-863b-5c995946bb5c.png)

I believe this addresses this issue! Will close this!

PR #820, opened 2021-10-17
Added raw stock data #817

colin99d:
Adds the ability to view, sort, and customize raw stock data. The data can also be exported.


jmaslek:
Nice.  I think it may be a simpler implementation if you just add this to the main stocks menu (where candle is) and you can just use the self.stock data.

jmaslek:
> Was about to say the same. I don't think TA is the right menu to contain this. I would just have it in candle directly and allow a --raw command. Think is the right thing to do. At least that's what we've been doing for all other commands when we want raw data from a plot we see or so

I think having a separate command would make sense.  That way you can access it without having to have a plot come up.

DidierRLopes:
> > Was about to say the same. I don't think TA is the right menu to contain this. I would just have it in candle directly and allow a --raw command. Think is the right thing to do. At least that's what we've been doing for all other commands when we want raw data from a plot we see or so
> 
> I think having a separate command would make sense. That way you can access it without having to have a plot come up.

We don't need to have an AND condition, I was suggesting a OR. I.e. if the user adds `--raw` there's no plot popping. Which is exactly the same as adding a new command and removes complexity

jmaslek:
> We don't need to have an AND condition, I was suggesting a OR. I.e. if the user adds `--raw` there's no plot popping. Which is exactly the same as adding a new command and removes complexity

I think the separate command makes more sense, since if you want to do raw and sort by a column, that is two flags instead of 1 not to mention on the plot command.  So could get confusing.

DidierRLopes:
> > We don't need to have an AND condition, I was suggesting a OR. I.e. if the user adds `--raw` there's no plot popping. Which is exactly the same as adding a new command and removes complexity
> 
> I think the separate command makes more sense, since if you want to do raw and sort by a column, that is two flags instead of 1 not to mention on the plot command. So could get confusing.

We already do that in other commands. I just don't like adding a new command for the same data. I think doing 
```
candle --raw -s o
```
where o is for open.

is almost the same as
```
raw -s o
```

This is assuming that this feature is used by let's say 1% of users. If we start having a lot usage for it, I think it makes sense being a command. And I think that command should live on quantitative analysis menu. What you think?

DidierRLopes:
> Agree. Could also add a raw flag to candle in addition

True. I'm happy with that :) 

So the OP will get this feature twice 😄

colin99d:
> So I personally see more value as it's own command (as per the issue being addressed)
> 
> Just right under candle in the menu and you can do as you say (I'd probably go with hist instead of raw)
> 
> "hist c,v -n 20 --export"
> 
> Any input from the issue creator (I can't tag them)

hist is already in use

PR #826, opened 2021-10-18
Cleanup options controller

jmaslek:
There were some misspellings and some incorrectly parsed variables as well as logic causing parser to fail on certain commands.


DidierRLopes:
This automatic "fix #" closed the PR 😅

PR #878, opened 2021-10-30
Initial implementation of the Markdown Linter

aia:


.github/workflows/test.yml
```diff
@@ -84,6 +84,16 @@ jobs:
...
+  markdown-link-check:
+    name: Markdown linting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        name: Check out the code
+      - name: Lint Code Base
+        uses: docker://avtodev/markdown-lint:v1
+        with:
+          args: "./gamestonk_terminal/options/README.md"
```
DidierRLopes, line 96: Do you want to check options README?

This file should be changed drastically to only have a high-level detail of what Options contains. Similar to this: https://github.com/GamestonkTerminal/GamestonkTerminal/tree/main/gamestonk_terminal/stocks

aia: In a separate PR, yes.

DidierRLopes: Why on the workflow do we not use as args "." the entire codebase?

aia: > Why on the workflow do we not use as args "." the entire codebase?

We have too many linting errors in our existing markdowns. I want to add and fix them one at a time.

DidierRLopes: Sounds good. Probably we can wait for the markdown refactoring that will happen soon for this.

DidierRLopes (approved):

.github/workflows/test.yml
aia: I keep looking at all the red that I have in VScode every time I open a markdown from the repo and I decided that this is the time to get it fixed :-)

PR #882, opened 2021-10-30
Improve terminal speed for stocks context

DidierRLopes:
Similarly to what was done before, this does the import of a controller module when the user selects such menu. This takes some load from the stocks context menu.

### DIFF
--- a/gamestonk_terminal/stocks/stocks_controller.py
+++ b/gamestonk_terminal/stocks/stocks_controller.py
@@ -17,21 +17,8 @@ from gamestonk_terminal.helper_funcs import (
     parse_known_args_and_warn,
 )
 from gamestonk_terminal.menu import session
-from gamestonk_terminal.options import options_controller
-from gamestonk_terminal.stocks.backtesting import bt_controller
-from gamestonk_terminal.stocks.behavioural_analysis import ba_controller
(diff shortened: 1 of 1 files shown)


gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -41,7 +28,7 @@
 )
 from gamestonk_terminal.common.quantitative_analysis import qa_view
 
-# pylint: disable=R1710
+# pylint: disable=R1710,import-outside-toplevel
```
jmaslek, line 31: Should we add the import outside toplevel and too many lines into the pylintrc to ignore for all?

jmaslek (approved):

aia (approved):

gamestonk_terminal/stocks/stocks_controller.py
DidierRLopes: too many lines I would add to pylintrc, the import outside I wouldn't because we only want this on the controllers and not on the rest.

One disadvantage of this import outside as well is that when something is wrong inside a menu, the terminal will only crash when we go into that same menu. 

We can probably add tests for this to try to go into each of the menus individually

PR #886, opened 2021-10-31
Added Report Text

colin99d:
Financial reports are useful because the provide a wealth of data, with meaningful explanations of what the data means. This PR adds those explanations, and also includes refactoring to improved future expansion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ exports/*
 .coverage
 htmlcov
 gamestonk_terminal/reports/stored/*
+gamestonk_terminal/portfolio/portfolios/*
 
 # Jupyter Lab Extension build files
 jupyterlab/*/*.bundle.*
--- a/gamestonk_terminal/portfolio/portfolio_model.py
+++ b/gamestonk_terminal/portfolio/portfolio_model.py
@@ -301,6 +301,9 @@ def get_return(df: pd.DataFrame, df_m: pd.DataFrame, n: int) -> pd.DataFrame:
     )
     comb["return"] = comb["return"].fillna(1)
     comb["return"] = comb["return"].cumprod() - 1
+    comb[("Market", "Return")] = (
+        comb[("Market", "Close")] - comb[("Market", "Close")][0]
+    ) / comb[("Market", "Close")][0]
     return comb
 
 
@@ -351,8 +354,43 @@ def get_rolling_beta(
     return final
 
 
-def get_main_text() -> str:
-    text = """
-    Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning.
+def get_main_text(df: pd.DataFrame) -> str:
+    """Get main performance summary from a dataframe with returns
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Stock holdings and returns with market returns
+
+    Returns
+    ----------
+    text : str
+        The main summary of performance
     """
-    return text
+    v_com = (
+        "greater" if np.var(df["return"]) > np.var(df[("Market", "Return")]) else "less"
+    )
+    d_debt = np.where(df[("Cash", "Cash")] > 0, 0, 1)
+    bcash = 0 if df[("Cash", "Cash")][0] > 0 else abs(df[("Cash", "Cash")][0])
+    ecash = 0 if df[("Cash", "Cash")][-1] > 0 else abs(df[("Cash", "Cash")][-1])
+    bdte = bcash / (df["holdings"][0] - bcash)
+    edte = ecash / (df["holdings"][-1] - ecash)
+    if sum(d_debt) > 0:
+        t_debt = (
+            f"Beginning debt to equity was {bdte:.2%} and ending debt to equity was"
+            f" {edte:.2%}. Debt adds risk to a portfolio by amplifying the gains and losses when"
+            " equities change in value."
+        )
+    else:
+        t_debt = "Debt was not used this year. This reduces this risk of the portfolio."
+    t = (
+        f"Your portfolio's performance for the period was {df['return'][-1]:.2%}. This was"
+        f" {'greater' if df['return'][-1] > df[('Market', 'Return')][-1] else 'less'} than"
+        f" the market return of {df[('Market', 'Return')][-1]:.2%}. The variance for the"
+        f" portfolio is {np.var(df['return']):.2%}, while the variance for the market was"
+        f" {np.var(df[('Market', 'Return')]):.2%}. Since our portfolio's variance was"
+        f" {v_com} than the market's variance, our returns should be {v_com} than those"
+        f" of the market. {t_debt} The following report details various analytics from the"
+        f" portfolio. Read below to see the moving beta for a stock."
+    )
+    return t
--- a/gamestonk_terminal/portfolio/portfolio_model.py
+++ b/gamestonk_terminal/portfolio/portfolio_model.py
@@ -349,3 +349,10 @@ def get_rolling_beta(
     final["total"] = final.sum(axis=1)
     final = final[final.index >= datetime.now() - timedelta(days=n + 1)]
     return final
+
+
+def get_main_text() -> str:
+    text = """
+    Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning. Hello this is some sample text that you can read to learn more about learning.
+    """
+    return text
--- a/gamestonk_terminal/portfolio/portfolio_model.py
+++ b/gamestonk_terminal/portfolio/portfolio_model.py
@@ -9,7 +9,11 @@ from typing import List
 import numpy as np
 import pandas as pd
 
-from gamestonk_terminal.portfolio import portfolio_view, yfinance_model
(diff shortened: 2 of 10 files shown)


gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -333,19 +342,95 @@ def get_rolling_beta(
...
+
+    Returns
+    ----------
+    t : str
+        The beta history for a ticker
+    """
+    betas = df[list(filter(lambda score: "beta" in score, list(df.columns)))]
+    high = betas.idxmax(axis=1)
+    low = betas.idxmin(axis=1)
+    return (
```
DidierRLopes, line 421: to be more explicit could we declares these variables above and then create the beta_text string and only then return it.

This would allow code to be more readable

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -119,7 +111,10 @@ def plot_overall_return(
     fig.set_facecolor("white")
     ax.set_title(
         "%s - %s"
-        % (df.index[:1][0].strftime("%Y/%m/%d"), df.index[-1:][0].strftime("%Y/%m/%d")),
+        % (
+            comb.index[:1][0].strftime("%Y/%m/%d"),
```
DidierRLopes, line 115: The date format should be %Y-%m-%d to match the rest of the codebase

.gitignore
```diff
@@ -14,6 +14,7 @@ exports/*
 .coverage
 htmlcov
 gamestonk_terminal/reports/stored/*
+gamestonk_terminal/portfolio/portfolios/*
```
DidierRLopes, line 17: Screenshot 2021-10-31 at 09 42 31

Can we have at least 1 example here that users can use to understand how to build their portfolios?

jmaslek:
small note:  if cash line is `Deposit`, we get:
```

(✨) (portfolio)> ar
Cash type must be deposit or withdrawal 
```
So you should just look at lower cases for that

jmaslek:
I think it would be nicer to see tables showing some of the key statistics.  Can have some for returns/variance etc.

It would also be pretty easy to include some metrics, including the Sharpe ratio at least.

gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -333,19 +342,97 @@ def get_rolling_beta(
...
+        t_debt = "Debt was not used this year. This reduces this risk of the portfolio."
+    string = (
+        f"Your portfolio's performance for the period was {df['return'][-1]:.2%}. This was"
+        f" {'greater' if df['return'][-1] > df[('Market', 'Return')][-1] else 'less'} than"
+        f" the market return of {df[('Market', 'Return')][-1]:.2%}. The variance for the"
+        f" portfolio is {np.var(df['return']):.2%}, while the variance for the market was"
+        f" {np.var(df[('Market', 'Return')]):.2%}. Since our portfolio's variance was"
+        f" {v_com} than the market's variance, our returns should be {v_com} than those"
+        f" of the market. {t_debt} The following report details various analytics from the"
+        f" portfolio. Read below to see the moving beta for a stock."
```
jmaslek, line 401: can you combine this into a single string using` f""" blah blah """`

gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -333,19 +342,97 @@ def get_rolling_beta(
...
+        f" at {df['total'][0]:.2f}. This went"
+        f" {'up' if df['total'][-1] > df['total'][0] else 'down'} to"
+        f" {portfolio_helper.beta_word(df['total'][-1])} at {df['total'][-1]:.2f} by the end"
+        f" of the period. The ending beta was pulled {'up' if df['total'][-1] > 1 else 'down'} by"
+        f" {portfolio_helper.clean_name(high[-1] if df['total'][-1] > 1 else low[-1])}, which had"
+        f" an ending beta of {df[high[-1]][-1] if df['total'][-1] > 1 else df[low[-1]][-1]:.2f}."
+        f" To {'lower' if df['total'][-1] > 1 else 'raise'} the beta of the portfolio and"
+        f" {'decrease portfolio risk' if df['total'][-1] > 1 else 'increase potential return'}"
+        f" invest more heavily in"
+        f" {portfolio_helper.clean_name(low[-1] if df['total'][-1] > 1 else high[-1])} which had"
```
jmaslek, line 435: Im not a fan of this.  You are basically just breaking down a single factor (also average beta is not 1 -- that just means your movements are exactly correlated to the market.  If you really wanted, you could include the individual stock betas on the graph or just some kind of table.  

This also kind of reads as investment advice.

gamestonk_terminal/portfolio/portfolio_model.py
colin99d: When I tried combining it into a single string like how we do for the terminal print outs it messes up the pdf's spacing. How do I get around this?

gamestonk_terminal/portfolio/portfolio_model.py
colin99d: I do believe the average of all market beta is 1: https://en.wikipedia.org/wiki/Beta_(finance). Also my goal is to expand this and break down lots of different factors in detail. My hope is that this is a report that new investors can use to learn more about finance. 
Also, I do see your point about investment advice. I think to avoid this we tell them the facts and don't tell them what to do. Good catch!

colin99d:
> I think it would be nicer to see tables showing some of the key statistics. Can have some for returns/variance etc.
> 
> It would also be pretty easy to include some metrics, including the Sharpe ratio at least.

I was going to start adding more analytics in the next PR. Would you rather see them in this one?

gamestonk_terminal/portfolio/portfolio_model.py
jmaslek: > I do believe the average of all market beta is 1: https://en.wikipedia.org/wiki/Beta_(finance). Also my goal is to expand this and break down lots of different factors in detail. My hope is that this is a report that new investors can use to learn more about finance.

Value weighted beta (by definition) is 1.  Thats not the same as average beta.  If average beta was 1, everyone would beat the market.

jmaslek: Maybe we can let the more in depth report be added with a flag.

gamestonk_terminal/portfolio/portfolio_model.py
jmaslek: Not sure.  Haven't played with the pdf library.  Not necessary if it doesn't format well.

jmaslek:
> I was going to start adding more analytics in the next PR. Would you rather see them in this one?

Can wait

gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -421,7 +421,7 @@ def get_beta_text(df: pd.DataFrame) -> str:
     low = betas.idxmin(axis=1)
     string = (
         "Beta is how strongly a portfolio's movements correlate with the market's movements."
-        " A stock with a high beta is considered to be riskier, with the average being one."
+        " A stock with a high beta is considered to be riskier. The market has a beta of one."
```
jmaslek, line 424: just get rid of "The market has a beta of one"

The statement above (Beta is how strongly correlated .... ) implies that correlation of 1 is the market

jmaslek:
Another thing is you should add some kind of checking regarding cash.  I input quantity=2000, at price =1 (makes more sense to me, no?). and this is my rmr graph:
Screen Shot 2021-11-01 at 10 25 16 AM

colin99d:
> I don't think this is a useful plot to have on a report. You should state statistics of the portfolio that is loaded. This shows you the current optimization options based off the last n periods. Maybe further down the line, but at this stage I want to see more of the performance
> 
> Screen Shot 2021-11-01 at 10 28 15 AM

Thanks for reminding me! I was going to add how we did on this. As a point on the graph. More in-depth reports will be added in a future PR.

colin99d:
> Another thing is you should add some kind of checking regarding cash. I input quantity=2000, at price =1 (makes more sense to me, no?). and this is my rmr graph (which just consists of apple and Microsoft): Screen Shot 2021-11-01 at 10 25 16 AM

You are right. I multiplied by quantity so that people can do it however they want and it will give the same result. Let me know if you have further issues.

jmaslek:
Any thoughts on how to make this not be 0 for the first few months?

Screen Shot 2021-11-01 at 10 07 52 PM

Other comments:

"Since our portfolio's variance was greater than the market's variance, our returns should be greater than those of the market"

Remember the other half of the efficient frontier.  Greater variance does not guarentee greater returns.

"Debt was not used this year"
Probably better to say that you did not exceed a leverage of 1 or something to that effect.

jmaslek:
Bug:

```
(✨) (portfolio)> add cash -p 1000
CASH successfully added

(✨) (portfolio)> rmr
Brokers require cash, input cash deposits 
```

doing `add cash` should probably default to a deposit, not buy.

gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -330,22 +341,91 @@ def get_rolling_beta(
...
-        )
-        df_var = pd.merge(df_var, df1, how="left", left_index=True, right_index=True)
-        df_var[f"beta_{col}"] = df_var[f"cov_{col}"] / df_var["var"]
-    final = pd.merge(res, df_var, how="left", left_index=True, right_index=True)
-    final = final.fillna(method="ffill")
-    final = final.drop(columns=["var"] + [f"cov_{x}" for x in uniques])
+        exog = sm.add_constant(comb["Close"])
+        rols = RollingOLS(comb[col], exog, window=252)
+        rres = rols.fit()
+        res[f"beta_{col}"] = rres.params["Close"]
```
jmaslek, line 351: Just curious the reason for doing this instead of what you had? The formula for a regression coefficient is the same cov(x,y)/var(x) you calculated before.

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -72,18 +78,14 @@ def show_df(df: pd.DataFrame, show: bool) -> None:
...
 
     Parameters
     ----------
-    df : pd.DataFrame
-        The dataframe to be analyzed
-    df_m : pd.DataFrame
-        The dataframe for historical market performance
-    n : int
-        The number of days to include in chart
+    comb : pd.DataFrame
```
jmaslek, line 87: Just a small comment that doesn't need to be addressed now:  comb is a very non-descriptive variable name.  If it is representing returns, I would expect the df to be named something more related to that.

gamestonk_terminal/portfolio/portfolio_model.py
colin99d: Yup its the exact same equation! Just thought that the package was a little more official and probably has built in error handling that I don't want to worry about doing myself. Also allows for more analysis in the future.

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -198,38 +189,107 @@ def plot_rolling_beta(
...
+    return ImageReader(imgdata)
+
+
+class Report:
+    def __init__(self, df: pd.DataFrame, hist: pd.DataFrame, m_tick: str, n: int):
+        self.df = df
+        self.hist = hist
+        self.m_tick = m_tick
+        self.df_m = yfinance_model.get_market(self.df.index[0], self.m_tick)
+        self.returns, self.variance = portfolio_model.get_return(df, self.df_m, n)
```
jmaslek, line 259: You should add a docstring either to the class of to __init__ that explains what all the inputs are.  

Maybe not as intense, but check out https://github.com/scikit-learn/scikit-learn/blob/1f8825c8dd6238355191e3d1c98f4e4d54cfbf16/sklearn/linear_model/_bayes.py#L24. as an example

jmaslek (approved):

jmaslek:
@DidierRLopes Im happy to merge.  tests are failing because fred site is down.

PR #891, opened 2021-11-01
Add a check in dd report to avoid crash

DidierRLopes:


gamestonk_terminal/reports/due_diligence.ipynb
```diff
@@ -114,7 +114,10 @@
    "source": [
     "from gamestonk_terminal.stocks.due_diligence import finviz_view\n",
     "\n",
-    "finviz_view.analyst(ticker=ticker, export='')"
+    "try:\n",
+    "    finviz_view.analyst(ticker=ticker, export='')\n",
```
jmaslek, line 118: The export = '' shouldn't be needed

jmaslek (approved):

gamestonk_terminal/reports/due_diligence.ipynb
DidierRLopes: I thought the same, but it failed when I tried.
Screenshot 2021-11-01 at 01 48 49

I'm about to do a PR to fix something else Colin reported, I'll also add an optional export="" to analyst

PR #894, opened 2021-11-01
Small tweaks to DCF and add thepassiveinvestor to requirements

JerBouma:
Noticed two things about DCF:

- When you rename the folder of GST, you can not export a DCF anymore. Fixed this by adding _os.getcwd()_ instead of 'Gamestonk Terminal'.
- This function also uses the financedatabase package. Noticed that it uses the industry name from yfinance which can be different because in the database a few adjustments are made to the names due to some characters returning an UNICODE when saved to JSON. Added these corrections to the function here as well. (an example: try running a DCF for MSFT, that doesn't work)

Furthermore, it looks like thepassiveinvestor was dropped out of the requirements, added this again.

**By the way, I see that all commits from an earlier pull requests are also included (while not affecting the files changed). No idea how to fix that.**


gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
```diff
@@ -77,8 +77,7 @@ def create_workbook(self):
             self.run_audit()
 
         trypath = os.path.join(
-            "..",
-            "GamestonkTerminal",
+            os.getcwd(),
```
DidierRLopes, line 80: I would recommend something like this instead: 
```python
os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
```

This is because with `os.getcwd()` you rely on the location of where you execute terminal. Which in 99% of the cases will be `python terminal.py` because you are on the same directory of `terminal.py`. However, let's say you are in an upper folder, you can run the terminal with `python GamestonkTerminal/terminal.py` in those cases `os.getcwd()` would retrieve directory above which will make this path to not exist.

Thus, to be 100% correct we can rely on the location of this file as the "anchor" and therefore regardless of what the user does, it should work

requirements-full.txt
```diff
@@ -224,6 +224,7 @@ termcolor==1.1.0
 terminado==0.12.1; python_version >= "3.6"
 testpath==0.5.0; python_version >= "3.7"
 textwrap3==0.9.2; python_version >= "3.6"
+thepassiveinvestor==1.0.9
```
DidierRLopes, line 227: @aia do you know why this keeps being dropped?

gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
JerBouma: Added this fix. Unfortunately it required a bit more than just this line. This is because with 
```
os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
```

You select the stocks directory while you want the main directory. This results in this line:
```
  os.path.abspath(
      os.path.join(
          os.path.dirname(os.path.dirname(os.path.dirname(__file__))), ".."
      )
  )
```

gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
```diff
@@ -77,8 +77,11 @@ def create_workbook(self):
             self.run_audit()
 
         trypath = os.path.join(
-            "..",
-            "GamestonkTerminal",
+            os.path.abspath(
+                os.path.join(
+                    os.path.dirname(os.path.dirname(os.path.dirname(__file__))), ".."
+                )
+            ),
```
JerBouma, line 84: This seems to do the trick, works in both cases when you open terminal.py from the main folder or from somewhere else.

gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
DidierRLopes: I said something like this because I used this Sunday on the discordbot :)

Something that would work in your case and is more elegant would be:
```
os.path.abspath(
   os.path.join(
      os.path.dirname(__file__),
      "..",
      "..", 
      ".."
   )
)
```

This is how it works:
*  os.path.dirname(__file__): directory of where the current file is (__file__), which is dcf_view.py and therefore fundamental_analysis.
* When you do os.path.join, you are just adding "/" or "\" depending on OS, therefore:
   * with 1 `..` you get into stocks (1 level above) 
   * with 2 `..` you get into gamestonk_terminal (2 levels above) 
   * with 3 `..` you get into GamestonkTerminal (3 levels above) 
* The absolute path just simplifies it to it's simplest form removing those `..` and effectively getting the absolute path

gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
DidierRLopes: Look into my other comment for a more elegant solution

DidierRLopes (approved):

DidierRLopes:
Niceee, thank you 😄

PR #899, opened 2021-11-02
Rename finbrain command to not use source and comment out forex/ba menu

jmaslek:
As per didiers request:

- rename crypto - fibrain command
- Comment out forex/ba menu

### DIFF
--- a/gamestonk_terminal/cryptocurrency/crypto_controller.py
+++ b/gamestonk_terminal/cryptocurrency/crypto_controller.py
@@ -60,7 +60,7 @@ class CryptoController:
     ]
 
     CHOICES_COMMAND = [
-        "finbrain",
+        "sentiment",
         "chart",
         "load",
         "find",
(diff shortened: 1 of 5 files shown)


aia (approved):

piiq (approved):

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -121,7 +121,7 @@ def print_help(self):
     load        load a specific cryptocurrency for analysis
     chart       view a candle chart for a specific cryptocurrency
     find        alternate way to search for coins
-    finbrain    crypto sentiment from 15+ major news headlines
+    sentiment    crypto sentiment from 15+ major news headlines [Finbrain]
```
DidierRLopes, line 124: Shouldn't this be "headlines" to match the same we have on "stocks"?

DidierRLopes: also, there's an additional space here, crypto isn't aligned with above

DidierRLopes (changes requested):
alignment of menu

DidierRLopes (approved):

PR #902, opened 2021-11-02
Add contributing document to discord

DidierRLopes:
@northern-64bit could you provide feedback if you have time?


jmaslek (approved):
Think this is looking good.  I will add to this once I go through and add some commands 👍

PR #904, opened 2021-11-03
Add bandit to dev requirements

aia:


piiq (approved):
bandit is useful, yeah

PR #906, opened 2021-11-03
Risk neutral options valuation

colin99d:
This PR lets the user add different estimates for a stock's price at the given expiration and the probability of that price happening. The user can use this to compare their estimated options value with current options values. 

**Note:** I have NOT added Hugo documentation yet because I am assuming you guys will have me chance stuff.

### DIFF
--- a/gamestonk_terminal/options/yfinance_view.py
+++ b/gamestonk_terminal/options/yfinance_view.py
@@ -496,17 +496,17 @@ def show_parity(
 
     Parameters
     ----------
-    ticker: str
+    ticker : str
         Ticker to get expirations for
-    exp: str
+    exp : str
         Expiration to use for options
-    put: bool
+    put : bool
         Whether to use puts or calls
-    ask: bool
+    ask : bool
         Whether to use ask or lastPrice
-    mini: float
+    mini : float
         Minimum strike price to show
-    maxi: float
+    maxi : float
         Maximum strike price to show
 
     """
@@ -567,8 +567,8 @@ def show_parity(
     if maxi is None:
         maxi = filtered.strike.quantile(0.75)
 
-    filtered = filtered.loc[filtered["strike"] > mini]
-    filtered = filtered.loc[filtered["strike"] < maxi]
+    filtered = filtered.loc[filtered["strike"] >= mini]
+    filtered = filtered.loc[filtered["strike"] <= maxi]
 
     show = filtered[["strike", diff]].copy()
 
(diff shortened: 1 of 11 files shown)


gamestonk_terminal/options/yfinance_view.py
```diff
@@ -590,3 +590,66 @@ def show_parity(
...
+    """
+    if put:
+        chain = get_option_chain(ticker, exp).puts
+    else:
+        chain = get_option_chain(ticker, exp).calls
+
+    r_date = datetime.strptime(exp, "%Y-%m-%d").date()
+    delta = (r_date - date.today()).days
+    vals = []
+    rf = get_rf()
```
jmaslek, line 623: I would allow this to be specified in the controller.

jmaslek:
I only had the one comment, but when you add the Hugo documentation, I would like to see a more in depth description of what the command does and how it goes about calculating the value.

gamestonk_terminal/options/yfinance_view.py
colin99d: The expiration?

colin99d: Ohhhhh. Can do!

website/content/options/rnval/_index.md
```diff
@@ -0,0 +1,17 @@
+```
+usage: rnval [-p] [-m MIN] [-M MAX] [-r RISK] [-h]
+```
+
+calculates the value of an option based on ending price estimates. this
+calculates the expected value of the option payoff and then compares it
+to current prices.
+learn more: https://en.wikipedia.org/wiki/Rational_pricing#Delta_hedging
```
jmaslek, line 8: This is not delta hedging.

Aren't you just calculating the expected value for the option assuming it grew at a risk free rate (i.e https://www.equitymethods.com/articles/real-world-vs-risk-neutral-understanding-the-difference-between-valuation-and-derived-service-period-assumptions/ )?

jmaslek: So I would envision an example (similar to something like the link I sent) that walks through the calculation and what the calculation actually means.  Also, unrelated but can this be applied to continuous distributions?

colin99d: Would you like me to just include this link? Also, this could be related to continuous distributions.

website/content/options/rnval/_index.md
```diff
@@ -2,10 +2,13 @@
...
-calculates the expected value of the option payoff and then compares it
-to current prices.
-learn more: https://en.wikipedia.org/wiki/Rational_pricing#Delta_hedging
+calculates the expcected value of a given option by multiplying
+the payoff at each predicted stock price by the probability of
+that stock price happening. This expected profit is divided by
+one over the risk free rate to determine the value of those
+future dollars today. The expected payoff for each strike price
+is then compared with the last traded price to determine if there
+are differences in the market.
```
jmaslek: Can you include a toy example of a calculation and maybe what the meaning is?

website/content/options/rnval/_index.md
```diff
@@ -0,0 +1,20 @@
...
+usage: rnval [-p] [-m MIN] [-M MAX] [-r RISK] [-h]
+```
+
+calculates the expcected value of a given option by multiplying
+the payoff at each predicted stock price by the probability of
+that stock price happening. This expected profit is divided by
+one over the risk free rate to determine the value of those
+future dollars today. The expected payoff for each strike price
+is then compared with the last traded price to determine if there
+are differences in the market.
```
jmaslek, line 11: Can you go a little further into an example calculation and what the user should take away from using the function?

jmaslek (approved):

PR #908, opened 2021-11-04
Improves load and candle relation + 3 bonus READMEs

DidierRLopes:
This can still be improved further, but is already an improvement.

Also allows to select an end date as @JerBouma mentioned that can be useful for someone to analyse a specific time in history.

PS: also improves 3 READMEs as a bonus :)


jmaslek:
```python
(✨) (stocks)> load amc -i 15
Loading Intraday 15min AMC stock with starting period 2021-09-06 for analysis.

(✨) (stocks)> candle
/opt/anaconda3/envs/gst/lib/python3.9/site-packages/mplfinance/_arg_validators.py:35: UserWarning: 

 ================================================================= 

   WARNING: YOU ARE PLOTTING SO MUCH DATA THAT IT MAY NOT BE
            POSSIBLE TO SEE DETAILS (Candles, Ohlc-Bars, Etc.)
   For more information see:
   - https://github.com/matplotlib/mplfinance/wiki/Plotting-Too-Much-Data
   
   TO SILENCE THIS WARNING, set `type='line'` in `mpf.plot()`
   OR set kwarg `warn_too_much_data=N` where N is an integer 
   LARGER than the number of data points you want to plot.

 ================================================================ 
  warnings.warn('\n\n ================================================================= '+
```
Idk if you want to add a warning filter?

jmaslek:
Interesting trend line
trend

DidierRLopes:
> ```python
> ```python
> (✨) (stocks)> load amc -i 15
> Loading Intraday 15min AMC stock with starting period 2021-09-06 for analysis.
> 
> (✨) (stocks)> candle
> /opt/anaconda3/envs/gst/lib/python3.9/site-packages/mplfinance/_arg_validators.py:35: UserWarning: 
> 
>  ================================================================= 
> 
>    WARNING: YOU ARE PLOTTING SO MUCH DATA THAT IT MAY NOT BE
>             POSSIBLE TO SEE DETAILS (Candles, Ohlc-Bars, Etc.)
>    For more information see:
>    - https://github.com/matplotlib/mplfinance/wiki/Plotting-Too-Much-Data
>    
>    TO SILENCE THIS WARNING, set `type='line'` in `mpf.plot()`
>    OR set kwarg `warn_too_much_data=N` where N is an integer 
>    LARGER than the number of data points you want to plot.
> 
>  ================================================================ 
>   warnings.warn('\n\n ================================================================= '+
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> Idk if you want to add a warning filter?
> ```

Can do. Was very unsure what to do there. The reason why I left it is because sometimes the plot didn't pop up and having the warning would let the user know about this.

jmaslek (commented):
To keep with your convention - shouldn't trendline_api get put into yfinance model since it relies on yfinance data.

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -407,22 +426,18 @@ def load(
         return [s_ticker, s_start, s_interval, df_stock]
 
 
-def display_candle(s_ticker: str, s_start: datetime, plotly: bool):
-    """Shows candle plot of loaded ticker
+def display_candle(s_ticker: str, df_stock: pd.DataFrame, plotly: bool):
+    """Shows candle plot of loaded ticker. [Source: Yahoo Finance]
```
jmaslek, line 430: Could be yfinance/iex or alpha vantage....

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -407,22 +426,18 @@ def load(
         return [s_ticker, s_start, s_interval, df_stock]
 
 
-def display_candle(s_ticker: str, s_start: datetime, plotly: bool):
-    """Shows candle plot of loaded ticker
+def display_candle(s_ticker: str, df_stock: pd.DataFrame, plotly: bool):
```
jmaslek, line 429: Also should we name poorly to something more intuitive like `interactive` (or `ion` or `it` like docker)

jmaslek:
>Was very unsure what to do there. The reason why I left it is because sometimes the plot didn't pop up and having the warning would let the user know about this.

If the plot doesn't come up then may be better to have it (or can we catch it and print something custom?  )

DidierRLopes:
> To keep with your convention - shouldn't trendline_api get put into yfinance model since it relies on yfinance data.

Yes, was being lazy to do that refactoring because the economy report relies on it 😄

DidierRLopes:
> > Was very unsure what to do there. The reason why I left it is because sometimes the plot didn't pop up and having the warning would let the user know about this.
> 
> If the plot doesn't come up then may be better to have it (or can we catch it and print something custom? )

Wasn't able to catch it :/ I tried a try-except where I did a type==line on the except, but it seems to be just an error message.

jmaslek (approved):

gamestonk_terminal/stocks/stocks_helper.py
DidierRLopes: It was actually not like that when I committed this :) There was a yf.download function within this candle function. But now I changed that, so now it makes sense to add the others

gamestonk_terminal/stocks/stocks_helper.py
DidierRLopes: Will do that

DidierRLopes:
@jmaslek should be better now!

PR #917, opened 2021-11-05
Add analytics beacons to hugo and readthedocs

piiq:
This PR integrates the readthedocs documentation with the privacy focused Cloudflare Analytics.

The beacon is placed in the template footer and the token is being put in place as a part of the build process at readthedocs.io.


piiq:
@aia I found and fixed a couple of issues with the dependencies that break readthedocs build process:

- Harmonized keras and tensorflow versions. This involved downgrading keras to 2.6 because 2.7 doesn't work with latest tf and the documentation for the prediction menu can't be built
- Added `math` and `binascii` to allowed c extensions list in `.pylintrc`
- I've noticed that the only way for the readthedocs to build documentation from `requirements-full.txt` is is the file is generated with a `--dev` flag that you were not using in the command that you've showed. So the command that generates a requirements file that works with documentation generator is:

```bash
poetry export -f requirements.txt --dev --extras prediction --without-hashes -o requirements-full.txt
```

requirements-full.txt
```diff
@@ -190,7 +190,7 @@ pytrends==4.7.3
 pytz-deprecation-shim==0.1.0.post0; python_version >= "3.6" and python_full_version < "3.0.0" or python_full_version >= "3.6.0" and python_version >= "3.6"
 pytz==2021.3; python_full_version >= "3.6.8" and python_version >= "3.6" and python_full_version < "4.0.0"
 pyupgrade==2.18.2; python_full_version >= "3.6.1"
-pywin32==301; sys_platform == "win32" and platform_python_implementation != "PyPy" and python_version >= "3.6"
+pywin32==302; sys_platform == "win32" and platform_python_implementation != "PyPy" and python_version >= "3.6"
```
aia, line 193: I keep running into issues with pywin32. I think we should keep it at 301.

aia: actually 302 is working again for me now. Windows inconsistencies are annoying.

DidierRLopes: Is this good to go then?

piiq: It is to me=) @aia ?

DidierRLopes (approved):

piiq:
I’ve updated the PR to include adding a beacon to the Hugo web-site. The token is already added to the repo secrets so we should see something as soon as we merge this.

DidierRLopes (approved):

aia (approved):

PR #919, opened 2021-11-05
Discord Bot: Added ta context

northern-64bit:
This PR adds the technical analysis context to the discord bot.


discordbot/stocks/technical_analysis/ta_menu.py
```diff
@@ -0,0 +1,916 @@
...
+    async def sma(
+        self,
+        ctx: discord.ext.commands.Context,
+        ticker="",
+        window="",
+        offset="",
+        start="",
+        end="",
+    ):
+        """Displays chart with sma of a given stock and sends it
```
DidierRLopes, line 78: I would rephrase all these sentences to:
`"""Displays chart with simple moving average [Yahoo Finance]"""`
i.e.:
* Avoid the `and sends it`
* Be more explicit what it does so when a user does !help they know what this command does, 
* Refer to the data source between brackets []

discordbot/stocks/technical_analysis/ta_menu.py
```diff
@@ -0,0 +1,916 @@
...
+
+        Returns
+        -------
+        Sends a message to the discord user with the commands from the stocks.ta context.
+        The user can then select a reaction to trigger a command.
+        """
+
+        if cfg.DEBUG:
+            print(f"!stocks.ta {ticker}")
+
```
DidierRLopes, line 639: Doing `!stocks.ta` should generate an error saying that a ticker is required

discordbot/stocks/technical_analysis/ta_menu.py
```diff
@@ -0,0 +1,916 @@
...
+        cols_temp = []
+        cols = []
+
+        if ticker != "":
+            stock = helpers.load(ticker, datetime.now() - timedelta(days=365))
+            if stock.empty:
```
DidierRLopes, line 645: To make it simpler you can check here for stock.empty or ticker == "", and that way there's no need for an extra else

discordbot/stocks/technical_analysis/ta_menu.py
```diff
@@ -0,0 +1,916 @@
...
+                )
+
+        else:
+            text = (
+                "\nAll commands aro only available when providing a ticker with:"
+                "\n!stocks.ta <TICKER>"
+            )
+            cols.append(text)
+
+        emoji_list = ["0️⃣", "1️⃣", "2️⃣", "3️⃣", "4️⃣", "5️⃣", "6️⃣", "7️⃣", "9️⃣"]
```
DidierRLopes, line 709: Do you have anything against command number 8? 😝

discordbot/stocks/technical_analysis/ta_menu.py
```diff
@@ -0,0 +1,916 @@
...
+                                    id="front",
+                                    style=discord_components.ButtonStyle.green,
+                                    disabled=True,
+                                ),
+                            ]
+                        ]
+                        await main_message.edit(components=components)
+                        return
+
+            except asyncio.TimeoutError:
```
DidierRLopes, line 885: In case of a TimeoutError we prob should consider sending an error message? What do you think?

northern-64bit: The timeout error happens after the selected timer runs out and makes that the user no longer can press the buttons of the message. So the "error" isn't an real error or problem.

DidierRLopes: True, my point is to send a message saying that. Rather than just not allowing user to select options say: "Error timeout - you snooze you lose! 😋 "

discordbot/stocks/technical_analysis/ta_menu.py
northern-64bit: This is checked on the part with if ticker != "":, so this won't be needed to check again.

discordbot/stocks/technical_analysis/ta_menu.py
northern-64bit: Should I append it to the menu message or send a new message?

DidierRLopes: Send a new message I'd say

discordbot/stocks/technical_analysis/ta_menu.py
DidierRLopes: Screenshot 2021-11-06 at 19 08 30

The problem is me, as a user, not getting a reply for this

DidierRLopes: I should get an error saying that a ticker is necessary

discordbot/stocks/technical_analysis/recom.py
```diff
@@ -21,6 +21,11 @@ async def recom_command(ctx, ticker=""):
             ticker, "america", "", ""
         )
 
+        cols = list(recom.columns)
+        a, b = cols.index("RECOMMENDATION"), cols.index("SELL")
+        cols[b], cols[a] = cols[a], cols[b]
+        recom = recom[cols]
```
DidierRLopes, line 27: if recommended is a data frame you can do
```
recom = recom[["BUY", "NEUTRAL", "SELL", "RECOMMENDATION"]]
```
which is much easier

DidierRLopes (approved):
Good work :)

PR #922, opened 2021-11-06
Performance Ratios

colin99d:
Added key performance ratios with a brief description on what they mean. This gives investors a way to compare their performance with other investors. This also fixes a glitch where a matplotlib graph would be pulled up when the ar command was run.


gamestonk_terminal/portfolio/portfolio_helper.py
```diff
@@ -60,3 +60,23 @@ def clean_name(name: str) -> str:
...
     """
     return name.replace("beta_", "").upper()
+
+
+def get_fraction(n: float, d: float) -> str:
+    """Turn two numbers into a fraction
+
+    Parameters
+    ----------
+    n : float
```
aia, line 70: I'd recommend more descriptive names.

numerator instead of n and denominator instead of d.

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -222,17 +230,16 @@ def plot_ef(
     ax.scatter(stds, rets, marker=".", c=sharpes, cmap="viridis_r")
     plotting.plot_efficient_frontier(ef, ax=ax, show_assets=True)
     # Find the tangency portfolio
-    rfrate = get_rf()
-    ret_sharpe, std_sharpe, _ = ef.portfolio_performance(risk_free_rate=rfrate)
+    ret_sharpe, std_sharpe, _ = ef.portfolio_performance(risk_free_rate=rf)
```
aia, line 233: I'd stick to rfrate as a name.

Descriptive names very much improve readability.

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -307,15 +319,27 @@ def generate_pg1(self, report: canvas.Canvas) -> None:
         )
         main_t = portfolio_model.get_main_text(self.returns)
         reportlab_helpers.draw_paragraph(report, main_t, 30, 410, 550, 200)
+        rp = self.returns["return"][-1]
```
aia, line 322: Similar comment here.

Very difficult to make sense out of rp, b, mar tnr.

I would spell them all out - current_return, beta, market_return, fraction_return.

gamestonk_terminal/portfolio/portfolio_helper.py
```diff
@@ -60,3 +60,23 @@ def clean_name(name: str) -> str:
...
+        The numerator
+    d : float
+        The denominator
+
+    Returns
+    ----------
+    text : str
+        A fraction as a string
+    """
+    if d > 0:
```
DidierRLopes: I think here should be if `d != 0` as you can still get the fraction if d is a negative number

DidierRLopes (commented):
I agree with Artem comment, and I think you should start doing that going forward. Code readability is super super important, even more so in an open source project.

Believe me, when you look into this code in a couple of months, you'll appreciate being super explicit and even adding comments to it. Let alone if it's a new person looking into the code.

You can look into this https://stackoverflow.com/questions/12106515/is-there-anything-to-be-gained-from-short-variable-names

Btw, it's ok to use super short variable names in this case:
- If you are in a loop and want to use that variable to iterate through a variable that has a descriptive name, like `for val in fibonacci_sequence`

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -317,23 +317,30 @@ def generate_pg1(self, report: canvas.Canvas) -> None:
...
-        mar = self.returns[("Market", "Return")][-1]
-        srp = portfolio_helper.get_fraction(
-            rp - self.rf, np.std(self.returns["return"])
+        main_text = portfolio_model.get_main_text(self.returns)
+        reportlab_helpers.draw_paragraph(report, main_text, 30, 410, 550, 200)
+        current_return = self.returns["return"][-1]
+        beta = self.betas["total"][-1]
+        market_return = self.returns[("Market", "Return")][-1]
+        sharpe = portfolio_helper.get_fraction(
+            current_return - self.rf, np.std(self.returns["return"])
```
DidierRLopes, line 326: much much nicer :)

gamestonk_terminal/portfolio/portfolio_helper.py
colin99d: This if is to prevent negative numbers from returning. The only value that can be negative is the Treynor ratio, and the Treynor ratio becomes nonsensical if the denominator is negative.

gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -429,3 +429,25 @@ def get_beta_text(df: pd.DataFrame) -> str:
...
     )
     return string
+
+
+def get_perm_text() -> str:
+    """Get description for performance text
+
+    Returns
+    ----------
+    t : str
```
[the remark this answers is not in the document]
colin99d: Would it be better if it was just a variable?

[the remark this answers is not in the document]
DidierRLopes: I think so yes. Unless you wanted this sentence to depend on the parameter given by the function it makes sense to have it here I'd say.

gamestonk_terminal/portfolio/portfolio_helper.py
DidierRLopes: Ok, so this is something that you should bear in mind. The name of the function is `get_fraction` and therefore if I was looking for a similar function I would think "Ohh, ok, look this function gives me the fraction". But in this case you are doing a specific manipulation (not accepting negative values to be returned).

So, I would proceed in 1 of 2 ways:
- Either rename the function so people know that this is for your specific case, and mention on the docstring that it doesn't accept negative denominators.
- Allow the function to just return the division value with 2 decimal places, and then when you use it in the context you are using have the check for negative denominator before calling the function.

gamestonk_terminal/portfolio/portfolio_helper.py
```diff
@@ -71,12 +73,16 @@ def get_fraction(numerator: float, denominator: float) -> str:
...
+    allow_negative_denominator : bool
+        If false, a negative denominator will return "NA"
 
     Returns
     ----------
-    text : str
+    str
         A fraction as a string
     """
+    if allow_negative_denominator and denominator != 0:
```
DidierRLopes, line 84: I think this is overcomplicating, why not letting as it was and calling the function
```
def get_fraction_with_positive_denominator(
     numerator: float, denominator: float
 ) -> str:
     """Turn into string the result of the numerator divided by the denominator, in which the denominator must be a positive value
     ....
```

colin99d: I only like this way better in case I ever add ratios that can have negative denominators. If you want me to change it I can.

DidierRLopes: I think this function is a bit overkill as is, IMHO because you can do:
```
(numerator/denominator):.2f if denominator > 0 else "N/A"
```
And is more readable than having the function itself

aia (approved):

gamestonk_terminal/portfolio/portfolio_helper.py
colin99d: I like that better.

PR #925, opened 2021-11-06
Fix spelling and add missing thepassiveinvestor

aia:


poetry.lock
```diff
@@ -3091,6 +3091,14 @@ category = "main"
 optional = false
 python-versions = "*"
 
+[[package]]
```
DidierRLopes, line 3094: We need to fix this one, it keeps on getting removed somehow...

gamestonk_terminal/options/chartexchange_view.py
```diff
@@ -26,7 +26,7 @@ def display_raw(
     call : bool
         Whether the underlying asset should be a call or a put
     price : float
-        The stike of the expiration
+        The stake of the expiration
```
DidierRLopes, line 29: should be *strike I think

gamestonk_terminal/portfolio/portfolio_optimization/po_controller.py
```diff
@@ -714,7 +714,7 @@ def call_yolo(self, _):
 
     @try_except
     def add_stocks(self, other_args: List[str]):
-        """Add ticker or Select tickes for portfolio to be optimized"""
+        """Add ticker or Select ticks for portfolio to be optimized"""
```
DidierRLopes, line 717: tickers*

DidierRLopes (approved):

PR #928, opened 2021-11-07
Validate loaded data

colin99d:
Small PR to validate loaded data.

### DIFF
--- a/gamestonk_terminal/portfolio/portfolio_model.py
+++ b/gamestonk_terminal/portfolio/portfolio_model.py
@@ -74,6 +74,28 @@ def load_df(name: str) -> pd.DataFrame:
         df["Name"] = df["Name"].str.lower()
         df["Type"] = df["Type"].str.lower()
         df["Date"] = pd.to_datetime(df["Date"], format="%Y/%m/%d")
+
+        for item in ["Quantity", "Price", "Fees", "Premium"]:
+            result = any(df[item] < 0)
+            if result:
+                print(
+                    f"The column '{item}' has a negative value. Ensure all values are positive."
+                )
+                return pd.DataFrame()
+
+        if len(df[~df["Type"].isin(["cash", "stock"])].index):
+            print("Warning: 'Type' other than 'cash' and 'stock' will be ignored.")
+
+        if len(
+            df[
+                ~df["Side"]
+                .str.lower()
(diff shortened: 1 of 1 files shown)


gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -74,6 +74,28 @@ def load_df(name: str) -> pd.DataFrame:
...
+        for item in ["Quantity", "Price", "Fees", "Premium"]:
+            result = any(df[item] < 0)
+            if result:
+                print(
+                    f"The column '{item}' has a negative value. Ensure all values are positive."
+                )
+                return pd.DataFrame()
+
+        if len(df[~df["Type"].isin(["cash", "stock"])].index):
+            print("Warning: 'Type' other than 'cash' and 'stock' will be ignored.")
```
DidierRLopes, line 87: I think you are missing a `return pd.DataFrame()` here

gamestonk_terminal/portfolio/portfolio_model.py
```diff
@@ -74,6 +74,28 @@ def load_df(name: str) -> pd.DataFrame:
...
+
+        for item in ["Quantity", "Price", "Fees", "Premium"]:
+            result = any(df[item] < 0)
+            if result:
+                print(
+                    f"The column '{item}' has a negative value. Ensure all values are positive."
+                )
+                return pd.DataFrame()
+
+        if len(df[~df["Type"].isin(["cash", "stock"])].index):
```
DidierRLopes, line 86: Another way I would have done this if it was me would have been:
```
if sum(df["Type"].isin(["cash", "stock"]).values) != len(df)
```

When you do `df["Type"].isin(["cash", "stock"]).values` these values become `True` or `False`, when you sum them `False = 0` and `True = 1`. Therefore, if the sum of these is equal to the length of the data frame it means all values were `cash or stock`.

But what you have works :)

gamestonk_terminal/portfolio/portfolio_model.py
colin99d: I still want this to work as normal, I just want the user to know other assets will not be analyzed.

DidierRLopes: Gotcha!

DidierRLopes (approved):

PR #929, opened 2021-11-07
Discordbot Heroku

colin99d:
Tired of relying on your ol' PC to handle your discord bot? Fear no more! This PR allows users to EASILY host their discord both for free without a credit card on Heroku. Simply follow the Readme to have a live bot in no time.

**Note**: Now that we are adding packages and features that are important, but that most users will not use, should we consider having a permanently separate branch for people looking to create discord bots?


poetry.lock
```diff
@@ -11,36 +11,22 @@ six = "*"
 
 [[package]]
 name = "aiohttp"
-version = "3.8.0"
+version = "3.7.4.post0"
```
DidierRLopes, line 14: Can you not update the poetry.lock? There are several updates that shouldn't happen here!

This is one of them, chardet is another one, ...

DidierRLopes (changes requested):
We separate discord from the terminal by not having any bot dependency on the requirements or poetry.lock of the terminal.

We can create a requirements.txt that lives on the discordbot folder.

So for this PR I suggest removing changes to poetry.lock and requirements.txt and just add instructions as you have on README

discordbot/README.md
```diff
@@ -52,6 +52,21 @@ https://discord.com/oauth2/authorize?client_id=CLIENTID&scope=bot, where **CLIEN
 
 **NOTE:** Registering is free for all open source projects and if your discord bot uses fewer than 1250 uploads per day.
 
+### HEROKU
+
+1. Add the following lines into requirements.txt
+    1. discord==1.7.3
```
DidierRLopes, line 58: In fact these already belong to initial installation some lines above, so no need to repeat

poetry.lock
colin99d: aiohttp does not work with one of the packages (I think it was Discord). This is the only thing that lets all the packages work together nicely.

discordbot/README.md
colin99d: Forgot to change that.

poetry.lock
DidierRLopes: The reason why we updated to 3.8.0 is to remove the dependency for chardet which has a LGPL license

colin99d: https://github.com/Rapptz/discord.py/blob/master/requirements.txt
Looks like this is going to conflict with discord.py requirements. Since this project is no longer actively maintained should fork it with and updated requirements.txt?

DidierRLopes: Hummm, yes, I think that's a good idea.

But in any case the discord requirements should be separated from the main terminal one as this is an "optional usage"

colin99d: Added a solution to this. User does not need to install new requirements, they just need to be present for Heroku.

discordbot/requirements.txt
```diff
@@ -0,0 +1,236 @@
+aiohttp==3.7.4.post0; python_version >= "3.6" and python_full_version >= "3.5.3"
```
DidierRLopes, line 1: What if we had here only the ones that we need on top of the default requirements.txt?

I.e. aiohttp, discord, discord_components, pymgur?

colin99d: My only worry would be if some of the other dependencies are also different versions, and if in the future we update other dependencies in the main requirements.txt, but that interferes with the discord.py. This discord.py package will be a continual pain since it is archived.

DidierRLopes: Sounds good to me, I agree!

DidierRLopes (approved):

PR #930, opened 2021-11-08
Yup, incoming NFT context

DidierRLopes:
Uses this website https://nftcalendar.io/ from a video recommended by @aia

### DIFF
--- a/gamestonk_terminal/nft/nft_controller.py
+++ b/gamestonk_terminal/nft/nft_controller.py
@@ -1,0 +1,259 @@
+import argparse
+from typing import List
+
+from prompt_toolkit.completion import NestedCompleter
+from gamestonk_terminal import feature_flags as gtff
+from gamestonk_terminal.menu import session
+from gamestonk_terminal.helper_funcs import (
+    EXPORT_ONLY_RAW_DATA_ALLOWED,
+    MENU_GO_BACK,
+    MENU_QUIT,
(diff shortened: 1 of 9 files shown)


piiq:
Lalalalalala 💃
Lalalalalala 🕺

piiq:
@DidierRLopes I think you forgot to enable the context in the main menu. Or was it on purpose so that the feature is like a placeholder or easter egg?

DidierRLopes:
> @DidierRLopes I think you forgot to enable the context in the main menu. Or was it on purpose so that the feature is like a placeholder or easter egg?

Nahh, I forgot to commit that file! 🙄

jose-donato:
@DidierRLopes Just tried it locally. All four features working as expected.
Since they are all related to drops we could simplify and call it `drops` and then pass arguments `drops --upcoming` or `drops --ongoing`. What do you think about that?

DidierRLopes:
> @DidierRLopes Just tried it locally. All four features working as expected.
> Since they are all related to drops we could simplify and call it `drops` and then pass arguments `drops --upcoming` or `drops --ongoing`. What do you think about that?

I agree in doing that refactoring later on, I think for now is best to have more commands since this menu is just getting started 😋

piiq (approved):

jose-donato (approved):
Everything looks good to me!

PR #935, opened 2021-11-09
Force usage of tensorflow 2.6.1 because 2.6.2 not available on windows

piiq:


piiq:
TF 2.6.2 failed during the Windows 10 Full Clean Build with ML test (that's only run when PR is merged and not when PR is created?) 
See: https://github.com/GamestonkTerminal/GamestonkTerminal/runs/4156422695?check_suite_focus=true
So this PR downgrades the tf version to the latest one available for windows.

P.S. Thanks, here I've used the poetry hack that you told me about

piiq:
@aia thanks for pointing out that I have forgotten to commit the requirements files. Will merge after the tests pass

PR #936, opened 2021-11-10
Adding the due diligence context

northern-64bit:
This pr adds the due diligence commands, which don't require api keys, to the discord bot.

### DIFF
--- a/discordbot/discordbot.py
+++ b/discordbot/discordbot.py
@@ -34,6 +34,7 @@ gst_bot.load_extension("generic_commands")
 gst_bot.load_extension("economy.economy_menu")
 gst_bot.load_extension("stocks.dark_pool_shorts.dps_menu")
 gst_bot.load_extension("stocks.technical_analysis.ta_menu")
+gst_bot.load_extension("stocks.due_diligence.dd_menu")
 
 # Get all command names
 all_cmds = gst_bot.all_commands.keys()
--- a/discordbot/economy/economy_menu.py
+++ b/discordbot/economy/economy_menu.py
@@ -177,11 +177,17 @@ class EconomyCommands(discord.ext.commands.Cog):
                 await msg.remove_reaction(emoji, ctx.bot.user)
 
         except asyncio.TimeoutError:
-            text = text + "\n\nCOMMAND TIMEOUT."
-            embed = discord.Embed(title=title, description=text)
-            await msg.edit(embed=embed)
             for emoji in emoji_list:
                 await msg.remove_reaction(emoji, ctx.bot.user)
+            embed = discord.Embed(
+                description="Error timeout - you snooze you lose! 😋",
+                colour=cfg.COLOR,
+                title="TIMEOUT Economy Menu",
+            ).set_author(
+                name=cfg.AUTHOR_NAME,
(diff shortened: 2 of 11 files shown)


jmaslek:
Awesome job :)

Just curiously, is it possible to print out tables?  Example attached where the output is kinda hard to read in raw text for

Screen Shot 2021-11-11 at 1 34 59 PM
m

northern-64bit:
> Just curiously, is it possible to print out tables? Example attached where the output is kinda hard to read in raw text for


It's not really possible due to discord having different line lengths on different platforms. I've tried to use tabulate, but discord destroyed the formating and made it nearly useless. This is how it looked, but it wouldn't look as good for the command that you mentioned:
image
I'm unsure what is the best option here, but I thought that the implementing way it was in the image could be confusing and unclear. What do you think @jmaslek ?

jmaslek:
>  I'm unsure what is the best option here, but I thought that the implementing way it was in the image could be confusing and unclear. What do you think @jmaslek ?

Im not sure.  But I agree that its better as it is.

piiq:
How does an html payload that's sent with the message look like? Technically we can craft static web pages with a table and have this thing as a message attachment. Discord should process the page and generate some sort of a preview that will be the image part and on click the user can access the tabular data

northern-64bit:
> How does an html payload that's sent with the message look like? Technically we can craft static web pages with a table and have this thing as a message attachment. Discord should process the page and generate some sort of a preview that will be the image part and on click the user can access the tabular data

Yes, this may be a possible solution, but I don't know enough about something like this to implement it. Another possibility is to put the table on a background image and attach the image and add a flag to the commands for raw imput. Now after some testing with it, I've seen that tabulate works if every value in a column has the same length so maybe it's possible to implement a function that makes them the same length. Otherwise we may need to create a "tabulate" specialised for discord.

northern-64bit:
I think I fixed this:
Skärmbild 2021-11-12 110514

piiq:
Since this started, can/should we do something to have the row fit a single line? Like rename the column names and maybe round the decimals?

jmaslek:
> I think I fixed this:
> Skärmbild 2021-11-12 110514


Personally I like everything on one line better.

northern-64bit:
> 
> 
> Since this started, can/should we do something to have the row fit a single line? Like rename the column names and maybe round the decimals?

I agree that it would be good to have everything in one line. There are many other commands that not always return one line. Therefore I think that it would be better to merge this pr and add a note/label with this issue in the discordbot project, since this is a new thing which has nothing to do with the due diligence commands specifically. To continue this conversation we could open a discussion here on github or talk about it in the discord server.

piiq:
> > 
> 
> > 
> 
> > Since this started, can/should we do something to have the row fit a single line? Like rename the column names and maybe round the decimals?
> 
> 
> 
> I agree that it would be good to have everything in one line. There are many other commands that not always return one line. Therefore I think that it would be better to merge this pr and add a note/label with this issue in the discordbot project, since this is a new thing which has nothing to do with the due diligence commands specifically. To continue this conversation we could open a discussion here on github or talk about it in the discord server.

Totally agree. Small incremental changes are the way to go. 
Great job 🙏🙌👍

PR #940, opened 2021-11-12
Add short data by exchange

jmaslek:
As per Danglewoods request, I have scraped https://ftp.nyse.com/ShortData/ to get data by exchange.

Since there is a lot of data, I only scrape 2021 data and I am currently storing it on mongoldb using my personal free tier (the atlas connection is read only), so I added a note on _view that users should ping me on discord if there are issues (since I can imagine a couple that pop up).

The database will update nightly at some random time (these seem to get posted around midnight).  Also took the approach of having an interactive chart (with a matplotlibflag)  :)

Screen Shot 2021-11-12 at 9 48 35 AM


aia (approved):

DidierRLopes (approved):
Looks good to me :)

PR #942, opened 2021-11-12
Added Market Cap categorisation for the fds function

JerBouma:
I made a change in the Finance Database that categorises the Market Cap number to either Small Cap, Mid Cap or Large Cap. This is based on [this article](https://www.investopedia.com/investing/market-capitalization-defined/). Could have also added Nano Cap, Micro Cap and Mega Cap but I think for now this is fine and makes more sense.

image

### DIFF
--- a/gamestonk_terminal/stocks/discovery/disc_controller.py
+++ b/gamestonk_terminal/stocks/discovery/disc_controller.py
@@ -955,6 +955,15 @@ Finance Database:
         )
 
         parser.add_argument(
+            "-mc",
+            "--marketcap",
+            default=["Large Cap"],
+            nargs="+",
+            dest="marketcap",
+            help="Specify the Equities selection based on Market Cap",
+        )
+
+        parser.add_argument(
             "-ie",
             "--include_exchanges",
             action="store_false",
@@ -991,6 +1000,7 @@ Finance Database:
             industry=ns_parser.industry,
             name=ns_parser.name,
             description=ns_parser.description,
+            marketcap=ns_parser.marketcap,
             include_exchanges=ns_parser.include_exchanges,
             amount=ns_parser.amount,
             options=ns_parser.options,
--- a/gamestonk_terminal/stocks/discovery/disc_controller.py
+++ b/gamestonk_terminal/stocks/discovery/disc_controller.py
@@ -955,7 +955,7 @@ Finance Database:
         )
 
         parser.add_argument(
-            "-mc",
+            "-m",
             "--marketcap",
             default=["Large"],
             choices=["Small", "Mid", "Large"],
--- a/gamestonk_terminal/stocks/discovery/disc_controller.py
+++ b/gamestonk_terminal/stocks/discovery/disc_controller.py
@@ -957,9 +957,11 @@ Finance Database:
         parser.add_argument(
             "-mc",
             "--marketcap",
-            default=["Large Cap"],
+            default=["Large"],
+            choices=["Small", "Mid", "Large"],
             nargs="+",
             dest="marketcap",
+            type=str.title,
             help="Specify the Equities selection based on Market Cap",
         )
 
--- a/gamestonk_terminal/stocks/discovery/financedatabase_view.py
+++ b/gamestonk_terminal/stocks/discovery/financedatabase_view.py
@@ -1,5 +1,6 @@
 """Finance Database view"""
 __docformat__ = "numpy"
+# pylint:disable=too-many-arguments
 
 import financedatabase as fd
 import pandas as pd
@@ -13,6 +14,7 @@ def show_equities(
     industry: str,
     name: str,
     description: str,
+    marketcap: str,
     amount: int,
     include_exchanges: bool,
     options: str,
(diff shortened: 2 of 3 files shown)


gamestonk_terminal/stocks/discovery/disc_controller.py
```diff
@@ -954,6 +954,15 @@ def call_fds(self, other_args: List[str]):
...
 
+        parser.add_argument(
+            "-mc",
+            "--marketcap",
+            default=["Large Cap"],
+            nargs="+",
+            dest="marketcap",
+            help="Specify the Equities selection based on Market Cap",
+        )
+
```
JerBouma, line 965: If I added options, it couldn't handle `" ".join(marketcap).title()` anymore. If there is a workaround please let me know.

gamestonk_terminal/stocks/discovery/financedatabase_view.py
```diff
@@ -64,6 +66,14 @@ def show_equities(
...
     if description is not None:
         data = fd.search_products(data, query=" ".join(description), search="summary")
+    if marketcap is not None:
+        marketcap = " ".join(marketcap).title()
+        if marketcap not in ["Small Cap", "Mid Cap", "Large Cap"]:
+            raise ValueError(
+                f"Invalid choice ({marketcap}) for -mc/--marketcap. Choose from 'Small Cap', "
+                f"'Mid Cap' or 'Large Cap'."
+            )
+        data = fd.search_products(data, query=marketcap, search="market_cap")
```
JerBouma, line 76: Because `" ".join(marketcap).title()` doesn't work if you add options, I added a custom ValueError.

gamestonk_terminal/stocks/discovery/disc_controller.py
jmaslek: 2 possible workarounds to include choices:

1) Have your market cap be one word (LargeCap,etc..)
2) Input Large/Mid/Small then do `mc = {input} Cap`

jmaslek: Apologies if this is the third time commenting, but my comment doesn't show up now.

2 possible workarounds:

1) Market cap to be one work (ie LargeCap)
2) Input just Large/Mid/Small and do something like `mc = f"{input} Cap"`

jose-donato: We could remove "cap" from the options and keep just `small` `mid` and `large` since it is implied

gamestonk_terminal/stocks/discovery/financedatabase_view.py
```diff
@@ -64,6 +66,14 @@ def show_equities(
         data = fd.search_products(data, query=" ".join(name), search="long_name")
     if description is not None:
         data = fd.search_products(data, query=" ".join(description), search="summary")
+    if marketcap is not None:
+        marketcap = " ".join(marketcap).title()
+        if marketcap not in ["Small Cap", "Mid Cap", "Large Cap"]:
```
jose-donato, line 71: same here. we could reduce cap word from the options since it is implied

jose-donato (commented):
Looks good to me.  Tried locally also. 

Would just make that little update

gamestonk_terminal/stocks/discovery/disc_controller.py
jose-donato: I removed my comment by mistake. I think we could update from `large cap` to `large` since it is implied no?

JerBouma: Don't like the first one because readability. Going to implement the 2nd one, good suggestion. I noticed that the `choices` arguments doesn't make the function as robust, I need to add the option `small` and `Small` unfortunately. By any chance also already encountered that?

JerBouma: Found the solution myself, will update now.

jmaslek: Nice.  Think you could also set it to str.lower and compare the lower case of each option.

JerBouma: True, but in the end in the Finance Database is written 'Large Cap'. You will always require 'large' to be set to 'Large'. Might as well do it right away (otherwise in `financedatabase_view.py` you will still need to add `.title()`)

jmaslek (approved):

gamestonk_terminal/stocks/discovery/disc_controller.py
```diff
@@ -954,6 +954,17 @@ def call_fds(self, other_args: List[str]):
             help="Specify the Equities selection based on the description (not shown in table)",
         )
 
+        parser.add_argument(
+            "-mc",
+            "--marketcap",
```
jmaslek, line 959: Also just caught this.  The convention we use is that single dashes gets a single letter, so it would be "-m"

JerBouma: Changed it!

PR #944, opened 2021-11-15
Fraud error

colin99d:
It looks like the Alpha Vantage API is returning a lot of Nones, where there should not be. This is a small patch to tell the user that data cannot currently be returned if we get Nones that we should not. Not sure why Alpha Vantage is having this issue.

### DIFF
--- a/gamestonk_terminal/stocks/fundamental_analysis/av_model.py
+++ b/gamestonk_terminal/stocks/fundamental_analysis/av_model.py
@@ -325,9 +325,6 @@ def get_fraud_ratios(ticker: str) -> Tuple[Dict[str, float], float]:
     df_cf = df_cf.set_index("fiscalDateEnding").iloc[:2]
     df_bs = df_bs.set_index("fiscalDateEnding").iloc[:2]
     df_is = df_is.set_index("fiscalDateEnding").iloc[:2]
-    pd.set_option("display.max_columns", None)
-    print(df_bs)
-    print(df_bs.columns)
 
     ar = df_bs["currentNetReceivables"].apply(lambda x: 0 if x else int(x)).values
     sales = df_is["totalRevenue"].apply(lambda x: 0 if x else int(x)).values
--- a/gamestonk_terminal/stocks/fundamental_analysis/av_model.py
+++ b/gamestonk_terminal/stocks/fundamental_analysis/av_model.py
@@ -325,23 +325,46 @@ def get_fraud_ratios(ticker: str) -> Tuple[Dict[str, float], float]:
     df_cf = df_cf.set_index("fiscalDateEnding").iloc[:2]
     df_bs = df_bs.set_index("fiscalDateEnding").iloc[:2]
     df_is = df_is.set_index("fiscalDateEnding").iloc[:2]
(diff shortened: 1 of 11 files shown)


jmaslek:
Whatever you changed is now failing a test :P

gamestonk_terminal/stocks/fundamental_analysis/av_model.py
```diff
@@ -325,23 +325,46 @@ def get_fraud_ratios(ticker: str) -> Tuple[Dict[str, float], float]:
...
-    ni = df_is["netIncome"].apply(lambda x: int(x)).values
-    ca = df_bs["totalCurrentAssets"].apply(lambda x: int(x)).values
-    cl = df_bs["totalCurrentLiabilities"].apply(lambda x: int(x)).values
-    ppe = df_bs["propertyPlantEquipment"].apply(lambda x: int(x)).values
-    cash = df_bs["cashAndCashEquivalentsAtCarryingValue"].apply(lambda x: int(x)).values
-    cash_and_sec = df_bs["cashAndShortTermInvestments"].apply(lambda x: int(x)).values
+    pd.set_option("display.max_columns", None)
+    print(df_bs)
+    print(df_bs.columns)
+
```
jmaslek, line 331: I assume this is for debugging?

gamestonk_terminal/stocks/fundamental_analysis/av_model.py
```diff
@@ -325,23 +325,46 @@ def get_fraud_ratios(ticker: str) -> Tuple[Dict[str, float], float]:
...
+        df_is["costofGoodsAndServicesSold"].apply(lambda x: 0 if x else int(x)).values
+    )
+    ni = df_is["netIncome"].apply(lambda x: 0 if x else int(x)).values
+    ca = df_bs["totalCurrentAssets"].apply(lambda x: 0 if x else int(x)).values
+    cl = df_bs["totalCurrentLiabilities"].apply(lambda x: 0 if x else int(x)).values
+    ppe = df_bs["propertyPlantEquipment"].apply(lambda x: 0 if x else int(x)).values
+    cash = (
+        df_bs["cashAndCashEquivalentsAtCarryingValue"]
+        .apply(lambda x: 0 if x else int(x))
+        .values
```
jmaslek, line 344: Is there a way to print out which variable causes the error?  All you show is `Data incomplete for this ticker. Unable to calculate risk`

colin99d: I am hoping this is a temporary issue with Alpha Vantage which is why I didn't go too much into detail. I could add more but not sure how useful it is for the end user.

jmaslek: I was thinking this would be more useful to you to see where the issues comes from lol

colin99d: It seems to vary based on the stock, which is why I believe it's an API issue. Unfortunately not much I can do with it since it just ruins the whole calculation.

jmaslek (approved):

PR #945, opened 2021-11-15
Add search function to the stocks menu

JerBouma:
I see value in being able to search a company ticker within the terminal. This helps keeping me (and other users) inside the terminal instead of needing to Google the company ticker first (and perhaps already finding the required info that way). This makes for a more fluent experience.

This functionality has a lot of similarities to the **stocks > disc > fds** function but serves a completely different purpose. This is merely used to find a company ticker and nothing else. Therefore, I find it valuable to include it this way.

image


jmaslek:
FEATURE # 500 🏆

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -189,6 +192,43 @@ def call_reset(self, _):
...
+            type=int,
+            dest="amount",
+            help="Enter the number of Equities you wish to see in the Tabulate window.",
+        )
+
+        if other_args and "-q" not in other_args and "-h" not in other_args:
+            other_args.insert(0, "-q")
+
+        ns_parser = parse_known_args_and_warn(parser, other_args)
+
```
jmaslek, line 228: Should include `if not ns_parser: return`

That way doing -h doesn't show the error that passes through

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -189,6 +192,43 @@ def call_reset(self, _):
...
+    def call_search(self, other_args: List[str]):
+        """Process search command"""
+        parser = argparse.ArgumentParser(
+            add_help=False,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+            prog="search",
+            description="Show companies matching the search query.",
+        )
+
+        parser.add_argument(
```
jmaslek, line 205: Also styling point moving forward (can blame @DidierRLopes ) : no newlines between parser arguments.

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -83,6 +86,60 @@ def clear(
...
+
+    Returns
+    -------
+    tabulate
+        Companies that match the query.
+    """
+    equities_list = (
+        "https://raw.githubusercontent.com/JerBouma/FinanceDatabase/master/"
+        "Database/Equities/Equities List.json"
+    )
```
jmaslek, line 110: Not sure if this is common practice or not, but having a space in the url can get tricky at some points (same with naming files - spaces can get confusing).  If you had this as EquitiesList.json,  you could skip the dictionary part and just load using `pd.read_csv(url)`.  But since there is a space there, pandas yells at us.

gamestonk_terminal/stocks/stocks_controller.py
JerBouma: Done!

gamestonk_terminal/stocks/stocks_controller.py
JerBouma: And Done!

gamestonk_terminal/stocks/stocks_helper.py
JerBouma: Hmm, I personally like readability and with the current method I haven't  had any issues whatsoever. The same code I have right now is also applied within the fundamentanalysis and financedatabase packages.

jmaslek: So I am not gonna make you change your personal code, and what you have does work (although you could do it without the intermediate dictionary using `pd.DataFrame.from_dict(request.json(),orient="index").reset_index(drop=False)`).

Heres a link about the naming conventions and why spaces can get tricky: https://webtech.training.oregonstate.edu/faq/what-are-good-file-naming-conventions

JerBouma: While I do agree with you that over time it might be valuable to remove the spaces, I don't think it is needed right now. This is because it is in line with the rest of the Finance Database code (and that matches the actual industry/sector names as also shown on Yahoo Finance). See for example here: https://github.com/JerBouma/FinanceDatabase/tree/master/Database/Equities/Industries. In time this could be changed to include underscores instead of spaces.

For the quicker way to set up the code, `pd.DataFrame.from_dict(request.json(),orient="index").reset_index(drop=False))`, here I again advocate readability. I'd  rather have an intermediate step and make it clear what is going on then have both the requesting of a JSON file + creating of a DataFrame on one line.

I think this is a style preference and perhaps the team needs to decide on what is best practice.

jmaslek: > For the quicker way to set up the code, here I again advocate readability. I'd rather have an intermediate step and make it clear what is going on then have both the requesting of a JSON file + creating of a DataFrame on one line.
> 
> I think this is a style preference and perhaps the team needs to decide on what is best practice.

Absolutely. My preference for the one liner is due to not needing to create an unused dictionary.  Granted the dictionary only takes up .5 Mb of memory, so its not slowing down much (I hope), but if we have those from multiple functions laying around and different imports that can add up and we end up back to where loading a menu takes up to 10 seconds.

jmaslek (approved):

Issue #952, opened 2021-11-16
[Bug] Raw Data from stocks->gov->topsells is Inaccurate


grahuloverthere:
**Describe the bug**
When running "topsells --raw" from the stocks/gov menu, the terminal output does not match the plot displayed. It looks like "topsells --raw" output is the same as "topbuys --raw".

**To Reproduce**
> python terminal.py
> stocks
> gov
> topsells --raw

**Screenshots**
<img width="1091" alt="Screen Shot 2021-11-16 at 09 01 16" src="https://user-images.githubusercontent.com/75821256/142000034-0f5af8d6-b115-42dd-873b-e14a8fe3c841.png">

**Desktop (please complete the following information):**
- OS: Mac Monterey 12.0.1 with Intel i7
- Python version: 3.8.8

**Additional context**
Add any other information that you think could be useful for us.

jmaslek:
Another good catch!  I was not using the upper bounds of the reported data and was sorting in reverse order (hence why it equated to the top buys)

piiq:
works now
![image](https://user-images.githubusercontent.com/11668535/142022025-9c6f6adb-8f61-4b04-ae4f-d7368f5ef901.png)

piiq:
Thanks, this has been addressed in https://github.com/GamestonkTerminal/GamestonkTerminal/pull/953

PR #953, opened 2021-11-16
bug where `topsells` was not plotting same as raw data

jmaslek:
This addresses #952 

The issue was that the plotted data was using upper bound and raw data was using lower bound.  We title our plot with upper bound, so I switched both to do that and added the abs() for sorting.


piiq:
sorry, wrong chat
image

piiq (approved):

Issue #939, opened 2021-11-12
[Bug] (stocks)>(fa)>key: Passing list-likes to .loc or [] with any missing labels is no longer supported. 


deeleeramone:
![Screen Shot 2021-11-11 at 6 35 45 PM](https://user-images.githubusercontent.com/85772166/141399406-3ddbc71c-eea4-4505-a99b-52702fe8ef7a.png)

(✨) (stocks)>(fa)> key
"Passing list-likes to .loc or [] with any missing labels is no longer supported. The following labels were missing: Index(['Payout ratio'], dtype='object'). See https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#deprecate-loc-reindex-listlike"

jmaslek:
Closing via #954

PR #954, opened 2021-11-16
Remove payout ratio from key

jmaslek:
Addressed #939.

Not sure why, but "Payout ratio" was not in the list of what was pulled down from AV.


piiq:
image

piiq (approved):

PR #961, opened 2021-11-19
Feature/price

CameronBeebe:
I am new to the GST, but very cool.  

Very simple feature: a price only option for quote in stocks.  

Example: "quote -t GME -p"
Output: "Price of GME 210.12"

One line, doesn't show whole df table.

Two files changed: 
gamestonk_terminal/stocks/stocks_helper.py
website/content/stocks/quote/_index.md

It could be this can be achieved some other ways I am unaware of.  Generally, I have the intuition (or wish) that I should be able to access the loaded df from the command line... that is, when I do "load GME" I should then be able to "GME.stuff" or "GME -stuff".  Perhaps I am just too new to GST and have yet to figure it out.  Gotta start somewhere!


DidierRLopes:
Hey @CameronBeebe, great to see some new blood over here eheh

So, I was wondering, given that with this feature you just want to be able to access the last price.

What about having this being displayed in load by default? E.g. currently we have:
```
(✨) (stocks)> load pltr
Loading Daily PLTR stock with starting period 2020-11-18 for analysis.
```

We could have:
```
(✨) (stocks)> load pltr
Loading Daily PLTR stock with starting period 2020-11-18 for analysis.
2021-11-19 17:00:00 - Last price: 33.33
```

This would be straightforward to implement given that when we load we actually obtain the data frame of the data from yahoo finance, iex cloud or alpha vantage.

CameronBeebe:
@DidierRLopes Or both?  I can also call quote without load first, in that case the load print will not show, and also it saves compute if I don't want to load historical data right?

DidierRLopes:
> @DidierRLopes Or both? I can also call quote without load first, in that case the load print will not show, and also it saves compute if I don't want to load historical data right?

In theory yes, in practice not so much. If you try it you'll see that our `load` function is actually much faster than the `quote` one.

So if we want to just be able to see latest price of a stock, I think having it on `load` is where it makes sense. Also because if you like what you see then you are ready to do your investigation on it.

CameronBeebe:
@DidierRLopes Interesting.  So loading historical data is quicker than a price quote?

I feel like perhaps I am an atypical user compared to how others are using GST.  I would like it to have at least the basic function to replace looking up prices on Google Finance.  

Sometimes I don't care to see the historical data or do analysis, I just want to check a bunch of prices in the terminal.  (I am not a day trader.)  In that case, a clean and simple one-line output in the terminal would be nice.

So, to me implementing both load and quote options seems ideal.  (It won't hurt my feelings if you tell me why I'm wrong... like I said I am new to GST.)

jmaslek:
> @DidierRLopes Interesting. So loading historical data is quicker than a price quote?
> 
> I feel like perhaps I am an atypical user compared to how others are using GST. I would like it to have at least the basic function to replace looking up prices on Google Finance.
> 
> Sometimes I don't care to see the historical data or do analysis, I just want to check a bunch of prices in the terminal. (I am not a day trader.) In that case, a clean and simple one-line output in the terminal would be nice.
> 
> So, to me implementing both load and quote options seems ideal. (It won't hurt my feelings if you tell me why I'm wrong... like I said I am new to GST.)

So the quote is actually ~30x slower:

Screen Shot 2021-11-23 at 12 14 20 PM

I would agree that having both options would be good.  Something else is that you could do multiple tickers to get a snapshot. So that if we do `quote AAPL,MSFT`, one would see both. 

Looks like you need to clean up the linter first.  you can run `black ./` from the main directory to pass that one, but you can also look into the pre commit hooks that will make your life easier :)

CameronBeebe:
@jmaslek Hmm.

So it looks like it should be straightforward to print out the last closing price from the loaded df, i.e. the last entry, but that would be yesterday's news.  And of course you want the current price to print out regardless of the time interval chosen to create the df.  

It might be a bit more of a change to grab the current price in load (code which, if I am not mistaken, is basically already in the quote function grabbing from yf.Ticker ).  The returned df entries are formatted based on complete day info (closing price, volume etc.), and so the df looks like it is created only up to yesterday.

This linting thing is new to me, so might take me a bit to figure out where exactly it is failing.  I thought it was automatically running before every commit, and it wasn't spitting anything out so I thought it was good to go.

jmaslek:
> @jmaslek Hmm.
> 
> So it looks like it should be straightforward to print out the last closing price from the loaded df, i.e. the last entry, but that would be yesterday's news. And of course you want the current price to print out regardless of the time interval chosen to create the df.

So the last entry in load corresponds to "now" (maybe after hours is confusing).  Note how the last price changes as I call it a couple times:

Screen Shot 2021-11-23 at 3 37 11 PM

The only case in which using `df.tail(1)["Close"]`  would not give you correct last price is if you specify an end date.  

> It might be a bit more of a change to grab the current price in load (code which, if I am not mistaken, is basically already in the quote function grabbing from yf.Ticker ). The returned df entries are formatted based on complete day info (closing price, volume etc.), and so the df looks like it is created only up to yesterday.

Not following this.  `load` is just using yf.download whereas the quote uses yf.Ticker.

> This linting thing is new to me, so might take me a bit to figure out where exactly it is failing. I thought it was automatically running before every commit, and it wasn't spitting anything out so I thought it was good to go.

Check out the pre-commit instructions (ie `pip install pre-commit` , `pre-commit install`).  They will run whenever you commit.  The way the repo is set up, one of us has to trigger the workflow on your first run, so the CI was not running on each commit.
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/CONTRIBUTING.md#pre-commit-hooks

CameronBeebe:
@jmaslek Thanks for your help.  Yeah, I followed the pre-commit instructions correctly I think.  `black ./` seems to reformat the file, hopefully that will work.

Maybe I am missing something, but `df_stock_candidate.tail(1)['Close']` can only grab yesterday's data: `2021-11-22    247.550003` (for GME from `load GME`).  If you print the whole df out the last entry is yesterday's (not sure about after hours / when current day is available).  

`df_stock_candidate` is the df we want, right?  If I understand all of the conditional logic in `load`, this df comes from whatever chosen data backend.  Perhaps not YF.

jmaslek:
So I need to double check the sources, but at least for yahoo, load should grab the latest date (if I call load at noon, then it will update with that days data).  I could be mistaken.

As for the relevant data frame, yes df_stock_candidate is the right thing, but the print should probably be on the controller after load and use the self.stock, which is the loaded price df.

CameronBeebe:
@jmaslek Just took a closer look, the df are truncated by end dates.  I think it also affects `candle` plotting.  

Without end date:
Screen Shot 2021-11-30 at 11 04 06 AM

With end date:
Screen Shot 2021-11-30 at 11 03 43 AM

Making changes to data source specific data frames so that `df_stock_candidate.tail(1)['Close']` spits out current market price would require changing df slicing throughout all of the conditional logic, likely breaking intended `load` functionality (or requiring a lot of extra code so it doesn't break).  At least thats what it looks like to me at the moment.

Alternatively, we just grab current market price (from e.g. `yf`) regardless of data backend... which is pretty much just what `quote` does.  Which makes me think that perhaps `quote GME -p` is sufficient for now.  Let me know what you think.

jmaslek:
Sorry em haven't looked further into this.  I am good with this solution for now.  Once you fix up the linter I will merge in

CameronBeebe:
The tests finally look good!  I guess I have an issue with my pre-commit install in conda env, perhaps not using the local config... didn't think it would pass.

jmaslek (approved):

PR #963, opened 2021-11-19
delete hugo content for deleted `screener` function

jmaslek:


jmaslek:
small change.  admin merging :)

PR #964, opened 2021-11-19
Fix export path for candle function

jmaslek:
The export function for candle was pointing to quantitative analysis export folder.  Fixed that to save to a stocks/raw_data folder.


piiq (approved):

piiq:
image
image

PR #966, opened 2021-11-20
Update fa to account for suffix,  add load

jmaslek:
Another bug/improvement.  This PR addresses 2 things

1.  `load` was listed in the menu but not implemented.  Now implemented.
2. yahoo finance allows looking at fa stuff using the suffix convention (ie glxy.to).  I now pass through the suffix so that just the yahoo finance can be used.  As shown in screenshot, I add a disclaimer and dim the unusable functions.
Screen Shot 2021-11-20 at 4 24 36 PM

### DIFF
--- a/gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
+++ b/gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
@@ -4,6 +4,8 @@ __docformat__ = "numpy"
 import argparse
 from typing import List
 from prompt_toolkit.completion import NestedCompleter
+from colorama import Style
+import pandas as pd
 
 from gamestonk_terminal.stocks.fundamental_analysis.financial_modeling_prep import (
     fmp_controller,
@@ -27,6 +29,7 @@ from gamestonk_terminal.helper_funcs import (
     try_except,
     system_clear,
 )
+from gamestonk_terminal.stocks.stocks_helper import load
 from gamestonk_terminal.menu import session
 
 # pylint: disable=inconsistent-return-statements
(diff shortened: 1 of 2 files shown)


piiq:
Nice one @jmaslek. 

I have noticed one important thing.

When loading data from different exchanges prices come in local currency. Here's and example of YNDX traded in NY and Moscow
image

This leads to discrepancies in representation of the some data like `shrs`. This is the last row of the `Mutual Fund Holders` for .ME suffix

```
│ iShares Core MSCI Emerging Markets ETF           │ 2.685 M  │ 2021-09-30 00:00:00 │ 0.84 %  │ 15.608 B  │
```

and here's for nasdaq

```
│ iShares Core MSCI Emerging Markets ETF           │ 2.685 M  │ 2021-09-29 00:00:00 │ 0.84 %  │ 213.987 M │
```

piiq (approved):
I'll approve this one as the PR scope is about handling the suffixes and the load function in FA menu, but we really need to look into the local currencies issue

jmaslek:
> I'll approve this one as the PR scope is about handling the suffixes and the load function in FA menu, but we really need to look into the local currencies issue

Indeed this pulls different currencies from yahoo.  Let me do a quick dig if anything in the Ticker object could indicate which currency (or if there is just a straight USD conversion hidden somewhere)

Screen Shot 2021-11-22 at 12 01 06 PM
)

jmaslek:
So @piiq.  At a quick glance, the yf Ticker object has the following: `'financialCurrency': 'RUB'`.

So this can be solved in one of two ways (IMO)

1. Just change the reporting from dollars to whatever that Currency is.
2. Do a quick forex call (I assume I can figure out how to do on yahoo).

Either way, should we merge then I can look into it further in a dedicated PR?

piiq:
@jmaslek yep, let's merge

Should we have both options? Like 1 by default and 2 as a separate command argument?

PR #971, opened 2021-11-23
General UI improvements with sorting/filtering on some functions

### LINKED ISSUE #818: [IMPROVE] Unusual Options Activity to include "and" with filter capabilities; also the ability to sort by expiry date and ticker..
Unusual Options Activity to include "and" with filter/sort capabilities; also the ability to sort by expiry date and ticker.

jmaslek:
General user experience upgrades based on discord chatwith some filtering/sorting abilities for the following functions

- options>unu (will close #818 )
- disc - arkord
- economy-performance

Also threw in a quick fix for ca where the user selects the same symbol.  Uses a set to only take unique inputs.


aia (approved):

piiq (approved):
Big one

PR #982, opened 2021-11-24
Add difflib to controllers

jmaslek:
Idk about all y'all but I type the wrong stuff all the time (when prompt is off that is).

I went through the controllers and added a suggestion just like the discord bot.

Screen Shot 2021-11-24 at 5 00 49 PM

### DIFF
--- a/gamestonk_terminal/cryptocurrency/crypto_controller.py
+++ b/gamestonk_terminal/cryptocurrency/crypto_controller.py
@@ -3,6 +3,7 @@ __docformat__ = "numpy"
 # pylint: disable=R0904, C0302, R1710, W0622
 
 import argparse
+import difflib
 import matplotlib.pyplot as plt
 import pandas as pd
 from colorama import Style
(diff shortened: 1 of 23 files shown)


DidierRLopes:
Can we do something like:
- if cutoff > 0.85 or so we just use that command directly
- if cutoff > 0.7 we say "did you mean X?"
- otherwise, we do nothing

jmaslek:
> Can we do something like:
> 
> - if cutoff > 0.85 or so we just use that command directly
> 
> - if cutoff > 0.7 we say "did you mean X?"
> 
> - otherwise, we do nothing

So it seemed that .7 was doing a good job all around.  Should I bump it up a little - maybe .75 then print what the terminal will do - add like a 1 second sleep then just run the command?

DidierRLopes:
> > Can we do something like:
> > 
> > * if cutoff > 0.85 or so we just use that command directly
> > * if cutoff > 0.7 we say "did you mean X?"
> > * otherwise, we do nothing
> 
> So it seemed that .7 was doing a good job all around. Should I bump it up a little - maybe .75 then print what the terminal will do - add like a 1 second sleep then just run the command?

Yup, seems like a good idea to me :)

aia (approved):

piiq (approved):

DidierRLopes:
@piiq you should have waited for the changes here. There was still some pendent work and this introduced bug #987.

Next time I'll add the "request changes" flag, I think is better practice.

PR #985, opened 2021-11-25
Enable full mypy

aia:


DidierRLopes (approved):

piiq:
@aia I have a question.

I've deleted the `type : ignore` in the *calculator_model.py* to test how the error would be cought.
When I tried to commit it. I was expecting that mypy precommit hook to stop the commit, but it didn't.

Although if I run the command that is being run in the action the error _is_ being raised.

The commands that I use to check are:
- `.git/hooks/pre-commit` to run pre-commit hooks
- `mypy --ignore-missing-imports --exclude '/setup\.py$' ` for the mypy check as in the GH action

Should the hook be modified somehow? It's not obvious from the docs of the hook repo

aia:
> Should the hook be modified somehow? It's not obvious from the docs of the hook repo

Most likely. I don't use pre-commit hooks myself because I have VSCode configured to run all linters for me on every file save and the errors/warnings are displayed very clearly.

PR #988, opened 2021-11-26
fix #987

### LINKED ISSUE #987: [Bug]
**Describe the bug**
Pressing ?/q to go back to previous menu results  in error and force quit.

**To Reproduce**
Pressing ?/q to go back to previous menu .

- OS: [Windows 11]
- Python version [e.g. 3.9.0]

**Additional context**

Maybe using the command "back" for only go back to previous step would be better then calling  main menu with ?/q.

DidierRLopes:
This should fix #987


piiq (approved):
image
image

PR #990, opened 2021-11-26
Hugo Server Update - November 26, 2021

deeleeramone:


.github/workflows/labels.yml
```diff
@@ -1,9 +1,9 @@
 name: Pull Request Labels
 "on":
-  pull_request:
```
DidierRLopes, line 3: This shouldn't have been changed

.github/workflows/labels.yml
```diff
@@ -1,9 +1,9 @@
...
+  pull_request: unlabeled
     types: [opened, reopened, labeled, unlabeled, synchronize]
 jobs:
-  label:
```
DidierRLopes, line 6: Neither this

DidierRLopes (changes requested):

deeleeramone:
Ya.. my bad. I thought I was directed to the file with the flag that was missing.

DidierRLopes:
@deeleeramone 

Can you change it back and commit? So we can merge this baby?

deeleeramone:
Hoping I did that right! 😆

DidierRLopes:
@deeleeramone yes! It looks good to me!

DidierRLopes (approved):
Btw no need to add the date on the title, it's better if you actually mention what sort of menus did you update so we can trace it back

PR #995, opened 2021-11-29
Sector and industry analysis

DidierRLopes:
S/O to @JerBouma and to his https://github.com/JerBouma/FinanceDatabase package and a previous notebook he has sent with some tips on how we could incorporate this on the terminal.

image

image

image

image


DidierRLopes:
@JerBouma can review this menu and your user experience with it?

Also, can you let me know what financials do you think would be useful to have by default. I have:
```
Financials
    roa           return on assets
    roe           return on equity
    qr            quick ratio
    rg            revenue growth
    rec           recommendation mean
    td            total debt
    ebitda        earnings before interest, taxes, depreciation and amortization
```

but I didn't think much about it to be honest, just picked some I found interesting. 

Let me know the ones you would find important from this list: 'ebitdaMargins', 'profitMargins', 'grossMargins', 'operatingCashflow', 'revenueGrowth', 'operatingMargins', 'ebitda', 'targetLowPrice', 'recommendationKey', 'grossProfits', 'freeCashflow', 'targetMedianPrice', 'currentPrice', 'earningsGrowth', 'currentRatio', 'returnOnAssets', 'numberOfAnalystOpinions', 'targetMeanPrice', 'debtToEquity', 'returnOnEquity', 'targetHighPrice', 'totalCash', 'totalDebt', 'totalRevenue', 'totalCashPerShare', 'financialCurrency', 'maxAge', 'revenuePerShare', 'quickRatio', 'recommendationMean'

DidierRLopes:
The only thing that should be missing here is the Hugo server update. I'll do that after we agree on the features (particularly financials).

Also, I've not incorporated the economy/valuation,performance,spectrum from finviz commands because @jmaslek mentioned that these provide an overall overview of the economy. And although I was inclined to move them at the start, upon further reflecting, I do think they add more value in economy than in this menu. Let me know if you disagree.

JerBouma:
@DidierRLopes 

First of all, it already looks pretty great! However, some pointers that you should consider:

**Data Plotting**
First of all, I would like a limitation on the amount of tickers you can graph. It doesn't make too much sense to select a sector (for example 'Real Estate') and then plot all Quick Ratios of this sector. Simply because you get wildly different companies in wildly different countries. Next to that, it takes a long time to load. The figure below already has United States + Large Cap restriction.

Figure_1

Some suggestions:
- Do not allow plotting of one whole sector (or perhaps only if you enter a certain command). In essence, you want the combination Sector + Industry pretty much always.
- Set default parameters, for example Large Cap and perhaps United States as country.

**Financials**
Per sector/industry it depends a lot what metrics you want to look at. For example inventory is not so important for Facebook but is very important for Airlines (high inventory usually means they sit on a lot of unused fuel which can expire and costs a lot of money). Therefore, I'd recommend to add in as much metrics as possible. It is up to the user to understand which ones makes sense.

Further, my main improvements would be:

1. Add ratios _over time_. It doesn't say much if your QR is 1 today if it was 10 last year. You need to be able to capture the trend over time to understand inventory management for (in this case) the airline industry. This is a difficult one and probably requires some premium data but it is nonetheless very important.
2. Maybe for the general ones I am missing at least the P/E ratio and EPS. The first comparison between companies is often done by their P/E ratio. That helps in valuing the companies properly.
3. It can be beneficial to add an industry benchmark (for example the median of all companies in the graph) so you understand which ones are performing above and which ones are performing below the benchmark.

DidierRLopes:
> @DidierRLopes
> 
> First of all, it already looks pretty great! However, some pointers that you should consider:
> 
> **Data Plotting** First of all, I would like a limitation on the amount of tickers you can graph. It doesn't make too much sense to select a sector (for example 'Real Estate') and then plot all Quick Ratios of this sector. Simply because you get wildly different companies in wildly different countries. Next to that, it takes a long time to load. The figure below already has United States + Large Cap restriction.
> 
> Figure_1
> 
> Some suggestions:
> 
> * Do not allow plotting of one whole sector (or perhaps only if you enter a certain command). In essence, you want the combination Sector + Industry pretty much always.
> * Set default parameters, for example Large Cap and perhaps United States as country.
> 
> **Financials** Per sector/industry it depends a lot what metrics you want to look at. For example inventory is not so important for Facebook but is very important for Airlines (high inventory usually means they sit on a lot of unused fuel which can expire and costs a lot of money). Therefore, I'd recommend to add in as much metrics as possible. It is up to the user to understand which ones makes sense.
> 
> Further, my main improvements would be:
> 
> 1. Add ratios _over time_. It doesn't say much if your QR is 1 today if it was 10 last year. You need to be able to capture the trend over time to understand inventory management for (in this case) the airline industry. This is a difficult one and probably requires some premium data but it is nonetheless very important.
> 2. Maybe for the general ones I am missing at least the P/E ratio and EPS. The first comparison between companies is often done by their P/E ratio. That helps in valuing the companies properly.
> 3. It can be beneficial to add an industry benchmark (for example the median of all companies in the graph) so you understand which ones are performing above and which ones are performing below the benchmark.

Regarding your first section, I did 2 things:
- Set a limiting parameter. Before I limit the Barnhart display, I order the dictionary from bigger metric value to smaller. This allows to show companies with those "strong"metrics, and if you want to see a lot of companies you can still do so increasing the limit parameter. I set by default to 10, does that sound reasonable?
- I set Large cap and US as default.

JerBouma:
@DidierRLopes 

Could you use `tight_layout` for the plot? I notice that everytime I run a graph I don't see the full names even in full screen. Furthermore, it might be interesting to not surpress the TQDM from yfinance so users can understand why it takes some time. I also notice it is currently printing out the values. Think that is a good thing but perhaps put it in a table.

Figure_1

DidierRLopes:
@JerBouma it should all be tackled now I believe.

Less the ratios over time, but we can have that as a TODO for next

jmaslek:
Nice.  1 complaint and 1 suggestion.

Complaint: pretty slow.  Probably not much to do about that.
Suggestion:  maybe a way to link tickers and go straight to >ca ?

DidierRLopes:
> Nice. 1 complaint and 1 suggestion.
> 
> Complaint: pretty slow. Probably not much to do about that. Suggestion: maybe a way to link tickers and go straight to >ca ?

Ups, forgot to do that. Was meant to do it. Will do in the morning when I wake up

JerBouma:
I would still suggest to show a progress bar so users understand why it is taking some time because it indeed can be very slow.

Furthermore, to speed up things I would save the last selection of data. So if I pick Airplanes in the US (my default example) and choose QR it would collect data (and thus be a bit slow) but if I then choose RoA it would re-use the collected data. It should clear the data when you switch menus to not fill up the memory too much.

piiq (changes requested):
This is a very important menu that I've wanted for some time now. Thank you very much! I am already having lots of fun poking into it and getting nice overview of the economy through those industry lenses.

A couple of comments on from my side:

1. `roa` hangs up if no industry is selected or is it just taking THAT long? If it's just long, as @JerBouma pointed out either a progressbar is needed or we should set a limit to some of the commands
image
*update while writing: I've tested a progressbar and it is just very slow*
image

1. In the `cpi` and `cps` charts I see that there are datapoints with extremely low values that lead to charts like this
image

1. the `cpi` and `cps` would require filtering of empty results
image

JerBouma:
@piiq Your implementation could be shortened. This should be enough:
image
image


@DidierRLopes 
Details:
- I noticed the message "_Limiting the amount of companies displayed to 10._" is always received. That shouldn't be the case if I select less than 10 companies to begin with.
- The benchmark should be a percentage for most of the ratios and could you include a table whenever I select a command like 'roa'. It helps seeing numbers as well.

gamestonk_terminal/stocks/sector_industry_analysis/financedatabase_view.py
```diff
@@ -0,0 +1,280 @@
...
+            else:
+                company_metric_to_do_median.append(metric[1][0])
+
+        company_metric_to_do_median += company_metric
+
+        if company_metric_to_do_median:
+            print(f"Limiting the amount of companies displayed to {limit}.")
+
+        for n, m, t in zip(
+            company_name[::-1], company_metric[::-1], company_ticker[::-1]
```
JerBouma, line 122: Perhaps this is personal preference but I dislike reading code that has a bunch of 0's and 1's in it like you are programming here. It makes it difficult to understand what is going on without debugging myself and thus reduces readability. It helps if these 1's and 0's could be changed to actual names of what you are selecting (and thus use dictionaries instead of lists for example).

gamestonk_terminal/stocks/sector_industry_analysis/financedatabase_view.py
```diff
@@ -0,0 +1,280 @@
...
+    )
+
+    # are there more industries than colors
+    if len(companies_per_industry) > len(colors):
+        companies_per_industry_sliced = dict(
+            list(companies_per_industry.items())[: len(colors) - 2]
+        )
+        companies_per_industry_sliced["Others"] = sum(
+            dict(list(companies_per_industry.items())[len(colors) - 2 :]).values()
+        )
```
JerBouma, line 253: And it helps to not perform too many actions in one line of code. You are here making a sub selection of the length of colors (minus 2) of a dictionary convert into a list converted into a dictionary. It also makes me wonder if it wouldn't be tidier to use a DataFrame in this case, then you still have the labels and can select like you are using lists.

All in all, I think I am looking for some sort of coding style standard (other than PEP8) for the project.

gamestonk_terminal/stocks/sector_industry_analysis/financedatabase_view.py
```diff
@@ -0,0 +1,280 @@
...
+    companies_per_industry = dict(
+        OrderedDict(
+            sorted(companies_per_industry.items(), key=lambda t: t[1], reverse=True)
+        )
+    )
+
+    # are there more industries than colors
+    if len(companies_per_industry) > len(colors):
+        companies_per_industry_sliced = dict(
+            list(companies_per_industry.items())[: len(colors) - 2]
```
JerBouma, line 249: And it helps to not perform too many actions in one line of code. You are here making a sub selection of the length of colors (minus 2) of a dictionary converted into a list converted into a dictionary. It also makes me wonder if it wouldn't be tidier to use a Series in this case, then you still have the labels and can select like you are using lists.

All in all, I think I am looking for some sort of coding style standard (other than PEP8) for the project.

DidierRLopes:
> I would still suggest to show a progress bar so users understand why it is taking some time because it indeed can be very slow.
> 
> Furthermore, to speed up things I would save the last selection of data. So if I pick Airplanes in the US (my default example) and choose QR it would collect data (and thus be a bit slow) but if I then choose RoA it would re-use the collected data.
> 
> It should then clear the data when you switch menus to not fill up the memory too much.

This is a great idea, will work on this.

Yup, I'll work on progress bar as well.

gamestonk_terminal/stocks/sector_industry_analysis/financedatabase_view.py
DidierRLopes: In this case the [::-1] just reverts the order of the list to be the opposite, I had to do this due to the implementation of plt.barh and me wanting to have the bigger elements on top.

This is a standard reverting operation on a list, hence why I didn't think it was worth creating a new variable just for this as it is only used here.

gamestonk_terminal/stocks/sector_industry_analysis/financedatabase_view.py
DidierRLopes: The reason why I went for this design is because the only thing I needed was to slice a dictionary based on the value. This is why I convert the items to list, slice and reconvert back to dictionary. The variable didn't change, only got sliced, this is also why the name of the variable is `companies_per_industry_sliced`. Would it help having:
`companies_per_industry_sliced: Dict = ...`?

The use of colours for the slicing could actually be better, because I rely on the size of `mcolors.CSS4_COLORS` which may not be ideal in case it changes in the future. Will improve this.

Yup, PEP8 is the coding style standard we have been using. Apart from that the code has the same "convention" through black linters and through code reviews.

DidierRLopes:
Will share some screenshots. Only thing missing is: taking tickers to comparison analysis and having a better display of roa/roe/debit ratio ....

DidierRLopes:
Incoming examples 😄 

Screenshot 2021-12-01 at 03 29 23
Screenshot 2021-12-01 at 02 42 01
Screenshot 2021-12-01 at 03 30 06
Screenshot 2021-12-01 at 02 53 37
Screenshot 2021-11-29 at 01 41 58
Screenshot 2021-12-01 at 00 37 08
Screenshot 2021-12-01 at 00 44 45
Screenshot 2021-12-01 at 02 41 07
Screenshot 2021-12-01 at 02 51 10

piiq (approved):

piiq:
This is great! 

Here are a couple of things that I have encountered while using the feature:
- I could use a better help message on the following commands:
   - `exchange`. I don't understand what it does, but without triggering it I couldn't get data from some international exchanges,
   - the `fcf` help message says "operating margins"
   - `rec`. It comes as a float. What does it mean?
   - In terms of formatting the help messages should probably be capitalized

here are all the help messages for the commands as they are now

```
clear : Clear all or a particular parameter
industry : See existing industries, or set industry if arg specified
sector : See existing sectors, or set sector if arg specified
exchange : Swap exclude exchanges flag
cpi : companies per industry in a country
cps : companies per sector in a country
cr : Current Ratio
qr : Quick Ratio
de : debt to equity
tc : total cash
tcs : total cash per share
tr : total revenue
rps : revenue per share
rg : Revenue Growth
eg : earnings growth
pm : profit margins
gp : gross profits
gm : gross margins
ocf : operating cash flow
om : operating margins
fcf : operating margins
td : Total Debt
ebitda : earnings before interest, taxes, depreciation and amortization
ebitdam : ebitda margins
rec : Recommendation Mean
```

- When I exit the menu and enter it back the Industry/Sector/Cap/Country filter is still there but no data can be retrieved. Here's my scenario:
```
(✨) (stocks)>(sia)> industry software - application
Industry 'software - application' does not exist.
Replacing by 'Software - Application'

(✨) (stocks)>(sia)> roe
Limiting the amount of companies displayed to 10.

(✨) (stocks)>(sia)> q
(✨) (stocks)> load adsk
Loading Daily ADSK stock with starting period 2020-11-30 for analysis.

(✨) (stocks)> candle

(✨) (stocks)> sia
```
After this whatever command I call in the menu returns
```
Not able to find any data with the combination of Country (United%20States), Sector (Technology) and Industry (Software—Application).
```
The way around this was to trigger `clear` and set the industry/sector/country/cap parameters anew

*edit while writing:* I've tried to replicate the last issue with AMAT and everything was OK, but on ADSK it breaks

*edit2 while writing:* The problem is with the industry names that have `space dash space ` in them. When we pass the ticker data into the sia menu the industry is filled in without the spaces. `Software - Application` is the industry that sia uses and `Software-Application` is what's filled in from the yfinance json (line 98 in sia_controller)

DidierRLopes:
> This is great!
> 
> Here are a couple of things that I have encountered while using the feature:
> 
> * I could use a better help message on the following commands:
>   
>   * `exchange`. I don't understand what it does, but without triggering it I couldn't get data from some international exchanges,
>   * the `fcf` help message says "operating margins"
>   * `rec`. It comes as a float. What does it mean?
>   * In terms of formatting the help messages should probably be capitalized
> 
> here are all the help messages for the commands as they are now
> 
> ```
> clear : Clear all or a particular parameter
> industry : See existing industries, or set industry if arg specified
> sector : See existing sectors, or set sector if arg specified
> exchange : Swap exclude exchanges flag
> cpi : companies per industry in a country
> cps : companies per sector in a country
> cr : Current Ratio
> qr : Quick Ratio
> de : debt to equity
> tc : total cash
> tcs : total cash per share
> tr : total revenue
> rps : revenue per share
> rg : Revenue Growth
> eg : earnings growth
> pm : profit margins
> gp : gross profits
> gm : gross margins
> ocf : operating cash flow
> om : operating margins
> fcf : operating margins
> td : Total Debt
> ebitda : earnings before interest, taxes, depreciation and amortization
> ebitdam : ebitda margins
> rec : Recommendation Mean
> ```
> 
>  * When I exit the menu and enter it back the Industry/Sector/Cap/Country filter is still there but no data can be retrieved. Here's my scenario:
> 
> ```
> (✨) (stocks)>(sia)> industry software - application
> Industry 'software - application' does not exist.
> Replacing by 'Software - Application'
> 
> (✨) (stocks)>(sia)> roe
> Limiting the amount of companies displayed to 10.
> 
> (✨) (stocks)>(sia)> q
> (✨) (stocks)> load adsk
> Loading Daily ADSK stock with starting period 2020-11-30 for analysis.
> 
> (✨) (stocks)> candle
> 
> (✨) (stocks)> sia
> ```
> 
> After this whatever command I call in the menu returns
> 
> ```
> Not able to find any data with the combination of Country (United%20States), Sector (Technology) and Industry (Software—Application).
> ```
> 
> The way around this was to trigger `clear` and set the industry/sector/country/cap parameters anew
> 
> _edit while writing:_ I've tried to replicate the last issue with AMAT and everything was OK, but on ADSK it breaks
> 
> _edit2 while writing:_ The problem is with the industry names that have `space dash space ` in them. When we pass the ticker data into the sia menu the industry is filled in without the spaces. `Software - Application` is the industry that sia uses and `Software-Application` is what's filled in from the yfinance json (line 98 in sia_controller)

Thanks Theo,

This is great! So if you see the entire menu you have:
```
...
    clear         clear all or one of industry, sector, country and market cap parameters
    industry      see existing industries, or set industry if arg specified
    sector        see existing sectors, or set sector if arg specified
    country       see existing countries, or set country if arg specified
    mktcap        set mktcap between small, mid or large
    exchange      revert exclude exchanges flag

Industry          : 
Sector            : 
Country           : United States
Market Cap        : Large
Exclude Exchanges : True
...
```

The "exchange" flag allows to swap the "exclude exchanges" from true to false. The default is to exclude exchanges, so when you do "exchange" one time, you will not exclude exchanges which will lead to more tickers.

What you described as a bug there is actually a feature. When you load a ticker on the "stocks_controller" and go into the "sia menu" what happens is that I added some intelligence to recognise the ticker and get it's corresponding sector and industry so you could analyse it's peers. See:
Screenshot 2021-12-01 at 12 34 46

piiq:
> The "exchange" flag allows to swap the "exclude exchanges" from true to false. The default is to exclude exchanges, so when you do "exchange" one time, you will not exclude exchanges which will lead to more tickers.

I'm still a bit confused, How can we make the help message a bit more clear? Does it mean that if I exclude exchanges it only queries the default exchange (US)? Maybe `Exclude international exchanges` in this case?

> What you described as a bug there is actually a feature.

Yeah, I understood this. The bug here is that the spaces are not set and the terminal can't find the `Software-Application` industry because it searches for `Software - Application`. I hope this clarifies the finding

DidierRLopes:
> > The "exchange" flag allows to swap the "exclude exchanges" from true to false. The default is to exclude exchanges, so when you do "exchange" one time, you will not exclude exchanges which will lead to more tickers.
> 
> I'm still a bit confused, How can we make the help message a bit more clear? Does it mean that if I exclude exchanges it only queries the default exchange (US)? Maybe `Exclude international exchanges` in this case?
> 
> > What you described as a bug there is actually a feature.
> 
> Yeah, I understood this. The bug here is that the spaces are not set and the terminal can't find the `Software-Application` industry because it searches for `Software - Application`. I hope this clarifies the finding

Exactly! Yes, I prefer that message. Will change!

Ok, I got what you meant. There isn't a 1:1 relation between yahoo finance and finance database in regards with naming. Will. fix this now!

gamestonk_terminal/stocks/sector_industry_analysis/financedatabase_model.py
```diff
@@ -0,0 +1,248 @@
...
+    -------
+    list
+        List of possible countries
+    """
+    return fd.show_options("equities", "countries")
+
+
+def get_sectors():
+    """
+    Get all sectors in Yahoo Finance data. [Source: Finance Database]
```
jmaslek, line 24: Not to be super paining the ass but wouldn't your convention put the doctoring after the """ :P

piiq: We have a docstring convention? 🤣 (why yes it looks like we do)

although as a side note I would mention that the `pydocstyle` convention requires multiline docstrings to start from a new line

DidierRLopes: I think this is what we do @piiq .

Short doctoring starts above and long (multiline) starts below.

jmaslek (approved):

PR #1004, opened 2021-11-30
(Finally) add prediction abilities to crypto

jmaslek:
Been losing money using the stock prediction?  Well now you can lose even more when the markets are closed or just cuz.

This PR integrates the common prediction functions into crypto stuff.

Some notes:

- The crypto loading is still a bit of a mystery to me, so I am defaulting to loading from CoinGecko when entering the prediction menu.  This is a clear TODO.
- All of the prediction views got a time_res added to them so that timestamps that are less than daily can be plotted.  Another TODO is to integrate these with intraday stock data.  Not in the scope here.
- Kind of roundabout, but to do smaller timestamps, the user can `load` in the prediction menu.  Since I limit to just cg, I didn't think it made sense to add this to load, also since other sources can specify a resolution, but for CG, I am just downloading then resampling into ohlc...

Some screens:
Loading intraday bitcoin data
Screen Shot 2021-11-30 at 2 21 21 PM

Predicting the next 24 hours of bitcoin
Screen Shot 2021-11-30 at 2 23 25 PM

### DIFF
--- a/gamestonk_terminal/cryptocurrency/crypto_controller.py
+++ b/gamestonk_terminal/cryptocurrency/crypto_controller.py
@@ -756,7 +756,7 @@ What do you want to do?
 
             pred = pred_controller.menu(
                 self.current_coin,
-                c_help.load_cg_coin_data(self.current_coin, "USD", 30, "1H"),
+                c_help.load_cg_coin_data(self.current_coin, "USD", 365, "1D"),
             )
             if pred is False:
                 self.print_help()
--- a/gamestonk_terminal/cryptocurrency/crypto_controller.py
+++ b/gamestonk_terminal/cryptocurrency/crypto_controller.py
@@ -120,7 +120,7 @@ What do you want to do?
 >   ov          overview of the cryptocurrencies,       e.g.: market cap, DeFi, latest news, top exchanges, stables{dim}
 >   dd          due-diligence for loaded coin,          e.g.: coin information, social media, market stats
 >   ta          technical analysis for loaded coin,     e.g.: ema, macd, rsi, adx, bbands, obv
->   pred        prediction techniques
+>   pred        prediction techniques                   e.g.: regression, arima, rnn, lstm, conv1d, monte carlo
 {Style.RESET_ALL if not self.current_coin else ""}
 >   onchain     information on different blockchains,   e.g.: eth gas fees, active asset addresses, whale alerts
 >   defi        decentralized finance information,      e.g.: dpi, llama, tvl, lending, borrow, funding
@@ -743,6 +743,12 @@ What do you want to do?
 
     def call_pred(self, _):
         """Process pred command"""
+        if not self.current_coin:
+            print(
+                "No coin loaded.  Please use `load <coin>` to access prediction menu\n."
+            )
+            return
+
         if self.source != "cg":
             print("Currently only supports CoinGecko source.\n")
             return
--- a/gamestonk_terminal/cryptocurrency/crypto_controller.py
+++ b/gamestonk_terminal/cryptocurrency/crypto_controller.py
@@ -61,7 +61,7 @@ class CryptoController:
         "find",
     ]
 
-    CHOICES_MENUS = ["ta", "dd", "ov", "disc", "onchain", "defi", "nft"]
+    CHOICES_MENUS = ["ta", "dd", "ov", "disc", "onchain", "defi", "nft", "pred"]
 
     SOURCES = {
         "bin": "Binance",
@@ -120,6 +120,7 @@ What do you want to do?
 >   ov          overview of the cryptocurrencies,       e.g.: market cap, DeFi, latest news, top exchanges, stables{dim}
 >   dd          due-diligence for loaded coin,          e.g.: coin information, social media, market stats
 >   ta          technical analysis for loaded coin,     e.g.: ema, macd, rsi, adx, bbands, obv
+>   pred        prediction techniques
 {Style.RESET_ALL if not self.current_coin else ""}
 >   onchain     information on different blockchains,   e.g.: eth gas fees, active asset addresses, whale alerts
 >   defi        decentralized finance information,      e.g.: dpi, llama, tvl, lending, borrow, funding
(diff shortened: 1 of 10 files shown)


jmaslek:
@JakubPluta feel free to leave me some feedback!

JakubPluta:
Hi, @jmaslek  overall PR looks great, I think it will very useful for crypto traders :). 
So it's a damn pretty huge feature that will add a lot of new possibilities for the crypto menu.

Regarding loading coins in the crypto menu, it's pretty complicated. The main issue there is that we have different sources for crypto data, and all of them have different identifiers of coins. It means that to load e.g. uniswap coin, in one data source you need to use id: uniswap, in other uni-uniswap, next other one UNI, 
and it makes some pretty hard to load coins in an easy way. 

It's not like in stocks that you have ticker AAPL and it's the same in all data sources :( That's why I tried to create some kind of translation table and map all used sources in the way that u can basically use only coingecko id or symbol as an entry point for all of these data sources, it's not working in 100% but I would say, that's it's not to bad. Anyway, in most cases, people are interested in coins from the top 100 market cap, so these ones are well mapped.  

Maybe we should think of some kind of script that will in e.g 1-day schedule fetch all coin ids 
from different sources and map it to our own translation table. But for sure it's something pretty complex, 
as we add a new data sources, new kind crypto ids formats that we will need to somehow add to this table.

Regarding pool request, I think that I would add
some description in the help menu in the main crypto controller, and also I would separate cases: 
- user didn't load coin -> in return information "Please load a coin to use pred menu",
- user loaded coin for different source then coingecko: "Only coingecko is supported"


But anyway I think that if you want to load data for prediction you should reuse function called 
`load_ta_data`, which returns dataframe with prices, and quoted currency. 
So the steps should look like call_load data `e.g. load btc --source cg, then in 
call_pred you should somehow reuse what is in call_ta function. This way I think it will be possible 
to support all data sources (binance, coingecko, coinpaprika). I could help with that. I mean you can merge your changes, 
and in a next steps I could try to add posiblity to use other source, or we can align on call and discuss it togather,
 maybe there are some better ideas/


Other things seems very good to me, I tried all of prediction commands and they worked very well.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -740,6 +741,32 @@ def call_dd(self, _):
                 "No coin selected. Use 'load' to load the coin you want to look at.\n"
             )
 
+    def call_pred(self, _):
+        """Process pred command"""
+        if self.source != "cg":
```
JakubPluta, line 746: I would add there additional case. When user haven't load any coin in previous step it should return information, that he should do that to access this pred menu.

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -120,6 +120,7 @@ def print_help(self):
 >   ov          overview of the cryptocurrencies,       e.g.: market cap, DeFi, latest news, top exchanges, stables{dim}
 >   dd          due-diligence for loaded coin,          e.g.: coin information, social media, market stats
 >   ta          technical analysis for loaded coin,     e.g.: ema, macd, rsi, adx, bbands, obv
+>   pred        prediction techniques
```
JakubPluta, line 123: Maybe it's good idea to add there some examples like in commands above. What do you think ?

jmaslek: Oops.  Yeah was being lazy.

gamestonk_terminal/cryptocurrency/crypto_controller.py
jmaslek: Yup.  Will do.

jmaslek:
> Regarding loading coins in the crypto menu, it's pretty complicated. The main issue there is that we have different sources for crypto data, and all of them have different identifiers of coins. It means that to load e.g. uniswap coin, in one data source you need to use id: uniswap, in other uni-uniswap, next other one UNI, and it makes some pretty hard to load coins in an easy way.
> 
> It's not like in stocks that you have ticker AAPL and it's the same in all data sources :( That's why I tried to create some kind of translation table and map all used sources in the way that u can basically use only coingecko id or symbol as an entry point for all of these data sources, it's not working in 100% but I would say, that's it's not to bad. Anyway, in most cases, people are interested in coins from the top 100 market cap, so these ones are well mapped.

Doing more loading is definitely a TODO, but I just picked CoinGecko since it didn't require a key

> Maybe we should think of some kind of script that will in e.g 1-day schedule fetch all coin ids from different sources and map it to our own translation table. But for sure it's something pretty complex, as we add a new data sources, new kind crypto ids formats that we will need to somehow add to this table.
> 
> Regarding pool request, I think that I would add some description in the help menu in the main crypto controller, and also I would separate cases:
> 
> * user didn't load coin -> in return information "Please load a coin to use pred menu",
> * user loaded coin for different source then coingecko: "Only coingecko is supported"
> 
> But anyway I think that if you want to load data for prediction you should reuse function called `load_ta_data`, which returns dataframe with prices, and quoted currency. So the steps should look like call_load data `e.g. load btc --source cg, then in call_pred you should somehow reuse what is in call_ta function. This way I think it will be possible to support all data sources (binance, coingecko, coinpaprika). I could help with that. I mean you can merge your changes, and in a next steps I could try to add posiblity to use other source, or we can align on call and discuss it togather, maybe there are some better ideas/

So the reason I did not do that for this PR was that I wanted the user to be able to change the time period, since crypto is not just during market hours, and I did not want to edit that function and break the  `ta` functionality.  I think that designing a good way of loading all sources will be important but also out of the scope of what I was trying to do.

JakubPluta:
> So the reason I did not do that for this PR was that I wanted the user to be able to change the time period, since crypto is not just during market hours, and I did not want to edit that function and break the `ta` functionality. I think that designing a good way of loading all sources will be important but also out of the scope of what I was trying to do.

Totally understand it, it needs for sure more design at loading step, to have strong fundamentals for other features.

DidierRLopes (approved):

PR #1005, opened 2021-12-01
Updating tests : stocks/due_diligence

Chavithra:
**Reasons**
- Adding `tests` for : `stocks/due_diligence`
- Adding helpers : `fixtures`

**Consequences**
- New `fixtures` to calculate the proper path for `data files` :
  - default_csv_path
  - default_txt_path
  - default_json_path
- New `fixture` to record `stdout` :
  - record_stdout
- Adding a `README.md` in `tests` folder
- New `tests` + `data files` for the package :
  - gamestonk_terminal/stocks/due_diligence
- Excluding additional `cassettes` from `detect-secrets` `pre-commit` :
Previous cassettes were already excluded.
I also excluded new cassettes.
The `gzipped request body` inside cassettes are detected as they were `secrets`.


aia:
> The binary data in the cassette yaml files. Is it possible to have that data in text format rather than binary? It may reduce pressure on the repo long term as git isn't good with doing diffs for binary blobs which would lead to a situation where all past versions of the binary blobs are stored in the repo history (instead of just the diffs)

@Chavithra and I have discussed Base64 encoded cassettes and agreed that keeping Base64 is a reasonable compromise.

aia (approved):
Amazing!

Chavithra:
> Hey @Chavithra 👋 This is a very nice PR. Great work 👏👏👏
> 
> I would ask one thing before diving deeper into the changes.
> 
> The binary data in the cassette yaml files. Is it possible to have that data in text format rather than binary? It may reduce pressure on the repo long term as git isn't good with doing diffs for binary blobs which would lead to a situation where all past versions of the binary blobs are stored in the repo history (instead of just the diffs)

That's possible but it will increase the `cassettes` size so I didn't.

Only the body of the response is `gzipped`.

I will add a command line tool to listing and decode `cassettes bodies`.

DidierRLopes (approved):
Great work!

I really liked the README file!

PR #1007, opened 2021-12-01
Screener controller big improvements

DidierRLopes:
@deeleeramone has brought to my attention that the screener could be improved in order for a more dynamic filtering which could allow for better content on our twitter.

I've improved the menu a bit by combining default presets (from finviz) and custom ones (from users) and allowing to sort by one of the columns of the table.

Next step could be taking these to comparison analysis menu. But I think that menu should be refactored a bit to not depend on a "main" ticker and rather perform the comparison of tickers provided.

Screenshot 2021-12-01 at 18 55 26


jmaslek:
Nice!

jmaslek (approved):

PR #1012, opened 2021-12-02
Improve comparison analysis controller to not depend on main ticker

DidierRLopes:
This PR refactors comparison analysis menu to not be dependent on a ticker, this allows that we can enter in there easily from screener (will implement after this gets merged) or sector and industry analysis.

Also, I made this menu a bit more robust and improved user experience. Towards a beast of a Gamestonk Terminal 1.0 🚀

### DIFF
--- a/gamestonk_terminal/stocks/comparison_analysis/ca_controller.py
+++ b/gamestonk_terminal/stocks/comparison_analysis/ca_controller.py
@@ -6,18 +6,21 @@ import difflib
 import random
 from typing import List
 
+from datetime import datetime, timedelta
+import yfinance as yf
 from colorama import Style
-import pandas as pd
 from matplotlib import pyplot as plt
 from prompt_toolkit.completion import NestedCompleter
 
 from gamestonk_terminal import feature_flags as gtff
 from gamestonk_terminal.helper_funcs import (
     check_non_negative,
+    check_positive,
     get_flair,
     parse_known_args_and_warn,
     try_except,
     system_clear,
+    valid_date,
 )
 from gamestonk_terminal.menu import session
 from gamestonk_terminal.portfolio.portfolio_optimization import po_controller
@@ -31,7 +34,6 @@ from gamestonk_terminal.stocks.comparison_analysis import (
     yahoo_finance_view,
     yahoo_finance_model,
 )
-from gamestonk_terminal.stocks.stocks_helper import load
 
 # pylint: disable=E1121
 
@@ -43,12 +45,13 @@ class ComparisonAnalysisController:
     CHOICES = ["?", "cls", "help", "q", "quit"]
 
     CHOICES_COMMANDS = [
-        "load",
+        "set",
         "getpoly",
         "getfinnhub",
         "getfinviz",
         "select",
         "add",
+        "rmv",
         "historical",
         "hcorr",
         "volume",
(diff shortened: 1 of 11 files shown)


jmaslek:
Title for historical can be long...
Screen Shot 2021-12-02 at 8 16 30 PM

jmaslek:
You might need to break up the market watch calls? 
Screen Shot 2021-12-02 at 8 18 14 PM

Screen Shot 2021-12-02 at 8 19 06 PM

jmaslek:
> You might need to break up the market watch calls? (Note that FB in the first call is all N/A, but there is data when called -- which there should be 2 in that. Used `set goog`, `select fb` and only got fb) Screen Shot 2021-12-02 at 8 18 14 PM
> 
> Screen Shot 2021-12-02 at 8 19 06 PM


This behavior is inconsistent for me ???  Now it doesn't happen

jmaslek:
From menu:
```
    set           set ticker to get similars from

Get similar: GOOG
    tsne          run TSNE on all SP500 stocks and returns closest tickers
```

Think get similar should be a more descriptive string.  Idk if we wanna call it `Main Ticker` or `Ticker to compare to` (too long for my liking)...

jmaslek:
```
(✨) (stocks)>(ca)> set aapl
a
(✨) (stocks)>(ca)> add amc
[Custom] Similar Companies: AMC 

(✨) (stocks)>(ca)> historical
"None of [Index(['AMC'], dtype='object', name='Date')] are in the [index]" 

(✨) (stocks)>(ca)>
```

Same happens for `select GME` but not after I do `getpoly`. (everything is with `set AAPL`)

jmaslek:
Think I found some buttons to break everything lol.  Main comments would be to not show all tickers in plot title and maybe rename the `Get Ticker` in menu.

piiq:
Niice! Great update!

I think I found some stuff:

- historical doesn't work if only 1 company added
image

- rmv doesn't clear the list of companies added
image
*edit while writing:* it does remove if I specify a ticker. Should we have an `rmv` with no args remove all tickers from a list?
- in the `rmv` the dosctring has a typo. "to" should be "from"

- I couldn't get getpoly to work (I have polygon key). Am I missing something?
image
*edit while writing:* I did make it work. it seems that "get similar companies" works only if you explicitly `set` the ticker to compare to.

- I am a bit confused with the `set`, `select` and `add` commands. They all do more or less the same thing but the results of what they do is different. I thing that `set` and `select` are the same, but they're not. Agree with @jmaslek that we might want to change it to "main ticker" or something.
*another edit while writing*: if I use `set` then `select` doesn't reset the `set` results (i hope this is not confusing).
try this:
```
stocks
ca
set aapl
select gme
help
```
you'll have both aapl and gme in the output, but based on the `select` docstring there shouldn't be aapl there

- If no ticker is specified the finviz commands are not blocked (`ownership` for example), and take super long to process and returns a table with 800+ rows. Should we add a progressbar or block them?

- sentiment chart doesn't show company names or tickers
image

that's all i've found for now

DidierRLopes:
@jmaslek @piiq I think you missed the purpose of `set` 😄 

I thought it was quite clear, but apparently it's not.

`set` only set ticker to get similars from. None of the commands like historical, balance, ... do that. Only the tsne, and getxxx do that.

Perhaps the `add` / `rmv` descriptions should have been improved to say "similar companies" and not talk about a currently selected.

Screenshot 2021-12-03 at 15 34 07

I will try to improve this on the next commit.

DidierRLopes:
@piiq I added capability of `rmv` to remove all, but doing just `select` would be the same.

DidierRLopes:
Screenshot 2021-12-03 at 16 43 43

@piiq regarding sentiment chart I was being silly. It's never a good idea to change a list while you are iterating it :) Should be all fixed now!

DidierRLopes:
because I had some free time and you took long to review I allowed to jump from screener menu to comparison analysis menu! 🚀 

So now you can screen some stocks, and take them to the comparison analysis menu to compare them in more detail!! 🚗  buuup buuup

jmaslek:
> because I had some free time and you took long to review I allowed to jump from screener menu to comparison analysis menu! 🚀

Got it.   Don't review your PRs so more gets done 👍

jmaslek (approved):

PR #1016, opened 2021-12-04
Bitquery feature branch

JakubPluta:
Hi there, 

Recently I have been working on integrating Bitquery (https://bitquery.io/) GraphQL APIs which allows me to query blockchain data. The possibilities offered by this API are huge, but it takes a lot of work to read the documentation and learn how to extract data. So far, I have only managed to touch the simplest things. However, I see a lot of potential in this. One of the disadvantages is that if we want to reach more historical data, the query often takes 30-40 seconds, and sometimes it is associated with timouts, so far I limited in the implemented features, to rather reach up to 3 months back, in this way inquiries is quite optimal :)

I also dumped most popular erc20 tokens to json file, and implement simple logic for mapping token symbol to blockchain addresses, which are required in some queries. 

Note: Bitquery requires API KEY


jose-donato:
Looks nice. Will take a look and try it tomorrow

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
```diff
@@ -849,6 +1383,14 @@ def print_help(self):
 
 Whale Alert:
     whales            check crypto wales transactions
+
+BitQuery (note: some commands execution can take > 10 seconds):
```
jose-donato, line 1387: do you think it is possible to add a progress indicator for the commands that take more time?
(`sia` menu has good progress indicators for some of the commands)

piiq (changes requested):
Hey @JakubPluta  this looks very nice! Great work and thank you!

There is a blocking error with loading of the .json with the erc tokens. The relative path cannot be resolved. I've left some comments in the code with regards to this error and added some other suggestions. Let me know what you think and if you need a helping hand.

Apes together strong 🦍

jose-donato:
Just tested everything locally also. Apart from the JSON path error that @piiq pointed everything looks good to me.

Great work 🎆

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
piiq: a progress makes sense if there are multiple calls made, which might not be the case here (but i don't know for sure)

JakubPluta: Agree with piiq, in case if we would scrape/query multi pages it would be good idea to do that, but in my case it's just a single query, but it's seems that it's expensive for bitquery server. I tried to limit somehow all queries in the way that time to recieve data will be pretty fast (like few seconds)

gamestonk_terminal/cryptocurrency/onchain/bitquery_model.py
```diff
@@ -0,0 +1,638 @@
...
+    """Helper method that loads ~1500 most traded erc20 token.
+    [Source: json file]
+
+    Returns
+    -------
+    pd.DataFrame
+        ERC20 tokens with address, symbol and name
+    """
+
+    with open("../data/erc20_coins.json") as f:
```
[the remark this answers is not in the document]
JakubPluta: Yes, will fix it :)  Thanks for that, this is something that I always forget to do. My bad.

gamestonk_terminal/cryptocurrency/onchain/bitquery_model.py
```diff
@@ -0,0 +1,638 @@
...
+            "transactions",
+            "avgGasPrice",
+            "medGasPrice",
+            "maxGasPrice",
+        ]
+    ]
+
+
+def get_most_traded_pairs(
+    network: str = "ethereum", exchange: str = "Uniswap", limit: int = 90
```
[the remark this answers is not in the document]
JakubPluta: So, for the query to bitquery api case sensitive exchange name is important, but I will try to make it possible that user don't need to take care about it.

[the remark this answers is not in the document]
piiq: Have a look at how Countries are selected in the `sector industry analysis` menu (it's still WIP but it looks and feels very nice). In the `call_country` function there is a code block that adds autocomplete for the country selection and substitutes all lowercase string with correct capitalized string.
The code is currently in a branch, but if it changes 'cause the PR is merged you'll find in it main
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/d4babe3dece5fcbd71f31fc0e1addc2ab1184869/gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py#L536

gamestonk_terminal/cryptocurrency/onchain/bitquery_model.py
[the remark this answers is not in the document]
piiq: Thanks for doing the heavy lifting, mate

gamestonk_terminal/cryptocurrency/onchain/bitquery_model.py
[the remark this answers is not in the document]
JakubPluta: Thanks for your comment, I implemented some stuffs with difflib in crypto menu, so I have some basic knowledge how to implement it. I took a look on `call_country` function, but I don't know if I understand correctly how this autocompleter works. Later I will try to check branch with this feature on my local computer, to see how it works. But in the meantime, if you have time you can take a look on my solution. :)

I think a little issue there is that we explicitly add list of possible exchanges into argparser choice parameter. So if we want to use difflib to get best match for what user wrote, we need to resign from choice parameter which is an array of possible exchanges. So I don't know what it will be a better experience for user: to have difflib in the background, but not see possible exchanges with -h parameter, or to have listed exchanges with lowercase, and in the background to have a mapping, that will translate lower case string into what api expects. For now I provide b) solution with lowercase and background mapping.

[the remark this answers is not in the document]
piiq: @DidierRLopes can you have a look here if it's possible to add autocomplete for exchanges the same way the country selection is done in sia?

DidierRLopes:
Same here with naming:
Screenshot 2021-12-06 at 18 22 09

Same for all commands regarding better column names.

```
(✨) (crypto)>(onchain)> mt -e decentrex
List of dex trades is empty {'dexTrades': []}
(✨) (crypto)>(onchain)>
```
here we could have a better error message and a new line.

Also no need for exchange here I believe
Screenshot 2021-12-06 at 18 26 27

I feel that these data is very nice and most of them could even be plots! To be more insightful, and then we could use `--raw` option to get this table format. What do you think?

gamestonk_terminal/cryptocurrency/onchain/bitquery_model.py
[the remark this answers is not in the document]
DidierRLopes: Yup, so this is how you want to do it:
Screenshot 2021-12-06 at 18 32 06

In your case the choices should be 
```
choices["mt"]= {c: None for c in possible_exchanges}
```

Note that this will only happen when you do `mt unis...` for instance but it's still very helpful.


Also, look into how I implemented this:
```
 def call_country(self, other_args: List[str]):
        """Process country command"""
        parser = argparse.ArgumentParser(
            add_help=False,
            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
            prog="country",
            description="See existing countries, or set country if arg specified",
        )
        parser.add_argument(
            "-n",
            "--name",
            type=str,
            dest="name",
            nargs="+",
            help="country to select",
        )
        if other_args:
            if "-" not in other_args[0]:
                other_args.insert(0, "-n")
         if not ns_parser:
             return

         possible_countries = financedatabase_model.get_countries(
             industry=self.industry, sector=self.sector
         )

         if ns_parser.name:
             if " ".join(ns_parser.name) in possible_countries:
                 self.country = " ".join(ns_parser.name)
             else:
                 print(f"Country '{' '.join(ns_parser.name)}' does not exist.")

                 similar_cmd = difflib.get_close_matches(
                     " ".join(ns_parser.name),
                     possible_countries,
                     n=1,
                     cutoff=0.75,
                 )
                if similar_cmd:
                    print(f"Replacing by '{similar_cmd[0]}'")
                    self.country = similar_cmd[0]
                 else:
                     similar_cmd = difflib.get_close_matches(
                         " ".join(ns_parser.name),
                         possible_countries,
                         n=1,
                         cutoff=0.5,
                     )
                     if similar_cmd:
                         print(f"Did you mean '{similar_cmd[0]}'?")

         else:
             for country in possible_countries:
                 print(country)
```

You are right when you say that the `choices=...` allows to retrieve an error straight if user doesn't provide correct argument. BUT with this implementation you test difflib similarity, if it is above 0.75 you go ahead and replace, If it is between 0.5 and 0.75 you ask "did you mean XXXX" and if it's below you can say something like "XXX not found. Use one of the following: .....". This way you are not as harsh saying that it doesn't exist even if the difference is just a capital letter or so. 

Does that make sense?

gamestonk_terminal/config_terminal.py
```diff
@@ -111,3 +111,6 @@
 
 # https://cryptopanic.com/developers/api/
 API_CRYPTO_PANIC_KEY = os.getenv("GT_API_CRYPTO_PANIC_KEY") or "REPLACE_ME"
+
+# https://bitquery.io/pricing
+API_BITQUERY_KEY = os.getenv("GT_API_BITQUERY_KEY")
```
DidierRLopes, line 116: or "REPLACE_ME"

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
DidierRLopes: Can we remove that note though?

I feel like a lot of commands have a waiting time, and if we want to notify the user about this should happen at the command level, like a print when we execute it. Rather than being in the menu adding noise.

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
```diff
@@ -849,6 +1383,14 @@ def print_help(self):
...
+    trades            last trades by dex or month
+    volp              daily volume for crypto pair
+    volt              token volume on DEXes
+    senders           unique ethereum addresses which made a transaction
+    mt                top traded crypto pairs on given decentralized exchange
+    spreads           bid, ask prices, average spread for given crypto pair
```
DidierRLopes, line 1393: So, one thing I try to do is have intuitive names that will suggest me straight the command I'm doing. In part, the command should make me think of the description. 

With that in mind I would have something along lines:
```
    lt                     last trades by dex or month
    dvcp               daily volume for crypto pair
     tv                   token volume on DEXes
     stran              specific transactions by ethereum addresses
     ttcp                top traded crypto pairs on given decentralized exchange
     baas               bid, ask and average spread for given crypto pair
```

What do you think?

DidierRLopes (changes requested):

gamestonk_terminal/cryptocurrency/onchain/bitquery_model.py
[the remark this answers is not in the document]
JakubPluta: Didier, thank you for comment. Will try to implement it for decentralized exchanges.

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
JakubPluta: > I would improve the "tradeAmount" column here, and drop a row that doesn't has an exchange name. Screenshot 2021-12-06 at 17 59 48
> 
> It says that default is `ETH` but when I do `volp` I get an error. You need to do `required="-h" not in other_args,'. Also here the column name `tradeAmountUSD` could be improved. Screenshot 2021-12-06 at 18 01 20
> 
> Same as above for `volt`. In addition given we specify a coin I don't think we need that column in here. Screenshot 2021-12-06 at 18 03 55
> 
> I'm still testing the other ones.

Thanks for all your comments, I tried to not change column names coming from source data, but I can improve it. What do you mean when you talk about better column naming ? 
e.g. `tradeAmount` -> you mean to change camel case names, or to change it more to `volume` ?
If you can give me few sugesstions with better naming will be much easier to me, to change it.

DidierRLopes: No, so I just mean something simple like "Trade Amount".

You don't even need to do this one by one, you can do the following:
```
import re

df.columns = [' '.join(re.findall('.[^A-Z]*', val)).capitalize() for val in df.columns]
```
Screenshot 2021-12-07 at 12 39 33

JakubPluta: @DidierRLopes thanks for an examples. Just wonder, how it will behave if you have it as in `sortby` argparser parameter?

So if user will try to write e.g. `volt -t 10 -s  Trade Amount` it will fail, because argparser will split space into seperate arguments like `Trade Amount` = ["Trade", "Amount"]`, is there any way to change this behavior ?

DidierRLopes: So I'm fine with keeping the `-s tradeAmount` and then we just improve the way the data frame is output. Basically just before displaying it to the user

jmaslek:
Can you add a test for the api in the keys function?  A couple recent crypto features haven't had this added.

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
```diff
@@ -1262,10 +1265,44 @@ def call_ttcp(self, other_args: List[str]):
             if not ns_parser:
                 return
 
+            exchange = "Uniswap"
```
JakubPluta, line 1268: @DidierRLopes I am not sure if I implemented this difflib feature :D But would be great if you can take a look. I set exchange parameter to not required, and I removed `choices`.

JakubPluta, line 1268: by default it's exchange is set to Uniswap as top 1 decentralized exchange.

gamestonk_terminal/cryptocurrency/onchain/onchain_controller.py
```diff
@@ -1447,3 +1485,6 @@ def menu():
 
         if process_input is True:
             return True
+
+
+menu()
```
JakubPluta: I removed it with next push. Didn't see that

DidierRLopes:
Yes, this is looking very nice!!

Last suggestion would be for `baas` to also have difflib, and implementing auto-completion for both baas and ttcp. See how @jose-donato has done it in this commit: 

https://github.com/GamestonkTerminal/GamestonkTerminal/pull/1028/commits/bffd31ae3710d313e64dfbbe5df6151ca3e6596c#diff-22903e095008bdb1c4865aa81eb7e9b7b52488d22bda59a9c3161cc7b1adae9fR1805

JakubPluta:
> Yes, this is looking very nice!!
> 
> Last suggestion would be for `baas` to also have difflib, and implementing auto-completion for both baas and ttcp. See how @jose-donato has done it in this commit:
> 
> [bffd31a#diff-22903e095008bdb1c4865aa81eb7e9b7b52488d22bda59a9c3161cc7b1adae9fR1805](https://github.com/GamestonkTerminal/GamestonkTerminal/commit/bffd31ae3710d313e64dfbbe5df6151ca3e6596c#diff-22903e095008bdb1c4865aa81eb7e9b7b52488d22bda59a9c3161cc7b1adae9fR1805)

Thanks :D I applied almost the same for `baas`. Needed to skip part with printing coins, as there are 1000 of them. If it's fine for you, we can merge it!

DidierRLopes (approved):
This looks good to me now!

Thank you @JakubPluta !!! ❤️

piiq (approved):
Thanks @JakubPluta this UX is very nice
image

PR #1017, opened 2021-12-04
Binomial Options Pricing

colin99d:
Binomial Options pricing first uses volatility to estimate the asset prices in the future, then uses these estimates to predict the value of options. This PR lets users use binomial options pricing on American/European calls and puts. The feature also allows users to chart the expected asset price on the expiration date, and allows them to export the binomial trees to excel. 

**Note:** currently the model seems to be consistently underpricing options. I am not sure whether this is a bug or if current market volatility is just significantly higher than the average over the past year. I am open to feedback in this area.
**Note2:** it looks like the file `.ipynb_checkpoints/terminal-checkpoint.py` got added to my PR. I have no clue what this is. Let me know how I should handle it.


piiq:
> Note2: it looks like the file .ipynb_checkpoints/terminal-checkpoint.py got added to my PR. I have no clue what this is. Let me know how I should handle it.

It's the cache that ipython notebooks save on disk. You can delete it and add .ipynb_ckeckpoints to gitignore

colin99d:
> > Note2: it looks like the file .ipynb_checkpoints/terminal-checkpoint.py got added to my PR. I have no clue what this is. Let me know how I should handle it.
> 
> It's the cache that ipython notebooks save on disk. You can delete it and add .ipynb_ckeckpoints to gitignore

Thanks for the insight! Update is in

jmaslek:
So whenever you implement a calculation like this, you need to include some helpful information about what the calculation is doing.  Check out pyportfolioopts docs:  https://pyportfolioopt.readthedocs.io/en/latest/MeanVariance.html. .  When I look there I know exactly where the numbers they give come from.

Ill gather a good percentage of users won't be familiar with a binomial tree, so you should have the documentation about the math involved so that they can look at it and understand we aren't just giving them a random number.

gamestonk_terminal/stocks/options/options_controller.py
```diff
@@ -1260,6 +1262,80 @@ def call_parity(self, other_args: List[str]):
...
+            prog="parity",
+            description="Shows whether options are over or under valued",
+        )
+        parser.add_argument(
+            "-s",
+            "--strike",
+            type=float,
+            default=0,
+            dest="strike",
+            help="Strike price for optiom shown",
```
jmaslek, line 1280: spelling

gamestonk_terminal/stocks/options/options_controller.py
```diff
@@ -1260,6 +1262,80 @@ def call_parity(self, other_args: List[str]):
...
         print("")
 
+    @try_except
+    def call_binom(self, other_args: List[str]):
+        """Process binom command"""
+        parser = argparse.ArgumentParser(
+            add_help=False,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+            prog="parity",
+            description="Shows whether options are over or under valued",
```
jmaslek, line 1272: This function gives the option value based on a binomial tree

gamestonk_terminal/stocks/options/options_controller.py
```diff
@@ -1260,6 +1262,80 @@ def call_parity(self, other_args: List[str]):
...
+            help="Value a put instead of a call",
+        )
+        parser.add_argument(
+            "-e",
+            "--european",
+            action="store_true",
+            default=False,
+            dest="europe",
+            help="Value a European option instead of an American one",
+        )
```
jmaslek, line 1297: Probably don't need to worry about European calls

gamestonk_terminal/stocks/options/options_controller.py
```diff
@@ -1260,6 +1262,80 @@ def call_parity(self, other_args: List[str]):
...
+        parser.add_argument(
+            "--export",
+            action="store_true",
+            default=False,
+            dest="export",
+            help="Export an excel spreadsheet with binomial pricing data",
+        )
+        parser.add_argument(
+            "-P",
```
jmaslek, line 1306: Personally the capital P confuses me, so I would leave --plot only.  Just personal

gamestonk_terminal/stocks/options/yfinance_model.py
```diff
@@ -117,9 +117,54 @@ def generate_data(
...
+    Returns
+    ----------
+    price : float
+        The price of the ticker
+    """
     ticker_yahoo = yf.Ticker(ticker)
     data = ticker_yahoo.history()
     last_quote = data.tail(1)["Close"].iloc[0]
     return last_quote
+
```
jmaslek, line 137: An actual question:  is this faster than doing yf.Ticker().info["regularMarketPrice"]?

gamestonk_terminal/stocks/options/yfinance_model.py
```diff
@@ -117,9 +117,54 @@ def generate_data(
...
+    ticker : str
+        The ticker to get the price for
+
+    Returns
+    ----------
+    price : List[float]
+        A list of closing prices for a ticker
+    """
+    tick = yf.Ticker(ticker)
+    return tick.history(period="1y")["Close"]
```
jmaslek, line 170: do you want Close or AdjClose ?

gamestonk_terminal/stocks/options/yfinance_view.py
```diff
@@ -662,3 +666,236 @@ def risk_neutral_vals(
...
+    und_vals : List[List[float]]
+        The expected underlying values at the expiration date
+    p : float
+        The probability of the stock price moving upward each round
+    ticker : str
+        The ticker of the option's underlying asset
+    expiration : str
+        The expiration for the option
+    """
+    ups = list(range(len(und_vals[-1])))
```
jmaslek, line 687: use a more descriptive variable name.  Your future self will thank you

gamestonk_terminal/stocks/options/yfinance_view.py
```diff
@@ -662,3 +666,236 @@ def risk_neutral_vals(
...
+    ups.reverse()
+    probs = [binom.pmf(r, len(ups), p) for r in ups]
+    fig, ax = plt.subplots(figsize=plot_autoscale(), dpi=cfp.PLOT_DPI)
+    ax.set_title(f"Probabilities for ending prices of {ticker} on {expiration}")
+    ax.xaxis.set_major_formatter("${x:1.2f}")
+    ax.yaxis.set_major_formatter(mtick.PercentFormatter())
+    plt.plot(und_vals[-1], probs)
+    fig.tight_layout()
+    plt.show()
```
jmaslek, line 696: Add in the USE ION feature flag along with the ion method

gamestonk_terminal/stocks/options/yfinance_view.py
```diff
@@ -662,3 +666,236 @@ def risk_neutral_vals(
...
+    div_yield = (
+        info["trailingAnnualDividendYield"]
+        if info["trailingAnnualDividendYield"] is not None
+        else 0
+    )
+    delta_t = 1 / 252
+    rf = get_rf()
+    exp_date = datetime.strptime(expiration, "%Y-%m-%d").date()
+    today = date.today()
+    days = (exp_date - today).days
```
jmaslek, line 842: Question (something I ran into when looking at black Scholes pricing):  If you look at 0dte options, then days will be 0.  This causes some things to diverge (maybe not here).  So should this actually be in hours?  That way you have decimal days til expiry?

Note that you don't need to change it, just a question I am not sure about.

website/content/stocks/options/binom/_index.md
```diff
@@ -0,0 +1,18 @@
+```
+usage: binom [-s STRIKE] [-p] [-e] [-E] [-P] [-v VOLATILITY] [-h]
+```
+
+Shows the value of an option using binomial option pricing. Can also show raw data and provide a graph with predicted underlying asset ending values.
+
+```
```
jmaslek, line 7: As I mentioned in a comment, you should add ALOT more description.

gamestonk_terminal/stocks/options/options_controller.py
colin99d: Is there a harm in keeping it? Particularly if someone wants to analyze data in excel.

jmaslek: I don't see any harm

gamestonk_terminal/stocks/options/yfinance_model.py
colin99d: Shocking results here. I ran code because I was generally curious and my way was 20 - 30 times faster. Not sure why.

`import yfinance as yf
from datetime import datetime

print("Running time for 10 requests")
print("Colin's Method:")

colin_start = datetime.now()
for _ in range(10):
    ticker_yahoo = yf.Ticker("TSLA")
    data = ticker_yahoo.history()
    last_quote = data.tail(1)["Close"].iloc[0]
colin_end = datetime.now()

colin_diff = (colin_end - colin_start)
print(colin_diff.total_seconds() * 1000)

print("\nJame's Method:")

james_start = datetime.now()
for _ in range(10):
    last_quote = yf.Ticker("TSLA").info["regularMarketPrice"]
james_end = datetime.now()

james_diff = (james_end - james_start)
print(james_diff.total_seconds() * 1000)`

gamestonk_terminal/stocks/options/yfinance_model.py
colin99d: Not sure why by that is not an option:
ndex(['Open', 'High', 'Low', 'Close', 'Volume', 'Dividends', 'Stock Splits'], dtype='object')

gamestonk_terminal/stocks/options/yfinance_model.py
jmaslek: So I actually did this previously.  Same results:

https://user-images.githubusercontent.com/18151143/143073050-794beea1-524b-423f-a914-f1ef32207cbc.png

My guess is getting all the dictionary entries takes longer than getting the time series.

gamestonk_terminal/stocks/options/yfinance_model.py
jmaslek: So I guess doing yf.download() gives the adj close and whatnot.

gamestonk_terminal/stocks/options/yfinance_model.py
DidierRLopes: @colin99d FYI, you can use timeit (https://docs.python.org/3/library/timeit.html) for this type of assessment

This doesn't surprise me though @jmaslek, as this is the reason why I mention having the last price from when we do `load` rather than with `quote`.

colin99d: Thats my guess too.

gamestonk_terminal/stocks/options/yfinance_view.py
colin99d: I do not have the data but I would venture most options on the date of expiration trade for pretty close to expiration because of they are expiring so soon. This means even with no additional branches being made we will still be decently close to the real value. I think focusing on a better measure of volatility will be much better for the tools overall predictive value.

website/content/stocks/options/binom/_index.md
```diff
@@ -0,0 +1,17 @@
...
+usage: binom [-s STRIKE] [-p] [-e] [-E] [-P] [-v VOLATILITY] [-h]
+```
+
+Shows the value of an option using binomial option pricing. Can also show raw data and provide a graph with predicted underlying asset ending values. The binomial options model calculates how big an up step or down step in the next time period will likely be. Then it creates a tree doing this at each period. The end results of this is a tree with possible asset values at each "step". For our calculations we use a day as our "step" time period. We then take all of the expected values at the finishing date and use this to begin a tree of option values at each step. The ending results is the value of the option today.
```
jmaslek, line 5: Provide more detail about the calculation for the probability of up vs down and whatnot.

gamestonk_terminal/stocks/options/yfinance_view.py
```diff
@@ -662,3 +666,238 @@ def risk_neutral_vals(
...
+    ws["A1"] = "U"
+    ws["B1"] = up
+    ws["A2"] = "D"
+    ws["B2"] = 1 / up
+    ws["D1"] = "q"
+    ws["E1"] = prob_up
+    ws["D2"] = "1-q"
+    ws["E2"] = 1 - prob_up
+    ws["D3"] = "Discount"
+    ws["E3"] = discount
```
jmaslek, line 776: More descript names here as well.  When I open up an xls and see q (which is usually denoted as 1-probability of success), I have no idea what it means

website/content/stocks/options/binom/_index.md
colin99d: Improved description

jmaslek (approved):

PR #1028, opened 2021-12-06
Adds commands for crypto withdrawal fees

jose-donato:
Adds commands to crypto `ov` (overview) menu from [withdrawal fees source](https://withdrawalfees.com/)

* Check overall withdrawal fees of cryptocurrencies ordered by market cap
* Search specific coin withdrawal fees per exchange
* Check exchange withdrawal fees

image

image


jose-donato:
Merged `wfc` and `wfcstats` command:
image

DidierRLopes (changes requested):
Looks like the name is made of COINcoinname. 

Can we have something like BTC-Bitcoin or so?
Screenshot 2021-12-06 at 17 37 31

Can we replace "nan" by an empty space?
Screenshot 2021-12-06 at 17 40 27

DidierRLopes:
```
(✨) (crypto)>(ov)> wfe -h
usage: wf [--export {csv,json,xlsx}] [-h]
```
the usage should match name of command

Can we have the menu like this?
```
WithdrawalFees:
    wf                 overall withdrawal fees
    ewf               overall exchange withdrawal fees
    wfpe             overal withdrawal fees per exchange
```

DidierRLopes:
As a default I would do 10/20 instead of 100. 100 just seems too much data to be shown!

DidierRLopes:
Can we combine `wfc` and `wfe` given they have the same index? We can just add the minimal withdraw amount in the end I think

jose-donato:
> Can we combine `wfc` and `wfe` given they have the same index? We can just add the minimal withdraw amount in the end I think

the index is different. wfc is a symbol (bitcoin, ethereum,...) and wfe is an exchange (binance, crypto.com, ...)

DidierRLopes (changes requested):
I think we are moving towards having a better user experience by doing 2 things:
- Auto-complete
- difflib to replace a command if user had a typo or if less sure suggest "did you mean"

Screenshot 2021-12-06 at 18 48 04

Have a look at this comment https://github.com/GamestonkTerminal/GamestonkTerminal/pull/1016#discussion_r763273073 and I think we can implement some of these here!

gamestonk_terminal/cryptocurrency/overview/withdrawalfees_model.py
```diff
@@ -0,0 +1,150 @@
...
+from gamestonk_terminal.helper_funcs import get_user_agent
+
+
+def get_overall_withdrawal_fees(top: int = 100) -> pd.DataFrame:
+    """Scrapes top coins withdrawal fees
+    [Source: https://withdrawalfees.com/]
+
+    Parameters
+    ----------
+    top: int
```
piiq, line 17: This threshold arg is called "limit" in sia. We might want to check how it's named in other menus and make it consistent

gamestonk_terminal/cryptocurrency/overview/withdrawalfees_model.py
```diff
@@ -0,0 +1,150 @@
...
+        Coin to check withdrawal fees. By default bitcoin
+    Returns
+    -------
+    List:
+        - str:              Overall statistics (exchanges, lowest, average and median)
+        - pandas.DataFrame: Exchange, Withdrawal Fee, Minimum Withdrawal Amount
+    """
+    crypto_withdrawal_fees = BeautifulSoup(
+        requests.get(
+            f"https://withdrawalfees.com/coins/{symbol}",
```
piiq, line 116: Check this out. The source uses a symbol, but when we issue a command to the terminal we use a full name and not the symbol
image

To me personally the symbol is preferable, although sometimes I don't know the symbol for a coin and i need to google it.

Is there an option to make the command accept either symbol or a full name? This is more of a question than a proposal

piiq (commented):
It would be super nice to have autocompletion for coins here. Is it possible? In addition to this I have left some comments around the code.


P.S. The data is funny
image

jose-donato:
Thanks for the review @piiq

Working on that, next push will have coins auto completion for `wfpe` command 👍

gamestonk_terminal/cryptocurrency/overview/withdrawalfees_model.py
jose-donato: Yea, the autocompletion will help here.

I agree that accepting both the symbol or full name would be beneficial not only for this but across some more crypto commands! It will need some work but it is possible

piiq: why doesn't github let me choose other emojis as a reaction? 🤷‍♂️

gamestonk_terminal/cryptocurrency/overview/overview_controller.py
```diff
@@ -1772,9 +1802,16 @@ def menu():
     while True:
         # Get input command from user
         if session and gtff.USE_PROMPT_TOOLKIT:
-            completer = NestedCompleter.from_nested_dict(
-                {c: None for c in overview_controller.CHOICES}
-            )
+            choices: dict = {c: {} for c in overview_controller.CHOICES}
+
+            choices["wfpe"]["-c"] = {
```
DidierRLopes, line 1807: @piiq @jmaslek we need to step up our UX overall on the terminal using this and difflib!

We need to go through our controllers 1 by 1 and tackle this

DidierRLopes (changes requested):
Ok, last thing I ask!

Can you add to the wfpe command the possibility to do:
```
wfpe bitcoin
```

since the `-c` flag is the only one (or even if it was the most important) we want user to be able to call this command faster. Thus, you need to do the following:
```
if other_args:
  if "-" not in other_args[0]:
      other_args.insert(0, "-n")
```
And since this will be possible then you can also add it as a capability of the auto-complete!

DidierRLopes (approved):

DidierRLopes:
Super!!

PR #1033, opened 2021-12-07
Fix finviz news

DidierRLopes:
@Chavithra


jmaslek:
Do we want to implement this function somewhere?

jmaslek (approved):

DidierRLopes:
> Do we want to implement this function somewhere?

Well, we could potentially implement on the `news` and add an argument of `--source` / `-s`

But didn't consider it a priority

PR #1034, opened 2021-12-07
refactor load command

DidierRLopes:
Improves load command

### DIFF
--- a/gamestonk_terminal/stocks/stocks_helper.py
+++ b/gamestonk_terminal/stocks/stocks_helper.py
@@ -123,12 +123,12 @@ def load(
     ----------
     ticker: str
         Ticker to get data
-    start: str
-        Start date to get data from with format YYYY-MM-DD
+    start: datetime
+        Start date to get data from with
     interval: int
         Interval (in minutes) to get data 1, 5, 15, 30, 60 or 1440
-    end: str
-        End date to get data from with format YYYY-MM-DD
+    end: datetime
+        End date to get data from with
     prepost: bool
         Pre and After hours data
     source: str
@@ -230,7 +230,7 @@ def load(
         else:
             s_start = start
 
-        s_interval = str(interval) + "min"
+        s_interval = f"{interval}min"
 
     else:
         if source == "av":
@@ -239,7 +239,7 @@ def load(
             df_stock_candidate, _ = ts.get_intraday(
                 symbol=ticker,
                 outputsize="full",
-                interval=str(interval) + "min",
+                interval=f"{interval}min",
             )
 
             df_stock_candidate.columns = [
--- a/gamestonk_terminal/stocks/stocks_helper.py
+++ b/gamestonk_terminal/stocks/stocks_helper.py
@@ -2,7 +2,6 @@
 __docformat__ = "numpy"
 import argparse
 import json
-import warnings
 from datetime import datetime, timedelta
 from typing import List, Union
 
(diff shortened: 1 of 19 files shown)


DidierRLopes:
@Chavithra this is failing due to tests.

I've improved the `load` function, can you let me know how I should handle it from the tests perspective?

Chavithra:
Changed a test this test:
- `tests/gamestonk_terminal/stocks/due_diligence/test_business_insider_view.py::test_price_target_from_analysts_plt`

Removed this test which I think doesn't make sense with the new code :
- `tests/gamestonk_terminal/stocks/due_diligence/test_dd_controller.py::test_call_load_no_args`

jmaslek:
Maybe not related to this specific pr, but `ta` should have a `load`

jmaslek:
When trying a foreign exchange:

```
(✨) (stocks)> load glxy.to
Loading Daily GLXY.TO stock with starting period 2020-12-07 for analysis.

not enough values to unpack (expected 2, got 1) 
```

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -140,206 +86,160 @@ def search(
...
-        Ticker
-    s_start : str
-        Start date
-    s_interval : str
-        Interval to get data for
-    df_stock : pd.DataFrame
-        Preloaded dataframe
+    ticker: str
+        Ticker to get data
+    start: str
```
jmaslek, line 126: You have this type hinted as datetime

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -352,79 +252,66 @@ def load(
...
+            if df_stock_candidate.index[0] > pd.to_datetime(start):
                 s_start = df_stock_candidate.index[0]
             else:
-                s_start = ns_parser.s_start_date
+                s_start = start
 
-        elif ns_parser.source == "yf":
-            s_int = str(ns_parser.n_interval) + "m"
+        elif source == "yf":
+            s_int = str(interval) + "m"
```
jmaslek, line 277: f string?

jmaslek:
looks good.  works on all the menus for me

DidierRLopes:
> looks good. works on all the menus for me

should be good now

jmaslek:
Still getting
```
(✨) (stocks)> load glxy.to
Loading Daily GLXY.TO stock with starting period 2020-12-07 for analysis.

not enough values to unpack (expected 2, got 1) 
```

piiq (commented):
It looks like a lot of code duplication is being introduces here. Each of the controllers that are touched by this PR has a load function that are really similar. If we are to change/refactor this again it would require making edits in a lot of places. Are we sure about this?

piiq:
> That's fixed.

All international exchanges that I've tried worked as expected.

alibaba in hk
image

blackberry in toronto
image

yandex in moscow
image

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -26,66 +25,13 @@
...
     plot_autoscale,
-    valid_date,
+    try_except,
 )
 
-warnings.simplefilter("ignore")
 # pylint: disable=no-member,too-many-branches,C0302
 
 
-def clear(
```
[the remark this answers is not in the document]
DidierRLopes: We aren't using this anywhere

DidierRLopes:
> It looks like a lot of code duplication is being introduced here. Each of the controllers that are touched by this PR has a load function that are really similar. If we are to change/refactor this again it would require making edits in a lot of places. Are we sure about this?

Yup, readability >> less lines.

Also, if you look into each of these loads they allowed user to specify intraday, interval, source, ... and then what we really wanted was just either a ticker name or a start date time...

And in terms of the usage on the reports now it's much nicer than what it was. Now you don't need to send a list of strings (other_args) which is something we moved away on our main refactoring couple weeks away, and load was just waiting to be refactored similarly

jmaslek (approved):

PR #1041, opened 2021-12-09
Big UX improvement

DidierRLopes:
UX Refactoring CHECKLIST:
- [ ] python scripting feature must be working
- [ ] 1 test per menu should be added that tries all functions in tests/simulator with name `test_menu_submenu.gst`
- [ ] python job should be working with `python terminal.py test/simulator/test_menu_submenu.gst`
- [ ] q/quit/reset/help are changed to cls/q/../exit/h/?/
- [ ] change `(stocks)` > to `/stocks/` and `(stocks)>(disc)>` to `/stocks/disc/`
- [ ] usage of export data on parse_known_args_and_warn rather than adding an argument (EXPORT_..)
- [ ] having difflib always replacing the command selected by the user if cutout is above 0.7
- [ ] when we can specify a limit of rows or so, always use `-l` or `--limit` 
- [ ] always have auto-completion on `choices=` . If it's static do it on the __init__ (see disc_controller) otherwise on the menu() (see sia_controller). Always do it in the most possible complete way:
```
choices["arkord"]["-s"] = {c: None for c in self.arkord_sortby_choices}
choices["arkord"]["--sortby"] = {
    c: None for c in self.arkord_sortby_choices
}
```
because user can do `arkord -s date` or `accord --sortby date`. If the user could do `arkord date` then we would have needed to add
```
choices["arkord"] = {c: None for c in self.arkord_sortby_choices}
```
- [ ] when there is a main argument always have:
```
if other_args and "-" not in other_args[0]:
      other_args.insert(0, "-n")
```
- [ ]  Allow for faster functions executing, e.g. if inside `stocks` menu I can do something like `disc/ugs -n 3/../load tsla`
- [ ]  Allow for faster functions executing through absolute path, e.g. if inside `crypto` menu I can do something like `/stocks/disc/ugs -n 3/../load tsla`
- [ ] show help menu if user goes manually into a menu but if he goes into a menu in scripting mode then no help should be shown
- [ ] These are the main command choices a user can select
```
CHOICES = [
        "cls",
        "h",
        "?",
        "help",
        "q",
        "quit",
        "..",
        "exit",
        "r",
        "reset",
    ]
```
- [ ] If a user goes from stocks/sia to stocks/ca when he presses quit the user must end on stocks and not sia
- [ ] If a user runs /stocks/disc/ugs then he must remain on the disc menu or where the last menu was
- [ ] Remove plt.close("all")


Chavithra:
Is it possible to apply the following command and see if the generated `txt` files are good ?
```python
pytest tests\test_terminal.py --record-mode=rewrite
```

DidierRLopes:
@Chavithra tests 😨  Will skip these for now if that's ok!

Chavithra:
> @Chavithra tests 😨 Will skip these for now if that's ok!

Alright

Chavithra:
Theodore and I discussed about the location of the `gst` files and we picked a solution :

Moving `gst` files from :
- `tests/e2e`

To :
- `scripts`

piiq (approved):
Let's goooooo! 🚀

DidierRLopes:
AMAZING WORK BY EVERYONE!! <3

@piiq @jmaslek @jose-donato

PR #1045, opened 2021-12-09
Discordbot: Adding screener preset

northern-64bit:
Adds screener preset to the discord bot.


jmaslek:
When I try to run, I get a bunch of errors similar to:

```
discord.ext.commands.errors.ExtensionFailed: Extension 'economy.economy_menu' raised an error: ImportError: cannot import name 'feargreed_command' from partially initialized module 'discordbot.economy.feargreed' (most likely due to a circular import) (/Users/james/PycharmProjects/GamestonkTerminal/discordbot/economy/feargreed.py)
```

```
discord.ext.commands.errors.ExtensionFailed: Extension 'stocks.screener.screener_menu' raised an error: ExtensionFailed: Extension 'economy.economy_menu' raised an error: CommandRegistrationError: The command economy.feargreed is already an existing command or alias.
```

I was able to address similar stuff in #1020

discordbot/stocks/screener/financial.py
```diff
@@ -0,0 +1,153 @@
...
+
+        limit = int(limit)
+
+        if limit < 0:
+            raise Exception("Number has to be above 0")
+
+        if ascend == "false" or ascend == "False" or ascend == "FALSE":
+            ascend = False
+        elif ascend == "true" or ascend == "True" or ascend == "TRUE":
+            ascend = True
```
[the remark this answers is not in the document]
northern-64bit: It assumes that descend is default and we currently don't have any system in place for flags, but I'll add it to the discord bot project in order to ensure that it isn't forgotten.

discordbot/stocks/screener/financial.py
```diff
@@ -0,0 +1,155 @@
...
+        else:
+            raise Exception("ascend argument has to be true or false")
+
+        # Output Data
+        df_screen = get_screener_data(
+            preset_loaded=preset,
+            data_type="overview",
+            limit=limit,
+            ascend=ascend,
+            signal=signal,
```
DidierRLopes, line 43: the tests are failing because this signal keyword doesn't exist anymore. For all of them :)

northern-64bit: I'm already fixing it!

northern-64bit: Eh after trying to fix this and testing the changes, I've realized that it doesn't work without the keyword and I only get errors.

DidierRLopes (changes requested):
So, my main 2 points are:

1. We should have a command to view all the presets a user can use, otherwise they don't know what there is.

2. Currently the printing is not great. What if we transposed the data frame and only showed 3-5 companies to fit the screen?

Screenshot 2021-12-12 at 12 54 23

northern-64bit:
> So, my main 2 points are:
> 
>     1. We should have a command to view all the presets a user can use, otherwise they don't know what there is.
I'll add it to the help text.
>     2. Currently the printing is not great. What if we transposed the data frame and only showed 3-5 companies to fit the screen?
I can make it that page one displays at what page every company is and every stock has each on page to create good formatting that also works on phones

DidierRLopes:
> > So, my main 2 points are:
> > ```
> > 1. We should have a command to view all the presets a user can use, otherwise they don't know what there is.
> > ```
> 
> I'll add it to the help text.
> 
> > ```
> > 2. Currently the printing is not great. What if we transposed the data frame and only showed 3-5 companies to fit the screen?
> > ```
> 
> I can make it that page one displays at what page every company is and every stock has each on page to create good formatting that also works on phones

1. I think a command would be even better to be honest as it would be more explicit.
2. Absolutely love that idea and that you think about the phone users :)

DidierRLopes:
@Chavithra do you know what happened here re. tests?

jmaslek (approved):

DidierRLopes (changes requested):
Nice work @northern-64bit . Like a lot what you've done here!

2 comments:
The error message here is not ideal:
Screenshot 2021-12-15 at 16 10 57

The custom presets are not being shown here:
Screenshot 2021-12-15 at 16 11 23

I think this is due to how we point towards the preset location where we use an absolute path. Do you think you can fix this on the main code to use relative path? 

If not don't worry because I can go ahead and do this :)

DidierRLopes (approved):
Amazing work, as usual 🚀

Issue #1044, opened 2021-12-09
[Bug] Stocks > IEX load


Chavithra:
**Describe the bug**
Can't load stocks with source `IEX`.

**To Reproduce**
```bash
(✨) (stocks)> load gme --source iex -s 2020-10-20 -e 2020-12-1
```

**Screenshots**
```
index 0 is out of bounds for axis 0 with size 0 

'NoneType' object has no attribute 'empty'
```

**Desktop (please complete the following information):**

- OS: windows
- Python version : 3.8

**Additional context**
The error seems to come from this file :
- gamestonk_terminal/stocks/stocks_helper.py

The data is received from `IEX`.

But this part seems to empty the DataFrame somehow :
```python
df_stock_candidate = df_stock_candidate[
    (df_stock_candidate.index >= start.strftime("%Y-%m-%d"))
    & (df_stock_candidate.index <= end.strftime("%Y-%m-%d"))
]
```

Chavithra:
The request made to IEX seems to only get last 30 days data :
```python
client.chartDF(ticker)
```

So filtering by date should only work if it's within 30 days range.

jmaslek:
I will fix that to specify start and end 👍

Issue #1042, opened 2021-12-09
[IMPROVE] Stocks > Helper


Chavithra:
**What's the feature that should be improved?**
These two functions seems unused :
 - gamestonk_terminal/stocks/stocks_helper.py::view
 - gamestonk_terminal/stocks/stocks_helper.py::plot_view_stock

Are they used for some purpose that I don't about ?

**Describe how you would like the feature improved**
Removing this functions if they are found to be unnecessary.

jmaslek:
Not that I can think of.  We scratched them in favor of candle

PR #1047, opened 2021-12-09
Clean up unused functions / iex load

jmaslek:
Address #1042  and #1044 

@Chavithra not sure if this will destroy some tests

- Removes the two unused plot functions
- Redefines the iex loading and eliminates intraday for non yf data


Chavithra (changes requested):
Apparently it can't find `s_start` variable in a print statement.

Here is the code I have run :
```python
from datetime import datetime
from gamestonk_terminal.stocks import stocks_helper

stock = stocks_helper.load(
    ticker="TSLA",
    start=datetime.strptime("2021-12-05", "%Y-%m-%d"),
    interval=1440,
)
```

Here is the displayed message :
```
local variable 's_start' referenced before assignment
```

gamestonk_terminal/stocks/stocks_helper.py
```diff
@@ -94,6 +93,7 @@ def load(
     end: datetime = datetime.now(),
     prepost: bool = False,
     source: str = "yf",
+    iexrange: str = "tyd",
```
Chavithra, line 96: I think it should be "ytd"

jmaslek:
> Apparently it can't find `s_start` variable in a print statement.
> 
> Here is the code I have run :
> 
> ```python
> from datetime import datetime
> from gamestonk_terminal.stocks import stocks_helper
> 
> stock = stocks_helper.load(
>     ticker="TSLA",
>     start=datetime.strptime("2021-12-05", "%Y-%m-%d"),
>     interval=1440,
> )
> ```
> 
> Here is the displayed message :
> 
> ```
> local variable 's_start' referenced before assignment
> ```

Fixed.  Had an incorrect tab

Chavithra (approved):
Seems to work, was able to :
- pass the test with `iex daily`
- get `candle graph` for `iex loaded`data`

gamestonk_terminal/stocks/stocks_controller.py
```diff
@@ -284,7 +284,7 @@ def call_load(self, other_args: List[str]):
             "--source",
             action="store",
             dest="source",
-            choices=["yf", "av", "iex"],
+            choices=["yf", "av", "iex"] if "-i" not in other_args else ["yf"],
```
DidierRLopes, line 287: nice one

DidierRLopes (approved):

PR #1049, opened 2021-12-10
UX Upgrade for Crypto controller(s)

jose-donato:
this is a WIP. expect more commits
UX Refactoring CHECKLIST:
- [x] python scripting feature must be working
- [x] 1 test per menu should be added that tries all functions in tests/simulator with name `test_menu_submenu.gst`
- [x] python job should be working with `python terminal.py test/simulator/test_menu_submenu.gst`
- [x] q/quit/reset/help are changed to cls/q/../exit/h/?/
- [x] change `(stocks)` > to `/stocks/` and `(stocks)>(disc)>` to `/stocks/disc/`
- [x] usage of export data on parse_known_args_and_warn rather than adding an argument (EXPORT_..)
- [x] having difflib always replacing the command selected by the user if cutout is above 0.7
- [x] when we can specify a limit of rows or so, always use `-l` or `--limit` 
- [x] always have auto-completion on `choices=` . If it's static do it on the __init__ (see disc_controller) otherwise on the menu() (see sia_controller). Always do it in the most possible complete way:
```
choices["arkord"]["-s"] = {c: None for c in self.arkord_sortby_choices}
choices["arkord"]["--sortby"] = {
    c: None for c in self.arkord_sortby_choices
}
```
because user can do `arkord -s date` or `accord --sortby date`. If the user could do `arkord date` then we would have needed to add
```
choices["arkord"] = {c: None for c in self.arkord_sortby_choices}
```
- [x] when there is a main argument always have:
```
if other_args and "-" not in other_args[0]:
      other_args.insert(0, "-n")
```
- [x]  Allow for faster functions executing, e.g. if inside `stocks` menu I can do something like `disc/ugs -n 3/../load tsla`
- [x]  Allow for faster functions executing through absolute path, e.g. if inside `crypto` menu I can do something like `/stocks/disc/ugs -n 3/../load tsla`
- [x] show help menu if user goes manually into a menu but if he goes into a menu in scripting mode then no help should be shown
- [x] These are the main command choices a user can select
```
CHOICES = [
        "cls",
        "cd",
        "h",
        "?",
        "help",
        "q",
        "quit",
        "..",
        "exit",
        "r",
        "reset",
    ]
```
- [x] If a user goes from stocks/sia to stocks/ca when he presses quit the user must end on stocks and not sia
- [x] If a user runs /stocks/disc/ugs then he must remain on the disc menu or where the last menu was
- [x] Remove plt.close("all")


DidierRLopes:
you can change this PR for "Draft Pull request" which basically means WIP :)

jose-donato:
checklist `disc_controller.py`: 
- [ ] remove export arg in each function
- [ ] add export int to `parse_known_args_and_warn` in each function
- [ ] update `top` to `limit` in each function
- [ ] update ns_parser validation in each function
- [ ] `return self.queue if len(self.queue) > 0 else []` in each function
- [ ] update commands + their functions
- [ ] update menu
- [ ] add `queue` to constructor

jose-donato:
Main refactor done across all crypto controllers

DidierRLopes (approved):
Amazing work!!

PR #1053, opened 2021-12-11
Dividend History Chart

deeleeramone:
Adds line graph to stocks-->fa-->divs

### DIFF
--- a/gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
+++ b/gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
@@ -77,7 +77,6 @@ class FundamentalAnalysisController:
 
     def __init__(self, ticker: str, start: str, interval: str, suffix: str = ""):
         """Constructor
-
         Parameters
         ----------
         ticker : str
@@ -109,7 +108,6 @@ Fundamental Analysis:
     q             quit this menu, and shows back to main menu
     quit          quit to abandon program
     load          load a new ticker
-
 Ticker: {self.ticker}
 {f"Note that only Yahoo Finance currently supports foreign exchanges{Style.DIM}{newline}" if self.suffix else ""}
     data          fundamental and technical data of company [FinViz]
@@ -147,7 +145,6 @@ Other Sources:
 
     def switch(self, an_input: str):
         """Process and dispatch input
-
         Returns
         -------
         True, False or None
@@ -460,15 +457,23 @@ Other Sources:
             add_help=False,
             formatter_class=argparse.ArgumentDefaultsHelpFormatter,
             prog="divs",
-            description="Get historical dividends for company",
+            description="Historical dividends for a company",
         )
         parser.add_argument(
             "-n",
             "--num",
             dest="num",
             type=check_positive,
-            default=12,
+            default=20,
             help="Number of previous dividends to show",
+        )
+        parser.add_argument(
+            "-p",
+            "--plot",
+            dest="plot",
+            default=True,
+            action="store_true",
+            help="Plots changes in dividend over time",
         )
         ns_parser = parse_known_args_and_warn(
             parser, other_args, export_allowed=EXPORT_ONLY_RAW_DATA_ALLOWED
@@ -476,7 +481,10 @@ Other Sources:
         if not ns_parser:
             return
         yahoo_finance_view.display_dividends(
-            ticker=self.ticker, num=ns_parser.num, export=ns_parser.export
+            ticker=self.ticker,
+            num=ns_parser.num,
+            plot=ns_parser.plot,
+            export=ns_parser.export,
         )
 
     @try_except
@@ -837,7 +845,6 @@ Other Sources:
 @try_except
 def key_metrics_explained(other_args: List[str]):
     """Key metrics explained
-
     Parameters
     ----------
     other_args : List[str]
@@ -867,7 +874,6 @@ def key_metrics_explained(other_args: List[str]):
 
 def menu(ticker: str, start: str, interval: str, suffix: str = ""):
     """Fundamental Analysis menu
-
     Parameters
     ----------
     ticker : str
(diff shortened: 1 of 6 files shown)


gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
```diff
@@ -460,23 +457,34 @@ def call_divs(self, other_args: List[str]):
...
-            description="Get historical dividends for company",
+            description="Historical dividends for a company",
         )
         parser.add_argument(
             "-n",
             "--num",
             dest="num",
             type=check_positive,
-            default=12,
+            default=20,
```
DidierRLopes, line 467: 20 is way too much as default IMO. Can we keep this to 10?

gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
```diff
@@ -460,23 +457,34 @@ def call_divs(self, other_args: List[str]):
...
             help="Number of previous dividends to show",
         )
+        parser.add_argument(
+            "-p",
+            "--plot",
+            dest="plot",
+            default=True,
```
DidierRLopes, line 474: I think you want default false here

gamestonk_terminal/stocks/fundamental_analysis/yahoo_finance_view.py
```diff
@@ -147,15 +143,16 @@ def display_calendar_earnings(ticker: str):
     print("")
 
 
-def display_dividends(ticker: str, num: int = 12, export: str = ""):
+def display_dividends(ticker: str, num: int = 12, plot: bool = False, export: str = ""):
```
DidierRLopes, line 146: num: int = 10

gamestonk_terminal/stocks/fundamental_analysis/yahoo_finance_view.py
```diff
@@ -165,6 +162,31 @@ def display_dividends(ticker: str, num: int = 12, export: str = ""):
...
+        ax.set_title(f"Dividend History for {ticker}")
+        ax.set_xlim(div_history.index[-1], div_history.index[0])
+        if gtff.USE_ION:
+            plt.ion()
+        fig.tight_layout()
+        plt.show()
+
     div_history.index = pd.to_datetime(div_history.index, format="%Y%m%d").strftime(
         "%Y-%m-%d"
     )
```
DidierRLopes, line 192: let's add an "else" here, in case the user is not plotting

DidierRLopes (changes requested):
@Chavithra do you know why the tests are failing here?

gamestonk_terminal/stocks/fundamental_analysis/yahoo_finance_view.py
deeleeramone: 20 represents five years of dividend history assuming quarterly payments. This is my reasoning for chosing this number.

gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
deeleeramone: 20 represents five years of dividend payments, assuming paid quarterly. If this number is still too high, then multiples of four work better here, IMO.

gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
deeleeramone: IMO, it's more telling to see the visual, gives a clear signal for strength of a dividend paying company.

deeleeramone:
I’m a newbie at this, so I’m not sure how to implement your ‘else’ request. 🙃

gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
DidierRLopes: That is a fair assumption!

Should we go with 12, 3 years?

I just think that every time I tried having 20 as default it was a bit too much. BUT in this case I was wondering if we should have dividends shown in terms of plot by default? and in that case having 20 is completely fine!

DidierRLopes:
@jmaslek What do you think about having the plot being displayed by default? and instead of having a `plot` options having a `raw` option?

jmaslek:
> @jmaslek What do you think about having the plot being displayed by default? and instead of having a `plot` options having a `raw` option?

Not a fan.  Think the numbers are better to look at then a plot for such small amounts.

deeleeramone:
Updated default to reflect

Chavithra:
Hello @deeleeramone, can you try to run the following commands :

```python
rm -r tests/gamestonk_terminal/stocks/fundamental_analysis/txt/test_fa_controller
pytest tests/gamestonk_terminal/stocks/fundamental_analysis/test_fa_controller.py --record-mode=once
```

It should update the expected result for `test_fa_controller`.

gamestonk_terminal/stocks/fundamental_analysis/yahoo_finance_view.py
```diff
@@ -165,6 +162,31 @@ def display_dividends(ticker: str, num: int = 12, export: str = ""):
...
+        )
+        ax.set_xlabel("Date")
+        ax.set_ylabel("Amount ($)")
+        ax.set_title(f"Dividend History for {ticker}")
+        ax.set_xlim(div_history.index[-1], div_history.index[0])
+        if gtff.USE_ION:
+            plt.ion()
+        fig.tight_layout()
+        plt.show()
+
```
DidierRLopes, line 189: Basically the `else:` should come on this line.

You should have something like:
```
if plot:
   do something for plotting
else:
   just show the raw data
```

DidierRLopes: @deeleeramone can you have a look at this so we can merge it?

deeleeramone:
The file has been updated.

DidierRLopes:
> The file has been updated.
> […](#)

It hasn't yet. You need to "add" then "commit" then "push".

DidierRLopes:
@Chavithra same here happening

deeleeramone:
Do I need to run those two commands again for the test?

jmaslek (approved):

DidierRLopes (approved):

Issue #1052, opened 2021-12-11
[Bug] Stocks > Discovery


Chavithra:
Here is a list of potential issues in this module :
- Module : `gamestonk_terminal.stocks.discovery`

1/ `cfg.API_KEY_QUANDL` is used in ` :
- Module : `gamestonk_terminal.stocks.discovery.nasdaq_model`
- Function : `get_retail_tickers`
- Suggestion : replace it with something like `--API_KEY_NASDAQ` ?

2/ `get_article_list` function seems to loop and never return :
- Module : `gamestonk_terminal.stocks.discovery.seeking_alpha_model`
- Function : `get_article_list`
- Code to runned :
```python
from datetime import datetime
from gamestonk_terminal.stocks.discovery import seeking_alpha_model

start_date = datetime.strptime("2020-12-10", "%Y-%m-%d")
seeking_alpha_model.get_article_list(start_date=start_date, num=2)
```

3/ `ratings_view` seems `broken` and `unused` :
- Module : `gamestonk_terminal.stocks.discovery.marketbeat_view`
- Function : `ratings_view`
- Code runned :
```python
from gamestonk_terminal.stocks.discovery import marketbeat_view

marketbeat_view.ratings_view(other_args=["--threshold=1"])
```
- Error generated :
```
ValueError: could not convert string to float: ''
```

4/ `"-ascend"` `argument` seems weird :
- Module : gamestonk_terminal.stocks.discovery.disc_controller
- Function : `call_arkord`
- Suggestion : replace it with  `"--ascend"` ?

jmaslek:
Will look at the last two points tomorrow but regarding the first one:

Nasdaq data link recently acquired quandl (or rebranded or something) and if you look at their api documentation it still calls the quandl package for python and uses the same quandl key.

So until they switch some documentation stuff up the quandl name still makes sense...kinda

jmaslek:
So the get_article_list website needs some human verification:

Im not sure if we should scrap or figure out the js needed to get past it.

Pushing fix for other comments 
<img width="931" alt="Screen Shot 2021-12-13 at 4 43 57 PM" src="https://user-images.githubusercontent.com/18151143/145893822-951e2d14-eb4a-4e53-890e-fa726ab905db.png">

PR #1058, opened 2021-12-13
DISC Cleanup

jmaslek:
Addressed comments in #1052 

I decided to delete market view since it is paywalled

@Chavithra make sure @DidierRLopes is good before merging.

Not sure what to do re: seeking alpha verification


DidierRLopes:
I'm happy with this.

I've tested Seeking Alpha commands and all work. Do you mean this "latest"?

If so I'm happy with removing this argument and just call this function `trending_news` or so.

```
def news(news_type: str, article_id: int, num: int, start_date: datetime, export: str):
    """Prints the latest news article list. [Source: Seeking Alpha]

    Parameters
    ----------
    news_type: str
        Select between 'latest' or 'trending'
    article_id: int
        Article ID. If -1, none is selected
    num: int
        Number of articles to display. Only used if article_id is -1.
    start_date : datetime
        Date from when to get articles dates. Only used if article_id is -1.
    export : str
        Export dataframe data to csv,json,xlsx file
    """
    # User wants to see all latest news
    if article_id == -1:
        if news_type == "latest":
            articles = seeking_alpha_model.get_article_list(start_date, num)
```

DidierRLopes (commented):
lgtm, let's merge once the test is fixed

Chavithra:
This should fix the current issue with tests :
```python
pytest tests/gamestonk_terminal/stocks/government/test_quiverquant_view.py -k test_call_func --record-mode=rewrite
```

DidierRLopes (approved):

Chavithra:
That's a mystery to me because the changes here don't seem to impact  `Quiverquant View`.

One think is sure the generated `txt` files now match the `cassettes` content.

PR #1061, opened 2021-12-16
Workflows + Stocks > Insider

Chavithra:
**Reason**
Removing unused file inside the package :
- `gamestonk_terminal.stocks.insider`


DidierRLopes:
@jose-donato the issue was with the versioning of mypy, Chavi fixed it here!

DidierRLopes (approved):

Chavithra:
I think the error message displayed by `mypy` is misleading.

But there is indeed a typing error.

Will make another PR to fix the typing.

Then will enable new version of mypy (v0.920) inside `github actions`.

PR #1069, opened 2021-12-17
Initial implementation of logging

aia:


gamestonk_terminal/config_terminal.py
```diff
@@ -15,6 +15,30 @@
...
 
+# Logging settings
+
+# 0 - INFO
+# 1 - DEBUG for terminal, INFO for libraries
+# 2 - DEBUG for terminal, DEBUG for libraries
+
+LOGGING_VERBOSITY = 2
+
+if tmp_verbosity := os.getenv("GT_LOGGING_VERBOSITY"):
```
DidierRLopes, line 26: Why not following our other convention here? Like
```
LOGGING_VERBOSITY = os.getenv("GT_LOGGING_VERBOSITY") or "2"
```

gamestonk_terminal/config_terminal.py
```diff
@@ -15,6 +15,30 @@
 # By default the jupyter notebook will be run on port 8888
 PAPERMILL_NOTEBOOK_REPORT_PORT = "8888"
 
+# Logging settings
+
+# 0 - INFO
```
DidierRLopes, line 20: Instead of magic numbers could we have these as consts? Something like?
```
LOGGING_INFO = 0
LOGGING_DBG_TERMINAL_INFO_LIBS = 1
LOGGING_DBG_TERMINAL_DBG_LIBS = 2
```

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
+"""Terminal helper"""
+__docformat__ = "numpy"
+import logging
+import os
+from pathlib import Path
+import sys
+import time
+import uuid
+
+import git
```
DidierRLopes, line 10: We may need to update requirements with this package.

I think we had at some point but I removed it because was giving issues, and manage to go without it.

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+def setup_file_logger() -> None:
+    """Setup File Logger"""
+
+    file_path = Path(__file__)
+    logger.debug("Parent dir: %s", file_path.parent.parent.absolute())
+    log_dir = file_path.parent.parent.absolute().joinpath("logs")
+    logger.debug("Future logdir: %s", log_dir)
+
+    if not os.path.isdir(log_dir.absolute()):
+        logger.debug("Logdir does not exist")
```
DidierRLopes, line 48:
```
logger.debug("Logdir does not exist. Creating one...")
```

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+        os.mkdir(log_dir.absolute())
+
+    log_id = log_dir.absolute().joinpath(".logid")
+
+    if not os.path.isfile(log_id.absolute()):
+        logger.debug("Log ID does not exist: %s", log_id.absolute())
```
DidierRLopes, line 54: Can we use f-string convention here?

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+        cfg.LOGGING_ID = f"{uuid.uuid4()}"
+        with open(log_id.absolute(), "a") as a_file:
+            a_file.write(f"{cfg.LOGGING_ID}\n")
+    else:
+        logger.debug("Log ID exists: %s", log_id.absolute())
```
DidierRLopes, line 59: Can we use f-string convention here?

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+        with open(log_id.absolute()) as a_file:
+            cfg.LOGGING_ID = a_file.readline().rstrip()
+
+    logger.debug("Log id: %s", cfg.LOGGING_ID)
+
+    uuid_log_dir = log_dir.absolute().joinpath(cfg.LOGGING_ID)
+
+    logger.debug("Current log dir: %s", uuid_log_dir)
```
DidierRLopes, line 67: Can we use f-string convention here?

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+
+    if not os.path.isdir(uuid_log_dir.absolute()):
+        logger.debug("UUID log dir does not exist: %s", uuid_log_dir.absolute())
```
DidierRLopes, line 70: Can we use f-string convention here?

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+        os.mkdir(uuid_log_dir.absolute())
+
+    cfg.LOGGING_FILE = uuid_log_dir.absolute().joinpath(f"{int(time.time())}.log")  # type: ignore
+
+    logger.debug("Current log file: %s", cfg.LOGGING_FILE)
```
DidierRLopes, line 75: Can we use f-string convention here?

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+    """Get Commit Short Hash"""
+
+    file_path = Path(__file__)
+    git_dir = file_path.parent.parent.absolute().joinpath(".git")
+
+    if os.path.isdir(git_dir.absolute()):
+        repo = git.Repo(search_parent_directories=True)
+        sha = repo.head.object.hexsha
+        short_sha = repo.git.rev_parse(sha, short=8)
+        cfg.LOGGING_VERSION = f"sha:{short_sha}"
```
DidierRLopes, line 139: Should we have an else condition here? Although that shouldn't happen

gamestonk_terminal/loggers.py
```diff
@@ -0,0 +1,181 @@
...
+            logging.getLogger().addHandler(handler)
+        elif a_handler == "noop":
+            handler = logging.NullHandler()  # type: ignore
+            formatter = CustomFormatterWithExceptions(fmt=LOGFORMAT, datefmt=DATEFORMAT)
+            handler.setFormatter(formatter)
+            logging.getLogger().addHandler(handler)
+        elif a_handler == "file":
+            setup_file_logger()
+        else:
+            logger.debug("Unknown loghandler")
```
DidierRLopes, line 175: To avoid code repetition we can have
```
for a_handler in cfg.LOGGING_HANDLERS.split(","):
         if a_handler == "stdout":
             handler = logging.StreamHandler(sys.stdout)
         elif a_handler == "stderr":
             handler = logging.StreamHandler(sys.stderr)
         elif a_handler == "noop":
             handler = logging.NullHandler()  # type: ignore
         elif a_handler == "file":
             setup_file_logger()
         else:
             logger.debug("Unknown loghandler")

        if a_handler in ("stdout", "stderr", "noop"):
             formatter = CustomFormatterWithExceptions(fmt=LOGFORMAT, datefmt=DATEFORMAT)
             handler.setFormatter(formatter)
             logging.getLogger().addHandler(handler)
```

gamestonk_terminal/menu.py
```diff
@@ -1,17 +1,21 @@
...
+
 
 def inputhook(inputhook_context):
     while not inputhook_context.input_is_ready():
         try:
             pyplot.pause(0.1)
         # pylint: disable=unused-variable
-        except Exception:  # noqa: F841
+        except Exception as exp:  # noqa: F841
+            logger.exception("%s", type(exp).__name__)
```
DidierRLopes, line 18: f-string here

gamestonk_terminal/menu.py
```diff
@@ -1,17 +1,21 @@
...
 from prompt_toolkit.history import FileHistory
 
+logger = logging.getLogger(__name__)
+
 
 def inputhook(inputhook_context):
     while not inputhook_context.input_is_ready():
         try:
             pyplot.pause(0.1)
         # pylint: disable=unused-variable
```
DidierRLopes, line 16: This probably can be removed as the variable `exp`  is now used

gamestonk_terminal/menu.py
```diff
@@ -23,6 +27,7 @@ def inputhook(inputhook_context):
     set_eventloop_with_inputhook(inputhook)
 # pylint: disable=unused-variable
 except Exception as e:  # noqa: F841
+    logger.exception("%s", type(e).__name__)
```
DidierRLopes, line 30: f-string

gamestonk_terminal/terminal_helper.py
```diff
@@ -389,6 +389,7 @@ def bootup():
             # pylint: disable=E1101
             sys.stdout.reconfigure(encoding="utf-8")
     except Exception as e:
+        logger.exception("%s", type(e).__name__)
```
DidierRLopes, line 392: f-string

gamestonk_terminal/terminal_helper.py
```diff
@@ -402,6 +403,7 @@ def bootup():
         try:
             thought.get_thought_of_the_day()
         except Exception as e:
+            logger.exception("%s", type(e).__name__)
```
DidierRLopes, line 406: f-string

DidierRLopes (commented):
minor stylistic comments!

Super excited about this 🚀

gamestonk_terminal/config_terminal.py
piiq, line 26: This is the first time I have seen a walrus operator in the wild lol. Since we have "officially" deprecated python 3.7 it's ok=)

piiq (changes requested):
The f-strings are more modern and are the way strings are formatted across the project. This is the only styling change request I have.

I would suggest lowering the default logging level to info to prevent overwhelming of the logs folder for active users. We don't yet have "log rotation" or log removal over time implemented here, so this might be a better default choice.

gamestonk_terminal/config_terminal.py
aia: 1. We don't want to introduce too many variables
2. Levels are ordered DEBUG+LIB > DEBUG > INFO. We want to follow logging convention - https://docs.python.org/3/howto/logging.html#logging-levels.

gamestonk_terminal/loggers.py
aia: Pylint explicitly advises against using f-strings with logging
https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html

gamestonk_terminal/menu.py
aia: Same comment logging should be done more explicitly per pylint - https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html

gamestonk_terminal/terminal_helper.py
aia: https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html

gamestonk_terminal/terminal_helper.py
aia: https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html

gamestonk_terminal/config_terminal.py
```diff
@@ -15,6 +15,30 @@
...
+# 1 - DEBUG for terminal, INFO for libraries
+# 2 - DEBUG for terminal, DEBUG for libraries
+
+LOGGING_VERBOSITY = 2
```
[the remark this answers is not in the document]
aia: We need 3 levels.

gamestonk_terminal/helper_funcs.py
```diff
@@ -186,6 +189,7 @@ def plot_view_stock(df: pd.DataFrame, symbol: str, interval: str):
         print(
             "Encountered an error trying to open a chart window. Check your X server configuration."
         )
+        logging.exception("%s", type(e).__name__)
```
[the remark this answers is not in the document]
aia: No

gamestonk_terminal/helper_funcs.py
```diff
@@ -784,7 +788,7 @@ def inner(*args, **kwargs):
         try:
             return f(*args, **kwargs)
         except Exception as e:
-            print(e, "\n")
+            logger.exception("%s", type(e).__name__)
```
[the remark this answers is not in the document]
aia: Prints are not needed. Either you are logging into a file and do not want to see extra, or you are logging into stdout and you will see logs.

gamestonk_terminal/menu.py
aia: > The f-strings are more modern and are the way strings are formatted across the project. This is the only styling change request I have.

https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html

Logging should be done without f-strings.

> I would suggest lowering the default logging level to info to prevent overwhelming of the logs folder for active users. We don't yet have "log rotation" or log removal over time implemented here, so this might be a better default choice.

Default logging goes into a file, the user does not see logging output.

gamestonk_terminal/menu.py
aia: https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html

gamestonk_terminal/config_terminal.py
aia: You probably don't have pylint enabled. If ":" is removed, pylint reports a warning of tmp_verbosity being undefined.

gamestonk_terminal/loggers.py
aia: We still have it:

```
$ grep git requirements.txt
gitdb==4.0.9; python_version >= "3.7"
gitpython==3.1.24; python_version >= "3.7"
```

gamestonk_terminal/loggers.py
aia: Fixed

gamestonk_terminal/config_terminal.py
aia: I am converting to int. Conversion to int may blow up.

gamestonk_terminal/menu.py
aia: > This probably can be removed as the variable `exp` is now used

Fixed

DidierRLopes: Very interesting. Thought you did it from force of habit or so... The reason is for backward compatibility and performance apparently, https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles.

The only other thing I'd say is should we let the user know through a message about the logging? We can couple this to that condition when the `logs` folder is created. Basically saying that from now on we will have logs so we can improve the usage of the terminal, and they can change the logging level allowed in the `config_terminal`.

gamestonk_terminal/loggers.py
piiq: Wow

gamestonk_terminal/menu.py
aia: > Very interesting. Thought you did it from force of habit or so... The reason is for backward compatibility and performance apparently, https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles.

My habit is to use f-strings.

> The only other thing I'd say is should we let the user know through a message about the logging? We can couple this to that condition when the `logs` folder is created. Basically saying that from now on we will have logs so we can improve the usage of the terminal, and they can change the logging level allowed in the `config_terminal`.

This is just a feature just like anything else. We add it to the docs.

gamestonk_terminal/loggers.py
aia: That looks less clear to me.

aia:
> minor stylistic comments!

43 comments on a 200-line pull request...

Why?

gamestonk_terminal/menu.py
piiq: TIL about lazy string formatting for logs

piiq:
> 43 comments on a 200-line pull request...
>
> Why?

Aaaahahahhahahaha

piiq (approved):

piiq:
> 43 comments on a 200-line pull request...
> 
> Why?

Because you don't chat with us on discord and we miss you

DidierRLopes (approved):
Because this is how we review at GST, every line of code matters! Also, 35/43 were f-strings 😂

PR #1076, opened 2021-12-20
Upgrades for the UX upgrade for the fa and fa/fmp menus

piiq:
- replaced --num with --limit
- excluded unreliable/non-working functions from the choices list
- improved docstring compliance with numpy convention
- added test scenarios in .gst files

### DIFF
--- a/gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
+++ b/gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
@@ -58,6 +58,8 @@ class FundamentalAnalysisController:
     CHOICES_COMMANDS = [
         "load",
         "analysis",
+        "score",
+        "warnings",
         "data",
         "income",
         "balance",
@@ -75,14 +77,12 @@ class FundamentalAnalysisController:
         "balance",
         "cash",
         "divs",
+        "earnings",
+        "fraud",
     ]
     # TODO: These commands either don't work or work inconsistently.
     # They have been excluded from the CHOICES list until they are fixed:
-    # "score",
-    # "fraud",
     # "dcf",
-    # "warnings",
-    # "earnings",
 
     CHOICES_MENUS = [
         "fmp",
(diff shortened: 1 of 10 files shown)


gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
```diff
@@ -222,32 +227,32 @@ def switch(self, an_input: str):
...
 
     def call_help(self, _):
-        """Process help command"""
+        """Process help command."""
         self.print_help()
         return self.queue
 
     def call_quit(self, _):
-        """Process quit menu command"""
+        """Process quit menu command."""
```
DidierRLopes, line 247: I don't think we need to add a "." here, we haven't been doing it anywhere else thus far I think

DidierRLopes (commented):
If the function doesn't work only for some of the tickers, I recommend for us to still have it and just display an error saying that it isn't available for that particular ticker.

piiq:
> If the function doesn't work only for some of the tickers, I recommend for us to still have it and just display an error saying that it isn't available for that particular ticker.

I'll revise this a bit later. I've figured out that the errors are most likely due to the limits of the alphvantage free tier api (5 requests per minute)
I'd like to add explicit handling of this situation so that the user is informed what exactly went wrong.

gamestonk_terminal/stocks/fundamental_analysis/fa_controller.py
piiq: There is a linter for docstrings that allows to check compliance to the numpy docstring convention. I have enabled it for the editor and will add it to the pre-commit hooks in a separate PR so that we can be in compliance with the best practices.

piiq: Here's what i'm talking about https://github.com/PyCQA/pydocstyle

piiq:
> If the function doesn't work only for some of the tickers, I recommend for us to still have it and just display an error saying that it isn't available for that particular ticker.

Brought most of them back (apart from dcf)

PR #1078, opened 2021-12-21
improve usage command

DidierRLopes:


jmaslek (approved):
Should this be the main "help" and not just usage?

DidierRLopes:
> Should this be the main "help" and not just usage?

What do you mean?

jmaslek:
> What do you mean?

So `help` on the main terminal message shows the usage and all the contexts, etc... Instead of having to do the usage separately.

The reasoning being we are on the 'main menu' so I would expect help to show everything and not need 2 'help' commands.

piiq:
> Should this be the main "help" and not just usage?

It can be incorporated into the main `help` then we can ditch the `usage`

gamestonk_terminal/terminal_helper.py
```diff
@@ -382,16 +382,18 @@ def usage_instructions():
...
+    cls             clear the screen
+    help / h / ?    help menu
+    quit / q / ..   quit this menu and go one menu above
+    exit            exit the terminal
+    reset / r       reset the terminal and reload configs from the current location
 
 Multiple jobs queue (where each '/' denotes a new command). E.g.
     /stocks $ disc/ugs -n 3/../load tsla/candle
 
+If you want to jump from crypto/ta to stocks you can use an absolute path. E.g.
```
piiq, line 394: `If you want to jump from crypto/ta to stocks you can use an absolute path that starts with a slash (/). E.g.`

gamestonk_terminal/terminal_helper.py
```diff
@@ -382,16 +382,18 @@ def usage_instructions():
...
+    /crypto/ta $ /stocks
```
piiq, line 395: Do you want to add the stars emoji like the terminal prints? Ir does it look so-so in the help printout?

piiq (approved):
I left a couple comments in the code, but they are just food for thought and non-blocking

gamestonk_terminal/terminal_helper.py
DidierRLopes: Done!

gamestonk_terminal/terminal_helper.py
DidierRLopes: Yea, I think formatting can be a bit weird because from what I saw from some users they don't even see the stars :/

DidierRLopes:
> Should this be the main "help" and not just usage?

Mainly because the usage shows the example of absolute and relative path and you can't put that on the main menu... Otherwise I would agree with this.

What do you think? @piiq

gamestonk_terminal/terminal_helper.py
piiq: Windows eh..

piiq:
I'd make it into the main help. Keep it simpler. Anywhere in the terminal you have a help and that's it. The root help will show some additional usage instructions. Nothing wrong with that

DidierRLopes:
Screenshot 2021-12-21 at 22 13 21

So something like this?

I'm not a fan, but if you and @jmaslek like it, let's go for it!

jmaslek:
If this makes sense, I would do 2-3-1

image

DidierRLopes:
Screenshot 2021-12-21 at 22 34 28

Will do. 

PS: @piiq I don't like the flair because it sounds confusing, looks like something happened there.

PR #1090, opened 2021-12-23
fix and improve options menu

DidierRLopes:


discordbot/stocks/options/options_menu.py
```diff
@@ -74,202 +73,95 @@ async def oi(
...
+        The user can then select a reaction to trigger a command.
+        """
+
         if cfg.DEBUG:
             print(f"!stocks.opt {ticker}")
-        cols_temp = []
-        cols = []
 
-        if ticker == "":
+        if not ticker:
```
piiq, line 88: I have hard times understanding some pythonic patterns
image

piiq (approved):

piiq:
Merged with playground without conflicts

discordbot/stocks/options/options_menu.py
DidierRLopes: @piiq so when you do `a = ""`, `a` becomes a string. So when you compare `a` with `None` and `False` those are always false because you are comparing a string with a different variable type, in this case, the special type None and the boolean False. "_The “is keyword” is used to test whether two variables belong to the same object._"

However, when you do `not a` what happens is that the `not` makes it so you do the following `bool(a)` which in turn because it's empty by convention is `False`, and therefore returns a `True`.

Note:
```
bool(a) is False -> True
not a is True -> True
a is False -> False
a is None -> False
a is 0 -> False
```

piiq: image
right

DidierRLopes: Screenshot 2021-12-23 at 17 05 39

piiq: yep

PR #1103, opened 2021-12-27
ETF context new and improved

DidierRLopes:
This PR adds a:
- load to ETF menu
- allows to see summary of ETF
- allows to see sector holdings of ETF
- a screener menu
- technical analysis
- prediction techniques
- allows to take holdings to comparison analysis


jmaslek:
finally getting around to this :)

jmaslek:
I don't understand the difference between `ln` (which my math background is already counting this command as taken :P)  and `ls`.  What is a situation I would use the latter over the former?

jmaslek:
Also looking at the menu: you deviate from how you require the other menus, ie
```
finance database:
   f1
   f2
StockAnalysis:
  g1 
  g2
```

jmaslek:
Functions:  I would name chold -> holdings and should sector.

jmaslek:
shold:  Should either include the percents on the chart or display them by default (there are only 11 to show, so doing so always won't be a crazy display on the terminal).  Also I don't get what the -m flag is?  If I do `shold --raw` it shows all 11 sectors (even when the percent is 2.5 which is less than the default).

jmaslek:
summary formats the string weird:
```
(✨) /etf/ $ summary
The investment seeks to provide investment results that, before expenses, correspond generally to the price and yield performance of the S&P 500Â® Index.
 The Trust seeks to achieve its investment objective by holding a portfolio of the common stocks that are included in the index (the âPortfolioâ), with the weight of each stock in the Portfolio substantially corresponding to the weight of such stock in the index. 

```

jmaslek:
title on candle  should not be stock:
Screen Shot 2021-12-29 at 2 20 03 PM

jmaslek:
Should `pir` and `compare`  be split with a newline in the menu since they allow for a variable input?  Also compare doesn't have a source :P

jmaslek:
I would argue that the wsj functions should be in a disc submenu - not screener.

jmaslek:
`ta` has a rogue print
Screen Shot 2021-12-29 at 2 25 31 PM

gamestonk_terminal/etf/financedatabase_model.py
```diff
@@ -0,0 +1,73 @@
...
+    """Return a selection of ETFs based on name filtered by total assets. [Source: Finance Database]
+
+    Parameters
+    ----------
+    name: str
+        Search by name to find ETFs matching the criteria.
+
+    Returns
+    ----------
+    df : pd.DataFrame
```
jmaslek, line 18: technically named data

gamestonk_terminal/etf/financedatabase_model.py
```diff
@@ -0,0 +1,73 @@
...
+    """Return a selection of ETFs based on description filtered by total assets. [Source: Finance Database]
+
+    Parameters
+    ----------
+    description: str
+        Search by description to find ETFs matching the criteria.
+
+    Returns
+    ----------
+    df : pd.DataFrame
```
jmaslek, line 37: same

gamestonk_terminal/etf/financedatabase_model.py
```diff
@@ -0,0 +1,73 @@
...
+    """Return a selection of ETFs based on category filtered by total assets. [Source: Finance Database]
+
+    Parameters
+    ----------
+    category: str
+        Search by category to find ETFs matching the criteria.
+
+    Returns
+    ----------
+    df : pd.DataFrame
```
jmaslek, line 56: boop

gamestonk_terminal/etf/financedatabase_view.py
```diff
@@ -1,56 +1,122 @@
...
-        data = fd.select_etfs(category=category, exclude_exchanges=include_exchanges)
+        print(tabulate_data_sorted.iloc[:limit].to_string(), "\n")
+
+    export_data(export, os.path.dirname(os.path.abspath(__file__)), "ld", data)
+
+
+def display_etf_by_category(
+    category: str,
+    limit: int,
+    export: str,
```
jmaslek, line 103: add default=''

gamestonk_terminal/etf/financedatabase_view.py
```diff
@@ -1,56 +1,122 @@
...
+    else:
+        print(tabulate_data_sorted.iloc[:limit].to_string(), "\n")
+
+    export_data(export, os.path.dirname(os.path.abspath(__file__)), "ln_fd", data)
+
+
+def display_etf_by_description(
+    description: str,
+    limit: int,
+    export: str,
```
jmaslek, line 59: same.  default export='' everywhere.  this makes life so much easier in Jupyter notebooks so one doesn't;t have to write `function(export=False)`

gamestonk_terminal/etf/technical_analysis/ta_controller.py
```diff
@@ -0,0 +1,1351 @@
...
+        if queue:
+            self.queue = queue
+        else:
+            self.queue = list()
+
+    def print_help(self):
+        """Print help"""
+        help_str = f"""
+
+ETF: {self.etf}
```
jmaslek, line 113: here is the rogue print.  I would think of naming this self.data instead of self.etf.  Looking at self.etf I would expect it to be the ticker.

gamestonk_terminal/etf/yfinance_view.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        main_holdings = holdings[holdings.values > min_pct_to_display].to_dict()[
+            holdings.columns[0]
+        ]
+        if len(main_holdings) < len(holdings):
+            main_holdings["Others"] = 100 - sum(main_holdings.values())
+
+        legend, values = zip(*main_holdings.items())
+
+        plt.subplots(figsize=plot_autoscale(), dpi=PLOT_DPI)
```
jmaslek, line 66: havent we switched to using fig, ax everywhere?

DidierRLopes:
> I don't understand the difference between `ln` (which my math background is already counting this command as taken :P) and `ls`. What is a situation I would use the latter over the former?

Screenshot 2021-12-30 at 00 19 20

`ln` is a lookup by name, so that means that you want that word on the name of what you are looking for. `ld` is lookup by description, so you get ETFs that have that word on the description but not necessarily on the name.

DidierRLopes:
> Also looking at the menu: you deviate from how you require the other menus, ie
> 
> ```
> finance database:
>    f1
>    f2
> StockAnalysis:
>   g1 
>   g2
> ```

So I'm using common sense here. I like 
```
FinanceDatabase:
    f1
    f2
    f3
```
when we have at least 3 functions (potentially 2). But think that it looks less noisier with the Source in the end. And think that with the rich package that will become even better... opinions?

DidierRLopes:
> n't get what the -m flag is?

This is actually intentional. As you said there are only 11 sectors, so displaying them is not crazy: Correct in a raw scenario. False in a pie chart. Therefore, the -m only applies to the pie chart in case we want to export the plot.

See what you get if we displayed all

Screenshot 2021-12-30 at 00 27 07

What is the train of thought to display raw on top of plot. I think most of users won't care about that, but only what sectors the ETF is focus towards. And if some want they can use raw on top of that.

DidierRLopes:
> summary formats the string weird:
> 
> ```
> (✨) /etf/ $ summary
> The investment seeks to provide investment results that, before expenses, correspond generally to the price and yield performance of the S&P 500Â® Index.
>  The Trust seeks to achieve its investment objective by holding a portfolio of the common stocks that are included in the index (the â��Portfolioâ��), with the weight of each stock in the Portfolio substantially corresponding to the weight of such stock in the index. 
> ```

Can't do much here as this comes from Yfinance :/

DidierRLopes:
> `ta` has a rogue print
> Screen Shot 2021-12-29 at 2 25 31 PM

LOL, was printing the dataframg instead of the name of the ETF ahah because my window was small never saw it haha

gamestonk_terminal/etf/technical_analysis/ta_controller.py
DidierRLopes: Gotcha

jmaslek:
> `ln` is a lookup by name, so that means that you want that word on the name of what you are looking for. `ld` is lookup by description, so you get ETFs that have that word on the description but not necessarily on the name.
You think there is enough to delineate between those two?  The oil example for description gives `PDBC` which has very little to do with oil (https://www.invesco.com/us/financial-products/etfs/product-detail?audienceType=Investor&ticker=PDBC)

gamestonk_terminal/etf/yfinance_view.py
DidierRLopes: I'm not returning fix, ax because I'm not using them anywhere after. Not sure what you mean here, may be missing something.

jmaslek:
> Can't do much here as this comes from Yfinance :/

Just train an NLP model to find the funky characters ;)

jmaslek:
> when we have at least 3 functions (potentially 2). But think that it looks less noisier with the Source in the end. And think that with the rich package that will become even better... opinions?

That's fair.  Just had to point it out :P.  Im fine with how you have it

DidierRLopes:
> > `ln` is a lookup by name, so that means that you want that word on the name of what you are looking for. `ld` is lookup by description, so you get ETFs that have that word on the description but not necessarily on the name.
> 
> You think there is enough to delineate between those two? The oil example for description gives `PDBC` which has very little to do with oil (https://www.invesco.com/us/financial-products/etfs/product-detail?audienceType=Investor&ticker=PDBC)

Don't hate the player, hate the game 🤣 I added it because FinanceDatabase had it. Should we have just as before, a `search` that can be by description if the user selects such flag?

gamestonk_terminal/etf/yfinance_view.py
jmaslek: We have just been using` fig, ax = plt.subplots()` throughout the code instead of just calling through plt.  Doesn't really mater, but can make life easier if one wants to return an axes object in a bigger plot down the road.

DidierRLopes: I haven't actually be using that because I didn't know. Can we have a call next week to discuss this and Matplotlib overall style? To make sure we have consistency across, and make some important decisions are made and we tackle them all at once. @piiq

jmaslek:
> Don't hate the player, hate the game 🤣 I added it because FinanceDatabase had it. Should we have just as before, a `search` that can be by description if the user selects such flag?

I would personally drop it unless presented a specific use case.... but won't hold the PR hostage lol

DidierRLopes:
> > Don't hate the player, hate the game 🤣 I added it because FinanceDatabase had it. Should we have just as before, a `search` that can be by description if the user selects such flag?
> 
> I would personally drop it unless presented a specific use case.... but won't hold the PR hostage lol

Screenshot 2021-12-30 at 00 57 23

This is the difference. Let's have @JerBouma commenting on this one, as he has played more with this data than us. The options are:
* dropping the `lookup by description`
* having a flag for when we want to lookup by description instead of name
* have 2 commands for lookup

jmaslek:
> This is actually intentional. As you said there are only 11 sectors, so displaying them is not crazy: Correct in a raw scenario. False in a pie chart. Therefore, the -m only applies to the pie chart in case we want to export the plot.

I see.  Missed that

> What is the train of thought to display raw on top of plot. I think most of users won't care about that, but only what sectors the ETF is focus towards. And if some want they can use raw on top of that.

I like numbers.  Looking at a pie chart, the difference between 30 and 40 % is negligible, so seeing the difference is my personal preference.    So I would just show table and plot (was initially envisioning percents in the legend)

DidierRLopes:
> > This is actually intentional. As you said there are only 11 sectors, so displaying them is not crazy: Correct in a raw scenario. False in a pie chart. Therefore, the -m only applies to the pie chart in case we want to export the plot.
> 
> I see. Missed that
> 
> > What is the train of thought to display raw on top of plot. I think most of users won't care about that, but only what sectors the ETF is focus towards. And if some want they can use raw on top of that.
> 
> I like numbers. Looking at a pie chart, the difference between 30 and 40 % is negligible, so seeing the difference is my personal preference. So I would just show table and plot (was initially envisioning percents in the legend)

Yea, adding a percentage on the legend I completely agree with!

gamestonk_terminal/etf/yfinance_view.py
jmaslek: lol yeah we should have a consistent format across.

DidierRLopes:
All has been fixed now @jmaslek

jmaslek (approved):

PR #1106, opened 2021-12-29
Mutual funds

jmaslek:
This adds a simple mutual fund menu using the investpy package (for investing.com) (and some yfinance commands).

After discussing with Didier, I decided to use this menu as a playground for using the open source rich package (https://github.com/willmcgugan/rich) for the terminal outputs as it allows for a more customizable interface.


jmaslek:
@JerBouma feel free to let me know your thoughts as well!

JerBouma:
> @JerBouma feel free to let me know your thoughts as well!

Scrolled through the code, I noticed _info_ requires a fund even if you selected one. Furthermore, the _search_ function doesn't  seem to return anything or does not have the proper message attached to it when it can't find anything.

About the usefulness, Mutual Funds are used by institutional investors. There are departments within investment firms that go through the "Manager Selection". This is a proces where the company looking to invest in a fund requests the LPA (Limited Partner Agreement) and ESG Policy (might be less relevant in US) and it will ask the manager (for example Vanguard) to fill in a Operational, Legal and Fiscal questionnaire. These go quite in-depth and can be 100+ pages long. Therefore, a very useful tool would be the ability to effectively scan these documents with NLP.

As for most retail investors, investing in mutual funds as opposed to ETFs is not so beneficial (in my view) thus I am wondering what the purpose is of this new menu. I can see value for researchers but not so much for potential institutional clients. What is your vision on what they would do with this?

gamestonk_terminal/mutual_funds/mutual_fund_controller.py
```diff
@@ -0,0 +1,546 @@
...
+        else:
+            self.queue = list()
+
+    def print_help(self):
+        """Print help"""
+        fund_string = f"{self.fund_name or None}"
+        fund_string2 = f" ({self.fund_symbol})" if self.fund_symbol else ""
+        fund_string += fund_string2
+        help_str = f"""
+[bold]Mutual Funds[/bold]:
```
DidierRLopes, line 101: We can remove this, if we are to follow the rest

gamestonk_terminal/mutual_funds/mutual_fund_controller.py
```diff
@@ -0,0 +1,546 @@
...
+
+Current Fund:[cyan] {fund_string}[/cyan]
+Current Country: [green]{self.country.title()}[/green][italic]
+
+Investing.com:[/italic]
```
DidierRLopes, line 106: I find this menu a bit confusing, can we have it like this:

```
Investing.com:
    overview    overview of top funds by country
    search        search for Mutual Funds
    load            load historical fund data

Current Fund: Vanguard International Shares Index Fund (0P00007962)

Investing.com:
    info          get fund information
    plot          plot loaded historical fund data

Yahoo Finance:
    sector        sector weightings
    equity        equity holdings
```

Basically the overview having as argument the country makes more sense. UNLESS, you explicitly need the country to load the data.

DidierRLopes (changes requested):
Info, sector, equity are not returning anything. I think something should be output.
Screenshot 2021-12-29 at 13 26 31

Can we replace the ID by the title?
Screenshot 2021-12-29 at 13 29 33

DidierRLopes:
@JerBouma is right about the search function, it's not really working. I had to use the ID I got from overview.

We can add the NLP to our roadmap, definitely! I think more than having something ready we are building foundations for a useful Mutual Funds menu, to show that we can be an investment research for anything. We don't need to focus only on crypto and stocks. In the future we can even have other forms of investing such as real estate or so :)

DidierRLopes:
`reset` is also not working properly:
```
(✨) /fund/ $ reset

resetting...

(✨) /fund/ $ load Vanguard International Shares Index Fund --name
```

gamestonk_terminal/mutual_funds/mutual_fund_controller.py
jmaslek: For whatever reason, the country does need to be specified :(

JerBouma:
@DidierRLopes 

While I agree that it always useful to think about new features, I would also suggest to take a step back and ask ourselves who is going to use these features. Not many institutional investors depend on public data to make their mutual fund decisions, that is what the whole Manager Selection process is for. We could also look at ways we can compliment that area.

jmaslek:
> @JerBouma is right about the search function, it's not really working. I had to use the ID I got from overview.
> 
> 
> 
> We can add the NLP to our roadmap, definitely! I think more than having something ready we are building foundations for a useful Mutual Funds menu, to show that we can be an investment research for anything. We don't need to focus only on crypto and stocks. In the future we can even have other forms of investing such as real estate or so :)

What search inputs are you using?  The search does go by country so is it possible you are using Australia like the above which has no matches?

I can search US easily with keywords like Technology or Finance etc

jmaslek:
> Info, sector, equity are not returning anything. I think something should be output.
> 
> Screenshot 2021-12-29 at 13 26 31
> 
> 
> 
> Can we replace the ID by the title?
> 
> Screenshot 2021-12-29 at 13 29 33
> 
> 

So I think those will not work outside of the US without some for of country code (like .TO)...

Also, why do exceptions no longer print?  They go to the logs but don't indicate any reason the function fails now.

gamestonk_terminal/mutual_funds/mutual_fund_controller.py
DidierRLopes: Got it, so I would reflect that on the display of the menu, having something like: 

```
Investing.com:
    country       set a country for filtering
 
Current Country: United States

Investing.com:
    overview    overview of top funds by country
    search        search for Mutual Funds
    load            load historical fund data

Current Fund: Vanguard International Shares Index Fund (0P00007962)

Investing.com:
    info          get fund information
    plot          plot loaded historical fund data

Yahoo Finance:
    sector        sector weightings
    equity        equity holdings
```

DidierRLopes:
> > @JerBouma is right about the search function, it's not really working. I had to use the ID I got from overview.
> > We can add the NLP to our roadmap, definitely! I think more than having something ready we are building foundations for a useful Mutual Funds menu, to show that we can be an investment research for anything. We don't need to focus only on crypto and stocks. In the future we can even have other forms of investing such as real estate or so :)
> 
> What search inputs are you using? The search does go by country so is it possible you are using Australia like the above which has no matches?
> 
> I can search US easily with keywords like Technology or Finance etc

Humm, but we should comment on that with like "No data found".

DidierRLopes:
> > Info, sector, equity are not returning anything. I think something should be output.
> > Screenshot 2021-12-29 at 13 26 31
> > Can we replace the ID by the title?
> > Screenshot 2021-12-29 at 13 29 33
> 
> So I think those will not work outside of the US without some for of country code (like .TO)...
> 
> Also, why do exceptions no longer print? They go to the logs but don't indicate any reason the function fails now.

Exceptions only print if you have the config set to True. We don't want exceptions to print because we want to protect the user from seeing this. We should be handling all of these ourselves. Exceptions should be the last layer of protection for terminal not to crash, but ideally they should never happen.

DidierRLopes:
> @DidierRLopes
> 
> While I agree that it always useful to think about new features, I would also suggest to take a step back and ask ourselves who is going to use these features. Not many institutional investors depend on public data to make their mutual fund decisions, that is what the whole Manager Selection process is for. We could look at ways we can compliment that area instead.

So, I disagree on this particular situation. If we were adding a feature abc to stocks that is not relevant/used, I think you are right. HOWEVER, in this case this is a new context and having something can help us say that we support mutual funds or even that we will have a strong support of mutual funds in the future. It basically shows what we can be, and allows contributors that are more keen on this area to develop it further. 

E.g. @jose-donato just created a PR for NFTs because there was already a very simple NFT menu. So it's very easy to build on top of something that we already have a concept for.

Also, since this is a new context, if 99% of users will not use it it doesn't matter because it's on this "mutual funds" World, and you don't really care about it. Same as Forex for myself as a user, it's there but I don't use it.

JerBouma:
@DidierRLopes 

Yeah that makes perfect sense. I agree with that!

piiq:
> Same as Forex for myself as a user

Forex has a neat events calendar that affect a specific currency pair. It's useful to see the events for a specific geographical region outside of the US (which is heavily covered in the economy menu)
Check it out (IIRC no API key needed):
image

jmaslek:
Curious what you guys think about the use of `rich` for the printing?  Looking at your green terminal @piiq I am curious if this messes with outputs.

jmaslek:
> `reset` is also not working properly:
> 
> ```
> (✨) /fund/ $ reset
> 
> resetting...
> 
> (✨) /fund/ $ load Vanguard International Shares Index Fund --name
> ```

needed to add the country.

jmaslek:
> Yeah that makes perfect sense. I agree with that!

As an example, one could add your finance database to this menu ;)

DidierRLopes (changes requested):
Screenshot 2021-12-30 at 02 29 54

Can you have the auto-complete to do `country Australia`? Instead of adding that flag, since this is the only flag.

Also, it should be `/funds/` and not `/fund`

Would it be possible to filter by a sentence? E.g.
Screenshot 2021-12-30 at 02 33 41

```
(✨) /fund/ $ load AU60SST00100
(✨) /fund/ $ ?
```
load is missing a newline

This formatting is a bit weird with the colours of the Sources not being selected being light
Screenshot 2021-12-30 at 02 34 42

Screenshot 2021-12-30 at 02 36 53
search should be above country because it doesn't depend on it.

Also, when we do load and there's an issue we should report it. And suggest that the country selected may be the issue.

Screenshot 2021-12-30 at 02 38 01
On info, can we drop rows that don't add anything? Empty ones and 0s (if that makes sense).

Screenshot 2021-12-30 at 02 38 48
newline missing and same for equity

Screenshot 2021-12-30 at 02 39 54
Can we have the first column more pretty? Instead of camel case having a space, e.g. priceToCashflow -> Price to cash flow

Screenshot 2021-12-30 at 02 40 41
Same here + sorting from bigger to lower + can we have default to have a pie-chart, find them much nicer to look at (you can copy what I did for ETFs)


Also, the exit needs another quit to be added!


Sorry for the many comments lol They should be all relatively fast in my defense.

DidierRLopes:
wrong version of mypy?

piiq:
> Curious what you guys think about the use of `rich` for the printing? Looking at your green terminal @piiq I am curious if this messes with outputs.

Everything is fine. The output is not messed up. I should grow up and ditch the green text in the terminal (it's almost 2022, not 1992 ffs lol).

piiq:
image

`exit` does not exit the application

jmaslek:
> Can you have the auto-complete to do `country Australia`? Instead of adding that flag, since this is the only flag.
> 
> Also, it should be `/funds/` and not `/fund`
> 
> Would it be possible to filter by a sentence? E.g. Screenshot 2021-12-30 at 02 33 41
> 
> ```
> (✨) /fund/ $ load AU60SST00100
> (✨) /fund/ $ ?
> ```
> 
> load is missing a newline
> 
> This formatting is a bit weird with the colours of the Sources not being selected being light Screenshot 2021-12-30 at 02 34 42
> 
> Screenshot 2021-12-30 at 02 36 53

All done (also when you search like you did above, you don't use the ISIN)

> search should be above country because it doesn't depend on it.

So I actually do filter on country.  The reason I do is because lets say I search for `Sachs`. that gives like 1000 results across different countries for all Goldman Sachs funds, so filtering makes it more helpful...

> Also, when we do load and there's an issue we should report it. And suggest that the country selected may be the issue.
> 
> Screenshot 2021-12-30 at 02 38 01
> 
> On info, can we drop rows that don't add anything? Empty ones and 0s (if that makes sense).

Ill add a dropna.  0 values could be information and not just missing data...

> Screenshot 2021-12-30 at 02 38 48
> 
> newline missing and same for equity
> Screenshot 2021-12-30 at 02 39 54
> 
> Can we have the first column more pretty? Instead of camel case having a space, e.g. priceToCashflow -> Price to cash flow
> Screenshot 2021-12-30 at 02 40 41
> 
> Same here + sorting from bigger to lower + can we have default to have a pie-chart, find them much nicer to look at (you can copy what I did for ETFs)
> Also, the exit needs another quit to be added!
> 
> Sorry for the many comments lol They should be all relatively fast in my defense.


👍

jmaslek:
> `exit` does not exit the application

Kinda confused why.  I have 2 quits for the queue.

jmaslek:
Everything was addressed.  now I will sit and cry while trying to fix these conflicts.

jmaslek:
@piiq @Chavithra i bumped mypy to ^.930 .  Will that break everything?

DidierRLopes:
Can overview default to 10 rows instead of 20? I think 20 is too much on the screen, and we should have top 10 everywhere in the code actually.

There's a big space after this load
Screenshot 2021-12-31 at 11 03 32

Empty rows are still displayed
Screenshot 2021-12-31 at 11 03 56

On top of 
```
(✨) /funds/ $ load Vanguard Growth Index Fund --name
```
can we allow for `-n` ?

Apart from that this is looking very nice!

piiq:
I don't understand how country filtering works
image

jmaslek:
> I don't understand how country filtering works

It's more that country is a required input to the api

image

I didn't add the choices to argparse because it acts up when a two word country is added.  There are checks on load and search that tell you to make sure your country is correctly specified.

Any insight as to why the windows build is continually failing?  The poetry matches the build yamls...

DidierRLopes:
> > I don't understand how country filtering works
> 
> It's more that country is a required input to the api
> 
> image
> 
> I didn't add the choices to argparse because it acts up when a two word country is added. There are checks on load and search that tell you to make sure your country is correctly specified.
> 
> Any insight as to why the windows build is continually failing? The poetry matches the build yamls...

Although I agree that the `choices` in argparse is not the right way because of acting up, I think once we go into argparse we should check whether country specified is correct

jmaslek:
> Can overview default to 10 rows instead of 20? I think 20 is too much on the screen, and we should have top 10 everywhere in the code actually.
> 
> There's a big space after this load Screenshot 2021-12-31 at 11 03 32
> 
> Empty rows are still displayed Screenshot 2021-12-31 at 11 03 56
> 
> On top of
> 
> ```
> (✨) /funds/ $ load Vanguard Growth Index Fund --name
> ```
> 
> can we allow for `-n` ?
> 
> Apart from that this is looking very nice!

All changes applied + country warning.  Just need to pass these damn windows builds...

DidierRLopes (approved):

PR #1109, opened 2021-12-29
NFT feature opensea collection statistics

jose-donato:
* adds new `stats` command to nft menu

image


jose-donato:
Should be good now, extra small feature to get the hands dirty with the new tests and improve NFT menu

Thank you for the help and the quick chat @Chavithra 🙌

Chavithra (approved):
The feature works on my side.

I don't thing it's related to this feature but when I do this I have an infinite loop :
```
(✨) /crypto/nft/ $ statts cryptopunks
```

jmaslek:
Can you add a float format to that table (based on the screenshot you included).

jose-donato:
> Can you add a float format to that table (based on the screenshot you included).

done 👍

DidierRLopes (changes requested):
3 minor things:
Screenshot 2021-12-30 at 02 20 02

Can we have the slurg to be required? You just need a `required=True`, because having "No data found" is misleading.

On the `Creation Date` row that date looks very ugly. Can we have just the day and then either remove the time or put on the next line?

Additionally, since you are at it can you remove `Non Fungible Tokens Menu:` ?

All the rest looks good to me!

DidierRLopes:
Can we add this function in one of the .gst files?

DidierRLopes (approved):

PR #1114, opened 2021-12-30
Discordbot/expiry reactions

tehcoderer:
@DidierRLopes was merged before I could push the changes we talked about haha. I worked on some other things to make it better.

Let me know what you think!

Activate Expiration Selection if date given is invalid
image

Auto Delete Expiration Reaction to reduce filling up the channel. 

**_Bot needs `Manage Messages` Permissions/ still functions as normal if it doesn't_**

image

If no ticker given
image

if ticker given
image

Display date on message title 
image
image

If strike price and c/p (call/put) not given with `!stocks.opt {ticker}` 
make 7️⃣  unselectable and display * Required on menu 

if given display both and 7️⃣  becomes clickable
image


DidierRLopes:
All works well, very nice!!

The only thing I'd say is that this is what I get when Trader API key is not provided. Can we compare the tradier_key against "REPLACE_ME" and if it is the same just reporta better error message so user knows he needs to update it?
Screenshot 2021-12-31 at 11 36 26

Not sure why the test is failing. But from my end it all worked.

tehcoderer:
> All works well, very nice!!
> 
> The only thing I'd say is that this is what I get when Trader API key is not provided. Can we compare the tradier_key against "REPLACE_ME" and if it is the same just reporta better error message so user knows he needs to update it? Screenshot 2021-12-31 at 11 36 26
> 
> Not sure why the test is failing. But from my end it all worked.

Oops, I forgot to check for token  in `hist.py` haha

Fixed
image

While I was at it I worked on some more error messages
image

DidierRLopes (approved):
Amazing work!!

Have a great 2022 🍾 🥂

PR #1115, opened 2021-12-30
Fix to dcf command

colin99d:
This fix updates an issue where a company with less than 10 years of historical data would cause issues. @piiq I saw the issue with AA, but it ran issue free with TSLA. My best guess is that the source website had temporary issues. If you have more insight feel free to share and I can dig in.

### DIFF
--- a/gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
+++ b/gamestonk_terminal/stocks/fundamental_analysis/dcf_view.py
@@ -400,9 +400,11 @@ class CreateExcelFA:
         self.ws2["A8"] = "Preferred Dividends"
         self.ws2["A9"] = "Free Cash Flows"
         r = 4
-        c1 = dcf_model.letters[self.len_data + 3]
-        c2 = dcf_model.letters[self.len_data + 4]
-        c3 = dcf_model.letters[self.len_data + 5]
+
+        # Switched len_data to len_pred
+        c1 = dcf_model.letters[self.len_pred + 3]
+        c2 = dcf_model.letters[self.len_pred + 4]
+        c3 = dcf_model.letters[self.len_pred + 5]
         for i in range(self.len_pred):
             j = 1 + i + self.len_data
             cols = dcf_model.letters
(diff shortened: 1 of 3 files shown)


piiq:
Hey @JerBouma 👋
You've asked about DCF on discord. Can you check if this fix works for you?

DidierRLopes (approved):
Tested several tickers, all worked. I see numbers and an excel sheet, I'm happy 😄 

Before merging, can you add `dcf` to the row before `exit` on `scripts/test_stocks_fa.gst`?

piiq (changes requested):
Hey! Thanks for the patch.

I've tested it with a number of tickers and the only one that failed was COIN
image
Can we have some feedback for the user if there was an error and no dcf sheet is being produced?

There is another problem that might only be visible to non-US ppl. The currencies in the XLS are local, what leads to spreadsheets like this
image

Can you explicitly set the currency type of the cells when you create the spreadsheet? Either the actual currency or USD if the data source is only in USD.

colin99d:
> Hey! Thanks for the patch.
> 
> I've tested it with a number of tickers and the only one that failed was COIN image Can we have some feedback for the user if there was an error and no dcf sheet is being produced?
> 
> There is another problem that might only be visible to non-US ppl. The currencies in the XLS are local, what leads to spreadsheets like this image
> 
> Can you explicitly set the currency type of the cells when you create the spreadsheet? Either the actual currency or USD if the data source is only in USD.

Looking into the currency issue right now. Also, COIN ran fine for me.

piiq:
> COIN ran fine for me.

On my end this was some issue with "sister companies" if I understand correctly.

colin99d:
> > COIN ran fine for me.
> 
> On my end this was some issue with "sister companies" if I understand correctly.

I see. For now I am just printing an error message. In the future this should just cause the excel to not have ratios. I will work on better error handling in the future.

DidierRLopes:
Let's merge this for now! It's @JerBouma birthday, so this will make him happy eheheh Let's go 🚀

DidierRLopes (approved):

DidierRLopes:
Needs @piiq approval 😬

piiq:
Thanks for addressing the issues @colin99d
image
Great job!

piiq (approved):

PR #1116, opened 2021-12-30
Example of adding logging

aia:


DidierRLopes (commented):
What do you think about adding a 
```
logger.info(an_input)
```
in the switch function of `disc_controller.py`, right above 
```
(known_args, other_args) = self.disc_parser.parse_known_args(an_input.split())
```

This would allow us to get directly the user input and could be a level of debug a bit higher level than the one done here.

Or are we trying to avoid having logging on the controller side, for some reason I'm unaware?

aia:
> What do you think about adding a
> 
> ```
> logger.info(an_input)
> ```
> 
> in the switch function of `disc_controller.py`, right above
> 
> ```
> (known_args, other_args) = self.disc_parser.parse_known_args(an_input.split())
> ```
> 
> This would allow us to get directly the user input and could be a level of debug a bit higher level than the one done here.
> 
> Or are we trying to avoid having logging on the controller side, for some reason I'm unaware?

We will be adding logging to every function call.

DidierRLopes (approved):

piiq (approved):

aia:
Will be doing this with decorators.

PR #1117, opened 2021-12-31
Adding Insider Trading Presets for Industries

deeleeramone:
Presets added for industry-specific trading by internal executives and directors


poetry.lock
```diff
@@ -3646,7 +3646,6 @@ content-hash = "fee0f4496b14e5fe0a0273280102cb83c50afa815fd81128cf83212974bda063
 
 [metadata.files]
 absl-py = [
-    {file = "absl-py-1.0.0.tar.gz", hash = "sha256:ac511215c01ee9ae47b19716599e8ccfa746f2e18de72bdf641b79b22afa27ea"},
```
DidierRLopes, line 3649: Was this intentional?

website/data/menu/main.yml
```diff
@@ -974,6 +974,8 @@ main:
           - name: robinhood
             ref: "/portfolio/brokers/robinhood"
             sub:
+              - name: login
+                ref: "/portfolio/portfolio/brokers/robinhood/login"
```
DidierRLopes, line 978: We need to remove this because this folder doesn't exist. Not sure why this didn't break in the tests...

DidierRLopes (changes requested):
From the presets.ini can you remove all lines that:
- Are either a comment, e.g. all sectors and industry
- Parameters that are empty and don't add any value.

This way when the user checks a preset from the terminal they can focus on what the preset does

poetry.lock
deeleeramone: I accepted the "incoming" changes to this, as Github told me I needed to merge before updating. So, yes..?

website/data/menu/main.yml
```diff
@@ -975,7 +975,7 @@ main:
             ref: "/portfolio/brokers/robinhood"
             sub:
               - name: login
-                ref: "/portfolio/portfolio/brokers/robinhood/login"
+                ref: "/portfolio/brokers/robinhood/login"
```
DidierRLopes, line 978: I don't have this folder. Do you?

deeleeramone:
Nope, that would have been an error when I was updating the directory for the refactoring of the menu.

DidierRLopes:
@deeleeramone can you remove the extra stuff from the .ini files as I mentioned above + remove that line from main.yml that invokes login folder that doesn't exist? Then we can merge this!

DidierRLopes:
Try the `view` command, the filtering is broken. 
Screenshot 2022-01-02 at 20 30 40


You need to:

* Keep the header:
```
# Author of preset: Sexy Year
# Description: The user that adds the preset can add a description for what type of insiders this is meant.
```

* Keep all "sections of filters" that are wrapped around square brackets, e.g.
`[General]` 

This is what you want to see:
Screenshot 2022-01-02 at 20 30 57

DidierRLopes (changes requested):
see above

deeleeramone:
Oh I see. I took out too much!

deeleeramone:
Hopefully I did that right.. I have a feeling I did something to muddle up the files somehow.

> 
Screen Shot 2022-01-02 at 5 19 19 PM

deeleeramone:
The files are not syncing properly between my local and this. Not working as intended.

PR #1119, opened 2021-12-31
Crypto menu refactor

jose-donato:
This PR tries to make the experience of crypto context similar to stocks context:
* refactor `load` command to fetch price data after the symbol is found. this way the user can enter `ta` or `pred` menus straightaway without the need of fetching price data (similar to stocks)
* `pred` menu works independently of source used
* all commands from `dd` menu work regardless of the source used. to achieve this a map of a certain coin to the multiple symbols from different sources is created and saved once `load` runs
* when crypto is loaded, the performance from the start (and in the interval selected) is displayed
unknown


jose-donato:
Need to improve documentation for load but apart from that i think this is done
Will detail in the first message what this PR does later today

discordbot/requirements.txt
```diff
@@ -1,4 +1,4 @@
-aiohttp==3.7.4.post0; python_full_version >= "3.5.3" and python_version >= "3.6"
+aiohttp==3.7.4.post0; python_version >= "3.6" and python_full_version >= "3.5.3"
```
DidierRLopes, line 1: Did you intended to change this file?

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -236,27 +244,84 @@ def call_load(self, other_args):
...
         )
+        parser.add_argument(
+            "-s",
+            "--start",
+            type=valid_date_in_past,
+            default=(datetime.now() - timedelta(days=366)).strftime("%Y-%m-%d"),
+            dest="start",
+            help="The starting date (format YYYY-MM-DD) of the crypto",
+        )
+
```
DidierRLopes, line 262: not important but in future would be nice to not have gaps between the `add_argument` cases

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -123,22 +130,23 @@ def print_help(self):
             if self.source != ""
             else "\nSource: ?\n"
         )
+        help_text += self.price_str
         help_text += """
+
```
DidierRLopes, line 135: This line can be removed

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -123,22 +130,23 @@ def print_help(self):
             if self.source != ""
             else "\nSource: ?\n"
         )
+        help_text += self.price_str
```
DidierRLopes, line 133:
```
Coin: bitcoin
Source: CoinGecko
Current Price: 46128.83 usd
Performance in interval (1day): 0.86%
Performance since specified start date (2021-01-04 00:00:00): 46.37%
```

could we have instead: 
```
Performance since 2021-01-04: 46.37%
```
without the time.

And also capitalize USD

gamestonk_terminal/cryptocurrency/crypto_controller.py
```diff
@@ -123,22 +130,23 @@ def print_help(self):
...
         help_text += """
+
     chart       view a candle chart for a specific cryptocurrency
     headlines   crypto sentiment from 15+ major news headlines [Finbrain]
```
DidierRLopes, line 137: also just noticed these lines don't have same indentation as above, would be good if they matched

gamestonk_terminal/cryptocurrency/due_diligence/dd_controller.py
```diff
@@ -167,16 +174,14 @@ def print_help(self):
    change          30d change of supply held on exchange wallets
```
DidierRLopes, line 174: I don't think we need to have `chart` again on this menu since it is right above!

DidierRLopes (changes requested):
`ta` doesn't work
Screenshot 2022-01-05 at 00 25 41

should be `/crypto` menu and not `stocks/options` menu
Screenshot 2022-01-05 at 00 26 00

Some stylistic changes, the rest looked good to me 🚀

DidierRLopes:
@JakubPluta 😬

gamestonk_terminal/cryptocurrency/crypto_controller.py
jose-donato: image

i tried to follow what was done in stocks context, should i move chart headlines etc to the beginning of the line?

discordbot/requirements.txt
DidierRLopes: Humm, if it was a change when you merged with main it shouldn't have shown any difference, since the diff compares with main.

I don't think you added any new package here, right? If so, personally, I would avoid updating the requirements since the dependencies can get quite tricky to fix. I always try to avoid touching these files as much as I can.

gamestonk_terminal/cryptocurrency/crypto_controller.py
DidierRLopes: No, I didn't mean putting them at the beginning. I meant that if you look at the indentation of chart/headlines it doesn't match the number of spaces of the ones below of the menus

jose-donato: oh my bad, should be good now

discordbot/requirements.txt
jose-donato: Yes you are right. Pycoingecko is not used on discordbot yet so this should not be changed. Reverted 👍

jmaslek:
So I note:

```
2022 Jan 05, 12:01 (✨) /crypto/ $ load btc --source bin

Loaded BTC against USDT from Binance source

2022 Jan 05, 12:01 (✨) /crypto/ $ load bitcoin --source bin
Couldn't find anything for symbol BITCOIN

2022 Jan 05, 12:01 (✨) /crypto/ $ load bitcoin

Loaded bitcoin against usd from CoinGecko source

```

Is there a full mapping available?  I know this is currently the hardest part.

jmaslek:
In disc:
```
Overview:
    coins             search for coins on CoinGecko, Binance, CoinPaprika
```

Should that be on the main -- along with 
```
find        alternate way to search for coins
```
?

jmaslek:
Does this happen always?

```
2022 Jan 05, 12:06 (✨) /crypto/disc/ $ cgvolume
HTTPSConnectionPool(host='www.coingecko.com', port=443): Max retries exceeded with url: /en/coins/high_volume (Caused by ResponseError('too many 503 error responses'))
Connection error. Couldn't connect to CoinGecko and scrape the data. Please visit CoinGecko site, and check if it's not under maintenance

Unable to retrieve data from CoinGecko.
```

JakubPluta:
> Does this happen always?
> 
> ```
> 2022 Jan 05, 12:06 (✨) /crypto/disc/ $ cgvolume
> HTTPSConnectionPool(host='www.coingecko.com', port=443): Max retries exceeded with url: /en/coins/high_volume (Caused by ResponseError('too many 503 error responses'))
> Connection error. Couldn't connect to CoinGecko and scrape the data. Please visit CoinGecko site, and check if it's not under maintenance
> 
> Unable to retrieve data from CoinGecko.
> ```

Unfortunately it happens since couple of weeks, the reason is that coingecko website needs some time to render page (load some dynamic content with js), but bs4 which we use for scraping expect that website will be instantly rendered. Only way to fix it is to use selenium but tbh I don't see it as a good approach. I discussed it with @jose-donato , and maybe we will need to drop all commands which use coingecko scraped data.

JakubPluta:
> Screen Shot 2022-01-05 at 12 05 01 PM
> 
> Maybe `fillna` with N/A or say not available

Agree, i think it make sense to replace it with some more meaningful value.

jmaslek:
Looks great.  Load is definitely a better experience.

I am curious if we should drop the cgXXX commands since they seem to not scrape?

Also is it possible to add your loading function to the `pred` menu -- so that one can select the source they want there?

We also need a load in ta.

JakubPluta:
@jose-donato agree that load has much better experience now, but unfortunately I spotted that it's not working so well in the background. I think it's something related to `load_data_data` function.
example:

I wrote `load -c Uniswap --source cg` - and it didn't load anything
image

But when I am debugging load function I see that coin was found in coingecko api:
image

And it seems that it failed at leve of `load_data_data` with Attribute error:
image

jose-donato:
> So I note:
> 
> ```
> 2022 Jan 05, 12:01 (✨) /crypto/ $ load btc --source bin
> 
> Loaded BTC against USDT from Binance source
> 
> 2022 Jan 05, 12:01 (✨) /crypto/ $ load bitcoin --source bin
> Couldn't find anything for symbol BITCOIN
> 
> 2022 Jan 05, 12:01 (✨) /crypto/ $ load bitcoin
> 
> Loaded bitcoin against usd from CoinGecko source
> ```
> 
> Is there a full mapping available? I know this is currently the hardest part.

This is a great point. I took notice of this and added as a [TODO](https://github.com/GamestonkTerminal/GamestonkTerminal/blob/213e8b0137ba5f9ffac449a7eb91e9932f92b6db/gamestonk_terminal/cryptocurrency/cryptocurrency_helpers.py#L325). For now it only works for CoinGecko as we can see on your message but I think we can easily make it possible for the other sources as well. Will take a look at this now

jose-donato:
> Does this happen always?
> 
> ```
> 2022 Jan 05, 12:06 (✨) /crypto/disc/ $ cgvolume
> HTTPSConnectionPool(host='www.coingecko.com', port=443): Max retries exceeded with url: /en/coins/high_volume (Caused by ResponseError('too many 503 error responses'))
> Connection error. Couldn't connect to CoinGecko and scrape the data. Please visit CoinGecko site, and check if it's not under maintenance
> 
> Unable to retrieve data from CoinGecko.
> ```
> 
> Every cg command I try fails.

Yea, i think it does now. They added some protection against scraping. I think we should drop these commands and don't rely on selenium. We will be okey, there are a lot of alternatives on the crypto space, we will found another useful info somewhere else and add more commands that use coingecko api

jose-donato:
> 

yep for sure! just to be sure i understood: add a `load` similar to crypto_controller command in both `pred` and `ta` menus right?

jmaslek:
> yep for sure! just to be sure i understood: add a `load` similar to crypto_controller command in both `pred` and `ta` menus right?

yup

jose-donato:
> @jose-donato agree that load has much better experience now, but unfortunately I spotted that it's not working so well in the background. I think it's something related to `load_data_data` function. example:
> 
> I wrote `load -c Uniswap --source cg` - and it didn't load anything image
> 
> But when I am debugging load function I see that coin was found in coingecko api: image
> 
> And it seems that it failed at leve of `load_data_data` with Attribute error: image
> 
> I think that issue occures there: image
> 
> It expects to have one coin value eg. `uni` but it found 3 coins with the symbol `uni`, so there is pandas Series object instead string value of coin. So I would say we need to ensure that only best match will be taken.
> 
> But anyway I think that we should totally change the way of loading coins. For now there is like 12k coins in mapping data frame, probably no one is interested in 99% of those coins. Maybe we should limit it to top 100 or top 300 based on market cap. Then it will be much easier to maintains mapping

hey @JakubPluta thanks for this detailed message and sorry for the delay. It helped a lot improving the load function and understanding what happened. I will try fix what you mentioned here. 

I tried to reuse the functions that you had already created (thank you for that) to get a simpler refactor and bring down the gap between crypto and stocks `load` asap.

I think we should probably finish this PR since the goal was to improve and match the load experience here with `stocks` context. After that we can have a chat and discuss how we can improve further and do some major refactors then. What do you think about this?

jose-donato:
The issues mentioned before should be fixed now

JakubPluta:
> The issues mentioned before should be fixed now

Nice, I check it for a few coins and worked perfectly :) Great job. Thanks for fixing this.

DidierRLopes (approved):

PR #1121, opened 2021-12-31
Handle KeyboardInterrupt on user input read

dakhouya:
Add keyboard interrupt exception in order to handle ctrl+c gracefully.
A typo in print_goodbye corrected at the same time.

### DIFF
--- a/gamestonk_terminal/terminal_helper.py
+++ b/gamestonk_terminal/terminal_helper.py
@@ -314,7 +314,7 @@ def print_goodbye():
         goodbye_msg[random.randint(0, len(goodbye_msg) - 1)] + goodbye_msg_time + "\n"
     )
 
-    logger.info("Terminal started")
+    logger.info("Terminal stopped")
 
 
 def sha256sum(filename):
--- a/terminal.py
+++ b/terminal.py
@@ -72,7 +72,6 @@ class TerminalController:
             "cmd",
             choices=self.CHOICES,
         )
-
         self.completer: Union[None, NestedCompleter] = None
 
         if session and gtff.USE_PROMPT_TOOLKIT:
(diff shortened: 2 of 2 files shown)


DidierRLopes:
Nice one! This only works on the main menu, but we can try to apply this to all controllers in the code!

@piiq @jmaslek @jose-donato

colin99d:
> Nice one! This only works on the main menu, but we can try to apply this to all controllers in the code!
> 
> @piiq @jmaslek @jose-donato

colin99d:
Sorry did not mean to close the pull request. Should we make a decorator or base class for all the context menus or is there a quicker way?

DidierRLopes:
> Sorry did not mean to close the pull request. Should we make a decorator or base class for all the context menus or is there a quicker way?

Yea, good point. We want to have a base class sooner or later since the controllers are very similar to each other apart from their main methods.

dakhouya:
> > Sorry did not mean to close the pull request. Should we make a decorator or base class for all the context menus or is there a quicker way?
> 
> Yea, good point. We want to have a base class sooner or later since the controllers are very similar to each other apart from their main methods.

This might be a good idea if the code is duplicated at multiple places. Is it something I can help with or there's already someone working on it?

DidierRLopes:
> > > Sorry did not mean to close the pull request. Should we make a decorator or base class for all the context menus or is there a quicker way?
> > 
> > 
> > Yea, good point. We want to have a base class sooner or later since the controllers are very similar to each other apart from their main methods.
> 
> This might be a good idea if the code is duplicated at multiple places. Is it something I can help with or there's already someone working on it?

Actually there isn't! We would be very much looking forward to such! Last week we had a massive effort to improve UX overall and rewrote most of the controllers, so now it's actually the perfect timing to do this!

dakhouya:
I'll close de PR en reopen it with a more complete solution.

DidierRLopes:
I would be ok to merge this one and then you could work on top of it :)

dakhouya:
> I would be ok to merge this one and then you could work on top of it :)

Ok then doing the changes in all the controllers sounds good to you?

DidierRLopes:
> > I would be ok to merge this one and then you could work on top of it :)
> 
> Ok then doing the changes in all the controllers sounds good to you for this one?

Yup, that would be perfect! Then we can work on a base controller as a next iteration! Thank you 🙌

PR #1122, opened 2021-12-31
Add whatif scenario feature

DidierRLopes:
```
(✨) /stocks/bt/ $ whatif
IPO date selected by default.
If you had acquired 1 share of TSLA on 2010-06-29 with a cost of 4.78.
This would be worth 1063.05. Which represents an increase of 22248.85%. 

(✨) /stocks/bt/ $ whatif -n 2.34
IPO date selected by default.
If you had acquired 2.34 shares of TSLA on 2010-06-29 with a cost of 11.18.
These would be worth 2486.98. Which represents an increase of 22243.83%. 

(✨) /stocks/bt/ $ whatif 2021-01-01
If you had acquired 1 share of TSLA on 2021-01-01 with a cost of 729.77.
This would be worth 1062.81. Which represents an increase of 145.64%. 
```

### DIFF
--- a/gamestonk_terminal/stocks/backtesting/bt_view.py
+++ b/gamestonk_terminal/stocks/backtesting/bt_view.py
@@ -3,6 +3,8 @@ __docformat__ = "numpy"
 
 import os
 
+from datetime import datetime
+import yfinance as yf
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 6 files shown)


gamestonk_terminal/stocks/backtesting/bt_view.py
```diff
@@ -18,6 +20,75 @@
...
+
+    if not data.empty:
+        data = data["Adj Close"]
+
+    ipo_date = data.index[0]
+    last_date = data.index[-1]
+
+    if not date_shares_acquired:
+        date_shares_acquired = ipo_date
+        print("IPO date selected by default.")
```
jmaslek, line 49: Maybe throw in what that is i.e.: IPO Date ( 10-30-1992 ) selected"

jmaslek (approved):

gamestonk_terminal/stocks/backtesting/bt_view.py
DidierRLopes: The reason I didn't it's because the date is shown straight below, and having it repeated looked ugly 😋

PR #1123, opened 2022-01-01
Refactored Reports

colin99d:
Changed the reports context into the Jupyter context. I then made sub menus for Reports and Dashboards. The reports is for papermill reports like the ones that already existed. The dashboards are for interactive pages like mine. I also added Voila support. Voila is much better for our non-coder friends, since it presents a much sleeker environment to get data, while leaving the jupyter-lab as default encourages tinkering, and PRS :) . Once this has been reviewed and changes requested are added I will update the documentation.


gamestonk_terminal/jupyter/dashboards/dashboards_controller.py
```diff
@@ -0,0 +1,252 @@
...
+        if session and gtff.USE_PROMPT_TOOLKIT:
+            choices: dict = {c: {} for c in self.CHOICES}
+
+            self.completer = NestedCompleter.from_nested_dict(choices)
+
+    @staticmethod
+    def print_help():
+        """Print help"""
+        help_str = """
+Dashboards:
```
DidierRLopes, line 64: No need for this

gamestonk_terminal/jupyter/dashboards/dashboards_controller.py
```diff
@@ -0,0 +1,252 @@
...
+            # Consume 1 element from the queue
+            an_input = dashboard_controller.queue[0]
+            dashboard_controller.queue = dashboard_controller.queue[1:]
+
+            # Print the current location because this was an instruction and we want user to know what was the action
+            if (
+                an_input
+                and an_input.split(" ")[0] in dashboard_controller.CHOICES_COMMANDS
+            ):
+                print(f"{get_flair()} /dashboard/ $ {an_input}")
```
DidierRLopes, line 200: should be `/jupyter/dashboard/`

gamestonk_terminal/jupyter/dashboards/dashboards_controller.py
```diff
@@ -0,0 +1,252 @@
...
+        # Get input command from user
+        else:
+            # Display help menu when entering on this menu from a level above
+            if an_input == "HELP_ME":
+                dashboard_controller.print_help()
+
+            # Get input from user using auto-completion
+            if session and gtff.USE_PROMPT_TOOLKIT and dashboard_controller.completer:
+                an_input = session.prompt(
+                    f"{get_flair()} /dashboard/ $ ",
```
DidierRLopes, line 211: should be `/jupyter/dashboard/`

gamestonk_terminal/jupyter/dashboards/dashboards_controller.py
```diff
@@ -0,0 +1,252 @@
...
+                    completer=dashboard_controller.completer,
+                    search_ignore_case=True,
+                )
+
+            # Get input from user without auto-completion
+            else:
+                an_input = input(f"{get_flair()} /dashboard/ $ ")
```
DidierRLopes, line 218: should be `/jupyter/dashboard/`

gamestonk_terminal/jupyter/jupyter_controller.py
```diff
@@ -0,0 +1,240 @@
...
+        "..",
+        "exit",
+        "r",
+        "reset",
+    ]
+
+    CHOICES_COMMANDS = [
+        "reports",
+        "dashboards",
+        "finra",
```
DidierRLopes, line 41: You don't want any of those below, only reports and dashboards. And it should be named `CHOICES_MENUS` instead

gamestonk_terminal/jupyter/jupyter_controller.py
```diff
@@ -0,0 +1,240 @@
...
+        if session and gtff.USE_PROMPT_TOOLKIT:
+            choices: dict = {c: {} for c in self.CHOICES}
+
+            self.completer = NestedCompleter.from_nested_dict(choices)
+
+    @staticmethod
+    def print_help():
+        """Print help"""
+        help_str = """
+Jupyter:
```
DidierRLopes, line 69: no need for this line

gamestonk_terminal/jupyter/jupyter_controller.py
```diff
@@ -0,0 +1,240 @@
...
+   reports       creates jupyter reports
```
DidierRLopes, line 70: This spacing is not the correct one and you are missing the ">" which distinguishes menu from command:
```
>   resources
    
(✨) / $ jupyter

Jupyter:
   reports       creates jupyter reports
   dashboards    shows interactive jupyter dashboards  
```

Should be:
```
>   reports       creates jupyter reports
>   dashboards    shows interactive jupyter dashboards  
```

gamestonk_terminal/jupyter/jupyter_controller.py
```diff
@@ -0,0 +1,240 @@
...
+        self.queue.insert(0, "quit")
+
+    def call_exit(self, _):
+        """Process exit terminal command"""
+        self.queue.insert(0, "quit")
+        self.queue.insert(0, "quit")
+
+    def call_reset(self, _):
+        """Process reset command"""
+        self.queue.insert(0, "resources")
```
DidierRLopes, line 147: `self.queue.insert(0, "jupyter")`

terminal.py
```diff
@@ -122,6 +123,7 @@ def print_help(self):
 >   forex
 >   portfolio
 >   reports
```
DidierRLopes, line 125: This one should be removed

terminal.py
```diff
@@ -61,6 +61,7 @@ class TerminalController:
         "etf",
         "reports",
```
DidierRLopes, line 62: This should be removed

DidierRLopes (changes requested):
On top of this:
* the notebook reports should be fixed because of the change of location
* the jupyter/reports controller should be updated to reflect new path + fix reset
* the `test_reports_base.gst` and `test_reports.gst` files need to be updated to reflect this change of location
* merge with main since `dd` notebook report was changed
* `(✨) /dashboard/ $ stocks -h` crashes
* Why not have Voila by default? It looks very nice!

Finally, @piiq has suggested for us to not open a port on the user's machine as this can lead to security issues. Hence why we deprecated such for the reports. I think here we would want to do the same, we can probably ask the user to run jupyter kernel behind the scenes or so... and then the user can say "Y" when he has done that, and we open the jupyter lab. Also, I don't like the fact that after we can enter in one of these, we can't keep using the terminal as "Ctrl+C" will make it crash (unless there's a smooth way to exit that I'm unaware).
Screenshot 2022-01-01 at 17 50 53
Screenshot 2022-01-01 at 17 53 54

jmaslek:
I fixed the build issue, so if you merge, the last test should pass.

colin99d:
> On top of this:
> 
> * the notebook reports should be fixed because of the change of location
> * the jupyter/reports controller should be updated to reflect new path + fix reset
> * the `test_reports_base.gst` and `test_reports.gst` files need to be updated to reflect this change of location
> * merge with main since `dd` notebook report was changed
> * `(✨) /dashboard/ $ stocks -h` crashes
> * Why not have Voila by default? It looks very nice!
> 
> Finally, @piiq has suggested for us to not open a port on the user's machine as this can lead to security issues. Hence why we deprecated such for the reports. I think here we would want to do the same, we can probably ask the user to run jupyter kernel behind the scenes or so... and then the user can say "Y" when he has done that, and we open the jupyter lab. Also, I don't like the fact that after we can enter in one of these, we can't keep using the terminal as "Ctrl+C" will make it crash (unless there's a smooth way to exit that I'm unaware). Screenshot 2022-01-01 at 17 50 53 Screenshot 2022-01-01 at 17 53 54

Could we have a prompt that explains jupyter-lab and what exactly happens and then prompt them with a y/n on whether they would like to proceed? That way they can understand what is happening while also getting time savings from using the terminal.

DidierRLopes:
> > On top of this:
> > 
> > * the notebook reports should be fixed because of the change of location
> > * the jupyter/reports controller should be updated to reflect new path + fix reset
> > * the `test_reports_base.gst` and `test_reports.gst` files need to be updated to reflect this change of location
> > * merge with main since `dd` notebook report was changed
> > * `(✨) /dashboard/ $ stocks -h` crashes
> > * Why not have Voila by default? It looks very nice!
> > 
> > Finally, @piiq has suggested for us to not open a port on the user's machine as this can lead to security issues. Hence why we deprecated such for the reports. I think here we would want to do the same, we can probably ask the user to run jupyter kernel behind the scenes or so... and then the user can say "Y" when he has done that, and we open the jupyter lab. Also, I don't like the fact that after we can enter in one of these, we can't keep using the terminal as "Ctrl+C" will make it crash (unless there's a smooth way to exit that I'm unaware). Screenshot 2022-01-01 at 17 50 53 Screenshot 2022-01-01 at 17 53 54
> 
> Could we have a prompt that explains jupyter-lab and what exactly happens and then prompt them with a y/n on whether they would like to proceed? That way they can understand what is happening while also getting time savings from using the terminal.

I'm fine with that as long as that message states what is gonna happen if they press `yes`

gamestonk_terminal/jupyter/dashboards/dashboards_controller.py
colin99d: Which part is unneeded?

DidierRLopes (changes requested):
Screenshot 2022-01-04 at 23 12 36

Still crashing, you just need to create that folder and add a `.gitkeep` file inside. `git` doesn't track folders, therefore you need to "trick" git by adding a meaningful hidden file in there

colin99d:
> Screenshot 2022-01-04 at 23 12 36
> 
> Still crashing, you just need to create that folder and add a `.gitkeep` file inside. `git` doesn't track folders, therefore you need to "trick" git by adding a meaningful hidden file in there

Done! Apologies good sir.

DidierRLopes:
eheh sorry to nag again, but you have added the wrong path it should be `jupyter/reports/stored`. Also, the name of the file should actually be `.gitkeep` - it's a convention thing AFAIK

DidierRLopes:
Last one hopefully (😄) the .html opening has not the correct path, it has 2 `gamestonk_terminal` when it should be a single one

DidierRLopes:
Screenshot 2022-01-05 at 00 47 12

exit is not exiting

colin99d:
> 

What do you mean by this?

jmaslek:
> What do you mean by this?

`exit` should leave the terminal.  In `call_exit` , you probably need an extra self.queue.insert(0,"quit")

colin99d:
> Last one hopefully (😄) the .html opening has not the correct path, it has 2 `gamestonk_terminal` when it should be a single one

Sorry meant this one

DidierRLopes:
Screenshot 2022-01-04 at 23 12 36


> > Last one hopefully (😄) the .html opening has not the correct path, it has 2 `gamestonk_terminal` when it should be a single one
> 
> Sorry meant this one
Screenshot 2022-01-05 at 00 59 34

So the exported file is correct, however my browser opened automatically with a different URL with a `gamestonk_terminal/gamestonk_terminal` in the path.

DidierRLopes (approved):
🦋 🚀  Flyiiinnggggg

PR #1126, opened 2022-01-03
improve usage of timezone in terminal

DidierRLopes:
This PR does 3 main things:
* When loading a ticker, retrieve exchange, timezone, currency and date time of exchange
Screenshot 2022-01-03 at 01 13 15

* Allows the user to specify it's timezone (either through a command or directly in file) and show it, e.g.
Screenshot 2022-01-03 at 02 43 27

* Adding the date time information before the flair (this is a feature flag)
Screenshot 2022-01-03 at 02 43 52

### DIFF
--- a/gamestonk_terminal/feature_flags.py
+++ b/gamestonk_terminal/feature_flags.py
@@ -19,6 +19,9 @@ USE_COLOR = strtobool(os.getenv("GTFF_USE_COLOR", "True"))
 # Select console flair (choose from config_terminal.py list)
 USE_FLAIR = os.getenv("GTFF_USE_FLAIR") or "stars"
 
+# Add date and time to command line
+USE_DATETIME = strtobool(os.getenv("GTFF_USE_DATETIME", "True"))
+
 # Enable interactive matplotlib mode
 USE_ION = strtobool(os.getenv("GTFF_USE_ION", "True"))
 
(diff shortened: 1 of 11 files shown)


jmaslek:
Loading the help/? command with the flag now takes way too long.  Can we default the use date time to False?

```
Exchange: NMS (in USD)
```

I would specify the exchange and currency separately. 

```
03 Jan 2022 16:00:16 [America/New_York]
Market: CLOSED

2022 Jan 03, 16:01:17 (✨) /stocks/ $
```

The date reporting is inconsistent.

Using help/? without a ticker has too many spaces

```
Daily Stock: 



    quote       view the current price for a specific stock ticker
```

jmaslek (changes requested):
comments in conversation

DidierRLopes:
@jmaslek I left flag on, but changed to ask it again. The time of additional info won't match, but it doesn't matter because at that time what you've got is the datetime corresponding to the loaded stock. Which makes sense.

jmaslek:
I will leave this to you -- but I think the time looks better if you have just HH:MM instead of HH:MM:SS (also could include a feature flag for 12 or 24 hour clock but that's overkill imo)

jmaslek (approved):

DidierRLopes:
> I will leave this to you -- but I think the time looks better if you have just HH:MM instead of HH:MM:SS (also could include a feature flag for 12 or 24 hour clock but that's overkill imo)

Agreed, and done!

PR #1132, opened 2022-01-03
Menu Decorator

colin99d:
The new menu function you guys have been working on is super useful, but its also big and scary. I made this decorator to make creating new contexts a little less scary. Currently marking menus that I am not sure how to integrate with `# Handle`.


Chavithra:
Looks like `self.completer` is not setup properly inside :
- `gamestonk_terminal.stocks.options.test_payoff_controller.PayoffController`

Also the `menu` function was not refactored inside this module:
- `gamestonk_terminal.stocks.options.options_controller`

DidierRLopes:
This is very nice!!! Can we handle the "Handle" cases on the wrapper? At least for the `from_submenu` cases, the ones with the dynamic choices is not as clear to me on how we can do that. 

The `sub_menu` var name is also not great. The idea is that when you do "quit" you go back to the menu straight above, e.g. if you are in `comparison_analysis` you want to go into `stocks`. However, due to the implementation of the queue, if you do `quit` on `ca` after you went there from `screener`, you will go back to screener. Therefore, we want to do the full path to go to `/stocks`. Does this make sense?

I think we could have something like: `go_into` and then do:
`go_into="quit"`
Screenshot 2022-01-04 at 09 48 48
`go_into="home/portfolio"`
Screenshot 2022-01-04 at 09 51 47

I can implement this if you prefer me to!

DidierRLopes (changes requested):
You are pushing .html stored jupyter, can you remove those?

Can we add them to .gitignore as well?

DidierRLopes:
For testing this to make sure everything still works, the scripts folder can be of great use, e.g. `python terminal.py scripts/test_*.gst`

colin99d:
Screen Shot 2022-01-04 at 11 57 28 AM
@DidierRLopes what are you thinking about this custom code in the options menu. Would it be better to remove or change our decorator to fit this specific case. Either with a parameter or only having the different code in the original function (like how the try except decorator runs the original function.

colin99d:
@DidierRLopes I cannot get this dang thing working. Whenever you have a free minute do you mind taking a look? You can search for this comment `THIS DOES NOT WORK, CANNOT FIGURE OUT WHY`. Once we get that figured out we just need to finish updating tests and then we should be good.

DidierRLopes:
This fixes the decorator: https://github.com/colin99d/GamestonkTerminal/pull/1

The menus with dynamic options are: stocks/sia; stocks/options; stocks/options/pricing. I think you may have deleted by accident last one.

Can we rename `choices_function` to `dynamic_completer` and `context` to `controller_class`? I think would be more easy to understand!

colin99d:
@Chavithra we are done changing things now and it looks like 8 failures and 1 error. I can try to fix these tests with some explanation from you, or you can do them. Either way works for me, just let me know what you want to do.

Chavithra:
> @Chavithra we are done changing things now and it looks like 8 failures and 1 error. I can try to fix these tests with some explanation from you, or you can do them. Either way works for me, just let me know what you want to do.

@colin99d sure we can plan call for that

colin99d:
I am closing this PR. Baseclass will negate the need for a decorator.

PR #1133, opened 2022-01-04
Add prediction to economy/fred

jmaslek:
Funsies.  I added to Fred and not main economy because I think all those AV datasets can be accessed in Fred and I wanted to avoid having 2 controllers.

Also changed to the rich console for the controllers :P

I do note that the NN predictions display price (in $) when the units for the Fred series are not $.  This would involve redoing the prediction functions, which is slightly outside the scope here.

Screen Shot 2022-01-03 at 9 18 20 PM


jmaslek:
@Chavithra I added some tests all on my own :D

DidierRLopes:
We should not crash in this scenario:
Screenshot 2022-01-04 at 10 23 13

Can the `rmv` be independent of upper or lower letter?
Screenshot 2022-01-04 at 10 23 35

Can you make the labels have 45 degrees, as the Montecarlo plot.
Screenshot 2022-01-04 at 10 25 03

Apart from these minor things all looks good to me :)

gamestonk_terminal/economy/fred/fred_view.py
```diff
@@ -131,15 +139,23 @@ def display_fred_series(
     plt.gcf().autofmt_xdate()
     fig.tight_layout()
     plt.show()
+    data.index = [x.strftime("%Y-%m-%d") for x in data.index]
     if raw:
         if gtff.USE_TABULATE_DF:
-            print(tabulate(data.tail(20), headers=data.columns, tablefmt="fancy_grid"))
+            t_console.print(
+                rich_table_from_df(
+                    data.tail(20),
```
DidierRLopes, line 147: Can we have this 20 customisable?

DidierRLopes (changes requested):
Apart from aforementioned changes, can you create a `scripts/test_economy_fred_pred.gst` file?

DidierRLopes:
Also, documentation is missing for website

jmaslek:
Data shape catching added:

Screen Shot 2022-01-04 at 10 46 56 AM

jmaslek:
> Also, documentation is missing for website

Isn't all the documentation in common?  There is no stocks/prediction website stuff

website/content/economy/fred/_index.md
```diff
@@ -0,0 +1 @@
+Menu for accessing Federal Reserve Economic Datat (FRED)
```
DidierRLopes, line 1: You need to add the TOC by default so users can expand on that!

jmaslek: ?

DidierRLopes (approved):

PR #1134, opened 2022-01-04
Step towards portfolio allocation analysis

northern-64bit:
Makes pa active and adds country to the df. The groupby command also gets percents of holding allocation. It also fixes warnings and prepares for a later pr that I'm currently working on.


DidierRLopes:
You should install black with `pip install black` and then run `black` on the the files or directory you are interested. E.g. `black gamestonk_terminal/stocks/....`

Or you can install `pre-commit`, `pre-commit install` which will run `black` once you commit!

DidierRLopes:
This PR should do: https://github.com/northern-64bit/GamestonkTerminal/pull/2

northern-64bit:
> This PR should do: [northern-64bit#2](https://github.com/northern-64bit/GamestonkTerminal/pull/2)

Thanks!

gamestonk_terminal/portfolio/portfolio_analysis/portfolio_model.py
```diff
@@ -50,6 +54,12 @@ def load_portfolio(
             axis=1,
         )
 
+    if country:
+        df["country"] = df.apply(
+            lambda row: yfinance_model.get_country(row.Ticker),
+            axis=1,
+        )
```
jmaslek, line 61: If the Dataframe gets large, I imagine this will take a while.  I think a better approach would be to create a dictionary using all unique tickers.  Something like `{tick: get_country(tick) for tick in df.Ticker.unique()}` and then you can do df["Country"] = df["Ticker"].map(country_dictionary}.  This will limit the number of times you get the country.

gamestonk_terminal/portfolio/portfolio_analysis/portfolio_view.py
```diff
@@ -16,13 +20,18 @@ def display_group_holdings(portfolio: pd.DataFrame, group_column: str):
     group_column : str
         Column to group by
     """
-
+    headers = [group_column, "value"]
     grouped_df = pd.DataFrame(portfolio.groupby(group_column).agg(sum)["value"])
+
+    if allocation:
+        total_value = grouped_df["value"].sum()
+        grouped_df["% allocation"] = grouped_df["value"] / total_value * 100
```
jmaslek, line 28: Lets try to keep columns in the Dataframe as single strings - so "pct_allocation".  The header Is perfect

northern-64bit: Done!

gamestonk_terminal/portfolio/portfolio_analysis/portfolio_model.py
northern-64bit: Also done!

jmaslek (approved):

PR #1135, opened 2022-01-04
Fixing tradier

DidierRLopes:
Still not sure what is going on, but it doesn't crash:
image

`grhist` is also compromised, but this is due to Syncretistm being down: https://api.syncretism.io/ops/TSLA220107C01000000

Not sure if this is temporary or not. Thoughts


DidierRLopes:
@Tyruiop solved `grhist`!! 🚀

DidierRLopes:
Put CE as default instead of Tradier

jmaslek (approved):

PR #1139, opened 2022-01-04
Qa update

jmaslek:
Couple changes:

1) Add a new `line` function that plots line for the target data
2) Allow said line function to show log scale
3) Add LogPrice as target column
4) Change everything to rich


DidierRLopes:
I think you just need to call `pytest` with `--record-mode=rewrite`

jmaslek:
> I think you just need to call `pytest` with `--record-mode=rewrite`

Think I changed a num to limit so I need to change that

DidierRLopes (approved):
Can you just remove the `Quantitative Analysis:` from the menu and allow to do `pick logprice` instead of having to specify the flag `-t`. It makes even more sense because that is the only argument.

jmaslek:
> Can you just remove the `Quantitative Analysis:` from the menu 

done

>and allow to do `pick logprice` instead of having to specify the flag `-t`. It makes even more sense because that is the only argument.

was already case.  changed to allow the autocomplete there

PR #1142, opened 2022-01-05
New alternative data context + first sub context (Covid)

jmaslek:
As @DidierRLopes and I talked about offline, having alternative data nowadays is all the rage.

Here I add the context + a COVID menu.  Currently this will just show highest/lowest slopes and deaths/cases/death rate for the selected country.

The data is obtained from https://github.com/CSSEGISandData/COVID-19. which is the Johns Hopkins repo.

### DIFF
--- a/gamestonk_terminal/alternative/alt_controller.py
+++ b/gamestonk_terminal/alternative/alt_controller.py
@@ -1,0 +1,225 @@
+"""Alternative Data Controller Module"""
+__docformat__ = "numpy"
+
+import argparse
+import difflib
+from typing import List, Union
+
+from prompt_toolkit.completion import NestedCompleter
+
+from gamestonk_terminal import feature_flags as gtff
+from gamestonk_terminal.helper_funcs import (
+    get_flair,
+    system_clear,
+)
+from gamestonk_terminal.menu import session
+
+# pylint:disable=import-outside-toplevel
+
+
+class AlternativeDataController:
+    """alternative Controller class"""
+
+    # Command choices
+    CHOICES = [
+        "cls",
+        "home",
+        "h",
+        "?",
+        "help",
+        "q",
+        "quit",
+        "..",
+        "exit",
+        "r",
+        "reset",
+    ]
+
+    CHOICES_COMMANDS: List[str] = []
+    CHOICES_MENUS = ["covid"]
+    CHOICES += CHOICES_COMMANDS + CHOICES_MENUS
+
+    def __init__(self, queue: List[str] = None):
+        """Constructor"""
+        self.alt_parser = argparse.ArgumentParser(add_help=False, prog="alternative")
+        self.alt_parser.add_argument(
+            "cmd",
+            choices=self.CHOICES,
+        )
+        self.queue = queue if queue else list()
+        self.completer: Union[None, NestedCompleter] = None
+
+        if session and gtff.USE_PROMPT_TOOLKIT:
+            choices: dict = {c: {} for c in self.CHOICES}
+
+            self.completer = NestedCompleter.from_nested_dict(choices)
+
+    @staticmethod
+    def print_help():
+        """Print help"""
+        help_str = """
+Alternative Data:
+>   covid
+        """
+        print(help_str)
+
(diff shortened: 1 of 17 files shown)


DidierRLopes:
crashes badly:
Screenshot 2022-01-06 at 01 56 28

Please, not this date format 🤣 
Screenshot 2022-01-06 at 01 58 14

Not even gonna use the argument that I hate it, but we have in our codebase YYYY-MM-DD pretty much everywhere. So let's stick with that for now, then we can work on a feature flag for the date format preferred by the user.

In any case, due to the type of data we have, we can probably even display like "Jun 2021, Oct 2021, ..". No?

Can we change ylabel to Millions/Thousands/... accordingly? Based on what makes sense for country in question.
Screenshot 2022-01-06 at 02 01 33

gamestonk_terminal/alternative/alt_controller.py
```diff
@@ -0,0 +1,228 @@
...
+        if session and gtff.USE_PROMPT_TOOLKIT:
+            choices: dict = {c: {} for c in self.CHOICES}
+
+            self.completer = NestedCompleter.from_nested_dict(choices)
+
+    @staticmethod
+    def print_help():
+        """Print help"""
+        help_str = """
+Alternative Data:
```
DidierRLopes, line 63: No need for this line, at least we removed it on the last UX iteration

gamestonk_terminal/alternative/alt_controller.py
```diff
@@ -0,0 +1,228 @@
...
+>   covid
```
DidierRLopes, line 64: add some examples of what the user can expect in front

gamestonk_terminal/alternative/covid/covid_controller.py
```diff
@@ -0,0 +1,415 @@
...
+
+    def call_exit(self, _):
+        """Process exit terminal command"""
+        self.queue.insert(0, "quit")
+        self.queue.insert(0, "quit")
+        self.queue.insert(0, "quit")
+
+    def call_reset(self, _):
+        """Process reset command"""
+        self.queue.insert(0, "covid")
```
DidierRLopes, line 164: the reset can also load the country here

gamestonk_terminal/alternative/covid/covid_controller.py
```diff
@@ -0,0 +1,415 @@
...
+            add_help=False,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+            prog="rates",
+            description="Show historical rates country.",
+        )
+        ns_parser = parse_known_args_and_warn(
+            parser,
+            other_args,
+            export_allowed=EXPORT_BOTH_RAW_DATA_AND_FIGURES,
+            raw=True,
```
DidierRLopes, line 234: I think this was a significant change to be in a different PR. Because it's not related at all with alternative data. Plus, in that PR we could go back and tackle all functions that needed such.

DidierRLopes (changes requested):
Can we also add a `.gst` file for this?

gamestonk_terminal/alternative/covid/covid_controller.py
jmaslek: This isn't really a significant change.  Since I added default arguments to the function, it won't break anything and we can go back and revisit where we see as needed.

jmaslek:
> crashes badly: Screenshot 2022-01-06 at 01 56 28

oops

> Not even gonna use the argument that I hate it, but we have in our codebase YYYY-MM-DD pretty much everywhere. So let's stick with that for now, then we can work on a feature flag for the date format preferred by the user.

Fair enough.

> In any case, due to the type of data we have, we can probably even display like "Jun 2021, Oct 2021, ..". No?

I don't think the MPL date formatter is that smart lol

> 
> Can we change ylabel to Millions/Thousands/... accordingly? Based on what makes sense for country in question. Screenshot 2022-01-06 at 02 01 33

Largest number in the entire df is basically 1Mil, so I think leaving this in thousands looks decent enough.

GST file added as well.

DidierRLopes (approved):

PR #1147, opened 2022-01-06
Remove all non-functioning interval options from Glassnode API

minhhoang1023:
## Summary: Remove all non-functioning interval options from Glassnode API.

The working options after the fix are:
- INTERVALS_HASHRATE = ["24h", "1w", "1month"]
- INTERVALS_ACTIVE_ADDRESSES = ["24h", "1w", "1month"]
- INTERVALS_NON_ZERO_ADDRESSES = ["24h"]
- INTERVALS_DISPLAY_EXCHANGE_NET_POSITION_CHANGE = ["24h"]
- INTERVALS_EXCHANGE_BALANCES = ["24h"]

## Diagnosis
Glassnode API does not support `1h` option despite having listed it in their documentation. 

image

If chosen the `1h` option, the following image would be shown:
image

After the fix, user will not longer be able to choose the `1h` option
image


jose-donato (changes requested):
`10min` does not work also on my end. can you confirm this and update accordingly?

minhhoang1023:
@jose-donato Indeed doesn't work for me too. 

I then went to test for other commands from the Glassnode model, e.g. nonzero, active. And each endpoint has a different set of working intervals. For nonzero, 1monh and 1w is not working, whereas active worked for 1w and 1m.

See the screenshot below. So I think we'd need to create a different INTERVAL variable for each of the endpoint. Wdyt?

image

jose-donato:
> @jose-donato Indeed doesn't work for me too.
> 
> I then went to test for other commands from the Glassnode model, e.g. nonzero, active. And each endpoint has a different set of working intervals. For nonzero, 1monh and 1w is not working, whereas active worked for 1w and 1m.
> 
> See the screenshot below. So I think we'd need to create a different INTERVAL variable for each of the endpoint. Wdyt?
> 
> image

yep! that looks perfect to me. don't forget to update those on `choices` for auto completion

minhhoang1023:
@jose-donato Added!

jose-donato (approved):
Great job. Thank you! Congrats on your first merge in GamestonkTerminal 🎉

PR #1148, opened 2022-01-06
Refactor common/ta for easier report usage

jmaslek:
This PR serves to take away the headache of writing a TA report.

Previously, we needed something like 

`overlap_model.ema("SMA", ticker, "1440min", 14, export = "")`

I added defaults to pretty much everything and left all ta commands of the form `_model(ticker, stocks_dataframe)`. 

Will make reports much simpler.


DidierRLopes:
Why not going further and removing the need of `s_interval` and instead of getting a DataFrame object getting a Series one. That way you can still access index and values, and therefore it will be able to do Intraday plots as well because it doesn't try to access AdjClose or so based on that interval.

jmaslek:
> Why not going further and removing the need of `s_interval` and instead of getting a DataFrame object getting a Series one. That way you can still access index and values, and therefore it will be able to do Intraday plots as well because it doesn't try to access AdjClose or so based on that interval.

Issue is that some functions require more than 1 input (some take OHLC) for example.  Can do this where applies, but would require logic in the controller instead...

DidierRLopes:
> > Why not going further and removing the need of `s_interval` and instead of getting a DataFrame object getting a Series one. That way you can still access index and values, and therefore it will be able to do Intraday plots as well because it doesn't try to access AdjClose or so based on that interval.
> 
> Issue is that some functions require more than 1 input (some take OHLC) for example. Can do this where applies, but would require logic in the controller instead...

Yea, I would add that intelligence on the controller. That way TA becomes available to any type of data regardless of it's index which is what makes sense from the beginning

piiq (approved):

jmaslek:
Have noticed that selecting intraday data has AdjClose = Close, so dropping the need to use ["Close"] if s_interval != "1440min" and always use Adj Close.

With how many places this has been in gonna start chipping away

jmaslek:
need tests and make sure discord is working still

jmaslek:
@DidierRLopes you can review now.

@northern-64bit Can you double check my discordbot :)

Tests now...

discordbot/stocks/technical_analysis/cci.py
```diff
@@ -41,13 +41,15 @@ async def cci_command(ctx, ticker="", length="14", scalar="0.015", start="", end
...
             end = datetime.strptime(end, cfg.DATE_FORMAT)
-
-        if not length.lstrip("-").isnumeric():
-            raise Exception("Number has to be an integer")
-        length = float(length)
-        if not scalar.lstrip("-").isnumeric():
-            raise Exception("Number has to be an integer")
-        scalar = float(scalar)
+        # pylint
+        try:
```
northern-64bit, line 45: I think that the previous solution is better than try except, but keep the int(lenght). Why did you just change it here to try except? The rest of the discord bot looks excellent!

jmaslek: So the previous way was not working with the default strings.

I went with this quote just to only make 1 call from string to number.  I was also confused why there was a strip("-") in there.

northern-64bit: Didn't realise that it wasn't working before :)

DidierRLopes (approved):

PR #1159, opened 2022-01-08
Dependency Audit

colin99d:
I ran through the current dependencies we have to see which ones were unused. It looks like the following were not needed:         
    - GitPython
    - sentiment-investor
    - pyrsistent
    - jsonschema


colin99d:
@Chavithra and I already agreed to the test skip in my baseclass PR. Just adding this here so the tests pass.

jmaslek:
Does the dps/volexch function work without dnspython?  I added it recently since it wasn't working without.  Also I think degiro-connector and pyally just get imported as different names...

colin99d:
> dps/volexch

How do I find dps/volexch

jmaslek:
> > dps/volexch
> 
> 
> 
> How do I find this.

stocks/dps

colin99d:
> > > dps/volexch
> > 
> > 
> > How do I find this.
> 
> stocks/dps

You are right.

jmaslek:
have you done through all the `gst` scripts and make sure the removed deps aren't hidden somewhere as a sub-dependency?

aia:
Gitpython is used here:
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/gamestonk_terminal/loggers.py#L10

colin99d:
It looks like these are used just with other names. My mistake. Will close. Thanks for the reviews.

aia:
This is a good though to keep an eye on used dependencies. Thank you Colin!

Even if we find out that all of our dependencies in use, regularly checks are important.

PR #1162, opened 2022-01-08
testing for pygments fix

jmaslek:
don't look at my branch name


colin99d (approved):
If this is all it takes imma cry.

jmaslek:
> If this is all it takes imma cry.

must be new here

jmaslek:
Future people trying to solve this again:  I noted that pygments 2.10.2 was being installed on the runner so I added that to poetry.

Now charset norm was at 2.0.10 in the lock file so I bumped that in the build/yml files

DidierRLopes (approved):

PR #1163, opened 2022-01-08
improve log plot

DidierRLopes:
This should fix btc plotting as well 😄


jmaslek:
Man this log plotting is killing me.  This is definitely better, but as an example off the bitcoin chart, we lose information that we get in the linear chart (like is it 65k or 62k -- which is a big enough difference).

Also on zooming it goes back to scientific notation


Figure_1

DidierRLopes:
> Man this log plotting is killing me. This is definitely better, but as an example off the bitcoin chart, we lose information that we get in the linear chart (like is it 65k or 62k -- which is a big enough difference).
> 
> Also on zooming it goes back to scientific notation
> 
> Figure_1

I don't think you have the last version. My zoom is working fine
Screenshot 2022-01-08 at 22 35 25

jmaslek:
well.  fuck me then lol

jmaslek (approved):

PR #1164, opened 2022-01-09
Added Interval

colin99d:
Added the ability to choose the interval to display. Not automatically filtering the interval options based on what yfinance allows because that makes the software buggy.


DidierRLopes:
Can you add "notebookreports" and "dashboards" to tests/labels since you ae at it 😋

piiq (approved):

PR #1167, opened 2022-01-10
Add logging decorators to simplify addition of logging on codebase

DidierRLopes:


DidierRLopes:
This should merge into #1116

This is what I was talking about @piiq :)

colin99d (changes requested):
I made a new file for decorators called decorators.py. Do you think this should be moved to that file?

piiq:
> I made a new file for decorators called decorators.py. Do you think this should be moved to that file?

@DidierRLopes I would also suggest not having the word "decorator" in the name of a decorator, if that's ok?
Maybe `@log_execution` instead of `@log_decorator`

DidierRLopes:
> I made a new file for decorators called decorators.py. Do you think this should be moved to that file?

This branch is not up-to-date with there existing a decorators.py file I think! Should we do that after this gets merged?

colin99d (approved):

DidierRLopes:
@aia you ok with merging this? So we can start implementing logging everywhere

gamestonk_terminal/loggers.py
```diff
@@ -12,7 +12,7 @@
 import gamestonk_terminal.config_terminal as cfg
 
 logger = logging.getLogger(__name__)
-LOGFORMAT = "%(asctime)s|%(levelname)s|%(name)s|%(funcName)s|%(lineno)s|%(message)s"
+LOGFORMAT = "%(asctime)s|%(levelname)s|%(name)s|%(funcName)s|%(message)s"
```
aia, line 15: I'd leave lineno in. It will not make sense for start and end logging, but for the rest of log statements it can be useful.

gamestonk_terminal/helper_funcs.py
```diff
@@ -810,3 +810,49 @@ def inner(*args, **kwargs):
 def system_clear():
     """Clear screen"""
     os.system("cls||clear")  # nosec
+
+
+def log_execution(func=None, log=None):
```
aia, line 815: Can we give it a more descriptive name?

Maybe log_start_end?

piiq: > Can we give it a more descriptive name?
> 
> Maybe log_start_end?

`log_execution_time`?

DidierRLopes:
This is replaced by #1173 which will merge to main

PR #1168, opened 2022-01-10
print_help when jump back to previous menu

devwolff:
Its nice to not having to type `?` to see the help string every time we jump back to previous menu

### DIFF
--- a/gamestonk_terminal/feature_flags.py
+++ b/gamestonk_terminal/feature_flags.py
@@ -44,4 +44,4 @@ ENABLE_QUICK_EXIT = strtobool(os.getenv("GTFF_ENABLE_QUICK_EXIT", "False"))
 OPEN_REPORT_AS_HTML = strtobool(os.getenv("GTFF_OPEN_REPORT_AS_HTML", "True"))
 
 # Enable auto print_help when exiting menus
-ENABLE_EXIT_AUTO_HELP = strtobool(os.getenv("GTFF_ENABLE_EXIT_AUTO_HELP", "True"))
+ENABLE_EXIT_AUTO_HELP = strtobool(os.getenv("GTFF_ENABLE_EXIT_AUTO_HELP", "False"))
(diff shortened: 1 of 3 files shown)


DidierRLopes:
Hey @devwolff, any reason you close this PR?

I think this is a good thing. The only thing I'd say is to have the default set to False

devwolff:
Hey @DidierRLopes, i wasn't sure it was worth it. Because if we want to it work as intended, we would have to have "if gtff.ENABLE_EXIT_AUTO_HELP:" at each menu command. Otherwise, say if we try to enter a ticker dependent menus before `load` ticker, it would print a menu right after the help text such as "use `load` before using this menu..". I dont think we would want that. Any suggestions?

DidierRLopes:
> Hey @DidierRLopes, i wasn't sure it was worth it. Because if we want to it work as intended, we would have to have "if gtff.ENABLE_EXIT_AUTO_HELP:" at each menu command. Otherwise, say if we try to enter a ticker dependent menus before `load` ticker, it would print a menu right after the help text such as "use `load` before using this menu..". I dont think we would want that. Any suggestions?

yea, this is definitely not straightforward to implement given how our architecture is laid out!

I created a PR https://github.com/devwolff/GamestonkTerminal/pull/1 in this branch with this implemented, check that you are happy with it and then we can merge it 😄

devwolff:
very neat! Must say structuring everything using self.queue and jobs_cmds is great

DidierRLopes:
> very neat! Must say structuring everything using self.queue and jobs_cmds is great

Does it work was you were expecting? If yes, let's merge it :)

I also tried with running a routine with a `quit` and it didn't show which is expected because in a routine you only want to see the outputs.

Thanks! It was a big effort but worth it IMO. In the future we need to make each controller aware of it's location so indirect jumps are more straightforward, like from "screener" to "comparison analysis" or "comparison analysis" to "portfolio optimisation".

In the future hopefully experienced users won't even need any help as they will know location of all commands, but for that we need to develop at a slower pace hahaha there are new commands every day 🙄

DidierRLopes (approved):

devwolff:
> > very neat! Must say structuring everything using self.queue and jobs_cmds is great
> 
> Does it work was you were expecting? If yes, let's merge it :)
> 
> I also tried with running a routine with a `quit` and it didn't show which is expected because in a routine you only want to see the outputs.
> 
> Thanks! It was a big effort but worth it IMO. In the future we need to make each controller aware of it's location so indirect jumps are more straightforward, like from "screener" to "comparison analysis" or "comparison analysis" to "portfolio optimisation".
> 
> In the future hopefully experienced users won't even need any help as they will know location of all commands, but for that we need to develop at a slower pace hahaha there are new commands every day 🙄

Yes, it does as intended. Thanks 

hahaha there are new commands every day indeed. Very impressive work maintaining them all

PR #1170, opened 2022-01-10
Add an API wrapper for terminal functions to be used in python

piiq:
This PR introduces an implementation of an API wrapper that make terminal functions accessible from plain python environment (script or notebooks).

The goals addressed here are:
- make the terminal contexts importable
- keep the same structure for all the commands as we have them in the terminal
- avoid importing controllers to avoid creating a prompt session
- having menu models available separately from the commands

With this functionality we can can have similar experience in terminal and python
image

Feedback and ideas are very welcome


colin99d (commented):
This is super interesting! I think the models in particular could be really helpful.

piiq:
The major "downsides" of this approach are:

- inability to use the same api to import a single function from some submenu
image

- when importing gst EVERYTHING is imported (no lazy imports)

@Chavithra maybe you have some ideas?

piiq:
> I would just add a folder with some examples of what you can do.

I was planning to update the existing reports to use this API

DidierRLopes:
> > I would just add a folder with some examples of what you can do.
> 
> I was planning to update the existing reports to use this API

even better 😄

piiq:
This covers pretty much everything in the stocks context. I would suggest merging it when tests pass and inviting more contributors to thread the api through other contexts in separate PRs.

Some things were not implemented because the controllers require refactoring and it was a blocker:
- ca (no propper mvc implemented)
- ins (no mvc related to presets management, openinsider needs refactoring)
- options (requires mvc refactoring and moving controllers to subfolders)
- screener (no mvc related to presets management, yfinance view is coupled with the model)

This type of refactoring is not in the scope of this PR

DidierRLopes:
> This covers pretty much everything in the stocks context. I would suggest merging it when tests pass and inviting more contributors to thread the api through other contexts in separate PRs.
> 
> Some things were not implemented because the controllers require refactoring and it was a blocker:
> 
> * ca (no propper mvc implemented)
> * ins (no mvc related to presets management, openinsider needs refactoring)
> * options (requires mvc refactoring and moving controllers to subfolders)
> * screener (no mvc related to presets management, yfinance view is coupled with the model)
> 
> This type of refactoring is not in the scope of this PR

I agree with this. Can you add some TODOs in the code for that, so we don't forget about it?

DidierRLopes (approved):

piiq:
> I agree with this. Can you add some TODOs in the code for that, so we don't forget about it?

Already there

piiq:
One thing to note when tying out this API is that the language server in VS Code provides a significantly less pleasant experience compared to Jupyter Lab

PR #1161, opened 2022-01-08
Terminal Wide Rich

colin99d:
This is my idea for how to implement rich terminal wide. @jmaslek would love to know what you think.

PR #1175, opened 2022-01-11
Adding the historical endpoint from sentiment investor

minhhoang1023:
Adding the historical endpoint and plot hourly-level data of RHI and AHI.
Note:
This command would break if we were to use the export function with the following error

FileNotFoundError: [Errno 2] No such file or directory: '/Users/minh/Desktop/DareData/GameStonkFork/GamestonkTerminal/exports/common/behavioural_analysis/his_20220111_112432.xlsx'

This is a common error across common commands since there's no common sub-folder within the exports folder. I was testing other commands under common and the same thing happened (e.g. headlines from Finbrain).

This is a known issue as I already mentioned it to Theodore.


jmaslek:
Hi @minhhoang1023!

Looking at your description, you should be able to replace "common" with "stocks" in the export path.  

Check out the qa_view.py in the common/quantitative analysis folder for an example.  If the exports/stocks/behavioral analysis folder doesn't exist, can you create one with a .gitkeep?

gamestonk_terminal/common/behavioural_analysis/sentimentinvestor_view.py
```diff
@@ -1,219 +1,71 @@
...
+        plt.gcf().autofmt_xdate()
+
+        if gtff.USE_ION:
+            plt.ion()
+
+        plt.show()
+
+        export_data(
+            export,
+            os.path.dirname(os.path.abspath(__file__)),
```
jmaslek, line 68: So this can be `os.path.dirname(os.path.abspath(__file__)).replace("common", "stocks") to export to the stocks folder.

gamestonk_terminal/stocks/behavioural_analysis/ba_controller.py
```diff
@@ -127,9 +131,11 @@ def print_help(self):
     metrics       core social sentiment metrics for this stock
     social        social media figures for stock popularity
     historical    plot the past week of data for a selected metric{res}
+    his           plot historical RHI and AHI data by hour
```
jmaslek, line 134: So these old SI functions can be removed from this string.  And the {res} should be moved to the end of the his function.

Also can we rename it `hist`

gamestonk_terminal/stocks/behavioural_analysis/ba_controller.py
```diff
@@ -890,6 +896,56 @@ def call_historical(self, other_args: List[str]):
...
+        other_args, EXPORT_BOTH_RAW_DATA_AND_FIGURES)
+
+        if ns_parser:
+            sentimentinvestor_view.display_historical(
+                    ticker=self.ticker,
+                    start=ns_parser.start,
+                    end=ns_parser.end,
+                    limit=ns_parser.number,
+                    export=ns_parser.export
+            )
```
jmaslek, line 947: So we should also allow for the user to print out some data to the terminal.  The parse_known_args_and_warn can add the following: (raw=True, limit=10).  This will add
`ns_parser.raw` and` ns_parser.limit` by default.

gamestonk_terminal/common/behavioural_analysis/sentimentinvestor_view.py
```diff
@@ -1,219 +1,71 @@
...
+        interval_locator = mdates.HourLocator(interval=6)
+        plt.gca().xaxis.set_major_locator(interval_locator)
+        plt.gca().xaxis.set_major_formatter(mdates.DateFormatter("%Y-%m-%d %H:%M"))
+        plt.gcf().autofmt_xdate()
+
+        if gtff.USE_ION:
+            plt.ion()
+
+        plt.show()
+
```
jmaslek, line 65: So here you can add a raw argument, then `if raw:  print data` as we do elsewhere.

minhhoang1023:
> 

Yep saw it. Indeed, it makes more sense for the user to export to the `stocks` folder, instead of creating a sub-folder called `common` inside `exports`.

As mentioned earlier, I notice that some other common commands are exporting to `common` instead of `stocks`. I could make a quick fix though it'd be better in another PR, and not this one

gamestonk_terminal/stocks/behavioural_analysis/ba_controller.py
minhhoang1023: what's our intention with the {res} and {dim} styling? With {dim} the text is greyed out when the ticker is not yet loaded.

jmaslek: So the {res} was just to reset the style so everything after wasn't dim (and the dim indicates you need a ticker to use the function as you say).  We are actually completely changing the styling in PR #1161 so feel free to ignore this comment for now

jmaslek:
> Yep saw it. Indeed, it makes more sense for the user to export to the `stocks` folder, instead of creating a sub-folder called `common` inside `exports`.
> 
> As mentioned earlier, I notice that some other common commands are exporting to `common` instead of `stocks`. I could make a quick fix though it'd be better in another PR, and not this one

Definitely can be addressed elsewhere.  We also need to change up because some common functions can be called from other contexts than 'stocks' (like crypto). So defaulting to exporting into stocks is not correct universally.

minhhoang1023:
@jmaslek Fixed your comments and also added the tests for the command, and remove tests for the deprecated commands.

jmaslek:
very nice work!  Especially with adding tests :)

I am getting the following error:
```
  File "/Users/james/PycharmProjects/GamestonkTerminal/gamestonk_terminal/stocks/behavioural_analysis/ba_controller.py", line 731, in call_hist
    default=(self.start + timedelta(days=7)).strftime("%Y-%m-%d"),
TypeError: can only concatenate str (not "datetime.timedelta") to str
```

We also did quite a big refactor last week and right as I type this, so the @try_excepts are removed and we changed the print style.  Let me know if you have any questions about the styling when merging!

minhhoang1023:
I fixed the Type Error bug and also added the requirement that `--start` and `--end` arguments cannot be specified individually. Tested with arguments and without arguments and it works fine.

jmaslek (approved):
Nice work!!

Once the tests pass I will merge this

minhhoang1023:
wohoo! It passed 🎉

PR #1186, opened 2022-01-14
Delete testdf.csv

colin99d:
Is there a reason this random CSV is on the main file branch? It is not referenced anywhere.


jmaslek:
Nope. I actually deleted it in the rich PR

PR #1187, opened 2022-01-14
Remember Contexts

colin99d:
Currently if you leave the stocks menu and come back you have to retype GME. That sucks. Now you don't have to. Currently I only changed the stocks menu because I my implementation to be reviewed before I change all of them.
Also, I changed the PATH to be a static variable so that it can be accessed without creating an instance of the class. Figured this could help for other features.


gamestonk_terminal/feature_flags.py
```diff
@@ -45,3 +45,6 @@
 
 # Enable auto print_help when exiting menus
 ENABLE_EXIT_AUTO_HELP = strtobool(os.getenv("GTFF_ENABLE_EXIT_AUTO_HELP", "False"))
+
+# Remember contexts during session
+REMEMBER_CONTEXTS = strtobool(os.getenv("REMEMBER_CONTEXTS", "True"))
```
DidierRLopes, line 50: On the getenv you want "GTFF_REMEMBER_CONTEXTS"

DidierRLopes (commented):
Nice work! I would recommend:
* Doing the `.menu()` outside parent classes, otherwise as a code reader it looks odd that you access a new menu through instantiating a class.

And then once you implement this for all, let's merge it 😄

DidierRLopes (approved):
@piiq @jmaslek you happy with this as well?

jmaslek (approved):

jmaslek:
niceee

PR #1188, opened 2022-01-14
Brotli.py check

colin99d:
No one seems to know they have brotli installed, yet it messes up a lot of tests. This hook does not allow tests to be run locally if brotli is installed to prevent the YAMLs from being ruined.


DidierRLopes:
Can you merge with main? I think this has been fixed!

.github/workflows/test.yml
```diff
@@ -240,6 +240,12 @@ jobs:
         run: |
           pip install --requirement requirements-full.txt
 
+      # Is there a better way to do this?
```
DidierRLopes, line 243: Can you add "# TODO: Is there a better.." that way we can search in the code for that keyword and see what there is to do

DidierRLopes (approved):

PR #1190, opened 2022-01-15
Fix + improve efficient frontier

jmaslek:
2 Things here:

1. Fixed a bug where the efficient frontier was throwing an instantiation error because we changed the constraints by adding the maxsharpe to the curves
2. Add labels to individual assets (cough @deeleeramone )
3. (Not really a feature but change all default times to 1yr not 3month)
4. Colin forgot to update requirements 

EF2

### DIFF
--- a/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
+++ b/gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
@@ -1,11 +1,11 @@
 """Optimization View"""
 __docformat__ = "numpy"
 
+import copy
 import math
 from typing import List
 
 import matplotlib.pyplot as plt
-
 from matplotlib.lines import Line2D
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
```diff
@@ -342,13 +342,20 @@ def display_ef(
     ef, rets, stds = optimizer_model.generate_random_portfolios(
         stocks, period, n_portfolios
     )
+    # The ef needs to be deep-copied to avoid error in plotting sharpe
+    ef2 = copy.deepcopy(ef)
```
DidierRLopes, line 346: interesting, had never used this before!

DidierRLopes (approved):

gamestonk_terminal/portfolio/portfolio_optimization/optimizer_view.py
jmaslek: me neither.  I had to dive into the portfolio source code

PR #1193, opened 2022-01-15
Custom data context

jmaslek:
As discussed, this will add the ability for a user to use their own custom data sets within the terminal.  

The test cdv I am using for examples is just 1000 samples from the sklearn California housing dataset (I spent way too long thinking of a sample dataset to use).  

Please suggest features.


DidierRLopes:
This looks the start of something great! Some things I would add:
* head to allow to select number of rows + sorting
* tail to do the same
* plot instead of having `--vs` could we have `-x` it's much more intuitive

Adding `qa` and `pred` from common here makes sense!

jmaslek:
> This looks the start of something great! Some things I would add:
> 
> * head to allow to select number of rows + sorting
> * tail to do the same
> * plot instead of having `--vs` could we have `-x` it's much more intuitive
> 
> Adding `qa` and `pred` from common here makes sense!

Got through these additions!  Let me know if you are able to break it while I try to add some tests/website etc

DidierRLopes:
It looks good to me!

DidierRLopes (approved):

PR #1204, opened 2022-01-17
Allocation Command

northern-64bit:
# Description

Adds the allocation command to the portfolio context. After loading a portfolio and using the command, a table (or pie chart if choosen) of the allocation gets displayed. The allocation command takes etf holdings into consideration and adds them to the table or directly to an existing stock position. This is also preparing for allocation handling on the standard portfolio report.

# How has this been tested?

It has been tested by using it on different portfolios with different etfs and using all parameters. It didn't break or output any errors.

# Checklist:

I don't know how to do this
I don't know what to do here, help? Isn't this some kind of code test?

# Others


gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -251,6 +251,39 @@ def plot_ef(
     return ImageReader(imgdata)
 
 
+def al(data: pd.DataFrame, graph: bool):
```
colin99d, line 254: This function name is a little ambiguous. I would rename to something like display_allocation

northern-64bit: Fixed :)

colin99d:
Screen Shot 2022-01-18 at 2 37 56 PM
Looks really good so far. Is there a way to make the pie chart prettier when the slices are really small?

northern-64bit:
> Looks really good so far. Is there a way to make the pie chart prettier when the slices are really small?

bild
Something similar to this may be a possible solution, but Idk if it's a good one

colin99d:
> ! Something similar to this (but with the portfolio) may be a possible solution and then show the small holdings on the side thing, but Idk if it's a good one

Is there a way to have the key on the side instead of in the graph?

DidierRLopes:
> > Looks really good so far. Is there a way to make the pie chart prettier when the slices are really small?
> 
> bild Something similar to this (but with the portfolio) may be a possible solution and then show the small holdings on the side thing, but Idk if it's a good one

we can probably use "Others" when slice is too small and then display in print all holdings.

Also, you can use parameter `explode` so the slices come off a bit if you want, not sure if it is helpful in this case though

northern-64bit:
> > ! Something similar to this (but with the portfolio) may be a possible solution and then show the small holdings on the side thing, but Idk if it's a good one
> 
> Is there a way to have the key on the side instead of in the graph?

bild
Having the keys in the labels is an option like this but with better formatting

> we can probably use "Others" when slice is too small and then display in print all holdings.

This is also an option, but it could maybe be solved by the approach in the picture above maybe.

jmaslek (approved):

PR #1216, opened 2022-01-19
Added Open MIT options course

deeleeramone:
This adds a link to the stocks/resources for an Open MIT course on equity options.


DidierRLopes:
> This adds a link to the stocks/resources for an Open MIT course on equity options.

The only thing needed is putting the link between < and >. Then the tests will pass and we can merge this!

DidierRLopes (approved):

PR #1220, opened 2022-01-20
make whatif works for decrease as well

devwolff:
Screen Shot 2022-01-20 at 1 05 24 PM

Currently the command doesn't work for decrease in value.

Screen Shot 2022-01-20 at 1 08 31 PM


DidierRLopes (approved):
Very wel spotted!

PR #1221, opened 2022-01-20
Update pre commit

colin99d:
# Description

This PR will allow us to check all config_terminal settings before a push. This is also for visibility so that we can determine if there is a better way to do this.

# How has this been tested?

Experimented with incorrect settings in the check.

# Checklist:

# Others


gamestonk_terminal/config_terminal.py
```diff
@@ -78,9 +78,8 @@
 DG_TOTP_SECRET = os.getenv("GT_DG_TOTP_SECRET") or None
 
 # https://developer.oanda.com
-OANDA_ACCOUNT_TYPE = (
-    os.getenv("GT_OANDA_ACCOUNT_TYPE") or "practice"
-)  # "live" or "practice"
+DEBUG_MODE = False
```
DidierRLopes, line 81: I think this was not intentional

jmaslek:
Just had a thought.  Instead of having this debug as a feature flag, can we just have a --debug option on the terminal.py ?

colin99d:
> Just had a thought. Instead of having this debug as a feature flag, can we just have a --debug option on the terminal.py ? So that I would run `python terminal.py --debug` without worrying about setting this.

This seems like the best option. How would I implement this?

jmaslek:
> This seems like the best option. How would I implement this?

Not 100% sure, but in terminal.py there is the `__main__`  routine.  You probably want to check if --debug is in that and then set some kind of variable to inform that debug should be true.

custom_pre_commit/check_config_terminal.py
```diff
@@ -0,0 +1,141 @@
...
+        The setting to check
+    value : str
+        The correct value of the setting
+
+    Returns
+    ----------
+    correct : bool
+        Returns whether the setting was already correct
+    """
+    print(setting)
```
DidierRLopes, line 108: I don't think you want this

custom_pre_commit/check_config_terminal.py
```diff
@@ -0,0 +1,141 @@
...
+def main():
+    base = os.path.dirname(os.path.dirname(__file__))
+    path = os.path.join(base, "gamestonk_terminal", "config_terminal.py")
+    with open(path) as file:
+        lines = file.readlines()
+        lines = [line.rstrip() for line in lines]
+
+    returns = [check_setting(lines, k, v) for k, v in settings.items()]
+
+    print(returns)
```
DidierRLopes, line 127: neither this

gamestonk_terminal/config_terminal.py
```diff
@@ -7,10 +7,10 @@
     load_dotenv(env_files[0])
 
 # Set to True to see full stack traces for debugging/error reporting
-DEBUG_MODE = False
+DEBUG_MODE = True
```
DidierRLopes, line 10: False

gamestonk_terminal/config_terminal.py
```diff
@@ -7,10 +7,10 @@
...
 
 # By default the jupyter notebook will be run on port 8888
-PAPERMILL_NOTEBOOK_REPORT_PORT = "8888"
+PAPERMILL_NOTEBOOK_REPORT_PORT = "7757"
```
DidierRLopes, line 13: 8888

gamestonk_terminal/config_terminal.py
```diff
@@ -78,9 +78,8 @@
 DG_TOTP_SECRET = os.getenv("GT_DG_TOTP_SECRET") or None
 
 # https://developer.oanda.com
-OANDA_ACCOUNT_TYPE = (
-    os.getenv("GT_OANDA_ACCOUNT_TYPE") or "practice"
-)  # "live" or "practice"
+OANDA_ACCOUNT_TYPE = os.getenv("GT_OANDA_ACCOUNT") or "REPLACE_ME"
```
DidierRLopes, line 81: GT_OANDA_ACCOUNT_TYPE

terminal.py
```diff
@@ -316,7 +318,12 @@ def terminal(jobs_cmds: List[str] = None):
 
 if __name__ == "__main__":
     if len(sys.argv) > 1:
-        if ".gst" in sys.argv[1]:
+        if "--debug" in sys.argv:
+            os.environ["DEBUG_MODE"] = "true"
+            sys.argv.remove("--debug")
+        else:
+            os.environ["DEBUG_MODE"] = "false"
+        if len(sys.argv) > 1 and ".gst" in sys.argv[1]:
```
DidierRLopes, line 326: Can you remove that first condition, it is already done in line 320, we just need the 2nd part

DidierRLopes (commented):
Some comments, but if it works it looks good to me

terminal.py
colin99d: You need it twice because if you remove "--debug" you may be left with only one argument.

DidierRLopes: Gotchaaa! Sry, missed that.

DidierRLopes (approved):

PR #1246, opened 2022-01-22
portfolio: Small feature + controller tests

jmaslek:
# Description

Added tests to the portfolio controller 

Added a new portfolio function: drawdown

mdd

Fixed a quick bug in prediction/mc where -n was not an int type

As an aside: this portfolio menu/ model needs to be refactored.  It is a nightmare to build off of.


gamestonk_terminal/portfolio/portfolio_controller.py
```diff
@@ -93,6 +85,7 @@ def print_help(self):
 [info]Graphs:[/info][cmds]
     rmr         graph your returns versus the market's returns
     al          displays the allocation of the portfolio
+    mdd         display portfolio drawdown[/cmds]
```
DidierRLopes, line 88: Why is the name `mdd`?

DidierRLopes (approved):
Looks good apart from the name `mdd` being confusing to me

gamestonk_terminal/portfolio/portfolio_controller.py
northern-64bit: I think it's the first letters of "maximum draw down"

northern-64bit (approved):
This looks good, but the command hasn't been added to the Hugo website in this pr so don't forget it. I agree that the portfolio menu and model needs to be refactored.

gamestonk_terminal/portfolio/portfolio_controller.py
jmaslek: > I think it's the first letters of "maximum draw down"

@DidierRLopes This. The common jargon

jmaslek:
Commenting that I bumped down pandas to 1.3.5. Pandas at 1.4 seems to broke a bunch a shit

PR #1250, opened 2022-01-23
Fix Pandas Warning

colin99d:
# Description

Fix the Pandas warning from the issue I posted. Unfortunately, I just had to mute the warning. See this StackOverflow issue: https://stackoverflow.com/questions/20625582/how-to-deal-with-settingwithcopywarning-in-pandas.

# How has this been tested?

Reran the table with no warning showing.

# Checklist:

# Others


jmaslek (approved):

gamestonk_terminal/stocks/discovery/ark_view.py
```diff
@@ -65,15 +65,15 @@ def ark_orders_view(
         df_orders = df_orders[df_orders.direction == "Buy"]
```
northern-64bit, line 65: The correct way to solve the warnings is to add .copy() to every slice here:
```suggestion
        df_orders = df_orders[df_orders.direction == "Buy"].copy()
```

gamestonk_terminal/stocks/discovery/ark_view.py
```diff
@@ -65,15 +65,15 @@ def ark_orders_view(
...
     if sells_only:
         df_orders = df_orders[df_orders.direction == "Sell"]
```
northern-64bit, line 67: Add .copy() here too

northern-64bit (changes requested):
The warning comes from making df_orders a slice of df_orders and then making a slice of this slice. Pandas doesn't like this, so one must add .copy() to the first slices so it is a df thus not making a slice of slice. This has to be added to line 63, 65, 67 and maybe some more that I have missed.

gamestonk_terminal/stocks/discovery/ark_view.py
colin99d: Thanks for the input! I was actually able to triangulate the issue to the ark_model. Somewhere in the code surrounded by the pandas warning mute functions. I tried adding this code to those but the warning still appeared. Do you have any further suggestions?

northern-64bit: I don't know the error here, but it's probably an error of the same type or due to using df[value][value2] insted of iloc or loc. The error is probably here I'm however not sure (line 110-113 or maybe in 102-103 due to not using iloc and loc):
bild

northern-64bit (approved):
I sadly can't find the cause of the warning and this is a good solution.

PR #1262, opened 2022-01-25
Add check_supported and trending endpoint from Sentiment Investor

minhhoang1023:
# Description

- Added the trending endpoint from Sentiment Investor & test suite
- Integrated the check_supported_ticker to verify if the ticker is supported to endpoints required a chosen ticker
- Add valid_hour function to helper_funcs to validate hour of the day in 24-hour notation


jmaslek:
The check ticker works as expected! (can we add a newline after ?)

```
2022 Jan 25, 20:32 (✨) /stocks/ba/ $ hist
Ticker BTC-USD not supported. Please try another one!
```

On the `trend`:  strptime error:
```
2022 Jan 25, 20:33 (✨) /stocks/ba/ $ trend
Error: strptime() argument 1 must be str, not datetime.datetime
```

Couple small other points:

- Can we get rid of the unused sentiment investor key in config_terminal ?
- This should get added to the Hugo server
- Can you add some tests for these functions

As always, let me know if you have any question!

minhhoang1023:
@jmaslek thanks for the review!  Pushed new commits to add:
- Hugo server documentation
- Remove the unused SI keys & update the its test
- Fix the datetime error

I'm looking into the tests and working on adding them for those functions!

minhhoang1023:
Tests done!

jmaslek:
Awesome!  1 last small comment.  The `trend` command does not need to be italicized out on the menu -- since it works with or without a ticker!

jmaslek (approved):

PR #1269, opened 2022-01-26
Adding VaR and Adjusted VaR

northern-64bit:
# Description
Adding value at risk (VaR) command and avar (VaR adjusted for skew and kurtosis; Cornish-Fisher Expansion for details check qa_model) to the qa command. Will add var and avar to portfolio soon, but not in this pr (since it isn't done and due to making sure everything is correct here).

# Testing
Tested this command several times and it worked for different stocks and values


northern-64bit:
Any idea why the tests are failing? @Chavithra @jmaslek

northern-64bit:
> > Any idea why the tests are failing? @Chavithra @jmaslek
> 
> Yes. Since you changed the menu, we need to update the tests. To do so run the following:
> 
> `pytest tests/gamestonk_terminal/stocks/quantitative_analysis --record-mode=rewrite`.
> 
> Then commit the new txt file

I just get this error then:
image

colin99d:
In the future use pytest tests/gamestonk_terminal/stocks/quantitative_analysis --rewrite-expected. Its a wonderful @Chavithra built to simplify the process.

jmaslek:
2 comments:

1) I think you can have the adjusted VaR be just a flag on the VaR
2). You should allow for a variable percentile.  So I could do VaR 92.533 if i wanted

jmaslek:
Can you add the Hugo documentation?

northern-64bit:
> Can you add the Hugo documentation?

Sure, there was no hugo documentation for any qa command so I thought that there should be non. So I'll add documentation for every command in the qa context. By the way I just realised that the help command for the new feature somehow doesn't work, any idea why?

jmaslek:
> Sure, there was no hugo documentation for any qa command so I thought that there should be non. So I'll add documentation for every command in the qa context. By the way I just realised that the help command for the new feature somehow doesn't work, any idea why?

Just checking this now.  All of the `qa` Hugo documentation lives in the common folder.  This is because the qa functions will be usable by other contexts (like custom or economy).  You should delete the stocks/qs and just create a file in common/qa for var

northern-64bit:
> Just checking this now. All of the `qa` Hugo documentation lives in the common folder. This is because the qa functions will be usable by other contexts (like custom or economy). You should delete the stocks/qs and just create a file in common/qa for var

Fixed! The tests currently fail since the dates don't match in the coin gecko tearcheat I think. So the code should be good to go or atleast I hope so.

jmaslek (approved):

PR #1270, opened 2022-01-27
Added logger to baseclass controller

LBolte29:
# Description

- added logger to baseclass controller
- refined sequential logic of switch
- fixed prefix / for multiple commands at terminal startup

# How has this been tested?

- checked logs
- checked that the gainers was still displayed
- checked multiple cmds input in debug mode

# Others


gamestonk_terminal/parent_classes.py
```diff
@@ -55,6 +61,7 @@ def __init__(self, queue: List[str] = None) -> None:
         self.check_path()
         self.path = [x for x in self.PATH.split("/") if x != ""]
 
+        # self.queue = self.switch("".join(queue))
```
DidierRLopes, line 64: Extra comment here

gamestonk_terminal/parent_classes.py
```diff
@@ -116,69 +124,75 @@ def switch(self, an_input: str) -> List[str]:
...
             if not actions[0]:
-                an_input = "home"
-            # Relative path so execute first instruction
-            else:
-                an_input = actions[0]
+                actions[0] = "home"
 
             # Add all instructions to the queue
-            for cmd in actions[1:][::-1]:
+            for cmd in actions[0:][::-1]:
```
DidierRLopes, line 137: If it starts at 0, then I don't think we need `[0:]`

terminal.py
```diff
@@ -75,11 +75,11 @@ def __init__(self, jobs_cmds: List[str] = None):
 
         if jobs_cmds:
             # close the eyes if the user forgets the initial `/`
-            if len(jobs_cmds) > 0:
-                if jobs_cmds[0][0] != "/":
-                    jobs_cmds[0] = f"/{jobs_cmds[0]}"
+            for cmd_i, job_cmd in enumerate(jobs_cmds):
+                if not job_cmd.startswith("/"):
+                    jobs_cmds[cmd_i] = f"/{job_cmd}"
```
DidierRLopes, line 80: I don't think we even need to have `cmd_i`, couldn't we just use `jobs_cmds.append(f"/{job_cmd}")`?

DidierRLopes (commented):
Minor comments, but this looks great!!

Screenshot 2022-01-28 at 00 15 29

There's a random "HELLOOO" on disc_controller ahahah can you remove that please? Seems like something I would have done when testing lol.

Also, can you not display the "home", e.g. instead of:
Screenshot 2022-01-28 at 00 17 14

Having:
Screenshot 2022-01-28 at 00 18 23

On top of that, having just one line would be great!

An additional thing that would be good is having this pattern on the main menu. Currently it keeps on displaying menu which is not ideal. Not in scope of this PR, but would be a good addition!
Screenshot 2022-01-28 at 00 23 58

gamestonk_terminal/parent_classes.py
LBolte29: ha ye, didnt clean properly

terminal.py
LBolte29: mhmm, I think in this case we need it. 
What about this command (just an example cmds don't necessarily make sense):

``` python terminal stocks/disc/gainers economy /stocks/exit```

we have 3 cmds:

1.`stocks/disc/gainers`
2.`economy`
3.`/stocks/exit`

On the first 2 we wanna add the / and on the last one we don't.

Lemme know what you think

Comment (DidierRLopes) [reply terminal.py:80]: I see! I missed that the same list was being used. All good for me!

You could also have done it through a one-liner:
```
jobs_cmds = [job_cmd if job_cmd.startswith("/") else f"/{job_cmd}" for job_cmd in jobs_cmds]
```
😄

**Review by DidierRLopes — APPROVED** [review]

**Comment by DidierRLopes** [comment]
Just make sure to cleanup that comment and the [0:] and happy for this to be merged :)

File: terminal.py (hunk 3 above)
Comment (LBolte29) [reply terminal.py:80]: Ah yes, even better!

**Comment by LBolte29** [comment]
Couldnt find the HELLOOOOOO! lol

**Comment by LBolte29** [comment]
@DidierRLopes also dont really know what you mean with ```An additional thing that would be good is having this pattern on the main menu. Currently it keeps on displaying menu which is not ideal. Not in scope of this PR, but would be a good addition!
```
This is the screen I get for stocks
![image](https://user-images.githubusercontent.com/97528701/151675796-17f59b16-0e5d-443f-be8e-4d0c18dcf8b4.png)

DidierRLopes:
> @DidierRLopes also dont really know what you mean with
> 
> ```
> An additional thing that would be good is having this pattern on the main menu. Currently it keeps on displaying menu which is not ideal. Not in scope of this PR, but would be a good addition!
> ```
> 
> This is the screen I get for stocks
> 
> image

I thought you wouldn't, not really well explained ahah I fixed it on #1285 ! This is what I meant:
Screenshot 2022-01-29 at 21 39 39

PR #1280, opened 2022-01-28
Portfolio class

jmaslek:
Improving the portfolio menu by implementing a class.


colin99d (approved):
Looks good!

gamestonk_terminal/portfolio/portfolio_controller.py
```diff
@@ -183,119 +223,162 @@ def call_add(self, other_args: List[str]):
...
+            console.print(
+                f"Portfolio successfully initiialized with {inputs['Name']}.\n"
+            )
+            return
+        self.portfolio.add_trade(inputs)
+
+        console.print(f"{inputs['Name']} successfully added\n")
+
+    def call_build(self, other_args: List[str]):
+        """Process build  ommand"""
```
northern-64bit, line 285: Spelling mistake here in the doc string

gamestonk_terminal/portfolio/portfolio_controller.py
```diff
@@ -18,25 +17,38 @@
...
+    CHOICES_COMMANDS = [
+        "load",
+        "save",
+        "init",
+        "show",
+        "add",
+        "build",
+        "ar",
+        "rmr",
+        "al",
```
northern-64bit, line 48: The allocation (al) command got removed in this pr so it needs to be removed here. Is there a command that replaces this one? I've tested the show command, but it just shows the transactions and not the entire portfolio.

jmaslek: I did remove it.  I can actually add it back this way and then you can add the improvements you had in mind: 
Screen Shot 2022-02-02 at 11 45 20 AM

northern-64bit: Sounds good! I'll add my improvements in another pr with other commands (VaR for portfolio and more risk metrics), but I also really like that you added this nice looking graph with the allocation over time. Great work!!!

northern-64bit (approved):
This new portfolio class will make it so much easier to add more features and simply more clarity to the code! It's excellent 👍

jmaslek:
> This new portfolio class will make it so much easier to add more features and simply more clarity to the code! It's excellent 👍

Thanks!  Look forward to seeing what you can do with it :)

PR #1282, opened 2022-01-28
decorated functions

LBolte29:
decorated functions and removed assertion tests that are now caught b…y decorator logs

# Description

decorated all functions in controller and view with logs

# How has this been tested?

checked the unit tests - saw logs


LBolte29:
god this was messy

Chavithra (approved):
Tests are passing.
Log is generated.

Question : is it normal that we have some memory address like the following ?
```
I|sha:0b34f904|2022-01-31T19:55:32+0100|INFO|gamestonk_terminal.parent_classes|call_help|0|START params: <__main__.TerminalController object at 0x000001925F91F220>, []|none|
```

LBolte29:
Ye, might need to tackle that in a follow up PR

PR #1299, opened 2022-02-02
improve initial commands in terminal

DidierRLopes:
# Description

This should fix the .gst scripts + using multiple commands at start.

Additionally removed the initial menu being displayed since if the user uses an argument to start they don't want to see that menu because they know what they are doing.

# How has this been tested?

Screenshot 2022-02-02 at 00 22 52

Screenshot 2022-02-02 at 00 23 19

### DIFF
--- a/terminal.py
+++ b/terminal.py
@@ -74,13 +74,7 @@ class TerminalController(BaseController):
         self.queue: List[str] = list()
 
         if jobs_cmds:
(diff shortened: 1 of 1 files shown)


colin99d (approved):
Great work!

DidierRLopes:
> Great work!

Thank you sir 🤝

PR #1311, opened 2022-02-04
funds + custom

jmaslek:
Hit these two easy menus

### DIFF
--- a/gamestonk_terminal/mutual_funds/investpy_view.py
+++ b/gamestonk_terminal/mutual_funds/investpy_view.py
@@ -101,7 +101,7 @@ def display_overview(country: str = "united states", limit: int = 10, export: st
 
 @log_start_end(log=logger)
 def display_fund_info(fund_name: str, country: str = "united states"):
-    """Display fund infomration.  Finds name from symbol first if name is false
+    """Display fund information.  Finds name from symbol first if name is false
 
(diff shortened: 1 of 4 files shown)


gamestonk_terminal/mutual_funds/investpy_view.py
```diff
@@ -136,19 +142,22 @@ def display_historical(data: pd.DataFrame, fund: str = "", export: str = ""):
...
     ax.set_xlim([data.index[0], data.index[-1]])
     ax.set_xlabel("Date")
     ax.set_ylabel("Close Price")
     ax.set_title(f"{fund.title()} Price History")
-    fig.autofmt_xdate()
-    fig.tight_layout(pad=1)
-    if gtff.USE_ION:
-        plt.ion()
-    plt.show()
+    ax.yaxis.set_label_position("right")
```
piiq, line 158: this is set with other style stuff in `theme.style_primary_axis`, so no need to have it in the view

piiq (changes requested):
Hey! Thanks! I left comments in the code.

One more question - does this require cleaning out non-trading days?

jmaslek:
> One more question - does this require cleaning out non-trading days?

It does not.  No intraday

piiq (approved):

PR #1312, opened 2022-02-04
Refactored terminal.py

colin99d:
# Description

We seem to like adding custom flags to terminal.py. This is getting increasingly complex. I simplified this by converting the main function into using argparse. Also, we are integrating the all_scripts command directly into the terminal. This will make our integration tests much easier from now on. Please note that just running -s results in all scripts being ran, while -s stocks results in all scripts with "stocks" in the file path being ran.

I have also integrated a pytest like display of the amount of successes and failures and scripts that failed are printed at the end. Running the scripts also automatically cleans them so that they can be used as integrated tests. Two items are changed :

- "Exit" is added to the end of every script to prevent the workflow from getting stuck 
- "Reset" is removed from the files, unfortunately testing reset adds a lot of complexity, and for now it is easier to run the tests without them.

**Note:** The files themselves are not changed by this process, only the lists generated based on the files.

Things I still need to do before this is ready to push: 

- [x] Don't break on scripts failure
- [x] Introduce a way to keep track of which commands had an error

# How has this been tested?

I ran this through about every edge case I thought of. But @jmaslek is willing to write a thesis for anyone that can find an error.

# Checklist:

# Others


terminal.py
```diff
@@ -305,28 +306,80 @@ def terminal(jobs_cmds: List[str] = None):
...
+            folder = os.path.join(
+                os.path.abspath(os.path.dirname(__file__)), "scripts/"
+            )
+            files = [
+                name
+                for name in os.listdir(folder)
+                if os.path.isfile(os.path.join(folder, name))
+            ]
+            for file in files:
+                if file[-4:] == ".gst":
```
jmaslek, line 373: if file.endswith(".gst")

jmaslek (commented):
Looks good to me.  I'll leave to the real devs to confirm :P

DidierRLopes:
If it has @jmaslek stamp on it, I'm fine with this!

colin99d:
@Chavithra and I are having an issue with capture output. Once we can resolve this we should be good to go.

Chavithra:
Do you it it would be interesting to automatically append `exit` to the queue after loading the script ?

colin99d:
> @colin99d Do you it it would be interesting to automatically append `exit` to the queue after loading the script ?

I think that would be very wise.

DidierRLopes:
Is this ready for review?

gamestonk_terminal/portfolio/portfolio_view.py
```diff
@@ -134,7 +134,7 @@ def display_allocation(portfolio: portfolio_model.Portfolio, export: str = ""):
         if gtff.USE_ION:
             plt.ion()
     except Exception as e:
-        print(e)
+        console.print(e)
```
DidierRLopes, line 137: If there's an error we don't want to show or export something.

In any case, I don't think we need this try-except block here

DidierRLopes (changes requested):
Getting an error with `test_stocks_disc.gst`
Screenshot 2022-02-08 at 22 15 46

If we have 2 errors on the same file, I think only one appears.
Screenshot 2022-02-08 at 22 18 25

If possible would be great to have a:
14/14
11/12
9/9

i.e. success/failures per commands within a file.

In addition, I think that "test" could be a flag and we could filter as you have it.. 
And someone could do:
* `python terminal.py scripts/test.gst --test`
* `python terminal.py scripts/ --test`
* `python terminal.py scripts/ --test -f disc`

colin99d:
@DidierRLopes so you would like to see a test flag and a filter flag separate, instead of the combined method I have used ie. `python terminal.py -t stocks`?

DidierRLopes:
> @DidierRLopes so you would like to see a test flag and a filter flag separate, instead of the combined method I have used ie. `python terminal.py -t stocks`?

Kind of. But even a path to the folder being given instead of having a pre-defined

DidierRLopes:
Path should allowed to be a file, and since `-t` flag is enabled it should output success/fail
Screenshot 2022-02-10 at 00 17 56

Apart from this things are looking great 😄

DidierRLopes (approved):
Let's goooo!

PR #1313, opened 2022-02-04
Implement keys menu

DidierRLopes:
# Description

Screenshot 2022-02-04 at 04 09 20

# How has this been tested?

* Tested a couple of commands
* Can be used by accessing /keys menu


colin99d:
It runs `defined, test passed` even if I type in a bad API key. Is that supposed to happen?

jmaslek:
Same.

DidierRLopes:
@colin99d @jmaslek no. I think this is touching in different env files. Need to dig a bit more.

(What happens when u coding at 4am 😄 )

DidierRLopes:
Should be fixed, can u try now? @jmaslek @colin99d

colin99d:
A key of 1 still works for AV. Also, this destroys my old .env with cherished settings. Is there any way around that?

DidierRLopes:
> Sorry the issues I find keep stacking up! Once we get this stuff resolved this will be a great add.
> 
> * A key of 1 still works for AV.
> * This destroys my old .env with cherished settings
> * I think the .env is also now tracked by git
> * I cannot seem to leave the keys menu with q, quit, exit, or cntrl + c

No need to apologize! This is what reviews are for, thanks!

1. I just checked AV doesn't need any key for the type of data we are accessing. I will change the key on .env to be "freekey"
2. I don't think there's much we can do with regards to this. I think we need to swallow this one in order to have better security going forward
3. I added it explicitly to gitignore. Hopefully this will do.
4. Can this be a problem of your setup? I seem to do everything fine:
Screenshot 2022-02-04 at 16 43 32

colin99d:
> > Sorry the issues I find keep stacking up! Once we get this stuff resolved this will be a great add.
> > 
> > * A key of 1 still works for AV.
> > * This destroys my old .env with cherished settings
> > * I think the .env is also now tracked by git
> > * I cannot seem to leave the keys menu with q, quit, exit, or cntrl + c
> 
> No need to apologize! This is what reviews are for, thanks!
> 
> 1. I just checked AV doesn't need any key for the type of data we are accessing. I will change the key on .env to be "freekey"
> 2. I don't think there's much we can do with regards to this. I think we need to swallow this one in order to have better security going forward
> 3. I added it explicitly to gitignore. Hopefully this will do.
> 4. Can this be a problem of your setup? I seem to do everything fine:
> 
Git is still tracking it. We ran into this earlier, which is why the terminal needs a --debug.

DidierRLopes:
> > > Sorry the issues I find keep stacking up! Once we get this stuff resolved this will be a great add.
> > > 
> > > * A key of 1 still works for AV.
> > > * This destroys my old .env with cherished settings
> > > * I think the .env is also now tracked by git
> > > * I cannot seem to leave the keys menu with q, quit, exit, or cntrl + c
> > 
> > 
> > No need to apologize! This is what reviews are for, thanks!
> > 
> > 1. I just checked AV doesn't need any key for the type of data we are accessing. I will change the key on .env to be "freekey"
> > 2. I don't think there's much we can do with regards to this. I think we need to swallow this one in order to have better security going forward
> > 3. I added it explicitly to gitignore. Hopefully this will do.
> > 4. Can this be a problem of your setup? I seem to do everything fine:
> 
> Git is still tracking it. We ran into this earlier, which is why the terminal needs a --debug. The no exit bug must have been me because its gone now.

Yea, if the file exists it will be tracked... Do u have any suggestion?

colin99d:
> > Git is still tracking it. We ran into this earlier, which is why the terminal needs a --debug. The no exit bug must have been me because its gone now.
> 
> Yea, if the file exists it will be tracked... Do u have any suggestion?

This is hacky but python could create the file if it does not exist. Then we never have to track it, Or we could just keep it how things were, and your command adds the keys as they go.

DidierRLopes:
> > > Git is still tracking it. We ran into this earlier, which is why the terminal needs a --debug. The no exit bug must have been me because its gone now.
> > 
> > 
> > Yea, if the file exists it will be tracked... Do u have any suggestion?
> 
> This is hacky but python could create the file if it does not exist. Then we never have to track it, Or we could just keep it how things were, and your command adds the keys as they go.

Ok, let's do that then! Thanks!

DidierRLopes:
> > > Git is still tracking it. We ran into this earlier, which is why the terminal needs a --debug. The no exit bug must have been me because its gone now.
> > 
> > 
> > Yea, if the file exists it will be tracked... Do u have any suggestion?
> 
> This is hacky but python could create the file if it does not exist. Then we never have to track it, Or we could just keep it how things were, and your command adds the keys as they go.

Yup it worked like a glove. Much better!

colin99d (approved):

PR #1315, opened 2022-02-04
Style common TA

piiq:
# Description

This updates the common ta functions to support external axes mode, exclusion of non-trading days and newer style helpers. The controllers that use the common TA in stocks, crypto, etf have been patched accordingly

# How has this been tested?

.gst file

Some tests (crypto/onchain) started failing locally, but they don't seem to relate to the parts of the code touched by this PR


jmaslek:
Everything looking good except the fibinocci doesn't have price on right (also that function doesn't seem to work -- idk if that's because of something with the data being changed or if the function is just broken)
Screen Shot 2022-02-05 at 12 38 12 PM

jmaslek (approved):

piiq:
@jmaslek i've fixed both the price label and the line in fib, thanks for pointing this out. it DID go wrong because of data reindexing
image

PR #1327, opened 2022-02-07
fix bug (shortterm) and refine error message if ticker missing

LBolte29:
# Description

short term bugfix of https://github.com/GamestonkTerminal/GamestonkTerminal/issues/1309
will be changed again in https://github.com/GamestonkTerminal/GamestonkTerminal/pull/1292/

added an error message if the ticker is missing in Jupyter/reports

Feel free to reject the PR


DidierRLopes (changes requested):
Screenshot 2022-02-08 at 22 11 29

Can you make sure all of them working since we are at it?

LBolte29:
Hey, just to make sure we are on the same page: 
Are you asking me to change the `4` command to due_diligience? 
image
image

LBolte29:
There is actually another bug I can have a look at
image

DidierRLopes:
@LBolte29 just saying for us to make sure the reports all work, you should be able to do:
```
due_diligence appl
4 appl
econ_data
...
```

LBolte29:
Yes, so report 1 also does not work. but it seems that there is a problem with the finvizfinance.quote.finvizfinance.TickerFundament function :/ Not sure what to do about this?

we are using it in finviy_compare_model.py in line 46

image

DidierRLopes:
> Yes, so report 1 also does not work and 5 either will take a look

Thanks for that!!

LBolte29:
okay so 2-5 should be working ok! 
1 doesnt work (no errors in terminal, but output output report is buggy). It seems that its actually because of a function from CoinGeckoAPI. Not sure what to do about it
image

DidierRLopes:
> okay so 2-5 should be working ok!
> 1 doesnt work (no errors in terminal, but output output report is buggy). It seems that its actually because of a function from CoinGeckoAPI. Not sure what to do about it
> image

It works for me. This means it is likely due to API key.
Screenshot 2022-02-10 at 00 21 29

So, my suggestion is to have a try-catch and if there's an issue just output on the report "Wrong API key provided" or "Error".

Thanks!

LBolte29:
For me it also completes until 100%, but the notebook shows this 
image
Is this the same for you?

DidierRLopes:
> > For me it also completes until 100%, but the notebook shows this
> > image
> > Is this different for you? Will implement the try-catch there then
> 
> This is because the ticker you are using is not recognised, we can have a check for this at the beginning. Good catch.

or is not listed*

LBolte29:
But does this make sense? I mean aapl should be listed actually, no?
image

In any case, Ill put in a try-catch to check that ;)

DidierRLopes:
> But does this make sense? I mean aapl should be listed actually, no? image
> 
> In any case, Ill put in a try-catch to check that ;)

Sorry, you mentioned an error in CoinGecko API function so I thought you were talking about crypto_report.

LBolte29:
So just did it with an if: .. else: print (Error) output because sometimes the code doesn't break (see last cells) but it then just prints out all stocks

DidierRLopes (approved):

PR #1341, opened 2022-02-09
Add StockAnalysis (multi-year) to the Sector & Industry Analysis menu

JerBouma:
# Description
To understand the differences between companies in the same sector/industry, it is very helpful to look over a larger horizon than the one year Yahoo Finance provides. Therefore, I have added support for StockAnalysis ([source](https://stockanalysis.com/)) that does have multi-year data. Currently the implementation focuses on the Balance Sheet, Income Sheet and Cash Flow statements. An improvement could (and should) be made at a later stage by also including the ratios from StockAnalysis.

**Annually**
SIA Example

**Quarterly**
image

**Trailing** (with Mega cap)
image

# How has this been tested?
This has been tested by including tests within the corresponding folder. See the modules _test_stockanalysis_view_ and _test_stockanalysis_model_.

# Checklist:

# Others


JerBouma:
@DidierRLopes Can you take a peek?
@Chavithra Could you look at the tests I have set-up?

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -38,13 +39,18 @@ class SectorIndustryAnalysisController(BaseController):
...
         "cpis",
         "cpcs",
         "cpci",
         "sama",
         "metric",
+        "bs",
+        "cf",
+        "is",
+        "visualise",
     ]
```
DidierRLopes, line 54: I would not have these 3 commands, they add a bit of noise. 

Instead, let's have a "satma       see all time metrics available" or similar, and then if the user does: `satma` they get a list of all the commands with some separation between balance sheet, cash flow or income statement. But then also allow the user to do `satma ic` to basically just see metrics associated with income statement.

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -38,13 +39,18 @@ class SectorIndustryAnalysisController(BaseController):
...
         "cpic",
         "cpis",
         "cpcs",
         "cpci",
         "sama",
         "metric",
+        "bs",
+        "cf",
+        "is",
+        "visualise",
```
DidierRLopes, line 53: visualise is a long word, is there anything shorter we can use? Also @jmaslek will kill you for writing it the European way 😄

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -119,8 +125,148 @@ class SectorIndustryAnalysisController(BaseController):
         "ev": ("defaultKeyStatistics", "enterpriseValue"),
         "fpe": ("defaultKeyStatistics", "forwardPE"),
     }
-    mktcap_choices = ["Small", "Mid", "Large", "small", "mid", "large"]
+
+    visualise_choices = [
```
DidierRLopes, line 129: In order to avoid duplicating stuff, i.e. if you change one side you need to change the other. Here you can declare this as something like: ```list(sa_keys["BS"].keys()) + list(sa_keys["CF"].keys()) + list(sa_keys["IS"].keys())```

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -244,6 +392,7 @@ def print_help(self):
 [param]Country           : [/param]{self.country}
 [param]Market Cap        : [/param]{self.mktcap}
 [param]Exclude Exchanges : [/param]{self.exclude_exchanges}
+[param]Period            : [/param]{self.period}
```
DidierRLopes, line 395: You forgot to add the "period" command on the menu so users know they can change it

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -724,6 +906,169 @@ def call_metric(self, other_args: List[str]):
                 self.stocks_data,
             )
 
+    def call_bs(self, other_args: List[str]):
```
DidierRLopes, line 909: Let's merge these 3 into one. Also, here you are falling into the same duplication issue. This is bad because let's say there was a new metric, you would have to update it in 3 places which increases a lot the chances of future bugs.

Instead, you can iterate through your dictionary and do something like:
```
for k, v in sa_keys["BS"].items():
   help_text += f"  {k} {(20 - len(k)) * ' '} {v}"
```

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -724,6 +906,169 @@ def call_metric(self, other_args: List[str]):
...
+        ns_parser = parse_known_args_and_warn(
+            parser, other_args, EXPORT_BOTH_RAW_DATA_AND_FIGURES
+        )
+        if ns_parser:
+            (
+                self.stocks_data,
+                self.tickers,
+            ) = stockanalysis_view.display_plots_financials(
+                ns_parser.metric,
+                self.sa_keys,
```
DidierRLopes, line 1058: my suggestion is to have these keys on the `_model` file if they are needed there. Then you can just access them through `stockanalysis_model.sa_keys` in the controller.

DidierRLopes (changes requested):
I added some comments. Nothing too major, some minor things here and there.

For a first feature, this is looking very very nice! Amazing work.

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -724,6 +794,121 @@ def call_metric(self, other_args: List[str]):
...
+        help_text = ""
+        statement_string = {
+            "BS": "Balance Sheet Statement \n",
+            "IS": "Income Statement \n",
+            "CF": "Cash Flow Statement \n",
+        }
+
+        if ns_parser.statement:
+            ns_parser.statement = str(ns_parser.statement).upper()
+            if ns_parser.statement == "BS":
```
DidierRLopes, line 825: If you take one step closer here you notice that the code is very much the same. This is because you compare on the if condition but then created this logic where you actually don't need an if/else. This can be simply:

```
if ns_parser.statement:
   ns_parser.statement = str(ns_parser.statement).upper()

  if ns_parser.statement not in ("IS", "BS", "CF"):
     console.print("This type of information doesn't exist").

   help_text += statement_string[ns_parser.statement]
   for k, v in stockanalysis_model.sa_keys[ns_parser.statement].items():
       help_text += f"  {k} {(20 - len(k)) * ' '} {v} \n"

else:
   for statement, statement_value in stockanalysis_model.sa_keys.items():
      help_text += statement_string[statement]
      for k, v in statement_value.items():
          help_text += f"  {k} {(20 - len(k)) * ' '} {v} \n"
```

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
JerBouma: Fixed!

scripts/test_stocks_sia.gst
```diff
@@ -21,6 +21,12 @@ exchange
 mktcap Large
 country United States
 industry Financial Data & Stock Exchanges
+bs
+cf
+is
+visualise se
```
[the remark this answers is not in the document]
JerBouma: Made the adjustment.

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -724,6 +794,121 @@ def call_metric(self, other_args: List[str]):
...
+            dest="statement",
+            help="See all metrics available for the given choice",
+            choices=["BS", "bs", "IS", "is", "CF", "cf"],
+        )
+
+        ns_parser = parse_known_args_and_warn(parser, other_args)
+
+        help_text = ""
+        statement_string = {
+            "BS": "Balance Sheet Statement \n",
```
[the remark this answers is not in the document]
JerBouma: Makes sense, indeed changed it around. Also added an additional \n, looks prettier now :)

gamestonk_terminal/stocks/sector_industry_analysis/sia_controller.py
JerBouma: Agreed that this is a better format!

DidierRLopes (approved):
Perfect! Great work!

PR #1349, opened 2022-02-11
Improve StockAnalysis functions

JerBouma:
# Description
Made some improvements to the StockAnalysis functions:

- Changed the notation to either thousands, millions or billions.
- Limit function wasn't working properly, it would exceed the default 10 if it could.
- Altered the data to have dates on the index, symbols on the columns. This also benefits data exporting and using the data within the Statistics section.
- Moved the export data functionality to _above_ the limit the amount of companies function. Because the limiting parameter just takes the first 10 (by default) it wouldn't make much sense to export just those 10. I know you can set the limit to 200 for example, but there is simply no scenario where you would want a limit in the first place when you are exporting data since the limitation is not based on any metric whatsoever. The period limitation still applies since that does have value.
- Moved legend to top-left so it at least doesn't hide most recent data.

**image**

image

# How has this been tested?
Ran tests, they worked.

# Checklist:

# Others


JerBouma:
@DidierRLopes Since you also reviewed it last time, shouldn't take too much of your time :)

jmaslek:
On displaying the chart:  that is a hella lot of zeros.  Is it possible to either show everything with the M/B/K, etc (or just pick a common unit like Million) and also fill the nan (I kinda like filling it with "-").  Exporting with all the zeros is what you want, just displaying doesn't look good.

JerBouma:
> On displaying the chart: that is a hella lot of zeros. Is it possible to either show everything with the M/B/K, etc (or just pick a common unit like Million) and also fill the nan (I kinda like filling it with "-"). Exporting with all the zeros is what you want, just displaying doesn't look good.

You mean the table right? So I was thinking that since it is marked as `raw` I shouldn't adjust the table. That is not an issue?

jmaslek:
> You mean the table right? So I was thinking that since it is marked as `raw` I shouldn't adjust the table. That is not an issue?

yeah.  table.  I was just thinking that displaying all those zeros is not all that great.  I won't hold this hostage over it though.

jmaslek (approved):

Issue #1350, opened 2022-02-11
[Bug]


DidierRLopes:
**Describe the bug**
```
scripts/test_crypto_disc.gst: cannot convert float NaN to integer

scripts/test_forex_oanda.gst: 'NoneType' object has no attribute 'request'

scripts/test_port.gst: pie() got an unexpected keyword argument 'normalize'

scripts/test_stocks_ba.gst: Shape of passed values is (6, 1), indices imply (6, 6)

scripts/test_stocks_ca.gst: '5-year trend' is not in list

scripts/test_stocks_fa.gst: Length of headers does not match length of DataFrame

scripts/test_stocks_fa_fmp.gst: Length of headers does not match length of DataFrame

scripts/test_stocks_gov.gst: DateFormatter found a value of x=0, which is an illegal date; this usually occurs because you have not informed the axis that it is plotting dates, e.g., with ax.xaxis_date()

scripts/test_stocks_ins.gst: Length mismatch: Expected axis has 12 elements, new values have 13 elements

scripts/test_stocks_options.gst: 'formatter' must be an instance of matplotlib.ticker.Formatter, not a str

scripts/test_stocks_options_payoff.gst: 'formatter' must be an instance of matplotlib.ticker.Formatter, not a str

scripts/test_stocks_pred.gst: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

**To Reproduce**
`python terminal.py scripts/FILENAME -t` where FILENAME is the name of one of the scripts that failed above.

PR #1357, opened 2022-02-14
Output Missing API Key Message to Console

### LINKED ISSUE #1307: [IMPROVE] Missing API Keys is not always clear
Within Stocks --> Behavioural Analysis the function 'stats' return 'No sentiment stats found.' both when there is no data and when you entered an invalid API key. Same goes for the Reddit and SentimentInvestor functionalities.

 When no API key is set, this should give a different message. It is unclear what is happening now.

**Possibly describe the ideal way to improve this**
If API_KEY variable has the default value, return an Error Message that API key is missing

image

minhhoang1023:
Saw @DidierRLopes's Keys Menu in this [PR](https://github.com/GamestonkTerminal/GamestonkTerminal/pull/1313) and it was awesome ✨ ! Very helpful for new users to debug and know why the requests might not be working.

I implemented this PR to extend the functionality of the Keys Menu. It is a decorator wrapping around the call function in the menu controller and printing a message to the console on the status of the API key and token, in case people forgot to use the keys menu. 

Closes #1307 

# Note
- Only the `call_hr` command was implemented. After your feedback, I'll implement to rest of the command
- To test this PR:
1. Remove the Glassnode API Key (if you have set it)
2. run `python terminal.py cryto/onchain/hr`

image

# Description

# How has this been tested?

# Checklist:

- [] If a feature was added make sure to add it to the corresponding [scripts file](https://github.com/GamestonkTerminal/GamestonkTerminal/tree/main/scripts).

# Others


DidierRLopes:
This is a great idea!!

May I give a suggestion? Instead of checking by actually doing an API call to check that, why not call these all these checks once we start the terminal and then here you just need to get the dictionary to check the status of the variable.

What do you think?

piiq:
And a suggestion from my side.
Should we decorate the view/model function rather than the function in the controller? Rationale for that - we have controller functions that accept a "source" flag as argument. Given that there might be a case when only 1 our of 3 api keys is missing/wrong. And we probably wouldn't want to cut out the whole function because 1 out of 3 sources is not available to the user

minhhoang1023:
> This is a great idea!!
> 
> May I give a suggestion? Instead of checking by actually doing an API call to check that, why not call these all these checks once we start the terminal and then here you just need to get the dictionary to check the status of the variable.
> 
> What do you think?

This would help eliminate waiting time when calling each function. I like the option of checking all of them at first. My only concern is that, when our API sources grow, let's say to dozens, this will cause several seconds of delays from the user side. This on top of the current delay when starting might not be the most optimal. What's your thoughts on this?

minhhoang1023:
> And a suggestion from my side. Should we decorate the view/model function rather than the function in the controller? Rationale for that - we have controller functions that accept a "source" flag as argument. Given that there might be a case when only 1 our of 3 api keys is missing/wrong. And we probably wouldn't want to cut out the whole function because 1 out of 3 sources is not available to the user

Good point! There are indeed some functions in the controller that the users can specify the source. When designing the solutions, I was skeptical between the three options, models vs views vs controllers. And the edge case you mentioned didn't come across, I decided to go for the earliest ones to be called. Thanks for raising this! 🙌

DidierRLopes:
> > This is a great idea!!
> > May I give a suggestion? Instead of checking by actually doing an API call to check that, why not call these all these checks once we start the terminal and then here you just need to get the dictionary to check the status of the variable.
> > What do you think?
> 
> This would help eliminate waiting time when calling each function. I like the option of checking all of them at first. My only concern is that, when our API sources grow, let's say to dozens, this will cause several seconds of delays from the user side. This on top of the current delay when starting might not be the most optimal. What's your thoughts on this?

So, if you are concerned about that what we could do is check if the dictionary of keys status has been updated. If it has you just take that status, if it hasn't you check for it's validity. And then next time you just get the status without needing to do any API calls (hence not introducing delay).

Regarding Theo comment I think it's a very good point. I would do it on the _view personally.

minhhoang1023:
After having discussed this issue at great length with @Chavithra , @piiq , and also Sam, it became obvious that there is a bigger underlying issue. Because of this, we still have not found a single best solution that would solve both the error message problem and logging. Not making this extremely clear would lead us to a less optimal solution, and end up costing us more time in the future. So I'm putting together the list of possible solutions that we have come up with.

## The issue

So the underlying problem here is that we are not following the best practice to handle HTTP requests status code. This is the conclusion we reach in all of my conversations.

This is an example:

image

Due to how we are handling this, **_it's not possible for us to know if the API key is wrong, the server is down or the request simply returns nothing._** 

This caused us not to output the correct error message, and this is a big problem. I have seen multiple people, including myself, get puzzled by this. Even as a developer, on one occasion, I have spent up to 10 mins of debugging to finally realize the errors due to me forgetting to set the key 😅 . By having a clear error message, this problem wouldn't exist in the first place.

Instead of the above, we should have a block of if-else statements with the corresponding error message for each status code. I'm pasting here one block of code done by James that handled this beautifully. 

image

With this implementation, we would not need to do an extra API call to know if the key is successful or not. Users would know exactly what to do next. Without this, it's extremely likely that users consider the errors as a bug from our product. Which is what we definitely don't want to happen!

## Possible solutions
With what we have right now, there are a couple of possible solutions so that we can (1) Output the relevant, user-friendly error message and (2) Obtain the API Key status for our logging purpose

### Solution 1: Do HTTP requests for **_all API keys_** at terminal start time
At start time, we would do a round of requests for all existing API keys, whether or not they will be used later on.

**Pros:**
- Nicely serve both the logging and error message objectives

**Cons:**
- Add ~10 seconds to the already existing terminal starting time. In total, it'd take around 15s to get the terminal start - which definitely not a good experience for users.

While we can use multi-thread to shorten the waiting time, my hypothesis is that it will only get worse as we scale. Once we integrate with hundreds of data sources, and this would be the blocking point.


### Solution 2: Do HTTP request only when a specific command is executed
Instead of checking the status for all API keys, we only check for the specific API keys required to run the command. 

**Pros:**
- No waiting time at the start

**Cons:**
- Double API calls (1st one is to validate the key status and 2nd is to fetch data)

This solution should do the job in the mean time. However, there is a risk that that the API request limit is reached (e.g. a user who do a command 3-4 times consecutively)

### Solution 3: Do HTTP requests for **_all API keys_** in the background using threading
Here we'd perform the API calls in the background to get the status of all keys.

**Pros:**
- No waiting time at start and still give us the key status for logging

**Cons:**
- Consume resources
- Require development time
- Do not solve the error message problem

### Solution 4: Refactor the codebase to handle status code
When doing the refactor, we can output the relevant error message as well as log the API keys status. 

**Pros:**
- Serve both error message and logging purpose

**Cons:**
- Require non-trivial time investment upfront


## Recommendation
The problem we are experiencing now is very normal. As an open-source project, we have a lot of people organically contribute to the terminal which gave us such an impressive pace of development. What is unfortunately meant is we might not follow the best practices all the time. 

But we can fix it now, and include it in the contributing guideline. That's why I would fully recommend solution 4. I am aware that this will require a non-trivial investment of time to refactor all the code. However, this is the right thing to do, and it will save us time in the future. As I am advocating for this big code refactor, and it's not the most fun task in the world, I'd volunteer to take this on. 

That's my recommendation on the matter. Having said that, I will fully support whatever decision is made and am committed to implement even if it's not the solution that I don't 100% agree with.

jmaslek:
@minhhoang1023 all great points!  And I would agree that 4 is the correct way to do it.

One thing I noticed as I was adding apis and whatnot was that not all APIs are as nice as twitter - which specifies a 401 error.  For some sites, if I have an incorrect API key -- I will get a 400 -- which I also get if I throw in random arguments to the query or make the wrong query.  Some 400 errors may come with an error message specifying that the api key was wrong, so going through and refactoring based on each call will likely give you the best handling.

minhhoang1023:
@jmaslek Indeed. Not all APIs have an exact status code. In that case, I am planning to quickly go through the existing API source and do a quick round of checks for edge cases. We should be able to group them into 2-3 categories, e.g. ones with correct status code, while for some we must rely on their error messages too. Will keep you updated!

minhhoang1023:
Update: The PR is ready for review @piiq @DidierRLopes @jmaslek 

In this PR, I aimed to output error messages for the following cases:

## Undefined API Keys
I created `check_api_key` decorator under `decorators.py` to check whether users have set their API key. This decorator should be placed either in `_view.py` or `_controller.py` (when there's no view file). 

If users have not set their API key, the function in `_model.py` won't be executed and the following error message is displayed. At times, it was not clear to me which API source or key I should be setting, so this makes it very clear for users to know.

image


## API Error Handling 
Four error types are being handled:
- Invalid API Keys
- API Keys not authorized for Premium feature
- Empty payload / dictionary
- Invalid arguments (Optional)

The image below provides a good example. However, it's not always possible to distinguish error types using `status_code`. So depending on the API provider, either error messages or exception type will be used instead. This can be referred to in this [Confluence Documentation](https://openbb.atlassian.net/wiki/spaces/ENG/pages/edit-v2/67665937).

image

## Next steps
Having refactored 20+ APIs, it becomes obvious that API error handling comes in multiple forms. There's no single helper function for all cases and developers need to have a look at that specific API. So next step for me is to update the Contributing Guidelines, specifying the best practices and providing some examples to follow.

Look forward to your review. This PR touches so many files so it'd be much appreciated to have more pairs of eyes. thank you! 🙌

piiq:
I want to note that this PR resolves A LOT of errors from #1350 
When I run the test I can see only 3 out of 12 reported errors

👏👏👏

DidierRLopes:
image

piiq (changes requested):
Hey, we have a problem.
image


If I understand correctly - we can't use `console.print` in the models because it pulls a prompt-toolkit session that requires an interactive terminal to function. This leads to an error message when using gst from notebooks.

Any ideas on how to solve this?

minhhoang1023:
hmm strange because before we also have `console.print` in _model.py. I'm looking into it!

minhhoang1023:
@piiq indeed! When I try to comment out the `from .dark_pool_shorts import dps_api as dps` menu, there's no longer any warning. Can you give it a try on your side? If it behaves the same, then we got ourselves a culprit

piiq (approved):
Awesome job done here!!!
I've left minor comments in the review, but in general this is looking very nice.

tests and linting passes locally, integrated tests show 2 or 3 errors (depending if you have an api key). These errors seem unrelated to the scope of this PR

piiq (approved):

PR #1393, opened 2022-02-18
Log exceptions

LBolte29:
logging exceptions
logging settings


DidierRLopes (approved):

Chavithra (approved):
I can see the following functions are used both inside and outside `try/except` :
- `logger.exception`
- `logger.warning`

I suggest using :
- `logger.exception` : only inside a `try/except`
- `logger.warning/error/fatal/critical` : whenever you are outside a `try/except`

PR #1394, opened 2022-02-18
Adding risk metrics to portfolio and qa

northern-64bit:
# Description

Adding risk metrics.

Done:
* Adding var to portfolio
* Adding es to qa
* Add es to portfolio
* Fix distributions of the es

# How has this been tested?
Running the new commands with various different functions


northern-64bit:
@jmaslek Please check out the distributions in the es functions since they don't work properly. Message me via discord if you find anything. Thanks!

gamestonk_terminal/common/quantitative_analysis/qa_view.py
```diff
@@ -947,7 +947,13 @@ def display_line(
...
 
 def display_var(
-    data: pd.DataFrame, use_mean: bool, ticker: str, adjusted_var: bool, percentile: int
+    data: pd.DataFrame,
+    use_mean: bool,
+    ticker: str,
+    adjusted_var: bool,
+    student_t,
+    percentile: int,
+    portfolio: bool,
```
jmaslek, line 956: We should use as many default kwargs as we can here (and is es).  For example, if I just had a Dataframe of prices, I should be able to run `display_var( data)` instead of `display_var(data,True,True,....`. Also what if percentile is not provided?

jmaslek:
Also looks like you are missing Hugo documentation and some tests.

gamestonk_terminal/common/quantitative_analysis/qa_view.py
northern-64bit: I added many default arguments. I won't reduce the number of parameters since they make the code way more usefull in a multiple of scenarios. The default percentile is set to 99.9%.

northern-64bit:
> Also looks like you are missing Hugo documentation and some tests.

This should now have been fixed.

jmaslek:
For implementing on portfolio, you should check the portfolio is not empty:

```
2022 Feb 25, 10:49 (✨) /portfolio/ $ var
Error: 'Portfolio' object has no attribute '_stock_tickers'

Error: 'Portfolio' object has no attribute '_crypto_tickers'

Error: 'Date'

Error: 'Portfolio' object has no attribute 'returns'

2022 Feb 25, 10:49 (✨) /portfolio/ $
```

jmaslek (approved):

PR #1401, opened 2022-02-20
extra ref in main

jmaslek:
didn't check the main.yml fix close enough


JerBouma:
Tests seem to fail? Not sure why this is the case for a Documentation change.

DidierRLopes:
It's not. It's a test :/

DidierRLopes (approved):

jmaslek:
> Tests seem to fail? Not sure why this is the case for a Documentation change.

Guess things just need to be re run sometimes ¯\_(ツ)_/¯

PR #1413, opened 2022-02-22
Bump all deps following breaking changes in multiple libraries

piiq:
tensorflow 2.7 -> 2.8
statsmodel 1.2 -> 1.3
scipy 1.7 -> 1.8
charset-normalizer 2.0.11 -> 2.0.12

pandas-market-calendars has breaking changes introduced by version 3.3 so the version is locked as "~3.2"


piiq:
And it appears that in a m1 rosetta conda env tensorflow 2.8 works if you install it with poetry. No need to pull it separately from conda

Chavithra (approved):
Looks good to me

PR #1419, opened 2022-02-26
Portfolio fix

### LINKED ISSUE #1405: [IMPROVE] Make the Portfolio menu work well with a lot of tickers
When using the Portfolio menu, I noticed that some functions struggle when using a lot of tickers. For example, see the graph below:
Figure_1

Furthermore, when you load in your portfolio it makes sense that when visiting the _portfolio optimisation section_, your portfolio's tickers are used over there. This helps in doing portfolio optimisation with the _current_ portfolio.

I would suggest using the CSV attached to explore if every function accurately depicts the results of the portfolio without graphs getting cluttered by the legend or too many lines.

[Top_30_S&P500.csv](https://github.com/GamestonkTerminal/GamestonkTerminal/files/8110746/Top_30_S.P500.csv)

northern-64bit:
# Description

PR fixes #1405 issue raised by @JerBouma

# How has this been tested?

Testing and using the al command

# Checklist:

- [ ] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/CONTRIBUTING.md).

# Others

### DIFF
--- a/gamestonk_terminal/portfolio/portfolio_controller.py
+++ b/gamestonk_terminal/portfolio/portfolio_controller.py
@@ -134,8 +134,12 @@ Loaded:[/info] {self.portfolio_name or None}
     @log_start_end(log=logger)
     def call_po(self, _):
         """Process po command"""
+        if self.portfolio.empty:
+            tickers = []
+        else:
+            tickers = self.portfolio._stock_tickers + self.portfolio._etf_tickers + self.portfolio._crypto_tickers
         self.queue = self.load_class(
-            po_controller.PortfolioOptimization, [], self.queue
+            po_controller.PortfolioOptimization, tickers, self.queue
         )
 
     # BUG: The commands in pa menu throw errors. First one says that it's related to
@@ -388,8 +392,10 @@ Loaded:[/info] {self.portfolio_name or None}
             parser, other_args, export_allowed=EXPORT_ONLY_FIGURES_ALLOWED
         )
         if ns_parser:
+            if self.portfolio.empty:
+                console.print("[red]No portfolio loaded.[/red]\n")
+                return
             portfolio_view.display_allocation(self.portfolio, ns_parser.export)
-        console.print()
 
     # def call_ar(self, other_args: List[str]):
     #     """Process ar command"""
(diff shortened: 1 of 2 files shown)


JerBouma:
I would suggest to give the user the option to select companies. Especially if companies do not have a similar stock price, it can become cumbersome to compare. The image from James doesn't tell a lot for most of the companies.

jmaslek:
> I would suggest to give the user the option to select companies. Especially if companies do not have a similar stock price, it can become cumbersome to compare. The image from James doesn't tell a lot for most of the companies.

This could be an interesting approach

> For example, it might also be interesting to select the first 10 companies and provide a message that the user can plot specific companies if he provides it.

I dont agree.

> Another addition, normalize each value and add a denomination within the legend. For example "TSLA [$ Thousands]" and "SHELL [$ Hundreds]". As long as it is on the same scale. In the end you don't care about absolute growth anyways, doesn't really matter.

I also dont agree.

I have 4 ideas:

1) We ditch the legend and try to add a hover option (not sure if this would work in MPL)
2) Move the Holdings axis to left and make the right axis labeled by the company (can sketch out if doesn't make sense)
3) Just leave by default and allow user to specify specific tickers.
4) Move legend outside of axis.

JerBouma:
Could you elaborate James? Perhaps selecting companies is indeed enough but plotting all tickers simply doesn't yield any value. If you invest in Amazon and a bunch of penny stocks you would be looking at a graph of Amazon with a colourful floor.

What I would see as value is being able to show the actual growth of stocks within the graph, that is where normalizing adds value.

JerBouma:
Yeah that doesn't make sense, I misinterpreted the graph thinking it was just stock prices.

Think we need to figure out if anyone would even be interested in graphing their entire portfolio like this. Most likely it would be divided in asset classes anyways (i.e. emerging markets equities, bonds, cash, real estate).

Out of scope for this item obviously.

northern-64bit:
> Agreed. So lets merge this for now and figure out what we can do from here?
Sounds good! I'll try to add sector/region/asset type plotting support latter. This was added to the pa menu sort of in the groupby command (but the pa menu has stopped working). The issue with covering the screen isn't all too big since one can just move the graph.

jmaslek (approved):

PR #1421, opened 2022-02-28
Sentiment News correlation with Stock price

DidierRLopes:
Display stock price and headlines sentiment using VADER model over time.

- [x] Tests
- [x] Hugo
- [x] API
- [x] Requirements

Screenshot 2022-03-02 at 03 16 20

But also take care of several bug fixes:
- Degiro colors
- Fix colors of grhist
- Fix granularity of FRED overlay
- Refactor news to model and view
- Defi menu help string


jmaslek (changes requested):
plz no sneak merge while I sleep

DidierRLopes:
> plz no sneak merge while I sleep

??

jmaslek:
Could you provide some guidance on how one would use this function?

jmaslek (changes requested):
Looks like you are missing some styling.  Dates aren't angled.  Is it possible for us to not plot the long lines on nontrading days like with mpf (maybe a @piiq question). I would say get rid of them, but weekends have lots of headlines.


Figure_1

Also:
```
2022 Mar 03, 11:14 (✨) /stocks/ba/ $ stats
No sentiment stats found.
```
Have a hard time imagining nothing for aapl?

JerBouma:
> Looks like you are missing some styling. Dates aren't angled. Is it possible for us to not plot the long lines on nontrading days like with mpf (maybe a @piiq question). I would say get rid of them, but weekends have lots of headlines.

Wouldn't it make sense to move the news of the weekend to the nearest next trading day? Because the influence of the news is not visible until the next trading day anyways. This does falsely report that some news articles are published on a different date but this could be an argument within the function if you want to disable that?

DidierRLopes:
`stats` has become premium, I will remove that function from the controller.

@jmaslek when you mean non trading days, you mean in the candle chart? I know how to do this with `mpl` not with `mpf` but can investigate. Will it not become too noisy with mpf candles?

I don't like the idea of moving around the bars from the news since I think this is what happened in real-life, and that can help you understand certain patterns that wouldn't otherwise. E.g. a news happening on a Friday night right at the close will have a different effect than a news happening on a Monday morning once market opens.

DidierRLopes:
Screenshot 2022-03-03 at 17 44 26

jmaslek (approved):

JerBouma:
I agree with that but couldn't it be an argument you can enable? A chart with straight lines or gaps is not as appealing and you could add a disclaimer somewhere in the graph stating news over the weekend is moved to the next trading day.

In the end the "impact" of news friday evening or saturday evening can hardly be measured anyways.

DidierRLopes:
> I agree with that but couldn't it be an argument you can enable? A chart with straight lines or gaps is not as appealing and you could add a disclaimer somewhere in the graph stating news over the weekend is moved to the next trading day.
> 
> In the end the "impact" of news friday evening or saturday evening can hardly be measured anyways and the impact of the news becomes visible the next trading day.

True but that is tampering with the data. And how can we push the sentiment to the next trading day? If there are 100 news those are 100 news push together into a single time instant and that will be impossible to read. The good thing about keeping it as it is that time is continuous and the likelihood of a different news about same stock being published at same time is very minimal. And for the similar news I'm filtering that out.

I don't mind adding a parameter, I just need to understand better your suggestion 😋

piiq:
the example of how non-trading days are handled can be seen in many common/ta view functions. Like:
https://github.com/GamestonkTerminal/GamestonkTerminal/blob/517e89c4e86128707b06e11e25df104f2997651c/gamestonk_terminal/common/technical_analysis/momentum_view.py#L26
Reindex the dataframe and use dates as x tick labels.
Copying what's done in ta for excluding non-trading days will not work here without moving the weekend news to either the last trading day before the news date or the next trading day after the news date 

I personally would be against moving the news dates as it distorts the media context around the instrument

DidierRLopes:
I will merge this for now since it is a big chunk of code and don't see yet the benefit neither the implementation of moving the news sentiment from weekends towards Monday.

I have some more stuff to do in the terminal, so if we decide on anything else to add here I can do this tomorrow 😄

PR #1423, opened 2022-02-28
Line custom markers to highlight events of interest

DidierRLopes:
image

If we wanted to replicate the plot above we could do:

image

or 

image

This feature is most important when we are building customised reports, where the user can then go and talk about the specific events on the cell below but highlighting certain time events.

We could potentially add annotations in the plot, but this could get messy real quick as per discussed offline with @jmaslek . We could do this through external legend, which may work, but is also not as clean. Thus, for time being this PR introduces a small feature that could potentially be further improved.


JerBouma:
I think it is nice that we can replicate this within the terminal but I am still really looking for news headlines within the terminal. I understand that this is not easily done but we are "competing" here with just exporting the data to Excel and applying your own styling. in the end, the graph I shared from Tesla will _always_ require custom styling.

Screenshot 2022-02-28 at 09 39 36

Think the PR is fine as it is but it has very limited uses in my opinion.

JerBouma (approved):

DidierRLopes:
Done 😄 

Screenshot 2022-02-28 at 10 51 10

PR #1442, opened 2022-03-07
Adjust logger

LBolte29:
adjust logger, so that bot can pass userId and add appName

### DIFF
--- a/gamestonk_terminal/loggers.py
+++ b/gamestonk_terminal/loggers.py
@@ -6,6 +6,7 @@ from pathlib import Path
 import sys
 import time
 import uuid
+from logging.handlers import TimedRotatingFileHandler
 from math import floor, ceil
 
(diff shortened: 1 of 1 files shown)


DidierRLopes:
Is the appName meant to be "gst", "discord", "slack"? If yes, then should this be the first?

LBolte29:
@DidierRLopes you mean it should be the first column?

Chavithra (approved):
Let's merge this

LBolte29:
@DidierRLopes or you mean we should pass the appName when starting the service, e.g. passing "gst" from the terminal() function - this would make more sense than making this an attribute actually

DidierRLopes:
I was referring at appearing first.

But completely agree with that you said of being coupled with the session at the start rather than being an attribute!

Issue #1441, opened 2022-03-07
[Bug]Stock Screener - No Ticker Found with several commands


minhhoang1023:
**Describe the bug**
<img width="444" alt="image" src="https://user-images.githubusercontent.com/40023817/156991749-e8df269d-1f83-4ef2-b95b-dade4aea7e40.png">

this is due to finvizfinance package is outdated, and not able to scrape the website.

**To Reproduce**
finvizfinance version: 0.9.10

**Screenshots**

**Desktop (please complete the following information):**

- OS: [e.g. Mac Sierra]
- Python version [e.g. 3.6.8]

**Additional context**
Add any other information that you think could be useful for us.

GabrieleRisso:
Im experiencing the same error, can l help to fix

piiq:
resolved by https://github.com/GamestonkTerminal/GamestonkTerminal/pull/1443

minhhoang1023:
@GabrieleRisso It's fixed now! Let me know if it still doesn't work for you

GabrieleRisso:
@minhhoang1023 I've just updated the terminal and the error is not coming out as before. There is no output showing in the terminal

minhhoang1023:
@GabrieleRisso can you share with me the command you used? Tested intensively and it worked for us

Issue #1420, opened 2022-02-26
[Bug]Stock Screener - (applicable everywhere) Sorting labels have spaces; spaces are incompatible.


deeleeramone:
Choices for sorting in ownership have a space in them; the inputs fields are not able to handle a space.

2022 Feb 25, 17:29 (✨) /stocks/scr/ $ ownership -s Float Short
ownership: error: argument -s/--sort: invalid choice: 'Short' (choose from 'Ticker', 'Market Cap', 'Outstanding', 'Float', 'Insider Own', 'Insider Trans', 'Inst Own', 'Inst Trans', 'Float Short', 'Short Ratio', 'Avg Volume', 'Price', 'Change', 'Volume')

piiq:
resolved by https://github.com/GamestonkTerminal/GamestonkTerminal/pull/1443

PR #1443, opened 2022-03-07
Fix bug Stocks Screener & Update finvizfinance package

minhhoang1023:
# Description

- Fixes issue #1441: Finfizfinance package is outdated and not able to scrape the website. Update to latest version and refactor affected code.
- Fixes issue #1420: Add custom class to allow for space between argsparse arguments

image

# How has this been tested?

# Checklist:

- [ ] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/CONTRIBUTING.md).

# Others


jmaslek:
Just looking at the screen shot: Finviz title is misspelled, we can't see the tickers there and it still says that Market Cap isn't a valid sort column.

minhhoang1023:
@jmaslek Nice spot on the misspellings!  

`Market Caps` should throw an error as only `Market Cap` is accepted. I intentionally checked for it as I have removed the choice option in argparse and replaced it with something else.

Ticker not clear is due to me maximizing the search sidebar on my 13' laptop 🤣 However, now several columns are not displaying full figures. Gonna add KMBTP magnitude to see if it's better.


image

minhhoang1023:
image

gamestonk_terminal/stocks/screener/finviz_view.py
```diff
@@ -163,15 +167,15 @@ def screener(
         df_screen = df_screen.dropna(axis="columns", how="all")
 
         if sort:
-            if " ".join(sort) in d_cols_to_sort[data_type]:
```
minhhoang1023, line 166: @DidierRLopes Here I updated that only one column can be selected for sorting. More than one column we'd lose the auto-complete functionality. And it would require more dev time for it to be compatible with white space in column names.

DidierRLopes (changes requested):
Screenshot 2022-03-10 at 10 23 08

It says that "Market Cap" is an option but then doesn't allow for it. My suggestion is to create a dictionary where keys correspond to 1 word per column and description is the name the column. This way we can use the dict.keys() on the auto-complete, and the then when sorting we can use dict[ns_parser.sort] value. Does that make sense?

Screenshot 2022-03-10 at 10 25 29

The sorting of the data frame shouldn't be necessary to output something. 

Also, can we mute the 
```
[Info] loading page ##############################################################################################################################################################
```
It's very noisy. If there was just one would be fine, but multiple makes it look bad.

You should be able to do this through an argument, but if not you can use:`with suppress_stdout():`

minhhoang1023:
technical:
image

performance:
image

ownership:
image

valuation:
image

jmaslek (approved):

DidierRLopes (approved):

PR #1455, opened 2022-03-09
Allow for .gst routines to have variable inputs

DidierRLopes:
This was a user request and allows for users to run routines with variable inputs. See attached.

Screenshot 2022-03-09 at 11 23 10

I added this piece of information to main terminal message:

Screenshot 2022-03-09 at 11 45 16

Note: I've used $ARGV[X] convention as a tribute to Perl 😄


jmaslek:
Nice.  I dont get the Perl reference haha but this works :P

Screen Shot 2022-03-09 at 1 12 27 PM

jmaslek (approved):

DidierRLopes:
@jmaslek this one: https://www.php.net/manual/en/reserved.variables.argv.php

PR #1473, opened 2022-03-11
Options binom percent

jmaslek:
Percents were misrepresented:  also cleaned up wording on controller.

Figure_1


colin99d (approved):
Beautiful!

PR #1489, opened 2022-03-15
Customisable export

DidierRLopes:
This should allow to export to any folder regardless of it being:
* Relative path to terminal, e.g. routine_monday
* Absolute path to terminal, e.g. /Users/DidierLopes/Documents/GamestonkTerminal/routine_monday
* Path containing environment var, e.g. ~/Documents/GamestonkTerminal/routine_monday

The export folder is then updated as feature flag and will last for next time terminal is booted up.

If the user provides this path, then the commands are saved like this:
`date_time_context_menu_command.TYPE`

if the user doesn't provide path and is default, commands are saved like this:
`exports/context/menu/command_date_time.TYPE`

This is because if a user selects a folder he is interested in having all data in order of creation in the same place.

In a next PR I will add the user selecting the name of the export file.


jmaslek:
Should we create if it doesn't exist?

jmaslek (approved):

DidierRLopes:
Also at this one @deeleeramone

deeleeramone:
Custom export and import, oh yes! 💌

PR #1504, opened 2022-03-17
allow exe command to provide absolute path or absolute with HOME env var

DidierRLopes:
Screenshot 2022-03-17 at 16 14 10
Screenshot 2022-03-17 at 16 17 32

@deeleeramone fear no more!


andrewkenreich (approved):
Works!

PR #1532, opened 2022-03-21
Add TA and test coverage for the Forex context

piiq:
This PR introduces 2 things:

- The TA in the forex. It works, although it lacks all volume indicators because we don't get volume data from AV. This should probably be changed once more currency sources are added to the context
- Tests for the Forex menu in general. There were no tests in this menu before. Nad now there are. Yay


gamestonk_terminal/forex/technical_analysis/ta_controller.py
```diff
@@ -0,0 +1,928 @@
...
+        """Print help"""
+        currency_str = f" {self.ticker} (from {self.start.strftime('%Y-%m-%d')})"
+        help_text = f"""[cmds]
+[param]Currency Pair Loaded: [/param]{currency_str}
+
+[info]Overlap:[/info]
+    ema         exponential moving average
+    sma         simple moving average
+    zlma        zero lag moving average
+    vwap        volume weighted average price
```
Chavithra, line 90: Maybe remove this line ?

I can't see any `call_vwap` method in this file.

Chavithra (approved):
Two comment :
- there is this little issue with `load` command.
```
2022 Mar 22, 09:54 (✨) /forex/ $ load some_test
The following args couldn't be interpreted: ['-t', 'some_test']
```
- `TechnicalAnalysisController` `help` method to update

Other than that looks good to me.

PR #1525, opened 2022-03-20
Create a packaged app bundle with Pyinstaller

piiq:
This PR adds the infrastructure required to build the terminal into an executable using pyinstaller.

TODOS:
- [x] Test if it's working on Windows
- [x] Test if a separate bundle is required to use the prediction menu
- [x] Add documentation

---

How to create the bundle:

- activate the gst environment
- run `poetry install -E installer` to install the `pyinstaller` package into the environment
- run `pyinstaller build/pyinstaller/terminal.spec` from the project root

As the result you should have a `terminal` executable appear in the dist folder

PR #1534, opened 2022-03-22
Error handling for custom files when folder not present.

jmaslek:
This adds some error handling to the import custom file.

This may help with #1525 and the exe so that load doesn't crash on


andrewkenreich (approved):
tested on windows installer - works!

PR #1592, opened 2022-03-29
Updating dependencies and putting kaleido into a shibajail

aia:


bots/requirements.txt
```diff
@@ -91,30 +91,29 @@ inflection==0.5.1; python_version >= "3.6"
...
-ipython-genutils==0.2.0; python_version >= "3.6"
-ipython==8.1.1; python_full_version >= "3.6.2" and python_version >= "3.8"
+ipython-genutils==0.2.0; python_version >= "3.7"
+ipython==8.2.0; python_full_version >= "3.6.2" and python_version >= "3.8"
 ipywidgets==7.7.0
 iso8601==0.1.16
 isort==5.10.1; python_full_version >= "3.6.2" and python_version < "4.0"
 jedi==0.18.1; python_version >= "3.8"
-jinja2==3.0.3; python_version >= "3.7"
+jinja2==3.1.1; python_version >= "3.7"
```
piiq, line 102: We need keep jinja at 3.0.3 for papermill and voila to work. We need to keep an eye on this until the jupyter ecosystem catches up

piiq (changes requested):

bots/requirements.txt
aia: ok. will lock it to 3.0.3

piiq (approved):

PR #1613, opened 2022-04-02
Fix logging for users that don't rename parent folder to OpenBBTerminal

DidierRLopes:


jmaslek (approved):

Chavithra:
Did the following :
- Move : `openbb_terminal.log` to `openbb_terminal.core.log`
- Refactored the way we use the following files : `.env`, `OPENBB_DEFAULTS.json`
- Fixed a bug on filtering PII information from `log files`

We should be able to call the `terminal.py` from outside the repository like this :
```bash
python GamestonkTerminal\terminal.py
python OpenBBTerminal\terminal.py
```

Chavithra:
@piiq Can you take a look ?

piiq (approved):

PR #1653, opened 2022-04-07
Adding Necessary files to fix the windows installer splash screen

andrewkenreich:
Fix here was to put the "OPENBB_DEFAULTS.json" file into the main folder and not the openbb_terminal folder on build.

I also created the template folder inside build/nsis for future builds.

@piiq Can you verify I didn't break anything on the Mac installer - I changed the name of some files on your side, I think I got them all. Also check that you might not be settings the proper defaults as well if this file isn't in the main folder on your end.

@Chavithra Verify that the windows install works now per the docs


Chavithra (approved):
Looks good to me.

piiq (approved):
While I was digging into it I saw a couple of things:

1. The defaults json can go away in favour of a `.env` file with defaults that bundles with the package. This `.env` file is bundled to the `REPO_PATH` and should be picked up by the terminal on bootup.
2. In packaged apps a `.env` file in `REPO_PATH` will be overwritten by each update. This would require setting up the keys through the settings menu after every update. To have some sort of transferability between releases we can look at creating a `.env` file outside of `REPO_PATH`. For example in user home. This already works on all platforms for prompt toolkit (the prompt toolkit history file is in user home).
3. If we set up the python environment identically (python 3.8 from conda) on Windows and macOS build machines, we can leverage the cross-platform compatibility features that are bundled into pyinstaller. The specfile will be identical for all supported platforms (win, mac, linux).

I'll make a separate draft PR with the proposed changes

PR #1515, opened 2022-03-18
Risk metrics for qa and portfolio

northern-64bit:
# Description
Implementation of risk metrics. If I have missed some useful risk metric or have some errors in my calculations then message me please.

Done:
- Sortino ratio (qa)
- Sharpe ratio (qa)
- Omega ratio (qa)
- Adding metrics to portfolio context
- Documentation + Hugo Website
- Tests

To do:
- Nothing

# How has this been tested?

Testing the commands. The implementation of the mathematics needs to be checked by someone.

# Checklist:

- [x] Update [our Hugo documentation](https://gamestonkterminal.github.io/GamestonkTerminal/) following [these guidelines](https://github.com/GamestonkTerminal/GamestonkTerminal/tree/main/website).
- [x] Update our tests following [these guidelines](https://github.com/GamestonkTerminal/GamestonkTerminal/tree/main/tests).
- [x] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/GamestonkTerminal/GamestonkTerminal/blob/main/CONTRIBUTING.md).
- [x] If a feature was added make sure to add it to the corresponding [scripts file](https://github.com/GamestonkTerminal/GamestonkTerminal/tree/main/scripts).

# Others
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My code passes all the checks pylint, flake8, black, ... To speed up development you should run `pre-commit install`.
- [x] New and existing unit tests pass locally with my changes. You can test this locally using `pytest tests/...`.

PR #1666, opened 2022-04-10
Adding everything from Risk Metrics PR based on new main branch

northern-64bit:
# Description

Adding sh, so, om commands. This pr is based on the changes from #1515 that broke due to renameing.

# How has this been tested?

Using the commands

# Checklist:

# Others


JerBouma:
Hey, thanks for the PR! It seems that the -w argument doesn't accept any integers. Perhaps it gets saved as a string?

Screenshot 2022-04-11 at 10 38 23

Could you also add a little bit more documentation about these ratios on the website? What I am mostly looking at is a quick explanation what the Omega Ratio is for example. You already link to an article within the code but it would be helpful if that explanation is briefly given within the Docs as well.

Furthermore, could you add in an example to each command? You can do so by going into your own Fork GitHub location and directly editing the .md files. This allows you to drag and drop images thus you can simply run the command, save the image and drop it into the .md

northern-64bit:
> Could you also add a little bit more documentation about these ratios on the website? What I am mostly looking for is a quick explanation what the Omega Ratio is for example. You already link to an article within the code but it would be helpful if that explanation is briefly given within the Docs as well.
> 
> Furthermore, could you add in an example to each command? You can do so by going into your own Fork GitHub location and directly editing the .md files. This allows you to drag and drop images thus you can simply run the command, save the image and drop it into the .md

Should I also add a quick explanation of the ratio:s in the command help part (-h)? I'm a bit unsure how detailed it should be.

jmaslek:
> Should I also add a quick explanation of the ratio:s in the command help part (-h)? I'm a bit unsure how detailed it should be.

Would be helpful to have some formulas and descriptions in the Hugo (which is what @JerBouma is suggesting)

JerBouma:
@northern-64bit 

The -h text doesn't have to be so precise. You could say for the sharpe ratio for example "_Determines the Sharpe Ratio over time which is the portfolio return minus the risk free return divided by the portfolio's standard deviation._"

In the Hugo docs you could give the formula, provide (a little bit of) background information and show one example. It doesn't have to be an entire paragraph and feel free to just link to your source, happy to approve this soon :)

northern-64bit:
> The -h text doesn't have to be so precise. You could say for the sharpe ratio for example "_Determines the Sharpe Ratio over time which is the portfolio return minus the risk free return divided by the portfolio's standard deviation._"
> 
> In the Hugo docs you could give the formula, provide (a little bit of) background information and show one example. It doesn't have to be an entire paragraph and feel free to just link to your source, happy to approve this soon :)

I've now added a description to the hugo. After having thought about it I think that the command help description is good as it is and instead I made the hugo documentation more precise. The command help shouldn't clutter up the entire terminal and be easy to use. Looking forward towards having this merged 👍

JerBouma (approved):

PR #1691, opened 2022-04-11
Requests using style fix 

PzaThief:
# Description
There are many json() function calls of requests.response repeatedly.
If it call json() once and store in variables, it is faster than call json() several times.
of coures most of waiting time is for request, but I expect many a little makes a mickle.

and stocktwits_model.get_messages and stocktwits_model.get_stalker had a bug about limit.
it takes array of limit+1 length not limit. so I fixed it and changed style for improve performance.


PzaThief:
> Nice! Looks like black is failing on the CI, so you can just fix that up by running `black ./`

thanks!. I applied black formatting. could you check if it can be merged again?

DidierRLopes (approved):
Nice one! 🚀

PR #1693, opened 2022-04-12
Default env for packaged apps

piiq:
Introduces 3 things

- A json with defaults is removed in favor of a single .env file that resides in `REPO_PATH`
- Added cross platform and cross env (both conda and venv are working) to the specfile
- Fixes a printout bug in integration tests reporting


piiq:
@Chavithra I've adjusted the `pathex` param in the specfile in a way that it is expected to work with both conda env and venv so there shouldn't be a problem with packaging of statsmodels and linearmodels datasets. Check it out

Note:
I am open to splitting mac and win specfiles. 

The pros for this:
- more control over each platform

The cons for this:
- need to maintain x2 (or x3 ) the amount of configs

the code is identical for the both platforms. The difference is only what's on this image
image

andrewkenreich (approved):
This worked for me on the build. Although I had to do this still on the .spec file.


import sys
sys.path.insert(0, 'C:\\Users\\Andrew\\Desktop\\gamestonk MASTER FILE\\GamestonkTerminal')

Then it could find "openbb_terminal.loggers" 

But both files for statsmodel and linearmodels came in.

Chavithra (approved):
Looks ok to me.

PR #1697, opened 2022-04-14
Trading hours stock feature

buahaha:
# Description

    - Add checking for available markets that are open, closed, list all and check one. Also you can load one security to know if it's currently been traded.
    - Add new context box (menu) to look for securities. Might work for all of OpenBB tickers if implemented in full.
Screenshot 2022-04-14 at 04 31 26

    - You should. know when you can trade, and when you can't.
    - None

# How has this been tested?

    - Manual checking of few possible use cases...
    - `/stocks/th/open`
    - `/stocks/th/closed`
    - `/stocks/th/all`
    - `/stockc/th/symbol MSFT`

# Checklist:

- [*] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [*] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).

# Others


jmaslek:
Nice one.

Couple comments.  First is that the `tradinghours` still lives on the main menu.  I think this should just be shown under `stocks`

Screen Shot 2022-04-14 at 9 15 07 AM

And in the stocks menu, I think we can put this up the top section (see image below). (thoughts @DidierRLopes )

Screen Shot 2022-04-14 at 9 17 41 AM


When I load in an exchange, we should do some capitalization on the input, so that I could enter nyse or NYSE and it would understand
```
2022 Apr 14, 09:20 (🦋) /tradinghours/ $ exchange nyse

No exchange data loaded.
Make sure you picked proper exchange symbol.

2022 Apr 14, 09:20 (🦋) /tradinghours/ $ exchange NYSE
*works*
```

Also for the exchange function, is it possible to auto complete that field like we have throughout the terminal?

And using -h flag throws a small error:
```
2022 Apr 14, 09:20 (🦋) /tradinghours/ $ exchange -h
usage: exchange [-n EXCHANGE] [-h]

Select the exchange you want see open hours for

options:
  -n EXCHANGE, --name EXCHANGE
                        Exchange short name (default: None)
  -h, --help            show this help message (default: False)

Make sure 'symbol' is selected.
```

jmaslek:
Thanks for making some changes!  I am still running into an issue with `exchange`

```
2022 Apr 15, 11:34 (🦋) /th/ $ exchange NYSE
Error: 'NYSE'

Error: 'list' object has no attribute 'empty'
```

Additionally, exit needs one more quit, as it just takes us back to the root.  (Also on your model and view, check line 1, one of them still says forex)

website/content/terminal/stocks/_index.md
```diff
@@ -1,5 +0,0 @@
----
```
Chavithra, line 1: Is there a reason to remove this file ?

openbb_terminal/stocks/tradinghours/data/bursa.py
```diff
@@ -0,0 +1,1093 @@
+from datetime import time
```
Chavithra, line 1: I would replace this python file by a `yaml` (or `.ini` or `json`...) file and load it with a function.

Since it's a data file and not code.

Plus this will allow allow not having to ignore this file from the `codespell` check.

Chavithra (changes requested):
Thanks for contributing !

Looks good overall.

I wonder why this file was deleted ?
`website/content/terminal/stocks/_index.md`

Chavithra (changes requested):
I think the only issue left is the linter complaining about some spagetti code here :

- `openbb_terminal/stocks/tradinghours/bursa_model.py::check_if_open`

https://github.com/OpenBB-finance/OpenBBTerminal/runs/6080772850?check_suite_focus=true

openbb_terminal/stocks/tradinghours/data/bursa.py
buahaha: done

Chavithra (approved):
Looks ready for merging.

DidierRLopes (changes requested):
Not yet, the colouring is not right.

Chavithra:
> Not yet. Before we merge can we have:
> 
> 1. AAPL shouldn't be loaded by default, should be empty.
> 2. Autocomplete with 10k tickers doesn't work well, it's too slow and not worthy - we had this discussion internally already.3.
> 3. The colouring scheme is wrong parameters and the value should be [param] and normal by default.4.
> 4. the menu text is not properly aligned and there are 2 white lines instead of 1.5.
> 5. When parsing an argument let's use `.upper()` so we can accept a ticker even with lowercase
> 
> Screenshot 2022-05-11 at 12 55 53

Changed were done.

DidierRLopes (approved):
Thanks Chavi!

PR #1709, opened 2022-04-18
Add terminal-wide report command

minhhoang1023:
# Description

Confluence page: https://openbb.atlassian.net/wiki/spaces/ENG/pages/80936995/Helpdesk+Implementation

- Add terminal-wide report command with auto-completion for the commands
- Connect with Airtable DB for storage

Note:
- If you want to test it, please send me a DM and I'll share with you the Airtable access keys
- As next time while collecting feedback, reports should be sent to an API gateway, which is then connected with Airtable. this is such that we don't expose the key with write permissions

# How has this been tested?

# Checklist:

# Others


jmaslek:
Not necessary, but is int possible to use nocoDB as an air table alternative?

(https://github.com/nocodb/nocodb)

DidierRLopes (changes requested):
This command shouldn't be "report" but "support".

We already use "reports" naming for other features and we don't want to have that confusion here.

minhhoang1023:
> Not necessary, but is it possible to use nocoDB as an air table alternative?
> 
> (https://github.com/nocodb/nocodb)

Ooh that looks nice! airtable is not cheap, so I could look into this once the engagement grows out of scope for Google Sheets

openbb_terminal/parent_classes.py
```diff
@@ -278,6 +309,70 @@ def call_resources(self, _) -> None:
...
+            "--command",
+            action="store",
+            dest="command",
+            required=True,
+            choices=["generic"] + self.support_commands,
+            help="Command that needs support",
+        )
+
+        parser.add_argument(
+            "--msg",
```
DidierRLopes, line 340: let's allow `-m`

openbb_terminal/parent_classes.py
```diff
@@ -278,6 +309,70 @@ def call_resources(self, _) -> None:
...
+            action="store",
+            type=str,
+            nargs="+",
+            dest="msg",
+            required=False,
+            help="Message to send. Enclose it with double quotes",
+        )
+
+        parser.add_argument(
```
DidierRLopes, line 349: and `-t`

DidierRLopes (changes requested):
Comments below.

On top of that can you do a quick thing for me?

This is my welcome screen
Screenshot 2022-05-10 at 11 27 35

Can we change it to this and put it inside:
Screenshot 2022-05-10 at 11 30 19

Here is the text:
```
[info]Get API keys from data providers to access more features.[/info]
    For more instructions use: 'keys' 
```

On top of that when displaying terminal version, mine is showing vREPLACE_ME, which is not ideal. Can you check if the str is different than REPLACE_ME and if it's not just not display any version?

Thanks!!

DidierRLopes:
> @DidierRLopes @minhhoang1023 the vREPLACE_ME will be fixed by the time we do the next release. This relates to poetry not installing the openbb_terminal package correctly into the local environment. Known issue. Let's keep it out of the scope of this PR, please

Thought it would be a 2 liner change and therefore worth the protection added here - seems sloppy currently. Since I use advanced mode I only saw this now.

piiq:
> > @DidierRLopes @minhhoang1023 the vREPLACE_ME will be fixed by the time we do the next release. This relates to poetry not installing the openbb_terminal package correctly into the local environment. Known issue. Let's keep it out of the scope of this PR, please
> 
> Thought it would be a 2 liner change and therefore worth the protection added here - seems sloppy currently. Since I use advanced mode I only saw this now.

It's 2 symbols if you hardcode the version number =) . I was trying to fetch the version number from pyproject.toml on boot and it fails in some environments

DidierRLopes (approved):
Amazing 🙌🏽🙌🏽🙌🏽 legooo

PR #1714, opened 2022-04-19
Options VaR at Expiry

northern-64bit:
# Description

I've added a command for options VaR for the expiry by a monte carlo simulation of the stock price and then calculating the results.
A base for options valuations has been added to the options class which will later be further built upon. Currently it has the black scholes for european options. No command has been added to access this, but with a later pr with american options and commands will be built on this base.

Please check that the math is correct and how it is done (is there any flaws?).

# How has this been tested?

Using the command and testing the values for the option valuation functions.

# Checklist:

# Others


JerBouma:
First of all, thanks for yet another PR! 

As the future is uncertain and difficult to predict due to changing market conditions, could you elaborate what an investor could get out of this Monte Carlo simulation? Furthermore, do you have any literature that confirms that your methodology has proven to be effective?

JerBouma:
The biggest issue I have with Monte Carlo simulations is that it is still a very rough estimation and assumes all other parameters do not change (ceteris paribus). Therefore, you often do not capture market shifts as well as you would like. I have worked in Asset-Liability Management where we simulated 1k-10k scenarios with 30+ investment beliefs attached to it. So it would accurately project the future situation _given the fund's expectations_ and even that still didn't account properly for major market shifts.

That said, you need some form of indication and if the industry professional says this is common practice within his/her firm, it is relevant to add this feature!

By the way, could I get in touch with this industry professional by any chance? It helps if we can pick his/her brain further on this topic and his job to get a better understanding of the wishes/practices of his department and/or firm.

northern-64bit:
> The biggest issue I have with Monte Carlo simulations is that it is still a very rough estimation and assumes all other parameters do not change (ceteris paribus). Therefore, you often do not capture market shifts as well as you would like. I have worked in Asset-Liability Management where we simulated 1k-10k scenarios with 30+ investment beliefs attached to it. So it would accurately project the future situation _given the fund's expectations_ and even that still didn't account properly for major market shifts.

Yes, this is true that it's extremely hard to predict the future (if not impossible), but the goal isn't a prediction. Just where the stock price can be and ca. it's probability. I can imagine that it gets way more difficult with counter-party risk and so on. This code could maybe be more accurate with fatter tails on the distribution for the stock price and added skew and kurtosis, but this may add instability to it since these properties change more easily and is less robust. Sometimes simpler is better.

> That said, you need some form of indication and if the industry professional says this is common practice within his/her firm, it is relevant to add this feature!
> 
> By the way, could I get in touch with this industry professional by any chance? It helps if we can pick his/her brain further on this topic and his job to get a better understanding of the wishes/practices of his department and/or firm.

I'll send my code to him for review and also ask him if he would be happy to get a message from you. If yes (he is very social, likes python a lot and a lecturer at a university so he'll like be intressted), then I'll send you the contact details to you in discord.

openbb_terminal/stocks/options/options_controller.py
```diff
@@ -1222,6 +1224,71 @@ def call_greeks(self, other_args: List[str]):
...
+            "-sim",
+            "--simulations",
+            dest="simulations",
+            type=float,
+            default=100,
+            help="The number of simulations",
+        )
+        parser.add_argument(
+            "-per",
+            "--percentile",
```
jmaslek, line 1253: super nitpick but we want to keep any arguments with "-" to one letter for unix convention

openbb_terminal/stocks/options/options_controller.py
```diff
@@ -1222,6 +1224,71 @@ def call_greeks(self, other_args: List[str]):
...
+            dest="percentile",
+            default=99,
+            type=float,
+            help="The risk free rate",
```
jmaslek, line 1257: percentile = risk free rate?

jmaslek:
So I am curious about a couple things in this implementation

First for your MC.  `change = np.random.randn() * stock_vol / 100`. This has no drift component.  You only assume that returns are zero mean, which kinda touches on what Jer was saying where nobody actually knows what will happen, but usually there is a historical 

And using the IV from the average method doesn't make intuitive sense to me.  If I look at a far out expiry, IV will be higher, and using this approach, you are saying that just because the IV is higher, we expect higher jumps day to day, which is not the case.  Maybe I am missing something, and feel free to correct me there.

Check out how I did a super simple MC in  prediction techniques.  Since this is yahoo, you can get a historical mean/realized volatility and I would gather that makes more sense for longer expirations?

northern-64bit:
> First for your MC. `change = np.random.randn() * stock_vol / 100`. This has no drift component. You only assume that returns are zero mean, which kinda touches on what Jer was saying where nobody actually knows what will happen, but usually there is a historical

I'll look into adding a drift component and a change, the least I'll do is add a drift like the risk free rate.

> And using the IV from the average method doesn't make intuitive sense to me. If I look at a far out expiry, IV will be higher, and using this approach, you are saying that just because the IV is higher, we expect higher jumps day to day, which is not the case. Maybe I am missing something, and feel free to correct me there.

Oh, I didn't think about that. However I'll probably create an average of all expiries (except those that are very close to the current date). I think that this is the way that nasdaq does it, but what do you think of this method? 

> Check out how I did a super simple MC in prediction techniques. Since this is yahoo, you can get a historical mean/realized volatility and I would gather that makes more sense for longer expirations?

The whole thing is based upon the priced stock volatility and therefore not using historical data, thus using realized volatility would destroy this method.

@jmaslek @JerBouma see this pr as a draft since I'll look into your valid points and try to improve the model. Thank you very much!

jmaslek:
> Oh, I didn't think about that. However I'll probably create an average of all expiries (except those that are very close to the current date). I think that this is the way that nasdaq does it, but what do you think of this method? 

Do you have a source?  This method doesn't make sense

> The whole thing is based upon the priced stock volatility and therefore not using historical data, thus using realized volatility would destroy this method.

Source?  Once again IV accounts for time and strike etc so using it to estimate a stocks daily jump through MC doesn't make any sense.

northern-64bit:
> > Oh, I didn't think about that. However I'll probably create an average of all expiries (except those that are very close to the current date). I think that this is the way that nasdaq does it, but what do you think of this method?
> 
> Do you have a source? This method doesn't make sense

I've now done several tests on this method and it works decently, but for it one needs very good options data which the terminal doesn't have. So maybe just the data quality destroys the method. Do you think that I should use historical stock vol (it should work due to the Girsanov theorem)? or should I use realized vol? @jmaslek 

> > The whole thing is based upon the priced stock volatility and therefore not using historical data, thus using realized volatility would destroy this method.
> 
> Source? Once again IV accounts for time and strike etc so using it to estimate a stocks daily jump through MC doesn't make any sense.

The thought of using the average is that it smoothes out this effect. IV could be viewed as a "price" of the option so it's what the one pays for and what one expects (see https://en.wikipedia.org/wiki/Implied_volatility#As_a_price) so it would be partly reasonable to use this in my mind.

I'm no expert in this subject so I thought that this would be a good entry point to learn more. Is this entirely useless so that I should close this pr and forget about it or is there something that else I can do in order to improve this model? @JerBouma @jmaslek

DidierRLopes:
@JerBouma and @jmaslek what's the status on this?

This has been opened for 2 months now, let's try to be more active on the review & conclusion of a PR. Whether it is to merge to the codebase or close.

jmaslek:
I would close

PR #1724, opened 2022-04-20
Add performance table on load (stocks)

jmaslek:
When loading a stock, there was a nice little description of where you were, but I thought adding a snapshot would be nice.  @JerBouma and @andrewkenreich told me what to do

Screen Shot 2022-04-20 at 5 31 51 PM


jmaslek:
> Why is the average trading volume 100D? Is that a typical measurement? I read 20 or 30 days here: https://www.investopedia.com/terms/a/averagedailytradingvolume.asp

I went with 10 days (2 weeks).   Arbitrary number that is enough to get an average (also there is a yfinance ticker.info field called 10dayvolumeaverage, so I used that as reasoning for 10 days)

piiq (approved):

PR #1730, opened 2022-04-23
Initial implementation of terminal supporting multiple languages

DidierRLopes:
- [ ] Need to add i18n to requirements and poetry
- [ ] Need to make the language to be selected a feature flag
- [ ] Finish implementing for all menus and commands

### DIFF
--- a/i18n/en.yml
+++ b/i18n/en.yml
@@ -1,0 +1,61 @@
+en:
+  home_info: >
+    [info]Multiple jobs queue (where each '/' denotes a new command).[/info]
+        E.g. '/stocks $ disc/ugs -n 3/../load tsla/candle'
+
+    [info]If you want to jump from crypto/ta to stocks you can use an absolute path that starts with a slash (/).[/info]
+        E.g. '/crypto/ta $ /stocks'
+
+    [info]The previous logic also holds for when launching the terminal.[/info]
+        E.g. '$ python terminal.py /stocks/disc/ugs -n 3/../load tsla/candle'
+
+    [info]You can run a standalone .openbb routine file with:[/info]
+        E.g. '$ python terminal.py routines/example.openbb'
+
+    [info]You can run a .openbb routine file with variable inputs:[/info]
+        E.g. '$ python terminal.py routines/example_with_inputs.gst --input pltr,tsla,nio'
+
+    [info]The main commands you should be aware when navigating through the terminal are:[/info]
+  cls: clear the screen
+  help: help menu
+  quit: quit this menu and go one menu above
+  exit: exit the terminal
+  reset: reset the terminal and reload configs from the current location
+  resources: only available on main contexts (not sub-menus)
+  about: about us
+  update: update terminal automatically
+  tz: set different timezone
+  export: select export folder to output data
+  exe: execute automated routine script
+  settings: set feature flags and style charts
+  keys: set API keys and check their validity
+  export_folder: Export Folder
+  timezone: Timezone
+  stocks/search: search a specific stock ticker for analysis
+  stocks/search_: Show companies matching the search query
+  stocks/search_query: The search term used to find company tickers
+  stocks/search_limit: Enter the number of Equities you wish to see in the table window
+  stocks/search_country: Search by country to find stocks matching the criteria
+  stocks/search_sector: Search by sector to find stocks matching the criteria
+  stocks/search_industry: Search by industry to find stocks matching the criteria
+  stocks/search_exchange: Search by a specific exchange country to find stocks matching the criteria
+  stocks/load: load a specific stock ticker and additional info for analysis
+  stocks/quote: view the current price for a specific stock ticker
+  stocks/candle: view a candle chart for a specific stock ticker
+  stocks/news: latest news of the company
+  stocks/options: options menu,                 chains, open interest, greeks, parity
+  stocks/disc: discover trending stocks,     map, sectors, high short interest
+  stocks/sia: sector and industry analysis, companies per sector, quick ratio per industry and country
+  stocks/dps: dark pool and short data,     darkpool, short interest, ftd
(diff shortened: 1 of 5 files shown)


DidierRLopes:
@piiq @jmaslek what was the issue with `pygmentize`? Do you remember?

piiq:
@DidierRLopes There was a Pygments update that was not pushed up the chain. Trying to fix it in #1737

piiq:
I've looked a bit into this, here are some exploration results

We have the following text that can be translated:

- Controller menus description and help text
- Everything wrapped into `console.print`
- Table headers in view functions (`print_rich_table`?)
- Chart and axis labels

A brief search revealed:
80 controllers
410 unique console.prints (excluding empty lines)
309 print_rich_tables
181 charts

The strings can be parsed with regexp 

```regex
console\.print\((.*?)\)
```

or xgettext

```bash
xgettext openbb_terminal/stocks/stocks_controller.py -o 1test.txt -L Python -j -a
```

It would be cool if
a) we could have the translatable string as a key in the translation dictionary. I've tested it with current `python-i18n` implementation and it works
b) add translation on the console.print level. All text that's passed to console.print gets translated

... but ... f-stings spoil all the fun with this scenario. with f-strings the keys to the translation dictionary would be dynamic.
Although there are ideas online [here](https://mail.python.org/pipermail/python-ideas/2018-September/053441.html) and [here](https://stackoverflow.com/questions/49797658/how-to-use-gettext-with-python-3-6-f-strings) to create wrappers that make f-strings translatable. Maybe we can come up with something?

In addition I don't have a clear understanding of how to maintain without significant effort. Any ideas here are welcome.

DidierRLopes:
> I've looked a bit into this, here are some exploration results
> 
> We have the following text that can be translated:
> 
> * Controller menus description and help text
> * Everything wrapped into `console.print`
> * Table headers in view functions (`print_rich_table`?)
> * Chart and axis labels
> 
> A brief search revealed: 80 controllers 410 unique console.prints (excluding empty lines) 309 print_rich_tables 181 charts
> 
> The strings can be parsed with regexp
> 
> ```regexp
> console\.print\((.*?)\)
> ```
> 
> or xgettext
> 
> ```shell
> xgettext openbb_terminal/stocks/stocks_controller.py -o 1test.txt -L Python -j -a
> ```
> 
> It would be cool if a) we could have the translatable string as a key in the translation dictionary. I've tested it with current `python-i18n` implementation and it works b) add translation on the console.print level. All text that's passed to console.print gets translated
> 
> ... but ... f-stings spoil all the fun with this scenario. with f-strings the keys to the translation dictionary would be dynamic. Although there are ideas online [here](https://mail.python.org/pipermail/python-ideas/2018-September/053441.html) and [here](https://stackoverflow.com/questions/49797658/how-to-use-gettext-with-python-3-6-f-strings) to create wrappers that make f-strings translatable. Maybe we can come up with something?
> 
> In addition I don't have a clear understanding of how to maintain without significant effort. Any ideas here are welcome.

Thanks for this @piiq. This all makes sense.

My initial idea was to just translate: controller menu, and command helper (and its arguments). That by itself would already be a very good step forward. Not sure about the ROI of modifying the f-string to `.format()` since that reduces readability of the code imho.

My initial idea was to have keys for these 3 type of cases as I mention in the `contributing.md` guidelines. This was mostly because of the maintenance would be easy. But after thinking about this further I think having the English string as key may work best for:
* readability of the code
* translating from English to other language as key is text to translate

And my biggest concern can be minimised by introducing a test that checks if all keys on `en.yml` are being used. If they are not it means that there was a refactoring and the keys need to be adapted. Thoughts?

Chavithra:
@piiq I will try to summarize the idea let me know if this is accurate.

You are suggesting using `English translation` as the `key` ?

piiq:
> You are suggesting using `English translation` as the `key` ?
@Chavithra , yep

like `i18n.t('search a specific stock ticker for analysis')`

rationale:
- translatable strings would make the code self-documented -> readability doesn't degrade
- we can set a fallback locale in `python-i18n` (`i18n.set('fallback', 'en')`) so no problem if a translation for a language is partial

on the other hand we would need to maintain a dictionary for english and I find it hard to come up with an automation for this at the moment

DidierRLopes:
> > You are suggesting using `English translation` as the `key` ?
> 
> @Chavithra , yep
> 
> like `i18n.t('search a specific stock ticker for analysis')`
> 
> rationale:
> 
> * translatable strings would make the code self-documented -> readability doesn't degrade
> * we can set a fallback locale in `python-i18n` (`i18n.set('fallback', 'en')`) so no problem if a translation for a language is partial
> 
> on the other hand we would need to maintain a dictionary for english and I find it hard to come up with an automation for this at the moment

check my comment above regarding maintainability. I think that's a 10min break task for @colin99d

openbb_terminal/rich_config.py
```diff
@@ -34,15 +37,118 @@
...
+    def add_cmd_translation(self, key: str, source: str = "", cond: bool = True):
+        if source:
+            source = f" [src][{source}][/src]"
+        spacing = (22 - (len(key) + 4)) * " "
+        if cond:
+            self.menu_text += f"[cmds]    {key}{spacing}{i18n.t(self.menu_path + key)}{source}[/cmds]\n"
+        else:
+            self.menu_text += f"[unvl]    {key}{spacing}{i18n.t(self.menu_path + key)}{source}[/unvl]\n"
+
+    def add_menu_translation(self, key: str, cond: bool = True):
```
jmaslek, line 142: no docstrings anywhere

DidierRLopes:
> This needs some serious documentation on the contributing guide as to how you add a new feature to the help menu text.



> This needs some serious documentation on the contributing guide as to how you add a new feature to the help menu text.

Yup, was missing it. Now it should be good.

Let me know what you think

jmaslek (approved):

PR #1736, opened 2022-04-25
updated Linux installation instructions

deeleeramone:
This update changes only the README.md file.

Additional instructions added to the section Python Installation section add steps that resolved user issues preventing them from installing, on both Linux Ubuntu and Raspberry Pi. Additionally, old links to the .git repository were changed to the OpenBB versions; and, ML installation for MacOS users is updated with instructions to avoid the Illegal Instruction Error when entering the Prediction Menu.


Chavithra (approved):
Looks ok to me.

PR #1745, opened 2022-04-27
Adding stock price comparison via -c flag

Avani1994:
# Adding stock price comparison via -c flag

Added a compare flag in twitter sentiment analysis to allow see the movements in stock price alongside the sentiment. It's just a small enhancement to the sentiment command under /stocks/ba/.

I just felt stock price movements along with sentiment will be really helpful from customer point of view.

None of the dependencies are required for this change.

# How has this been tested?
Couldn't run tests, as it was asking to uninstall brotlipy and conda uninstall didn't do the job, hopefully CI should catch failing tests if any

`$/stocks/ba sentiment -c True` 

Output:
Screen Shot 2022-04-26 at 11 30 44 PM

* Please also list any relevant details for your test configuration. - None

# Checklist:


openbb_terminal/common/behavioural_analysis/twitter_view.py
```diff
@@ -152,52 +153,111 @@ def display_sentiment(
...
-        if len(external_axes) != 2:
-            logger.error("Expected list of one axis item.")
-            console.print("[red]Expected list of one axis item./n[/red]")
-            return
-        (ax1, ax2) = external_axes
+        closing_price_df = get_closing_price(ticker, n_days_past - 1)
+        # This plot has 3 axis
+        if external_axes is None:
+            _, axes = plt.subplots(
+                3, 1, sharex=False, figsize=(12, 7), dpi=cfg_plot.PLOT_DPI
```
DidierRLopes, line 210: Here the fig size should be `plot_autoscale()`

openbb_terminal/common/behavioural_analysis/twitter_view.py
```diff
@@ -152,52 +153,111 @@ def display_sentiment(
     df_tweets = df_tweets.sort_values(by="date")
     df_tweets["cumulative_compound"] = df_tweets["sentiment"].cumsum()
 
-    # This plot has 2 axis
-    if external_axes is None:
-        _, axes = plt.subplots(
-            2, 1, sharex=True, figsize=plot_autoscale(), dpi=cfg_plot.PLOT_DPI
+    # get stock end price for each corresponding day if compare == True
+    if not compare:
```
DidierRLopes, line 157: Because the ax1 and ax2 do not change regardless of whether we display stock price or not. We should do something like the following pseudo-code:

```
if compare:
    ax1,ax2,ax3 are created
else:
   ax1,ax2 are created

populate ax1 and ax2 (this avoids repeated code)

if compare:
   populate ax3 with stock price

set theme
```

This allows to minimize duplicated code and makes it easier for debug

openbb_terminal/stocks/behavioural_analysis/ba_controller.py
```diff
@@ -717,6 +717,15 @@ def call_sentiment(self, other_args: List[str]):
             default=6,
             help="number of days in the past to extract tweets.",
         )
+        parser.add_argument(
+            "-c",
+            "--compare",
+            action="store",
```
DidierRLopes, line 723: For true/false flags you can use here "store_true" as action and remove both the type and the default

DidierRLopes (commented):
Nice one @Avani1994 !!

I put my comments below 😄

DidierRLopes:
@Avani1994 for the tests to pass you need to run black linting to the `helper_funcs`

@piiq this is also failing in the `compose_export_path` do you know why?

piiq:
thank you @Avani1994 

re: `compose_export_path`
The error says `NameError: name 'Tuple' is not defined`. 
Add `Tuple` on line 7 in openbb_terminal/helper_funcs.py to resolve, although the import is available in main so it that line wasn't edited it should work fine once the PR branch is synced with main.

openbb_terminal/helper_funcs.py
```diff
@@ -4,8 +4,8 @@
 import argparse
 import logging
 from pathlib import Path
-from typing import List, Tuple, Union
```
piiq, line 7: this is why tests are failing

openbb_terminal/common/behavioural_analysis/twitter_view.py
Avani1994: I see, I tried this, but then with three figs, the individual plots are quiet small. Wouldn't that be a problem?

openbb_terminal/common/behavioural_analysis/twitter_view.py
Avani1994: I see yeah this makes sense to me, we have to just write `if compare` twice, i thought that wouldn't be good idea, but yeah we can do this.

openbb_terminal/stocks/behavioural_analysis/ba_controller.py
Avani1994: That's good advice :)

openbb_terminal/common/behavioural_analysis/twitter_view.py
DidierRLopes: That may be because our default is not ideal, but this should be able to be configurable by the user

openbb_terminal/common/behavioural_analysis/twitter_view.py
DidierRLopes: Makes it easier from the readability perspective and maintenance as well since there isn't repeated code

DidierRLopes:
@Avani1994 you just need to fix the test for the `-c` flag 😄

DidierRLopes (approved):
Nice work @Avani1994 !

PR #1753, opened 2022-04-30
Improve OpenBB API for investment research reports

DidierRLopes:
Improved a LOT of small things on the API:
* Some typos
* Some bad imports due to the logging decorator which broke implemented logic to get models funcs from notebook
* Create a very nice and slick equity report (can be improved on, it's still a POOC)

@piiq @jose-donato this is silly but how can we have the OpenBB_reports_logo.png image appearing on the report when this is run from the reports controller?


jose-donato:
The image should be good now. Converted to base64 and then rendered the image.

 also fixed some other HTML stuff

openbb_terminal/forex/quantitative_analysis/qa_api.py
```diff
@@ -47,11 +48,9 @@
 )
 
 # Models
-# NOTE: The raw function is used here to point to the commons path where all the
```
piiq, line 50: I don't remember why I've implemented it like this 😂

openbb_terminal/jupyter/reports/reports_controller.py
```diff
@@ -28,7 +28,7 @@ class ReportController(BaseController):
     reports_folder = os.path.dirname(os.path.abspath(__file__))
 
     report_names = [
-        notebooks.strip(".ipynb")
+        notebooks[:-6]
```
piiq, line 31: not critical to me, but why?

openbb_terminal/stocks/dark_pool_shorts/finra_view.py
```diff
@@ -207,7 +186,7 @@ def darkpool_otc(
     num: int,
     promising: int,
     tier: str,
```
piiq, line 188: Is there a way to have a default kwarg for the `tier`? It's a string and giving a default kwarg would make it easier to understand what it looks like

piiq (changes requested):
We'll need to get rid of the `.ipynb_checkpoints` that accidentally slipped into the PR

I've left a couple of comments additionally, but nothing critical

openbb_terminal/jupyter/reports/reports_controller.py
DidierRLopes: Took me some time to figure out that the strip was problematic. I had misunderstood it's usage. See
Screenshot 2022-05-03 at 11 10 27

DidierRLopes:
@piiq should be good now :)

piiq:
@DidierRLopes no, it's not
image

piiq (approved):

PR #1761, opened 2022-05-03
stock search error fix

### LINKED ISSUE #1758: [Bug] Stock Search Fails To Search By Industry Because Of Phrases.
Same problem that was brought up with the Stock Screener and sorting categories with multiple words. 

https://user-images.githubusercontent.com/85772166/166408857-b7b669a6-2eb1-423c-9870-2805f5c845b9.mov

PzaThief:
#1758 error fix.
parser choices option cannot be used for space separated string. because it checks before action about join string.
so I made a new action that check choices after join string args.
Just after fix that issue, I found another error about japan stock exchange "S" is not in market_coverage_suffix variable. So I added it .


openbb_terminal/stocks/stocks_controller.py
```diff
@@ -670,3 +674,13 @@ def call_pred(self, _):
                 "Predict is disabled. Check ENABLE_PREDICT flag on feature_flags.py",
                 "\n",
             )
+
+
+def choiceCheckAfterAction(choice):
```
piiq, line 679: wow, a metaclass in the wild

piiq (changes requested):
@PzaThief thank you for the PR

The integration tests pass and the command arguments that have spaces work in both terminal and .openbb scripts

As a change request I would ask you to add a docstring to `choiceCheckAfterAction` and `ActionClass`

Do you think it would be beneficial to make the `choiceCheckAfterAction` available to other controllers by moving it out of the stocks_controller? @Chavithra have a look 👀, will these ideas help resolving the slash issue?

PzaThief:
@piiq 
> The integration tests pass and the command arguments that have spaces work in both terminal and .openbb scripts

Thank you for check it.
> As a change request I would ask you to add a docstring to `choiceCheckAfterAction` and `ActionClass`

Okay. I'll change it soon.
> Do you think it would be beneficial to make the `choiceCheckAfterAction` available to other controllers by moving it out of the stocks_controller?

before your comment I couldn't think about it. but as your comment, this issue is not just for stocks_controller. So I'll move it out like AllowArgsWithWhiteSpace in openbb_terminal.helper_classes

piiq (approved):

PR #1772, opened 2022-05-06
bump version on nsis installer and change execution type

andrewkenreich:
Changing version on Nsis installer script to 1.2 and also changing the execution level to user instead of admin

### DIFF
--- a/build/nsis/setup.nsi
+++ b/build/nsis/setup.nsi
@@ -10,7 +10,7 @@
   !define NAME "OpenBB Terminal"
   !define COMPANY "OpenBB"
   !define APPFILE "OpenBBTerminal.exe"
-  !define VERSION "1.0.0"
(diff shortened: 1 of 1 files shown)


build/nsis/setup.nsi
```diff
@@ -10,15 +10,15 @@
   !define NAME "OpenBB Terminal"
   !define COMPANY "OpenBB"
   !define APPFILE "OpenBBTerminal.exe"
-  !define VERSION "1.0.0"
+  !define VERSION "1.2.0"
```
DidierRLopes, line 13: Do we want this to be 1.2.1?

piiq: Next step - we include this version change into the build script

piiq (approved):

Issue #1729, opened 2022-04-22
IEX API: Error: 'str' object has no attribute 'strftime'[Bug]


ginmaster:
**Describe the bug**
If a stock is loaded via IEX API, e.g. via

`load --source iex -t amzn`

or

`load --source iex -t amzn --iexrange 6m`

Nearly every terminal command, except `candle` that follows lead to the following error:

Error: 'str' object has no attribute 'strftime'

To get the terminal being usable one has to log out of the current session.

**To Reproduce**
Steps(from the start) and commands to reproduce the behavior

`stocks`
`load --source iex -t amzn`
`h`

**Screenshots**

**Desktop (please complete the following information):**

- OS: macOS 12.3.1 on MBP 2021 (M1 Max)
- Python 3.8.13

**Additional context**
Add any other information that you think could be useful for us.

DidierRLopes:
Has been solved!

PR #1779, opened 2022-05-07
controller start set in datetime when load data from iex source

PzaThief:
#1729 fix
self.stock.index has pandas datetime and it can be changed to native datetime using to_pydatetime().
becuase of self.start.strftime('%Y-%m-%d'), controller.start must to be set in datetime not str. So I fixed it and confirmed issue's error is resolved.


DidierRLopes (approved):
Thank you once again 🙌

Issue #1784, opened 2022-05-08
[Bug] /crypto/defi/vaults throws error


minhhoang1023:
<img width="1060" alt="image" src="https://user-images.githubusercontent.com/40023817/167305189-4cb198cd-8e81-438a-868c-d0293d30a7fb.png">

DidierRLopes:
@PzaThief saved the day 😄

PR #1786, opened 2022-05-08
fix coindix api call error

PzaThief:
#1784 fix
Coindix api call has 2 errors. first is SSL error in the issue's screenshot and second is Coindix check user-agent to block python requests(with 403 Forbidden code).
So I changed SSL verify option and user-agent in headers.
SSL error occurred by coindix ssl certificate that only certified api.coindix.com not apiv2.coindix.com. you can check it in your browser.


DidierRLopes (approved):
Thank youuuu!!! 🚀

Issue #1789, opened 2022-05-10
[Bug] stocks/ta/VWAP: VWAP artificially limits the user to the most recent trading session.


deeleeramone:
VWAP artificially limits the user to the most recent trading session despite intraday data being loaded. 

![Screen Shot 2022-05-09 at 9 00 12 PM](https://user-images.githubusercontent.com/85772166/167540122-917a1a77-b68f-4f5c-bc51-15985b610bd6.png)
![Screen Shot 2022-05-09 at 9 14 48 PM](https://user-images.githubusercontent.com/85772166/167541424-1004529b-4db2-4bf1-b5a0-851f6ecd6d08.png)

piiq:
resolved by #1790

PR #1790, opened 2022-05-10
change view_vwap to specify date range

PzaThief:
#1789 fix
now we can specify date range in vwap(include crypto menu) using --start and --end. (ex. vwap --start 2020-05-01 --end 2020-06-01)
allow daily interval when using date range conditions
add console log about loaded data's date or no data found notify
change vwap parameter name to just 'df' from 'day_df' because of changed spec of this commit


minhhoang1023:
you're the bug assassin @PzaThief ! 🚀

DidierRLopes (approved):

PR #1804, opened 2022-05-12
Improve overall UX

DidierRLopes:
Do not display vREPLACE_ME. Don't kill me @piiq but I have OCD and couldn't not add a protection against that.
Also added documentation on the first line so users can see more documentation
Screenshot 2022-05-12 at 01 58 01

Website added in bottom right of panel:
Screenshot 2022-05-12 at 01 21 34

About has documentation and website
Screenshot 2022-05-12 at 01 59 03

Exit message is uniform and says:
Screenshot 2022-05-12 at 01 59 30

### DIFF
--- a/openbb_terminal/alternative/alt_controller.py
+++ b/openbb_terminal/alternative/alt_controller.py
@@ -29,8 +29,7 @@ class AlternativeDataController(BaseController):
 
         if session and obbff.USE_PROMPT_TOOLKIT:
             choices: dict = {c: {} for c in self.controller_choices}
-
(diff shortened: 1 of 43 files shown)


jmaslek (approved):

andrewkenreich:
we need version somewhere still pls :)

Troubleshooting this would be very annoying.

how about in the bottom right "OpenBB Terminal v1.2.1 ( https://openbb.co)"

DidierRLopes:
> v1.2.1

The version will appear if it exists when logging on the terminal in any case.

But I like your suggestion actually, will do.

PR #1818, opened 2022-05-14
Portfolio improvements

DidierRLopes:
* Improve portfolio controller UX
* Shadow commands when these are unavailable
* Allow custom reset when loaded portfolio and benchmark for faster development
* bug when bench was loaded before portfolio, this fixes it
* allow to see sum of portfolio holdings (stacked plot)
* Add R-square command between portfolio returns and benchmark returns for all periods
* Add overall stats command for portfolio and benchmark (mean, std, percentiles, count) for a specific period
* Add kurtosis command for all periods
* Add skewness command for all periods
* Add period parameter to perf command
* Add yearly returns to cumulative return plot
* Add rolling volatility
* Add rolling sharpe
* Add rolling sortino
* Add rolling beta
* Allow to select period for cumulative returns vs EOY returns
* Add volatility, sharpe ratio, sortino ratio and maximum drawdown ratio

@JerBouma just fails on missing docs. Waiting for your approval on this so I can add the docs if you are happy with it.

 -------

Additions since @JerBouma review:

* remove sqrt(N) and N from sharpe and sortino calculations
* allow hold to export raw data from tail
* automatically add space before and after table
* add portfolio holdings in percentage
* fix relative periods to be more accurate
* refactor metric command to allow to select a metric of interest and see multiple periods for it
* fix cumulative return (in percentage)
* implement new yearly return command
* add daily returns graph
* add distribution of daily returns command and stats for raw
* add monthly returns command
* add summary command with multiple metrics for a specific period
* calculate yearly (out)performance

--------

Since 2nd @JerBouma review:

* fix show
* beta with benchmark of 1
* improve meet style


JerBouma:
Hey, looking good so far! You've been busy this weekend while i was sitting in the sun ;)

Some notes:

-  Could you transform the relevant numbers to percentages with 1 or 2 decimals? E.g. `stats` (since returns), `vol`, and possibly `rsquare`. 
- Could `hold` also give me a table (--raw) with the absolute values and % of the portfolio? I know `alloc` also does the percentages but I think it is useful to see this result in a table regardless.
- `return`:
    - Could `return` command also provide me with a table that gives me returns for each year and benchmark and depict outperformance? This combined with the periods I mentioned. Here I do want to point out that I see a clear difference in cumulative returns and the method used within `perf` (and I would go with `perf`). The difference here is that the `perf` method makes an attempt to mimic the scenario that you bought the ETF (benchmark) instead of the asset _at every order moment_ you bought and then determines if that would have led to outperformance. This also makes intuitive sense since I want to understand if I wasn't better off just buying the ETF. 
    - I am unsure if the cumulative returns make sense if you are buying on a periodic basis. Why is the cumulative return negative for the portfolio? That makes very little sense looking at how much return you were able to achieve with those early day trades. E.g. look at the return `perf` gives you.
- I think `maxdd` and the `rolling` graphs could all use a table with values.
- `rbeta` only gives portfolio value but does it make sense we show the beta of 1 for benchmark as well? Just overall consistency and making the graph more clear (??). Optional.
- I'd like `stats` to also include skewness and kurtosis. I think it makes little sense to have those as separate commands, they have little meaning.

JerBouma (changes requested):

DidierRLopes:
> Hey, looking good so far! You've been busy this weekend while i was sitting in the sun ;)
> 
> Some notes:
> 
> * Could you transform the relevant numbers to percentages with 1 or 2 decimals? E.g. `stats` (since returns), `vol`, and possibly `rsquare`.
> * Could `hold` also give me a table (--raw) with the absolute values and % of the portfolio? I know `alloc` also does the percentages but I think it is useful to see this result in a table regardless.
> * `return`:
>   
>   * Could `return` command also provide me with a table that gives me returns for each year and benchmark and depict outperformance? This combined with the periods I mentioned. Here I do want to point out that I see a clear difference in cumulative returns and the method used within `perf` (and I would go with `perf`). The difference here is that the `perf` method makes an attempt to mimic the scenario that you bought the ETF (benchmark) instead of the asset _at every order moment_ you bought and then determines if that would have led to outperformance. This also makes intuitive sense since I want to understand if I wasn't better off just buying the ETF.
>   * I am unsure if the cumulative returns make sense if you are buying on a periodic basis. Why is the cumulative return negative for the portfolio? That makes very little sense looking at how much return you were able to achieve with those early day trades. E.g. look at the return `perf` gives you.
> * I think `maxdd` and the `rolling` graphs could all use a table with values.
> * `rbeta` only gives portfolio value but does it make sense we show the beta of 1 for benchmark as well? Just overall consistency and making the graph more clear (??). Optional.
> * I'd like `stats` to also include skewness and kurtosis. I think it makes little sense to have those as separate commands, they have little meaning to look at individually.
> 
> Regarding `rbeta`, see QuantStats: image

Thanks for the thorough review! I'll try to have something for you by tomorrow. I will update the description to reflect all the changes I've done since this review, so we can make sure I didn't overlook anything!

JerBouma:
New day, new review:

- `show` seems to be broken:
```
2022 May 17, 03:04 (🦋) /portfolio/ $ show
Error: 'Portfolio' object has no attribute 'index'
```
- Do we need both `holdv` and `holdp`? Can't it be an argument and it defaults to absolute value?

- I'd still suggest to base the returns off my method used within `perf` (make a trade in the benchmark every time you make a trade in the portfolio, then base daily, monthly and yearly returns off that), this is because:
  - I still don't have the feeling `cret` is correct. How did the benchmark outperform times 6? Why does this differ so much with `perf`? I'd fully believe that the portfolio outperformed (not accounting for additional risk) because it holds such a little set of companies that all grew exponentially.
  - Looking at `yret` also makes little sense to me. How come the Portfolio dropped more than 75%. Similarly for the Benchmark even though you bought in the earlier days. Just looking at VTI for example just shows that is pretty much impossible. How come I lost 50% when I invested since 2010? Some of those purchases have 4 times their value now and would counterweight this large drop. Especially since YTD is only -17% too.
  Screenshot 2022-05-17 at 09 20 38
  Calculation my return for the year 2010 gives me a value of 39.04% (based on the portfolio at that time and end of year prices of Google and Amazon). Note that the yearly return is nothing else than **(end of year value / begin of year value) - 1** Looking at the table and graph, this is clearly off by a lot (and the table and graph also seem to show different results).
  Screenshot 2022-05-17 at 09 32 52
- I think you should be grouping `yret`, `mret` and `dret` in one command.
- `rbeta` should still have a line at 1 to represent the benchmark. Especially when you have a portfolio that moves around the beta a lot.
- Some styling:
   - `mret`:
 Screenshot 2022-05-17 at 09 13 07
  - `distr`:
    Screenshot 2022-05-17 at 09 14 44
- Check formatting for `metric` and adjust values that should be represented as percentage:
  ```
  2022 May 17, 03:19 (🦋) /portfolio/ $ metric volatility
  
  Volatility for Portfolio and Benchmark
  ┏━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━┓
  ┃     ┃ Portfolio ┃ Benchmark ┃
  ┡━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━┩
  │ mtd │ 0.03      │ 0.02      │
  ├─────┼───────────┼───────────┤
  │ qtd │ 0.03      │ 0.02      │
  ├─────┼───────────┼───────────┤
  │ ytd │ 0.02      │ 0.01      │
  ├─────┼───────────┼───────────┤
  │ 3m  │ 0.03      │ 0.02      │
  ├─────┼───────────┼───────────┤
  │ 6m  │ 0.02      │ 0.01      │
  ├─────┼───────────┼───────────┤
  │ 1y  │ 0.02      │ 0.01      │
  ├─────┼───────────┼───────────┤
  │ 3y  │ 0.01      │ 0.01      │
  ├─────┼───────────┼───────────┤
  │ 5y  │ 0.01      │ 0.01      │
  ├─────┼───────────┼───────────┤
  │ 10y │ 0.01      │ 0.01      │
  ├─────┼───────────┼───────────┤
  │ all │ 0.01      │ 0.01      │
  └─────┴───────────┴───────────┘
  ```

jmaslek:
In the case bench isn't supplied, should we just use SPY as default:

```
2022 May 17, 18:47 (🦋) /portfolio/ $ yret --raw
Please first define the benchmark (via 'bench')
```

DidierRLopes:
> In the case bench isn't supplied, should we just use SPY as default:
> 
> ```
> 2022 May 17, 18:47 (🦋) /portfolio/ $ yret --raw
> Please first define the benchmark (via 'bench')
> ```

Good idea, let me add that

JerBouma:
I disagree with adding a default benchmark. When you use `bench` you get countless of options already and setting a default benchmark makes little sense if your portfolio is composed of stocks that aren't even in the benchmark. What are you measuring then?

DidierRLopes:
> I disagree with adding a default benchmark. When you use `bench` you get countless of options already and setting a default benchmark makes little sense if your portfolio is composed of stocks that aren't even in the benchmark. What are you measuring then?
> 
> Also, I should go to bed. 🙃

Gotcha, will leave as is then! Will start working on the docs - @jmaslek is helping debug returns

JerBouma (approved):

PR #1832, opened 2022-05-17
Tidy up autoloading in places

jmaslek:
Instead of adding the smart attempt at loading everywhere that `load` exists, only restrict it to stocks, crypto, funds and etfs.

### DIFF
--- a/openbb_terminal/etf/etf_controller.py
+++ b/openbb_terminal/etf/etf_controller.py
@@ -80,6 +80,7 @@ class ETFController(BaseController):
         self.etf_name = ""
         self.etf_data = ""
         self.etf_holdings: List = list()
+        self.TRY_RELOAD = True
 
         if session and obbff.USE_PROMPT_TOOLKIT:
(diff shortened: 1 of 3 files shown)


DidierRLopes:
> Instead of adding the smart attempt at loading everywhere that `load` exists, only restrict it to stocks, crypto, funds and etfs.

aren't you missing crypto and forex?

jmaslek:
> aren't you missing crypto and forex?

crypto is done in base controller and  forex doesn't work like `load eur-usd`, you have to have the `to` and `from` defined as is

DidierRLopes:
> > aren't you missing crypto and forex?
> 
> crypto inherits from stocks controller and forex doesn't work like `load eur-usd`, you have to have the `to` and `from` defined as is

Meant stocks and crypto* not forex. Just because I only see a TRY_LOAD on the etf_controller and mutual_fund_controller

DidierRLopes (approved):

PR #1838, opened 2022-05-19
refactoring load, changed chart to candle

jose-donato:
**DRAFT: WILL BE UPDATED SOON**

* load simplified
* changed command `chart` to `candle` to match stocks

image


minhhoang1023:
Nice work! 🚀 This is a much-needed one.

Some comments:

1. We should remove the auto-completion with source
image

jose-donato:
Nice :) will go through your comments today

minhhoang1023:
@jose-donato Let me know when it's ready and I'll review it. Let's get this bad boy out!

Now the load is not working at all..

image

minhhoang1023:
When I loaded ADA in upper case, it said coin is not found. But in lower case that'd work

image

minhhoang1023:
When the starting date is specified, an error was thrown

image

jose-donato:
thanks for the help @Chavithra

minhhoang1023:
When I use the load multiple times, coingecko would limit # of API calls. We should have an error message telling them that they should try again in a bit. Other than that, work perfectly! ✨ 


image

minhhoang1023 (approved):

Issue #1836, opened 2022-05-19
[Bug] stocks/dps/volexch - missing venues: only ARCA is displayed


deeleeramone:
![Screen Shot 2022-05-19 at 12 09 10 AM](https://user-images.githubusercontent.com/85772166/169232238-cdd50a1a-3671-4d70-b658-c42a47e0fce7.png)

PR #1840, opened 2022-05-19
Deprecate stocks/dps/volexch

jmaslek:
As pointed out in #1836, the volexch function is broken.

The reason for the breaking is that the data was stored on my personal mongodb account which has limited free tier storage.  I have simply run out of storage to keep updating it.

This PR deprecates that function by commenting it out.  This can be added back when a better storage option is found.


colin99d (approved):
Sad. Is this feature used enough to justify a paid db?

jmaslek:
> Sad. Is this feature used enough to justify a paid db?

ehhhhh

PR #1869, opened 2022-05-25
Timeseries Forecasting: Theta Method

martinb-bb:
# Description

Newly updated Theta Method deterministic forecasting utilizing Darts library
(Found inside `stocks/pred/theta`)
- automated historical backtesting and retraining 
- modifiable sliding window and forecasting look-ahead 
- mean average precision error (MAPE) over backtest 
- supports N number of forecasting days; best results are 3-7 days in the future. 

image
image

Standardization, flexibility and readability of a common library. 
Improved performance and features. 

torch, darts

# How has this been tested?

```
stocks
load gme
pred
theta
theta -n 20 -s A -p 4 -w 0.8 -f 7
```

# Checklist:

# Others


jmaslek:
Tried this with a polygon source:

```
2022 May 26, 19:02 (🦋) /stocks/pred/ $ theta
[2022-05-26 19:02:55,598] ERROR | darts.timeseries | ValueError: Could not correctly fill missing dates with the observed/passed frequency freq='B'. Not all input time stamps contained in the newly created TimeSeries. For more information about frequency aliases, read https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases
Error: Could not correctly fill missing dates with the observed/passed frequency freq='B'. Not all input time stamps contained in the newly created TimeSeries. For more information about frequency aliases, read https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases

Error: not enough values to unpack (expected 6, got 0)
```

to get stock data:
```
load aapl --source polygon -s 2020-01-01
```

martinb-bb:
Since timeseries data can be of any frequency and have many variables to it, I do not want edge cases causing the kind of issues you faced (UTC daylight saving time). 

For now I have converted the TS models to go based on steps and users will now be able to forecast t+N steps into the future. This will allow them to use sec/min/hour/day/random times to do forecasting - no longer do we have to guess offset-aliases depending on the data provider or personal datasets.

Until we design a more elegant and fault proof method of loading timeseries with varying unknown frequencies, this will do for the newer timeseries models.
One thing that needs to change is the table that still mentions `Datetime`.  I will modify that if you agree with this decision, else we can chat more about future proofing. 

image
theta
eps

martinb-bb:
eps

theta

jmaslek:
> Since timeseries data can be of any frequency and have many variables to it, I do not want edge cases causing the kind of issues you faced (UTC daylight saving time).
> 
> For now I have converted the TS models to go based on steps and users will now be able to forecast t+N steps into the future. This will allow them to use sec/min/hour/day/random times to do forecasting - no longer do we have to guess offset-aliases depending on the data provider or personal datasets.
> 
> Until we design a more elegant and fault proof method of loading timeseries with varying unknown frequencies, this will do for the newer timeseries models. One thing that needs to change is the table that still mentions `Datetime`.
> 
> I will modify that, if you agree with this decision, else we can chat more about future proofing. (For instance, when someone wants to integrate 2 datasets with varying timescales but still needs a time reference.)
> 
> image

Hmmm.  Is it possible to re-index the data frame, then train on the "step number", then go back to the initial date time to display?

Then for t+N, we could default to business days, but if we detect a time delta, we can use that, or let the user specify the time steps?

martinb-bb:
Update: I have scrapped the idea of using an index. Instead it will work with any sort of `YYYY-MM-DD HH-MM-SS` and convert to `YYYY MM DD`
Everything should work now.
`load aapl --source polygon -s 2022-03-01`

image

image

I think the real trick in the future will be to convert all timestamps to unix seconds, forecast and then reconvert back to a timestamp. This will allow us to preserve the essence of `real time` and it's importance when we utilize multiple covariates to improve forecasting. This will also be important for our forecasting menu so that the user can use any kind of time to do forecasting and mix/match varying series. 



Let me know if it breaks again 😄

jmaslek (approved):

PR #1872, opened 2022-05-26
Bottom toolbar with parameters hint - PROOF OF CONCEPT

DidierRLopes:
This is in no state to be merged.

Just wanted to share with the team something that we can do, and looking for feedback on the feature.


DidierRLopes:
@jmaslek @minhhoang1023 @JerBouma do we close this? Or is this something we should pursue a bit more?

minhhoang1023:
@DidierRLopes We could just close this one. You have already merged the toolbar so that's working well ✨ . If there are other iterations, we can open a new one.

PR #1888, opened 2022-05-31
Fix tests and bump pandas version

piiq:
This fixes our broken CI


Chavithra:
This error seems to be because of `pandas 1.4.2` :
`ValueError: Specified \n as separator or delimiter. This forces the python engine which does not accept a line terminator. Hence it is not allowed to use the line terminator as separator.`

I don't have the issue with `pandas 1.3.5`

Curious about why this is only on the following test :
`tests/openbb_terminal/alternative/covid/test_covid_controller.py`

Chavithra (approved):
Looks ok to me.

PR #1895, opened 2022-06-03
Make documentation link front and center in readme

jmaslek:
I think we should make the Hugo documentation link more clear in the README

This could also be a hyperlink, but I just did this as an idea.

Screen Shot 2022-06-03 at 11 37 18 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@
   </a>
 
   <h3 align="center">OpenBB Terminal 🚀</h3>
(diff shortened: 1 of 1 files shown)


JerBouma (approved):
Agreed.

JerBouma:
And the one here should actually be openbb.co in my opinion. We want to promote traffic to both platforms don't we.

Issue #1898, opened 2022-06-03
[Bug] Generating a report fails due to not having a `pred` attribute in the `stocks_api` module


stkerr:
**Describe the bug**

I tried to generate a report with the commands:
```
reports
1 TSLA
```

The report ran for a while and then failed mid-way through to due the error:

`AttributeError: module 'openbb_terminal.stocks.stocks_api' has no attribute 'pred'`

It appears to be failing on this [line](https://github.com/OpenBB-finance/OpenBBTerminal/blob/6aa487b590d3be178531ce83ab501a8b286525ea/openbb_terminal/reports/equity.ipynb?short_path=7b29bec#L535).

**To Reproduce**
Steps(from the start) and commands to reproduce the behavior

Open the terminal.
Go to the reports menu.
Use the text `1 TSLA` (or another ticker) to generate a report.
Notice the error message.

**Screenshots**

Screenshot below.

![image](https://user-images.githubusercontent.com/69191/171931875-0c77b831-d2b8-45fc-95c6-7006b77d71ba.png)

**Desktop (please complete the following information):**

- OS: Ubuntu 20.04
- Python version 3.8.13

**Additional context**
Add any other information that you think could be useful for us.

jmaslek:
Do you have the prediction feature flag enabled?  You can check with `featflags` on the main menu.  (see below).

The prediction packageson also needstop be installed (if using poetry: `poetry install -E prediction` )

<img width="691" alt="Screen Shot 2022-06-03 at 4 08 01 PM" src="https://user-images.githubusercontent.com/18151143/171943470-732bdc96-348a-474c-8255-6b9744b7b1ce.png">

stkerr:
@jmaslek thanks for pointing me at that. I did have the feature flag enabled and could have sworn I'd already installed the prediction package with poetry. I re-ran poetry though and after doing so the bug went away, so I must have missed it before 🤷 

We can close now. I'll try to put a PR together to catch the error and print a string to suggest folks enable the flag & install the package, rather than a stack trace.

PR #1900, opened 2022-06-04
Skip predictions in reports if prediction package is missing.

stkerr:
# Description

This PR does two main changes:
* It adds logic to detect if the prediction package is installed and skips all predictions steps if it is not.
* It explicitly opens the report file as UTF-8, which helps with OS like Windows.

This enables the report to still be generated and used, even if no prediction capabilities are present.

# How has this been tested?

I set up a local build with no prediction packages installed and with the feature flag disabled. I then ran the report with `reports/equity TSLA` and several other tickers. I confirmed that the reports generated successfully and that the prediction tab had a message about the prediction capabilities being missing.

# Checklist:

# Others


stkerr:
@jmaslek following up to the [bug](https://github.com/OpenBB-finance/OpenBBTerminal/issues/1898) I opened yesterday - this PR adds logic to check if the prediction module is present and disables the steps that need it if it is not present.

Can you review this PR? Thanks!

openbb_terminal/reports/equity.ipynb
```diff
@@ -33,7 +33,17 @@
...
+    "\n",
+    "# Detect if prediction capabilities are present. If they are not, disable prediction in the rest of the script\n",
+    "# so that the report can still be generated without prediction results.\n",
+    "predictions = True\n",
+    "try:\n",
+    "    openbb.stocks.pred.models\n",
+    "except Exception as e:\n",
+    "    predictions = False\n",
+    "    from openbb_terminal.rich_config import console, MenuText\n",
+    "    console.print(\"Failed to import predictions package. Please ensure the prediction package is installed and the feature flag is enabled.\", \"\\n\")"
```
jmaslek, line 46: To do a quick test, I uninstalled tensor flow and re-ran the equity report.  This line didn't actually trigger (namely I don't see this text in the terminal), but the rest ran without throwing any errors.

Screen Shot 2022-06-02 at 5 02 56 PM
s

jmaslek:
> @jmaslek following up to the [bug](https://github.com/OpenBB-finance/OpenBBTerminal/issues/1898) I opened yesterday - this PR adds logic to check if the prediction module is present and disables the steps that need it if it is not present.
> 
> Can you review this PR? Thanks!

Thanks @stkerr!  Great catch here.  I tested it and the console.print doesn't print to the terminal (which presumably makes sense since its through paper mill).  But the report runs without TF installed, so this bug is squashed.

@martinb-bb and @piiq : we should improve the pred api so that certain models are still available in certain cases.  For example, the mc techniques are just numpy arrays so it should be usable even if tf is not installed.

martinb-bb:
That's a great idea @jmaslek! Will look into this.

openbb_terminal/reports/equity.ipynb
stkerr: @jmaslek when this notebook runs, it looks like all output is hidden behind a progress/loading bar. I believe it should show up fine if the notebook is done directly though. That's probably not the main use case though, so let's strike these lines for now.
```suggestion
```

piiq (approved):

PR #1913, opened 2022-06-07
Add argument to remove masking for heatmap in hcorr and print correlations to command line

stkerr:
# Description

This PR adds an argument to the `hcorr` command to toggle displaying either the masked correlation matrix (as it does today) or a full, unmasked correlation matrix. The behavior defaults to off, which means the command will behave as it does today without the flga.

This PR also prints the correlation matrix to the command line. This is helpful if someone is using the terminal over SSH or otherwise doesn't want to view the results only graphically. The print logic will wrap large matrices across multiple lines, rather than truncating the column.

This is after the change and using the flag:
image

# How has this been tested?

Several tests were run with different tickers.
Pytest was run and all tests passed.

# Checklist:

# Others


jmaslek (changes requested):
So this is definitely a personal style thing, but I am not a fan of showing the whole correlation matrix.  

I am fine with this being included behind an argparse argument that allows the user to specify whether they want the whole matrix or not

jmaslek:
Also the table crops off for me

Screen Shot 2022-06-07 at 12 05 38 PM

stkerr:
@jmasley fair enough on the style point. I added a flag to allow for switching between either mode, defaulting to the current state. Picture below. I'll see what I can do about the text cropping for large matrices.

image

stkerr:
Updated print logic to prevent truncating when there are lots of symbols. It should wrap to newlines to display all the values now, as shown in this picture:

jmaslek:
Thanks for adding the flag.  The print is still off.  We have a custom method `print_rich_table` that does this formatting.  Its in `helper_funcs.py`

I would also hide the console printing behind the `raw` flag.  This one is actually easier to add.  You can just add `raw=True` to `parse_known_args_and_warn` and it will auto add the --raw flag to the ns_parser namespace.

Screen Shot 2022-06-07 at 4 09 02 PM

stkerr:
@jmaslek is this sort of what you had in mind? This feels more unreadable to me than the printed values before unless I misunderstood your previous comment.

image

jmaslek:
> @jmaslek is this sort of what you had in mind? This feels more unreadable to me than the printed values before unless I misunderstood your previous comment.
> 
> image

Yeahhh that is what I had in mind.  The actual terminal will be the limiting factor in how readable any table will be, so to keep with consistency across the code, can we keep the formatted table and then the user can change how many tickers they have loaded in.

May be overkill, but I could also envision a case where we only show print pairs that meet a criteria (--min-corr .5 --max-corr .9) but that is out of the scope here

stkerr:
@jmaslek got it, I'll go ahead and patch up the code I used to make the picture and get it submitted for review in a bit then.

> May be overkill, but I could also envision a case where we only show print pairs that meet a criteria (--min-corr .5 --max-corr .9) but that is out of the scope here

That is an interesting idea. This would be helpful if it was integrated into some sort of scanner and also to look at variables beyond just the various parts of a candle - perhaps I would want to use it to scan for stocks that had a high volume correlation but had an inverse adjusted close correlation, as an example, to find a shortlist of stocks to dig into more.

stkerr:
@jmaslek can you take a look now? There is one failing Windows test, but that doesn't seem related to this change?

jmaslek (approved):

PR #1928, opened 2022-06-10
Preferred data source for a given command and help docs for any command

stkerr:
# Description

This PR adds a new feature settings under `/settings` called `preferred_data_source_file`. It is defaulted to `sources.json`. This PR updates the `/stocks/load` command to use the preferred data source if one is not specified explicitly.

This PR also introduces a new helper function called `get_preferred_source` which can be used to retrieve the preferred value for a given command. If a specific value is not present, the function wlll return the most specific preference as possible, eventually resulting in the default value. In practice, this means a call like `get_preferred_source("stocks.subcommand.subsubcommand")` will return first the value for `stocks.subcommand.subsubcommand`, then `stocks.subcommand["default"]`, then `stocks["default"]`, then eventually the root-level `["default"]` value.

This is intended to allow different parts of Terminal to know what data source the user wants to use and then they can take action based on that.  The individual commands implemented will be responsible for examining and respecting this value.

In the future, this can be used to change the logic inside of various functions depending on which data source the user prefers, rather than hardcoding to Yahoo finance as is currently done in several parts of the Terminal. As various other data sources, such as new brokers, are added, users can select those as their preferred data source instead of Yahoo.

# How has this been tested?

Existing pytests were run successfully.

Several manual runs to change the values in the source file and then run the load command were done.

# Checklist:

# Others


DidierRLopes:
Hey @stkerr ,

I am currently away until next week, but this is something I wanted for us to tackle in a more generic way.

I was thinking about doing something along the lines of having a file with:
```
{
    'stocks/load': 'yf',
    'crypto/load`: 'cg',
    ....
}
```

And adding a command to access this file and change it in settings. Something where we could do:
`source --cmd stocks/load --value av` to change 1 value
`source` to display the dictionary in a nice way

Thoughts?

tests/openbb_terminal/test_settings_controller.py
```diff
@@ -17,6 +17,11 @@ def controller(mocker):
     return SettingsController()
 
 
+@pytest.mark.parametrize("other", [["sources.json.default"], ["-v", "sources.json.default"]])
+def test_preferred_data_source_file(controller, other):
+    controller.call_preferred_data_source(other)
+
```
stkerr, line 23:
```suggestion
```
Remove double blank line

stkerr:
@DidierRLopes @jmaslek I changed the approach based on the feedback and our discussion in Discord. Can you take a look at this PR? Thanks!

DidierRLopes:
> @DidierRLopes @jmaslek I changed the approach based on the feedback and our discussion in Discord. Can you take a look at this PR? Thanks!

On it!

DidierRLopes:
@stkerr see my PR on your branch: https://github.com/stkerr/OpenBBTerminal/pull/1

Let me know your thoughts, and we can have a chat to discuss my changes 😄

DidierRLopes (approved):

PR #1968, opened 2022-06-20
Display the help menu whenever users move to a different menu

montezdesousa:
# Description

As the title suggests the point of this PR is just to display the help menu when users move to a new menu either going forward or back. This might help avoiding users feeling lost when they return to previous menus and nothing is displayed.

My suggestion is just to put feature flag OPENBB_ENABLE_EXIT_AUTO_HELP as True by default. Also changed the call_home function to print help in the case home is called from root already. This way the behavior of home command is consistent whether it is called from root or from any other context.

# How has this been tested?

# Checklist:

# Others


Chavithra (approved):
Looks all good on my side !

minhhoang1023:
Tested for the following edge cases and it's working well :) 

- Call the menu using paths such as `../../stocks`
image

- Tested with routines - there should not be any help menu printed to avoid noise.
image

Well done for the idea & quick implementation!

PR #1974, opened 2022-06-21
Allow loading of portfolio from arbitrary paths

stkerr:
# Description

This PR allows for loading of files that are located anywhere on a user's system, rather than just those limited to the pre-determined load path.

It also adds the default load path to the help message to help the user understand where files will be loaded from first.

# How has this been tested?

Tried local test workflows several times successfully. Pytest all passes.

# Checklist:

# Others


DidierRLopes (commented):
Can you try to replicate more this style?
Screenshot 2022-06-22 at 08 55 09

E.g. you can set the `default` variable on argparse, no need to handle that as a parameter which makes it more complicated.

Then when you enter in the condition you just check if the path is valid if it is you return that, otherwise you return the same value which will have the default selected on argparse. If the path doesn't exist you should do this `log_and_raise` func

DidierRLopes:
I think this may be what you are looking for:
Screenshot 2022-06-25 at 02 15 07

The `os.getcwd()` I think could get messy really fast by having all files dropped at the root of where program is executed.

For now what you add is already an improvement, so I would be happy to merge. The only thing I may ask is if you can add this for Mac as well. Because of the paths in Mac using forward slash this breaks with the "queue" that we have set at OpenBB - we are currently working on a fix for this.

Something as easy as this should work (probably make sure that forward slash exists when provided),
```
        parser.add_argument(
            "-f",
            "--folder",
            type=str,
            dest="folder",
            help="Folder where to export data. 'default' redirects to OpenBB Terminal 'exports'",
            default="default",
        )
        if other_args and "-" not in other_args[0][0]:
            other_args.insert(0, "-f")
        ns_parser = parse_simple_args(parser, other_args)

        if ns_parser:
            if other_args or self.queue:
                if other_args:
                    export_path = ""
                else:
                    # Re-add the initial slash for an absolute directory provided
                    export_path = "/"

                export_path += "/".join([ns_parser.folder] + self.queue)
                self.queue = []
```

stkerr:
@DidierRLopes ahh ok, I see what you're saying about the lambda now. I moved the logic into a function that gets returned to replace the lambda.

For the Mac topic, I don't have a Mac to test on but do have a Linux box, which should be ok since it uses forward slashes for paths as well. Let me know if there is something Mac specific that wouldn't show up on Linux though. I think I see what you mean about the queue too - I'll see if I can figure that out!

image

DidierRLopes:
Thanks @stkerr !

Just ping me when this is ready to be reviewed and I can test it 😄

DidierRLopes:
Hey @stkerr,

Yes, that is an issue that we are currently aware of. There are 2 PRs that currently have a POOC that addresses this, and we are discussing internally the best solution to move forward.

We have been avoiding this using an "hack". Basically we just take the queue and concatenate back to the path, the issue is that if there's a file path followed by a command that will be considered all the file path. See for example `alternative/oss/sh`.

FYI @piiq @Chavithra we should bump this up on the priority queue.

DidierRLopes:
Hey @stkerr,

Can you check this https://github.com/OpenBB-finance/OpenBBTerminal/pull/2064 and give your feedback please?

@piiq had a crack at the forward slash issue

stkerr:
@DidierRLopes will do! Going to write a comment with my thoughts on that PR.

DidierRLopes:
Hey @stkerr ,

The team has internally decided to go forward with the forward slash. Is this PR still relevant?

Cheers!

montezdesousa:
I will close this PR, seems to be fixed by #2091.

PR #1996, opened 2022-06-26
Improve economy menu

DidierRLopes:
I have refactored a bit economy menu, waiting for feedback.

### DIFF
--- a/i18n/en.yml
+++ b/i18n/en.yml
@@ -786,6 +786,8 @@ en:
   etf/pred/lstm: Long-Short Term Memory
   etf/pred/conv1d: 1D Convolutional Neural Network
   etf/pred/mc: Monte-Carlo simulations
+  economy/_database_: Databases
+  economy/_stored: Stored datasets
   economy/overview: show a market overview of either indices, bonds or currencies
   economy/futures: display a futures and commodities overview
   economy/map: S&P500 index stocks map
@@ -796,7 +798,6 @@ en:
   economy/treasury: obtain U.S. treasury rates
   economy/ycrv: show sovereign yield curves
   economy/plot: plot data from the above commands together
-  economy/options: show the available options for 'plot' or show/export the data
   economy/rtps: real-time performance sectors
   economy/valuation: valuation of sectors, industry, country
   economy/performance: performance of sectors, industry, country
--- a/openbb_terminal/economy/economy_controller.py
+++ b/openbb_terminal/economy/economy_controller.py
@@ -55,7 +55,6 @@ class EconomyController(BaseController):
         "index",
         "treasury",
         "plot",
-        "options",
         "valuation",
         "performance",
         "spectrum",
@@ -145,6 +144,9 @@ class EconomyController(BaseController):
     }
     ycrv_sources = ["FRED", "investpy"]
     PATH = "/economy/"
+
+    stored_datasets = ""
+
     FILE_PATH = os.path.join(os.path.dirname(__file__), "README.md")
 
     def __init__(self, queue: List[str] = None):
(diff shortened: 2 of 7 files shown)


JerBouma:
Was already wondering when you would do this, will have a look tomorrow :)

DidierRLopes:
> Was already wondering when you would do this, will have a look tomorrow :)

Had a lot to catchup this week, but managed to have some time for this today! 😄 It wasn't as big of a change as I thought it would be as well tbf!

DidierRLopes:
Also, nvm the tests, I'll fix them after we agree on some layout for the menu :)

JerBouma:
Think you want this on the same line (and a maximum amount of options to prevent overflowing):
```
│ Stored datasets:                                                                                                                                                                     │
│   macro    : United_States_CPI                                                                                                                                                       │
│   macro    : United_States_GDP                                                                                                                                                       │
│   macro    : United_States_RGDP                                                                                                                                                      │
│                                     
```
What if I want to plot GDP and S&P 500 together in an instant, can we give the ability to allow for that without needing to call both `macro` and `index` first? That was my initial idea with `--store` (and helps within a .openbb script). I get that `--store` is undesired though.

**Idea**: you can also plot data directly from the `plot` command if you correctly enter the stored dataset names. E.g. let's say I want to see the CPI of United States and the index of S&P 500. I know that `macro`'s database name is `United_States_CPI` and I know that the index name is `sp500` so I input this into `plot` and it automatically recognises where it needs to get the data from. Bit advanced, but we can expand on that in the guide so it makes sense to all and in due time allow for auto-filing.

DidierRLopes:
> Think you want this on the same line (and a maximum amount of options to prevent overflowing):
> 
> ```
> │ Stored datasets:                                                                                                                                                                     │
> │   macro    : United_States_CPI                                                                                                                                                       │
> │   macro    : United_States_GDP                                                                                                                                                       │
> │   macro    : United_States_RGDP                                                                                                                                                      │
> │                                     
> ```
> 
> What if I want to plot GDP and S&P 500 together in an instant, can we give the ability to allow for that without needing to call both `macro` and `index` first? That was my initial idea with `--store` (and helps within a .openbb script). I get that `--store` is undesired though.
> 
> **Idea**: you can also plot data directly from the `plot` command if you correctly enter the stored dataset names. E.g. let's say I want to see the CPI of United States and the index of S&P 500. I know that `macro`'s database name is `United_States_CPI` and I know that the index name is `sp500` so I input this into `plot` and it automatically recognises where it needs to get the data from. Bit advanced, but we can expand on that in the guide so it makes sense to all and in due time allow for auto-filing.

I was thinking that that is fine, what you are doing is just adding more data to the ram so you have a faster access to it.

I don't understand what you meant around the plotting question!

JerBouma:
What I am looking for is the ability to plot data without needing to plot it first with the other functions, e.g. `macro`, `fred`, `index`. Right now, I first need to use those commands (and get two separate plots) before I can combine a graph with`macro` and `fred` data. That is kinda annoying.

Ideally, you want to plot both without needing to run the commands separately first. E.g. I say `plot --y1 United_States_CPI --y2 sp500` and it figures out the data comes from `macro` and `index`.

DidierRLopes:
> What I am looking for is the ability to plot data without needing to plot it first with the other functions, e.g. `macro`, `fred`, `index`. Right now, I first need to use those commands (and get two separate plots) before I can combine a graph with`macro` and `fred` data. That is kinda annoying.
> 
> Ideally, you want to plot both without needing to run the commands separately first. E.g. I say `plot --y1 United_States_CPI --y2 sp500` and it figures out the data comes from `macro` and `index`.

But that is a different feature than what we have right now?

JerBouma:
Think it might be good as it is right now and we can create a PR later with more features? Then we can also start writing Guides on the menu.

JerBouma:
Oui oui!

DidierRLopes:
Stamp? 😄 

Do we want to create tickets for the tasks to improve this?

JerBouma (approved):

PR #1999, opened 2022-06-27
Build workflow 1

simmonsj330:
# Description
Creates workflow that builds openbb package. Saves .openbb folder, which can be used to extend the workflow to include integration tests. Currently set to run manually by clicking 'run workflow' until we decide later on when would be an appropriate time to trigger the workflow. 

# Others


.github/workflows/intel_macos_build.yml
```diff
@@ -0,0 +1,53 @@
...
+          auto-update-conda: true
+          channels: conda-forge,defaults
+          show-channel-urls: true
+          channel-priority: flexible
+          environment-file: build/conda/conda-3-8-env-full.yaml
+          activate-environment: build_env
+          use-only-tar-bz2: true # Needed for caching some reason
+      - name: Run Poetry
+        shell: bash -l {0}
+        run: poetry install
```
piiq, line 38: you need to add a `poetry install -E prediction` to install the ML libs

piiq (dismissed):
added a single comment
💪🚀

.github/workflows/intel_macos_build.yml
piiq: like you install extras for pyinstaller

PR #2007, opened 2022-06-28
replace pip in message

LBolte29:
making sure there are no pipe characters in the message


DidierRLopes:
Do we want to merge this or not?

Chavithra (approved):
Seems to work.

PR #2010, opened 2022-06-29
Add in a new GIF that better demonstrates the terminal

JerBouma:
The point of the new one is:

- It's faster (x2) making it easier for users to see everything that is available.
- It shows more functionalities. It does skip Crypto though but the previous one also had limited options over there! @minhhoang1023, might need to re-do if you got some suggestions from commands that look interesting in a GIF? Preferences?
- It is completely up to date with the latest release.

Added this one since it's also going to be in the presentation for Open Source Finance Forum 2022.

GIF

Also did some misc fixes:

- README of installation on GitHub was outdated with broken links
- Refreshed the "First step is to star" image

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@
 
 <p align="center">
    <a href="https://www.youtube.com/watch?v=fqGPK8OVHLk" rel="OpenBB Terminal Demo">
-      <img src="" alt="OpenBB Terminal Illustration" width="1000"/>
+      <img src="images/openbb_terminal_illustration.gif" alt="OpenBB Terminal Illustration" width="1000"/>
    </a>
 </p>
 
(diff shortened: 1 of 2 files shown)


JerBouma:
Hi Theo, I wanted to upload via GitHub itself but that didn't work, that is why. I feel like this GIF does add in a lot of understanding about the terminal when you visit our repository.

DidierRLopes:
My bad on this, I didn't see the file being committed when scrolling through the code...

Now that is committed and will be on our git history, should we still remove it?

Also, can we add a test that checks if a PR adds more than xxx kB size and fails if yes?

piiq:
it's there forever and ever and ever=)

yes, we should definitely add a test for that

PR #2027, opened 2022-07-01
Add in guide to Scripts & Routines

JerBouma:
image

Also section in Getting Started:
image

Also the command `exe` now has a page:
Screenshot 2022-07-01 at 14 11 58


routines/example_with_inputs.openbb
```diff
@@ -1,5 +1,5 @@
 # This script requires you to use arguments. This can be done with the following:
-# exe routines/example_with_inputs.openbb -i TSLA,AAPL,MSFT
```
DidierRLopes, line 2: Do we not need the `routine` folder location here?

JerBouma: We do not, it automatically checks that. Buuuuut we will need to change the code when we move around the user folders to the `Documents` folder (WIP).

DidierRLopes (approved):

PR #2032, opened 2022-07-03
Fix sorting of candle data

### LINKED ISSUE #2031: [Bug] Candle: Sorting by column does not change the date using --raw tables, which is the key value.
Values change, but the date does not. The reason to use these arguments is to quickly reference the interval in question. This bug creates confusion because the interval is not correctly identified.

```
(🦋) /stocks/ $ candle --sort Volume --raw -l 5
(🦋) /stocks/ $ candle --sort Volume --raw -l 5 -d
```
Screen Shot 2022-07-03 at 1 31 08 AM

DidierRLopes:
* Fixes #2031 

FYI @deeleeramone 

Screenshot 2022-07-03 at 10 18 47


deeleeramone:
Hate to break it to you, but now the columns are not linked by sorting. 

```
load tlt -s 2002-08-01 -m
```

Screen Shot 2022-07-03 at 10 59 29 PM

Screen Shot 2022-07-03 at 11 00 10 PM
Screen Shot 2022-07-03 at 11 06 02 PM

deeleeramone:
Still funky

Screen Shot 2022-07-03 at 11 18 51 PM
Screen Shot 2022-07-03 at 11 19 54 PM
Screen Shot 2022-07-03 at 11 22 05 PM

JerBouma (approved):

DidierRLopes:
@deeleeramone the sorting doesn't work for the chart. Only for the raw data

deeleeramone:
I know this, I was simply showing that the raw data is not correct.  Look at the `adjusted close` value.

PR #2033, opened 2022-07-03
Add labels automatically when issue is created

DidierRLopes:


piiq (approved):
aaaahahahaha let's do it!

PR #2035, opened 2022-07-03
Update defaults.ini

### LINKED ISSUE #2024: [Bug] /portfolio/po/params - default.ini contains invalid parameters - i.e. 5% instead of 0.05
Loading the `deafult.ini` file in the portfolio optimization menu creates the error:
```
error: argument -mn/--maxnan: invalid float value: '5%'
```
Screen Shot 2022-06-30 at 1 59 14 PM

DidierRLopes:
Fixes #2024 

FYI @deeleeramone

### DIFF
--- a/portfolio/optimization/defaults.ini
+++ b/portfolio/optimization/defaults.ini
@@ -2,7 +2,7 @@
 historic_period=3y
 log_returns=0
 return_frequency=d
-max_nan=5%
+max_nan=0.05
 threshold_value=0.3
 nan_fill_method=time
-significance_level=5%
+significance_level=0.05


DidierRLopes:
@JerBouma that is not a regression, seems the issue was already there.

Screenshot 2022-07-06 at 13 59 16

We should have tests around this to catch these issues, seems the problem is on the setting up of these values.

JerBouma:
Agreed but I figured we should be fixing that too while we are at it. I approve of the changes you made so far.

JerBouma (approved):

DidierRLopes:
@JerBouma can you please have a go at this? Very constrained time-wise, hopefully all works well

PR #2039, opened 2022-07-03
Fix volatility of crypto

### LINKED ISSUE #1658: [FR] Annualize volatility
Typically finance folk are more interested in the annualised volatility rather than purely the standard deviation of (log) returns.

Within the `rolling` function (or wherever else volatility is displayed to the user), there should be an annualisation parameter.

To convert from the std of daily returns to an annualised volatility, you simply multiply by sqrt(252), since there are 252 trading days in a year.

For a cryptoasset that trades 24/7, you would annualise by multiplying the std of daily returns by sqrt(365).

To annualise the std of hourly returns, you would multiply the std by sqrt(365*24), etc.

**Describe the solution you would like**

Within `rolling` (or wherever else volatility is displayed to the user) there should be an additional parameter e.g `-ann 252` to specify the annualisation.

For stocks, this should default to 252. For cryptos, it should default to 365 (assuming data is daily).

DidierRLopes:
This fixes #1658.

Crypto:
Screenshot 2022-07-03 at 14 34 49

Stocks:
Screenshot 2022-07-03 at 14 35 19


robertmartin8:
Is there a reason why you're using 364 instead of 365 in all the calculations?

DidierRLopes:
> Is there a reason why you're using 364 instead of 365 in all the calculations?

Silly mistake 🥲 Thanks. Should be good now

JerBouma (approved):
Looks good, dem negative returns tho.

PR #2055, opened 2022-07-07
Updates the documentation for the candle command

deeleeramone:
Replaces the current documentation with a comprehensive guide for users.


JerBouma (commented):
Hey @deeleeramone, I would have appreciated it if we discussed together whether we also wanted to make changes to the actual command pages. This should have been documented on Jira as well.

So, I actually do not approve of this at the moment.

deeleeramone:
The intention is not to go through every individual command. The candle command represents a huge part of the basic functionality of the terminal, so this is merely a part of the entire `getting started` theme.

DidierRLopes:
@JerBouma what do we do here?

JerBouma:
@DidierRLopes I'd like to keep the individual commands the same structure and not write detailed documentation on just a few. We can think of other type of guides that can live in a separate section. This is especially true if we wish to automate the command documentation at some point.

Issue #2054, opened 2022-07-06
[Bug] Forward slash pattern got broken


DidierRLopes:
<img width="1743" alt="Screenshot 2022-07-06 at 15 35 52" src="https://user-images.githubusercontent.com/25267873/177579194-3e70a9b4-ae3b-466f-b273-102e54ed8b7e.png">

PR #2058, opened 2022-07-07
Stop consuming first element of list

jmaslek:
#2054

This code served no purpose but to mess with us
```
 if obbff.ENABLE_EXIT_AUTO_HELP and len(t_controller.queue) > 1:
       t_controller.queue = t_controller.queue[1:]
```

Also:
Screen Shot 2022-07-07 at 2 54 53 PM


DidierRLopes (approved):
Lemme approve before anyone sees this 🙃

Issue #1866, opened 2022-05-25
[IMPROVE] Add fmp, plotting and ratios to the fa menu


JerBouma:
I'd like to see the following improved:
- Adding in the source fmp ([FinancialModellingPrep](https://github.com/JerBouma/FundamentalAnalysis)) that is currently an unnecessary sub-menu (can be an option where alpha vantage is the default)
- The ability to plot data from things like income  and balance over time
- The ability to show ratios over time (and plot them)

This makes the menu compact and allows you to see metrics that matter. It means a lot more if you look at revenue growth than the individual items of the financial statements.

northern-64bit:
This issue can be closed @JerBouma

PR #2060, opened 2022-07-08
Integrating the fmp submenu into the fa menu + added plotting

northern-64bit:
# Description

Integrated the fmp submenu into the fa menu and added plotting for the income and balance. This pr is addressed at fixing the #1866 issue.

To do:

- [ ] Fix tests
- [ ] Add plotting to more commands
- [ ] Fix conflicts

# How has this been tested?

Using the cammands with various inputs

# Checklist:

# Others


jmaslek:
Curious if this is just me, but every time I run a command, I get:

```
API_KEY_FINANCIALMODELINGPREP not defined. Set API Keys in config_terminal.py or under keys menu.
```
All my other keys in the env file load in just fine.

Since I can't test that one right now, I noticed you missed this in the menu text.

Screen Shot 2022-07-08 at 6 19 19 PM

jmaslek:
> Curious if this is just me, but every time I run a command, I get:
> 
> ```
> API_KEY_FINANCIALMODELINGPREP not defined. Set API Keys in config_terminal.py or under keys menu.
> ```
> 
> All my other keys in the env file load in just fine.

For some reason there  was a file called `.env` in my folder and our `dotenv` only looks at the first one in the list.  fixed

jmaslek:
Can we add function descriptions?

```
│     profile            stocks/fa/profile                                               [FMP]                                                                                                                         │
│     quote              stocks/fa/quote                                                 [FMP]                                                                                                                         │
│     enterprise         stocks/fa/enterprise                                            [FMP]                                                                                                                         │
│     metrics            stocks/fa/metrics                                               [FMP]                                                                                                                         │
│     ratios             stocks/fa/ratios                                                [FMP]                                                                                                                         │
│     growth             stocks/fa/growth                                                [FMP]  
```

jmaslek:
In the functions that have a limit argument, can we have the limit be more than 1 (ratios, metrics,...).  5 is my happy number.

jmaslek:
Thanks for making the changes!

For plotting, can we get the possible plot columns into the auto-complete, so when I do `income -p` a list of possible columns will drop down.  This will help understand what we can plot:
Screen Shot 2022-07-09 at 12 48 26 PM

jmaslek:
Should figure out a way to have units on the axis. Plotting brings back the scientific notation, which is buried under the title :/
Screen Shot 2022-07-09 at 12 54 07 PM

northern-64bit:
> For plotting, can we get the possible plot columns into the auto-complete, so when I do `income -p` a list of possible columns will drop down. 
The problem with this is that the possible columns to plot all change with the source. The columns to plot is just the column name from the rich table all lowercase with undercores instead of backspace: net_income

Maybe it is enough to state this clearly in the help command or what do you think?

jmaslek:
> The problem with this is that the possible columns to plot, all change with the source. The columns to plot is just the column name from the rich table all lowercase with undercores instead of backspace: net_income
> 
> Maybe it is enough to state this clearly in the help command or what do you think?

Fair point.  I think we can do a workaround by using the default source from the json file (data_sources_default.json) to load in the possible titles?

northern-64bit:
> Fair point. I think we can do a workaround by using the default source from the json file (data_sources_default.json) to load in the possible titles?

Sure, I just think that will confuse users in case they want to use an other source, however not many use many other sources than the default one. But I'll implement it as mentioned above.

jmaslek:
> Anything still required to be build here?

Looking good to me.  Just need to get the conflicts resolved then we can merge.

PR #2063, opened 2022-07-09
Fix failing exit

### LINKED ISSUE #2053: [Bug] Exit fails because of the way the we are trying to get parameters form controller below
Screenshot 2022-07-06 at 15 36 17

@colin99d you implemented this solution, could you have a look at what's going on? The issue is in that area and has to do on how you get parameters from controllers that have been previously loaded

colin99d:
# Description

There is no need to manually set the attributes of the class. The dictionary remembers them. Fixes #2053.

# How has this been tested?

Reran `r` and did not get an error.

# Checklist:

# Others

### DIFF
--- a/openbb_terminal/portfolio/portfolio_optimization/po_controller.py
+++ b/openbb_terminal/portfolio/portfolio_optimization/po_controller.py
@@ -12,7 +12,6 @@ from typing import List, Dict
 from prompt_toolkit.completion import NestedCompleter
 
 from openbb_terminal import feature_flags as obbff
-from openbb_terminal import parent_classes
 from openbb_terminal.decorators import log_start_end
 from openbb_terminal.helper_funcs import (
     check_non_negative,
@@ -530,13 +529,6 @@ class PortfolioOptimizationController(BaseController):
             self.params,
             self.current_model,
         )
-        self.current_file = parent_classes.controllers[
-            "/portfolio/po/params/"
-        ].current_file
-        self.current_model = parent_classes.controllers[
-            "/portfolio/po/params/"
-        ].current_model
-        self.params = parent_classes.controllers["/portfolio/po/params/"].params
 
     @log_start_end(log=logger)
     def call_show(self, other_args: List[str]):
--- a/tests/openbb_terminal/portfolio/csv/test_portfolio_model/test_common_sense_ratio.csv
+++ b/tests/openbb_terminal/portfolio/csv/test_portfolio_model/test_common_sense_ratio.csv
@@ -11,14 +11,14 @@ dtype: float64"
 3m,"TotalHoldings    0.68
 dtype: float64","Adj Close    0.613
 dtype: float64"
-6m,"TotalHoldings    0.731
-dtype: float64","Adj Close    0.699
+6m,"TotalHoldings    0.733
+dtype: float64","Adj Close    0.693
 dtype: float64"
-1y,"TotalHoldings    0.571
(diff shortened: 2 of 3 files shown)


DidierRLopes (changes requested):
This doesn't fix the issue, but masks it.

In this case we can't get the data from the submenu to the `po` menu. This is not the goal here.

colin99d:
Because I think we just need to change the way we do call_reset. Although I might be wrong.

DidierRLopes:
> Every or just when you reload?

So the idea is that if you change stuff on the submenu (params) that affects the menu above (po). By removing those lines of code you are not getting anything that was updated on the submenu.

You avoid the warning but lose the capability of getting that data.

BTW, it's not only on "reset" the issue. Also on "exit"

colin99d:
@jmaslek feel free to use your "dirty" fix here. I never intended for this to be a functionality we had, so it would take me some thinking to come up with an answer.

DidierRLopes:
> @jmaslek feel free to use your "dirty" fix here. I never intended for this to be a functionality we had, so it would take me some thinking to come up with an answer.

@colin99d I was pretty certain that this functionality was to access variables within a controller. Hence why you created a dictionary of the controllers, to access them. 

Otherwise, what was the use case? I can't recall

colin99d:
> > @jmaslek feel free to use your "dirty" fix here. I never intended for this to be a functionality we had, so it would take me some thinking to come up with an answer.
> 
> @colin99d I was pretty certain that this functionality was to access variables within a controller. Hence why you created a dictionary of the controllers, to access them.
> 
> Otherwise, what was the use case? I can't recall

I created it so that if you left a menu and then came back, it would still be saved.

DidierRLopes:
> > > @jmaslek feel free to use your "dirty" fix here. I never intended for this to be a functionality we had, so it would take me some thinking to come up with an answer.
> > 
> > 
> > @colin99d I was pretty certain that this functionality was to access variables within a controller. Hence why you created a dictionary of the controllers, to access them.
> > Otherwise, what was the use case? I can't recall
> 
> I created it so that if you left a menu and then came back, it would still be saved.

Yup. So that means that that controller in that dictionary should have access to those variables because they were "saved" in advance.

This sounds like it's the use case you added. The main thing to understand is why you can't access those variable from the dictionary directly? Do we need to somehow load it first or something? Because if you quit the menu, and then do it again, I believe it works

colin99d:
> Yup. So that means that that controller in that dictionary should have access to those variables because they were "saved" in advance.
> 
> This sounds like it's the use case you added. The main thing to understand is why you can't access those variable from the dictionary directly? Do we need to somehow load it first or something? Because if you quit the menu, and then do it again, I believe it works

I imagine we need to do modify the reset function somehow. Also I am confused about exit, doesnt exit leave the entire terminal?

DidierRLopes:
> > Yup. So that means that that controller in that dictionary should have access to those variables because they were "saved" in advance.
> > This sounds like it's the use case you added. The main thing to understand is why you can't access those variable from the dictionary directly? Do we need to somehow load it first or something? Because if you quit the menu, and then do it again, I believe it works
> 
> I imagine we need to do modify the reset function somehow. Also I am confused about exit, doesnt exit leave the entire terminal?

Yes. The issue is the same in both reset/exit. The issue is in the fact that the controller doesn't exist on the dictionary when we leave the params menu. Hence why I asked you to have a look at it since it should have been "saved".

DidierRLopes:
Check #2090

PR #1713, opened 2022-04-18
Fix forward slash used for navigation and for parameters

DidierRLopes:
This fixes #1620. This is an alternative to #1634.

Let's discuss this this week to make a decision regards this issue.

@piiq @jmaslek @Chavithra @JerBouma @jose-donato @minhhoang1023

PR #1634, opened 2022-04-04
Replaces '/' with ';' for running multiple commands

bolshoytoster:
This is probably not complete (documentation hasn't been updated), and it breaks things like `../dd` but it's better than not being able to input paths. Fixes #1620. 

- [x] parents_classes
- [x] scripts folders
- [x] routines folders
- [x] call_exe and call_sh funds that have the hacky version to parse the "/" - there's prob 1 more functionality like this

# How has this been tested?

* Please describe the tests that you ran to verify your changes. 
I can run all `*.openbb` files excluding `openbb_terminal/timezone.openbb`
* Provide instructions so we can reproduce. 
Run `exe ./routines/example.openbb`
* Please also list any relevant details for your test configuration.

# Checklist:

- [x] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [x] Update our tests following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/tests).
- [x] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).
- [x] If a feature was added make sure to add it to the corresponding [scripts file](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/scripts).

# Others
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My code passes all the checks pylint, flake8, black, ... To speed up development you should run `pre-commit install`.
- [x] New and existing unit tests pass locally with my changes. You can test this locally using `pytest tests/...`.

Issue #1620, opened 2022-04-03
[Bug] slash issue


andr0id3000:
**Describe the bug**
A "bad" load from the econometrics menu will crash the terminal. "Bad" loads have included so far: not giving an alias along with the filename, and setting the filename directly without using the auto-complete feature (which is dope, ty.) 

**To Reproduce**
from the terminal, enter the econometrics menu and try to explicitly load a filepath (i used one i just stored from FRED data) - `load -f /home/<user>/code/OpenBBTerminal/exports/icsa-test.csv` with or without an `alias` it fails, `load: error: argument -f/--file: expected at least one argument` which it not accurate, as it was given an argument. 

**Screenshots**
<img width="1728" alt="Screen Shot 2022-04-03 at 12 06 07 AM" src="https://user-images.githubusercontent.com/102889846/161410966-62b940e1-f490-438a-9b38-b6742a421d1e.png">

**Desktop (please complete the following information):**

- OS:  (Host) Mac OS X 12.2.1 running Parallels running (Guest) Ubuntu 20.04.2 ARM64
- Python version = 3.8.13

**Additional context**
Should be easy to reproduce, hmu in Discord otherwise, happy to help troubleshoot.

bolshoytoster:
@andr0id3000 Looks like an issue here:
https://github.com/OpenBB-finance/OpenBBTerminal/blob/5cc2ee9e142837b5d92750c5bd43a82d1419e1c8/openbb_terminal/parent_classes.py#L160-L171

If it finds the `/` character anywhere in your input it splits your command into seperate commands by `/` (if I explained it properly)

For example, your command (`load -f /home/q/code/OpenBBTerminal/exports/icsa-test.csv`) gets changed to:
```sh
$ load -f
gives the error
$ home
changes to home
$ q
quit the program
```
I don't know why it does this, but we should probably change the character used to something like `;` so we can enter paths.

You could go into the afformentioned piece of code and change:
https://github.com/OpenBB-finance/OpenBBTerminal/blob/5cc2ee9e142837b5d92750c5bd43a82d1419e1c8/openbb_terminal/parent_classes.py#L161

to
```py
        elif ";" in an_input:
```
and
https://github.com/OpenBB-finance/OpenBBTerminal/blob/5cc2ee9e142837b5d92750c5bd43a82d1419e1c8/openbb_terminal/parent_classes.py#L162

to
```py
            actions = an_input.split(";")
```
[...]

PR #2064, opened 2022-07-10
Add parser for input command arguments (The Forward Slash PR)

piiq:
**Problem:**

Our input parsing behaves incorrectly when there is a forward slash in the command queue.
It is known to affect unix style paths and sorting arguments for screeners.

There are issues reporting this and PRs proposing fixes.
https://github.com/OpenBB-finance/OpenBBTerminal/issues/1620
https://github.com/OpenBB-finance/OpenBBTerminal/issues/2119
https://github.com/OpenBB-finance/OpenBBTerminal/pull/1634
https://github.com/OpenBB-finance/OpenBBTerminal/pull/1713
https://github.com/OpenBB-finance/OpenBBTerminal/pull/2068
https://github.com/OpenBB-finance/OpenBBTerminal/pull/1974

**Solution:**

This PR introduces a filtering function that searches for paths and special arguments and replaces them with placeholders for the time it is splitting the input into individual commands.

This should resolve the problem of unix paths or github repo names to be considered individual commands.

**Along with the fix this PR introduces the following rules:**

1. The `-f` argument is reserved for file paths
2. When loading data from a file by providing a full path the `-f` or `--file` argument is mandatory

`load /mydir/myfile.xls/ta/ema`  <- will NOT work
`load -f /mydir/myfile.xls/ta/ema`  <- will work

Custom per-controller argument filters allow parsing edge cases in specific controllers. For example in `alternative/oss` this:
`sh openbb-finance/openbbterminal/../covid`  <- will work

**Linting:**
To support the rules a linter is added to the pre-commit hooks that checks that the short reserved command arguments are used with correct long command arguments

**Additional fixes:**

- fixed portfolio model tests
- fixed finviz screener tests


piiq:
There was a problem with tests related to dep version mismatch in the runner envs that forced recreating the test data. Sorry for that 🤷‍♂️

openbb_terminal/settings_controller.py
```diff
@@ -467,7 +467,6 @@ def call_export(self, other_args: List[str]):
             description="Select folder where to export data",
         )
         parser.add_argument(
-            "-f",
             "--folder",
```
DidierRLopes, line 470: If you remove the "-f" from here, you should replace the `-f` below in line `478` for `--folder`.

And this should be checked in the other scenarios

openbb_terminal/stocks/behavioural_analysis/ba_controller.py
```diff
@@ -411,7 +411,6 @@ def call_reddit_sent(self, other_args: List[str]):
             help="time period to get posts from -- all, year, month, week, or day; defaults to week",
         )
         parser.add_argument(
-            "-f",
             "--full_search",
```
DidierRLopes, line 414: Since we are at it can we remove the "_", we need to try to make it simple

openbb_terminal/econometrics/econometrics_controller.py
```diff
@@ -322,8 +323,6 @@ def call_load(self, other_args: List[str]):
             dest="examples",
         )
 
-        if other_args and "-" not in other_args[0][0]:
```
DidierRLopes, line 325: Why was this removed?

This is quite important for me to stay as it makes for a much better UX

DidierRLopes (changes requested):
A couple of minor things.

With what functionality are you testing?

openbb_terminal/econometrics/econometrics_controller.py
piiq: The rationale here is that usage of the `-f` flag is the same in all the menus of the terminal.

If in the econometrics you can load a dataset without specifying that it comes from a file then one can likely assume the same behaviour in other menus where we don't have such argument pre-pending

openbb_terminal/settings_controller.py
piiq: good catch, thanks

DidierRLopes (approved):

stkerr:
I'm going to take some time to review the code to provide additional comments but I wanted to start a discussion on the approach in general first.

OpenBB takes the opinionated stance that `/` is the single way to separate commands. This means all commands must acknowledge this and anything that requires forward slashes must respect this and accommodate it, such as using the new `-f` syntax. Examples are filepaths and fractions, though their could be others in the future. It is fine to be opinionated.

My question is _why_ is this the opinion that is held for the best way to separate commands? Shells (like Bash, TCL, etc) typically use semicolons, `&&`, or `||` to separate commands so it is surprising to me that forward slashes are used. My personal opinion is that using semicolons such as that proposed in #1634 would be simpler and wouldn't require as much special processing for things as common as filepaths. I would wager many people coming to OpenBB would expect something like a semicolon to be used and would be surprised that a forward slash is the command separator (that was my personal experience when I discovered it.)

piiq:
Thanks for the question @stkerr 

Back in the days (a year ago 😂) we had a lengthy discussion on the separator and the slash was choses because the contexts/menus structure resembles the folder tree. The use case would be to jump into contexts and menus from the root of the terminal as if calling a command from a nested folder in a shell. `/stocks/disc/gainers` would get you to the `gainers` command in the `disc` "folder" in the `stocks` "folder". And a single `/` takes to the terminal root.

```
2022 Jul 18, 16:16 (🦋) /stocks/disc/ $ gainers
```

The options that we had on the table were slash `/`, semicolon `;` and the pipe symbol `|` and after a discussion on discord and a vote we sticked to the slash for simplicity. At that time we didn't support importing or exporting data to arbitrary paths and the controllers looked different to what we have right now. So the slash has got entrenched in the functionality of the controllers, the routines we write to test and use the terminal and the docs.

Having a custom arbitrary separator (`;` or `&&` or anything else) is totally possible. Nowadays most of the terminal-wide customizations are done through feature flags. Do you think the command separator should be customizable like that? I'm open to jump on a call on discord to discuss a customization PR

piiq:
Hey folks! I've added a few fixes and updated this PR so it's ready to merge once the tests pass.

After the merge it would be great to look into enabling loading of portfolio files from arbitrary paths #1974

PR #2069, opened 2022-07-11
Creates a controller to specify sources

DidierRLopes:
Screenshot 2022-07-11 at 23 12 58


JerBouma:
Can't this be under the `settings` menu? I'd like to keep the Home menu small.

DidierRLopes:
> Can't this be under the `settings` or `featflags` menu? I'd like to keep the Home menu small.

This one is quite different from settings and featflags, and some users have been asking about this. So I would prefer to have it in it's own menu for more visibility :)

DidierRLopes:
> Couple of points.
> 
> Can't these either share the `stocks/load` structure or otherwise what we also use within Econometrics (`stocks.load`). This looks a bit odd.
> 
> image
> 
> Could you add some information within this menu how to set a different source? E.g. a text stating "_Use `get <command>` to figure out what sources are available. Use `set <command> -s <source>` to alter the default source. E.g. `set stocks_load -s iex`_"

The "/" and the "." unfortunately break the auto-completer.

I improved the documentation when doing "-h"!

PR #2072, opened 2022-07-12
Fill missing static data (sector, industry, country and region) on portfolio creation

montezdesousa:
# Description

This PR populates the loaded orderbook with sector, industry, country and region if they are not provided by user. This will be useful to allow integration with broker accounts that just provide transaction data (e.g. price, quantity, ticker...) and not much static data. There's a PR already trying to start this integration by exporting a formatted transaction file. See https://github.com/OpenBB-finance/OpenBBTerminal/pull/2065

# Others


JerBouma:
So, when I don't fill anything in the Orderbook, but do keep the columns it doesn't work. Should we remove those altogether? I think we should so there is less manual input required.

image

If we do, however, it takes longer to load. Could you add in a progression bar or, otherwise, an option to export to .csv so it loads quicker?

JerBouma:
@montezdesousa How far are you with this PR?

montezdesousa:
> @montezdesousa How far are you with this PR?

I'll be on it this afternoon

JerBouma:
Is it ready?

montezdesousa:
> Is it ready?

yes

DidierRLopes:
@JerBouma is this good to be merged?

montezdesousa:
> @JerBouma is this good to be merged?

@JerBouma as I mentioned to you, my implementation of returns was very poor performance wise, so I added a new commit that I believe fixes this (if ok calculating returns went from 25s to 0.6s) and the progress bar does not need to be that long

JerBouma (approved):
Sounds great!

PR #2080, opened 2022-07-13
Corrected wording

ghbcode:
# Description 

In the 'portfolio' context for the 'po' category it read "optimal your portfolio weights efficiently". It would be better/more clear to say "optimize your portfolio weights efficiently". There is no issue or bug related to this as it is purely a wording update.

# How has this been tested?

* All three locations/files where this wording occurred have been checked and confirmed to be correct and matching.
* Ran tests/portfolio/test_portfolio_controller.py and all tests (most importantly test_print_help) passed

# Checklist:

* No features were added. Only text change. 

# Others

* None


DidierRLopes (approved):
Nice one, thank you 🙏🏽

PR #2083, opened 2022-07-13
Add in DeFi, DD and NFT guides from the Crypto menu

JerBouma:
Also added some minor changes to the Getting Started guide.

image

image

image


JerBouma:
@jmaslek As a crypto lover, knew you would come through for me <3

PR #2068, opened 2022-07-11
DRAFT : Fixes using arguments /, chaining multiple commands and exe bug

pftp-jz:
Possible error fixes/ suggestions / tests
Fix error where arguments with / would be interpreted as commands
Fix error that exe function / scripts were not running properly
Make possible to chain commands using many different controllers from any place
Possible could be used as a base to run exe from a any path inside terminal

# Description

- [x] Summary of the change / bug fix.
- [ ] Link # issue, if applicable.
- [ ] Screenshot of the feature or the bug before/after fix, if applicable.
- [x] Relevant motivation and context. 
- [x] List any dependencies that are required for this change.

# How has this been tested?

passed 100% > pytest tests
tested using exe in a few routine files
tested using long commands -> home/crypto/nft/today/upcoming/../nft/../ov/hm/../nft/today/home/stocks/th/closed/
home/etf/disc/home -> in home, and inside others paths

# Checklist:

- [ ] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [ ] Update our tests following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/tests).
- [ ] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).
- [ ] If a feature was added make sure to add it to the corresponding [scripts file](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/scripts).

# Others
- [X] I have performed a self-review of my own code.
- [+-] I have commented my code, particularly in hard-to-understand areas.
-                   -> needs more reviewers to verify if its clear enough or more explanation needed
- [ ] My code passes all the checks pylint, flake8, black, ... To speed up development you should run `pre-commit install`.
- [X] New and existing unit tests pass locally with my changes. You can test this locally using `pytest tests/...`.

PR #2093, opened 2022-07-15
Able to change the data range in some stock controllers

pftp-jz:
# Description

new feature

Adding a way to change the range of the data being used without fetching it again, and reverse the change to the original range.

For stocks_controller, dps_controller, ba_controller,  dd_controller, fa_controller, gov_controller, pred_controller, qa_controller, ta_controller

For stocks_controller changes in range affect all controllers below it 

For dps_controller, ba_controller,  dd_controller, fa_controller, gov_controller, pred_controller, qa_controller, ta_controller - data ranges revert for stocks_controller current range once you exit the controller

Useful to perform analysis with different range in data and see the variation of the analysis

One example is : This way you can load data for training in ml them simple loadrange -e -30, and spare the last 30 days for testing your model with the most recent data.

# How has this been tested?

passed all pytest tests
Tested new functionality in terminal

still needs to be develop tests using it, but it's a simple function, it only slices a df into a new df and revert the change.

Weirdly one test is failing, from a function i didn't modify
-> comparison_analysis -> test_finviz_compare_view.py -> test_screener

# Checklist:

# Others

### DIFF
--- a/i18n/en.yml
+++ b/i18n/en.yml
@@ -86,6 +86,8 @@ en:
   stocks/SEARCH_industry: Search by industry to find stocks matching the criteria
   stocks/SEARCH_exchange: Search by a specific exchange country to find stocks matching the criteria
   stocks/load: load a specific stock ticker and additional info for analysis
+  stocks/loadrange: change the range of currently loaded data
+  stocks/loadfulldata: goes back to the fully loaded data
   stocks/_ticker: Stock
   stocks/quote: view the current price for a specific stock ticker
   stocks/QUOTE: Current quote for stock ticker
@@ -220,6 +222,8 @@ en:
   stocks/sia/_returned_tickers: Returned tickers
   stocks/sia/ca: take these to comparison analysis menu
   stocks/dps/load: load a specific stock ticker for analysis
+  stocks/dps/loadrange: change the range of currently loaded data
+  stocks/dps/loadfulldata: goes back to the fully loaded data
   stocks/dps/shorted: show most shorted stocks
   stocks/dps/ctb: stocks with highest cost to borrow
   stocks/dps/hsi: show top high short interest stocks of over 20% ratio
@@ -283,6 +287,8 @@ en:
   stocks/gov/qtrcontracts: quarterly government contracts analysis
   stocks/gov/toplobbying: top corporate lobbying tickers
   stocks/gov/load: load a specific ticker for analysis
+  stocks/gov/loadrange: change the range of currently loaded data
+  stocks/gov/loadfulldata: goes back to the fully loaded data
   stocks/gov/_ticker: Ticker
   stocks/gov/gtrades: show government trades
   stocks/gov/contracts: show government contracts
@@ -314,6 +320,8 @@ en:
   stocks/ca/technical: brief technical comparison
   stocks/ca/po: portfolio optimization for selected tickers
   stocks/ba/load: load a specific stock ticker for analysis
+  stocks/ba/loadrange: change the range of currently loaded data
+  stocks/ba/loadfulldata: goes back to the fully loaded data
   stocks/ba/_ticker: Ticker
   stocks/ba/headlines: sentiment from 15+ major news headlines
   stocks/ba/snews: stock price displayed over sentiment of news headlines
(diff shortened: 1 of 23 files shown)


pftp-jz:
loadfulldata will very likely be merged into loadrange, with a argument like, loadrange -r --reset,

DidierRLopes:
Hey @pftp-jz 

Thanks for this!

Instead of adding 2 new commands, why not have:
```
loadrange =  load TICKER --rangestart -30 --rangeend -10
```
and
```
loadfulldata =  load TICKER -a
```

That makes it easier since we are talking about the same function, and just a different way to have access to its data

DidierRLopes (changes requested):
Comment above

pftp-jz:
After making it, i realised it was kinda dumb to make an entire function for loadfulldata

I talked about it in discord about merging  into loadrange, and also add more functionality to it later.
I was afraid of crowding Load since it already has 11 arguments.

New functionalities like adding Len, split by % of data, split by specific date (see effect of news), split by specific quarter (companies reports)  & etc...

DidierRLopes:
> After making it, i realised it was kinda dumb to make an entire function for loadfulldata, loadfulldata could simple be -rd --reloaddata
> 
> I talked about it in discord about merging into loadrange, and also add more functionality to it later. I was afraid of crowding Load since it already has 11 arguments.
> 
> New functionalities like adding Len, split by % of data, split by specific date (see effect of news), split by specific quarter (companies reports) & etc...
> 
> So what should i do, merge both to load anyway, or keep loadrange, to add more functionality to it later?
> 
> @DidierRLopes

Good points.

Let's get some more feedback here: @jmaslek @JerBouma @minhhoang1023 @piiq

piiq:
My only comment on this particular implementation is that -s (--start) and -e (--end) are used as date (datetime) arguments everywhere in the terminal and to me it seems that it's better to keep it this way. 

On the conceptual side top of my head (I'm just juggling ideas here, not sure) I would store the a `self.range` instead of the second dataframe and use the range to slice the `self.stock`. That always gives me all the data in the controller and a way to select a subset for analysis.

In addition since the slash issue will be resolved by this https://github.com/OpenBB-finance/OpenBBTerminal/pull/2064 PR it will open opportunities to use own data loaded from anywhere on the computer and IMO that gives a lot more control over what data I do the analysis on.

@pftp-jz can you post an end-to-end command sequence that utilizes these new functions (like an .openbb routine file)? Maybe I'm missing something and your example will help understand more of the context?

pftp-jz:
@piiq nice catch, to be consistent would be better to have -s -e reserved for the datetime arguments

It probably would be better to store only the range than a copy of the subset of a dataframe, But, all functions already used self.stock as a base, this way you can simple change the range and execute all the functions as usual, if you want to store range and use a subset, you will have to either add argument to all functions using the different data to use the subset or the full data + the logic in each of them. any addition/modification to different types of arguments could result in the same work being done for extra functionality. 
if you refer to it only as range and not a copty like df2 = df1[-30:], any changes to in /stocks/ going to another stocks/ta/ will throw a error, self.stock need to be an df not slice of one

i also posted a PR https://github.com/OpenBB-finance/OpenBBTerminal/pull/2068 for that problem with the slashs + other problems too, the regex i used match for any argument using / in paths not only load we probably could mix our solutions to get something better, like --export function for instance. -> there is also other solutions here like queue commands from any path, like /home/stocks/load AMZN/fa
But this way you don't have to reload data from source every time, polygon/ other sources have some limits  to the amount of API calls you can make, and there is no reason to make 4 calls to get data you already have

There is lots of examples, like i mentioned you can spare a part of your to train using regression and use the last N rows as a test dataset for the result. its mostly useful for PRED, QA and TA, you can go back and forth from past "events" (news quarters & etc...) . We can make weekly / month analysis of the data, last 22 days analysis & etc...

piiq:
@pftp-jz all these very good points

@JerBouma will you have time to look at this PR and chip in? The use case is:
- To have a subset(s) of loaded data available to pass into individual commands without the need to reload the data from the source every time. Do you see any solution usage-wise?
Let's brainstorm this together

JerBouma:
This is an interesting idea but I wonder, why is it such a problem that you need to fetch the data again? It doesn't take a tremendous amount of time and even though you need to call the API again, how often do you really change this range? I get the ML example so in that case include this option within the `load` command? I find having two additional commands just for a, to me, quite specific use-case not the ideal solution.

Furthermore, I don't really like the way the amount of days is modelled. The market is not open 24/7 so the user could assume he is looking at the last month when he inputs `-30` days while he is actually (most likely) looking at the last 1.5 months. Or do you properly model for this?

DidierRLopes:
I am closing this one as the use-case may be too specific. We may look into this in the future, and we can then re-open this PR. Thanks!

PR #2097, opened 2022-07-16
Fix API

northern-64bit:
# Description

Implement a new and improved API

# How has this been tested?

# Checklist:

# Others


openbb_terminal/api.py
```diff
@@ -62,59 +98,131 @@ def __call__(self, *args, **kwargs):
...
+            spec = None
+
+        if spec is None:
+            return None
+        else:
+            module = importlib.import_module(module_path)
+
+            return module
 
     @staticmethod
```
Chavithra, line 172: If you need the object attributes/methods it's a normal method.
If you need to use the class properties/functions inside the function it's better to turn it into a `@classmethod`. 
If you don't need anything related to the object or the class you can turn it into a `@staticmethod`.

Here I would turn it into a `@classmethod`.

Chavithra (changes requested):
Maybe we can start using that on reports to see how it behaves.

We might also need to find more explicit names for the following classes:
- `APIFactory` 
- `Item`

openbb_terminal/api.py
northern-64bit: Done!

minhhoang1023:
Will we be able to get code introspection as discussed in the call?

Even though less desirable, docstring would still contain the information users need. 
 
The disadvantage I see is that we would need to put in more effort to maintain docstrings. Previously, we could get away with having an extra or missing param in the docstring. Without the code introspective, docstring is the only information users have without having to go through the source code. So we really need to make sure docstring is correct and up-to-date.

northern-64bit:
> Is there anything we can do @Chavithra @northern-64bit to expose the args and kwargs that the children have through the factory?

This is now fixed
image

PR #2113, opened 2022-07-19
Remove defirate scraping

piiq:
3 functions in the defi menu that are reliant of scraping of defirate.com are not working.

funding, lending and borrowing

This PR removes them as a part of the general move away from being reliant on scrapping data. In addition to this the layout at defirate.com changed in a way fixing is not straightforward.

The defirate commands were last updated 9 months ago and the integration tests were failing for a month or so


DidierRLopes (approved):
"killing spreeee" ☠️ 🔫

PR #2127, opened 2022-07-21
Fix ycrv to allow raw display and export when source is FRED

montezdesousa:
# Description

This PR fixes the --raw and --export outputs when `ycrv --source FRED --raw --export csv` is used. Currently, those flags only work for the default source `investpy`. If user picks FRED he cannot display raw data or export it to csv. 

image

Test:
ycrv --source FRED -d 2021-01-04  --raw --export csv

# Others


jmaslek:
Should we include logic  that if a country is selected AND fred source, that it won't work.  Namely 

``
ycrv -c canada --source FRED
``

gives US yield curve

jmaslek (approved):

PR #2145, opened 2022-07-23
Added a Historical Beta Regression Plot

ricwillis98:
# Description
Added beta regression plot to stocks/qa/beta

Added the ability to view the regression plot where beta comes from (with respect to any stock or ETF). 

N/A

image

This feature is particularly useful for checking the beta of a stock relative to it's sector ETF e.g. TSLA with respect to XLK to put on a sector hedge/hedged trade.

None

# How has this been tested?
Added unit tests

N/A 

# Checklist:

# Others


DidierRLopes:
Not sure if I would put this feature on `portfolio` menu.

@JerBouma @jmaslek thoughts?

colin99d:
> Not sure if I would put this feature on `portfolio` menu.
> 
> @JerBouma @jmaslek thoughts?

I know I wasn't asked, but a lot of people use beta as a way to measure market risk. So when creating a lot of portfolios beta is a big factor. So I think it would make sense to have it here.

ricwillis98:
The use case I had in mind, other than just looking up betas, is to check the beta of a single stock with respect to a sector ETF e.g. if I want to trade AMD with a sector hedge, I need to trade ~1.3x more XLK than the size of my AMD trade.

It doesn't seem like putting this in the portfolio menu would (easily) allow this because I can't find a way to create a simple portfolio of one stock. This would be a great feature for the portfolio menu as well though because @colin99d is right beta is important to portfolio construction also but I think the use cases are slightly different. 

Up to you guys though, let me know how to proceed. 

image

JerBouma:
It makes sense to have this somewhere. We already have rolling betas for the `portfolio` menu so we are good there. That is specifically meant for a larger collection of tickers.

To have this in the `ca` menu sounds good to me.

DidierRLopes:
> It makes sense to have this somewhere. We already have rolling betas for the `portfolio` menu so we are good there. That is specifically meant for a larger collection of tickers.
> 
> To have this in the `ca` menu sounds good to me.

My view is that comparison analysis is to compare tickers in terms of historical, volume, balance sheet, ... the use case here seems to be very distinct. This shouldn't include any ETFs. Plus you can only select a stock and ETF which seems completely out of style of `ca`.

Portfolio menu should allow to do this as the use case that @ricwillis98 is mentioning is having just 1 stock in said portfolio. Plus, it's looking into an ETF which is something that `portfolio` already supports.

As a matter of fact @JerBouma, `portfolio optimisation` used to support creating your portfolio by adding tickers. Should we have it there instead?

jmaslek:
Im team @JerBouma here.  Ive actually had this on a long list of TODOs -- having a function for beta in `ca`.  `ca` should absolutely not be stock tickers.  You even wrote the code to link ETF -> ca.  If I wanted to compare historical performance of SPY to QQQ, I would do so in a comparison menu.

Portfolio doesn't make sense, as I am researching a specific tickers beta.  But beta should also be an optimization factor as many Strats will look to keep a certain beta range.

DidierRLopes:
> Im team @JerBouma here. Ive actually had this on a long list of TODOs -- having a function for beta in `ca`. `ca` should absolutely not be stock tickers. If I wanted to compare historical performance of SPY to QQQ, I would do so in a comparison menu.
> 
> Portfolio doesn't make sense, as I am researching a specific tickers beta. But beta should also be an optimization factor as many Strats will look to keep a certain beta range.

But we are talking about `stocks/ca` here. We should also have a `etf/ca` for sure, but that's another story - and something I had in my TODO list. I just never worked on it because passive investor spreadsheet is pretty neat and you can compare multiple ETFs in there.

My only (big) problem with having this in `stocks/ca` is that it breaks the entire pattern of what we have currently:
* It brings `ETFs` to the mix, which breaks the point of being within `/stocks` menu.
* All the other functionalities have no limitation in the number of tickers you select, since the more you select the more you will be comparing. This one would have a limitation AFAIK.

This is why I think portfolio is more recommended since the whole menu is built allowing you to load a portfolio of any asset and an ETF as a benchmark.

jmaslek:
What if I wanna look at beta between SBUX and F?

JerBouma:
I get your point @DidierRLopes. It isn't in line with our structure we have in place. However, being able to calculate a beta to some ETF is common practice to gauge risk of that asset. As you will be doing that for one asset always, the portfolio menu doesn't make sense either.

I think it's fine you have a command that takes one stock and one ETF, you don't have to load in the ETF, you just have to select it in the command. It could even be an index.

DidierRLopes:
> What if I wanna look at beta between SBUX and F?

That's a different use case from the one presented above:

> This feature is particularly useful for checking the beta of a stock relative to it's sector ETF e.g. TSLA with respect to XLK to put on a sector hedge/hedged trade.

But if that was the case, I still wouldn't think that `stocks/ca` is the right place because you are not comparing them but you need both to do the beta calculation.

DidierRLopes:
> I get your point @DidierRLopes. It isn't in line with our structure we have in place. However, being able to calculate a beta to some ETF is common practice to gauge risk of that asset. As you will be doing that for one asset always, the portfolio menu doesn't make sense either.
> 
> I think it's fine you have a command that takes one stock and one ETF, you don't have to load in the ETF, you just have to select it in the command. It could even be an index.
> 
> Don't forget that this is in essence just calculating a component of CAPM.

I'd be way more open to create a new menu where we add this. @northern-64bit actually added a `capm` command to `qa`:

Screenshot 2022-07-23 at 22 10 06

northern-64bit:
> I'd be way more open to create a new menu where we add this. @northern-64bit actually added a `capm` command to `qa`:
> 
> Screenshot 2022-07-23 at 22 10 06

Sorry for jumping into this discussion, but I think that this should be in the qa menu since this is more of a quantitative feature and we already have VaR, CVaR and CAPM there. The best thing is if all of these kind of features is in one menu. Otherwise a new menu for this also sounds solid.

ricwillis98:
I think it could fit in qa as well @northern-64bit. What if I just put it there as a command (not a menu) and use an optional arg for the reference ticker e.g. you could call just "beta" to get beta to SPY or "beta --ref XLK" or "beta --ref SBUX" to get beta of the loaded stock w.r.t another stock or ETF. Does that keep everyone happy? @DidierRLopes @JerBouma @jmaslek

jmaslek:
> I think it could fit in qa as well @northern-64bit. What if I just put it there as a command (not a menu) and use an optional arg for the reference ticker e.g. you could call just "beta" to get beta to SPY or "beta --ref XLK" or "beta --ref SBUX" to get beta of the loaded stock w.r.t another stock or ETF. Does that keep everyone happy? @DidierRLopes @JerBouma @jmaslek 

At least makes me happy :)

DidierRLopes:
> I think it could fit in qa as well @northern-64bit. What if I just put it there as a command (not a menu) and use an optional arg for the reference ticker e.g. you could call just "beta" to get beta to SPY or "beta --ref XLK" or "beta --ref SBUX" to get beta of the loaded stock w.r.t another stock or ETF. Does that keep everyone happy? @DidierRLopes @JerBouma @jmaslek

Perfecto!! 😄

ricwillis98:
All finished - I moved to qa, added tests and comments, and updated docs. Could I get a code review please @DidierRLopes @jmaslek ?

JerBouma:
> I think it could fit in qa as well @northern-64bit. What if I just put it there as a command (not a menu) and use an optional arg for the reference ticker e.g. you could call just "beta" to get beta to SPY or "beta --ref XLK" or "beta --ref SBUX" to get beta of the loaded stock w.r.t another stock or ETF. Does that keep everyone happy? @DidierRLopes @JerBouma @jmaslek

Sounds great!

openbb_terminal/stocks/quantitative_analysis/beta_model.py
```diff
@@ -0,0 +1,53 @@
+from scipy import stats
+from openbb_terminal.stocks import stocks_helper
+
+
+def beta_model(stock_ticker, ref_ticker, stock=None, ref=None):
+    """Calculate beta for a ticker and a reference ticker.
+
+    Parameters
+    ----------
+    stock_ticker : str
```
DidierRLopes, line 10: can we add a sentence below like `stock` and `ref` have?

openbb_terminal/stocks/quantitative_analysis/beta_model.py
```diff
@@ -0,0 +1,53 @@
...
+    ref_ticker : str
```
DidierRLopes, line 11: can we add a sentence below like `stock` and `ref` have?

DidierRLopes (changes requested):
Some minor asks! Plus, in case the user provides a reference that doesn't exist, we should fail gracefully with a message saying that.

Also, minor thing can we have a whitespace after? A single `console.print()` at the end of `_view` is enough.

Screenshot 2022-07-25 at 14 26 39

DidierRLopes (approved):

DidierRLopes:
Nice work! @ricwillis98 🙏🏽 🙏🏽

PR #2154, opened 2022-07-24
Updates crypto load docs to match last load refactor

jose-donato:


DidierRLopes (approved):
@deeleeramone created a ticket to add more granularity on the data. FYI 😋

DidierRLopes (changes requested):
Why did you change the `closes[:-365]` to `closes[:365]`

jose-donato:
> Why did you change the `closes[:-365]` to `closes[:365]`

i think there was an error. that column was displaying NaN. why would it be -365?

DidierRLopes:
> > Why did you change the `closes[:-365]` to `closes[:365]`
> 
> i think there was an error. that column was displaying NaN. why would it be -365?

There's actually an issue, it should be `closes[-365:]` - this is so that we take the last 365 days from the data to calculate volatility

DidierRLopes (approved):

PR #2174, opened 2022-07-25
Survey command and info, more documentation and bugfixes

JerBouma:
This PR does the following:
- Add a command `survey` that opens the most recent survey.
- Add a link within the documentation to fill in the most recent survey.
- Add a link in the exit message about filling in the survey.
- For the bugfix for `about`, I noticed that within the `ta` menu typing `about` doesn't work since it adds in "terminal" twice. This was a very minor fix from changing an if-statement to elif.
- Also update documentation for `settings`, `featflags` and `sources` which now has a proper section in the Getting Started page.

The link, https://openbb.co/survey, still needs to be defined.

image

image

image

image


minhhoang1023:
Thanks for setting this up Jer! 🙌 

@piiq @andrewkenreich A quick question about the installer! My idea is that we can add the Survey link in the exit message for the Terminal. @JerBouma mentioned that it will only appear when people type `q` to exit. Is there any way we can still display the message when people exit in a different way?

piiq:
> Is there any way we can still display the message when people exit in a different way?

If the user closes the window the window just closes and there is no exit per se. We can display survey link in the start message, or in some other place

minhhoang1023:
It throws an error when I launched the command.

image

terminal.py
```diff
@@ -246,6 +249,10 @@ def call_guess(self, other_args: List[str]) -> None:
             )
             console.print(f"[red]{e}[/red]")
 
+    def call_survey(self) -> None:
```
minhhoang1023, line 252: since we are not using any class attribute, it can be a static class here.

minhhoang1023 (changes requested):

JerBouma:
Fixed the survey command not working.  Does it matter much it isn't a static method?

minhhoang1023:
@JerBouma it's software engineering best practice. So that an instance doesn't have to be instantiated each time you call it.

It would take less than 1 min to fix it, and we can keep the code base to a high-quality standard.

minhhoang1023:
Awesome! Working now 🚀

JerBouma:
> 

Could you suggest how?

minhhoang1023:
For the survey, we could add "Take 2 mins to fill in our survey" or "Fill in our 2-minute survey". Letting people know that this is extremely QUICK to do!

Looks great overall

JerBouma:
image

image

Better?

minhhoang1023:
> > @JerBouma it's software engineering best practice. So that an instance doesn't have to be instantiated each time you call >it.
> > It would take less than 1 min to fix it, and we can keep the code base to a high-quality standard.
> 
> Could you suggest how?

Just need to add the @staticmethod and remove the self. Like this

```
    @staticmethod
    def call_survey(_) -> None:
        """Process survey command"""
        webbrowser.open("https://openbb.co/survey")
```

JerBouma:
Done done

minhhoang1023 (approved):
LGG

piiq (approved):

Issue #2186, opened 2022-07-26
[Bug] Keys - Coinbase function is broken. Key can only be set through .env


deeleeramone:
The function to add the keys for Coinbase are broken. This is for both the installer package and GitHub repo.

![Screen Shot 2022-07-26 at 2 05 08 PM](https://user-images.githubusercontent.com/85772166/181111820-1dbfdce1-5362-4fb3-921a-6d333861988d.png)

deeleeramone:
The help dialogue is now being printed. 

```
(🦋) /keys/ $ coinbase -h
usage: coinbase [-k KEY] [-s SECRET_KEY] [-p PASSPHRASE] [-h]

Set Coinbase API key.

optional arguments:
  -k KEY, --key KEY     Key (default: None)
  -s SECRET_KEY, --secret SECRET_KEY
                        Secret key (default: None)
  -p PASSPHRASE, --passphrase PASSPHRASE
                        Passphrase (default: None)
  -h, --help            show this help message (default: False)
  ```

Removing my keys from the `.env` file is reflected in the Coinbase status. 

<img width="522" alt="Screen Shot 2022-07-29 at 10 26 07 AM" src="https://user-images.githubusercontent.com/85772166/181812213-d5187bf9-bec8-4b57-ba37-871399a81421.png">

```
defined, test passed
```
<img width="564" alt="Screen Shot 2022-07-29 at 10 27 43 AM" src="https://user-images.githubusercontent.com/85772166/181812467-a5829dc5-77d1-4eaf-ac8a-fce7617e9082.png">


Attempting to use a Coinbase function results in a new bug:

```
(🦋) /crypto/ov/ $ cbpairs
Error: "['base_min_size', 'base_max_size', 'max_market_funds'] not in index"

Error: list indices must be integers or slices, not str
```

colin99d:
Is there a way to get coinbase keys for free?

PR #2205, opened 2022-07-29
Fix coinbase function in crypto/ov

jmaslek:
Beat you @colin99d 

This was pointed out in #2186 comments.  Looks like it was just a change in the cb api response. oops


colin99d (approved):

colin99d:
Bloomberg better watch out.

PR #2208, opened 2022-07-29
Update latest allowed python version to 3.10.x

lotia:
# Description

This PR raises the latest allowed python version 3.10.x which is the latest minor version.

# How has this been tested?

I have run the test suite using `pytest tests` and although tests were skipped, there were no failures. This was done on macOS 12.4 on which python is managed using asdf and the python plugin for asdf.

To reproduce

1. Install a current 3.10.x version of python, in my case I use the [asdf version manager](https://asdf-vm.com/) and the associated python [plugin](https://asdf-vm.com/manage/plugins.html).
1. Confirm you are using that python version in your current shell using `python --version`
1. [Install poetry](https://python-poetry.org/docs/#installation)
1. Install all dependencies including optional ones

   ```shell
   poetry install --no-root -E prediction -E bots -E jupyterlab
   ``` 

1. Activate the poetry shell

   ```shell
   poetry shell
   ```

1. Run test suite

   ```shell
   pytest tests
   ```

This should enable you to confirm (hopefully) that all dependencies install cleanly using just poetry and also that the test suite does not have any failures, although some tests are skipped.

# Others


lotia:
Can confirm that above steps were also successful on Arch Linux under python 3.10.5

deeleeramone:
I attempted this by installing Python 3.10.5 from my base Conda environment. I made the change to the file shown, then after confirming that Python 3.10.5 was the active version I ran `poetry install --no-root -E prediction -E bots -E jupyterlab`

`poetry shell`

Then launching the terminal, and attempting to enter the prediction menu, with a M1 Mac Pro, I received the `error: Illegal instruction 4`

Python version is 3.9.9, despite my starting point at 3.10.5

deeleeramone:
Screen Shot 2022-07-31 at 12 55 47 AM
Screen Shot 2022-07-31 at 12 56 32 AM
Screen Shot 2022-07-31 at 12 56 45 AM

Screen Shot 2022-07-31 at 12 57 06 AM

lotia:
Hello @deeleeramone thanks for looking into this. Unlike conda, poetry is installed per env. It is entirely possible that poetry was attempting to run against `3.10.5` despite not being installed for that executable. I'm also very happy to go over this on a pairing session if you'd like so we can get to the bottom of this issue. I can do discord or we can attempt to pair using [Pop](https://pop.com/) which I've found to be an excellent tool for pairing.

deeleeramone:
Hi @lotia, I deleted the Poetry environment, but I'll have another go at installing this way. 

I can assure you that between @piiq, @martinb-bb, and myself, we have spent a lot of time messing around with the dependency chain, figuring out what package distributions do not work. @piiq is probably the best person you could connect with to attempt to iron out a viable solution.

On the plus side, it is entirely possible to run the Terminal from Python 3.10.5 & 3.10.7, on a Mac M1 as well as a legacy Mac Pro system running OS Catalina. This was accomplished by installing with `conda install -c conda-forge python==3.10.x`

I have not tested for errors, but on the surface things appear to work.

lotia:
I'll close the PR for now and continue with the latest stable 3.9.x. Thanks for the time you have put in and apologies if it was annoying.

colin99d:
It worked for me. I wouldn't about `-E prediction` because we are changing how that works anyway. Pytest did generate 3 errors locally, but I do not think this is due to the switch to python 3.10. @martinb-bb, would be interesting to see how this interacts with our menu.

martinb-bb:
It shouldnt affect too much. Will have to do testing for sure !

lotia:
Please freel free to open or ask me to rebase if/when we can sort out. I'm excited by the push the PSF has towards performance improvments in 3.10 and the upcoming ones in 3.11 specially since you folks have so much cool prediction/simulation functionality available

colin99d (approved):

PR #2243, opened 2022-08-03
Fixes #2220 by replacing Nan with 0 to allow rounding

### LINKED ISSUE #2220: Bug] Crypto/OV/HM - limit of 50 returns Error: cannot convert float NaN to integer
When the limit touches 50, this error message displays.

```
 /crypto/ov/ $ hm stablecoins -l 50
Error: cannot convert float NaN to integer
```

When a `category` is selected, the % of market cap should then be calculated as relative to the % of the chosen category, and not the entire market cap of all cryptos. This is why everything is crowded into one tiny corner while the biggest pieces are fractions of one percent. 

Screen Shot 2022-07-31 at 8 04 59 PM

Screen Shot 2022-07-31 at 8 06 41 PM

soggyomelette:
# Description

Fixes #2220  was due to not being able to round missing values in daily percentage change taken from coingecko. Replacing Nan with 0 fixes the issue. Also adds a note to heatmap help clarifying that the percentage displayed represents the daily percentage change in the currency


colin99d (approved):

colin99d:
Nice work man!!

PR #2266, opened 2022-08-07
Remove portfolio optimisation shortcut

### LINKED ISSUE #2257: [Bug] When entering the comparison analysis menu with a loaded ticker containing a suffix, the suffix is not passed through to the `ticker` in the menu.
When entering the comparison analysis menu with a loaded ticker containing a suffix, the suffix is not passed through to the `ticker` in the menu.  The user must first `remove` and then `add` the primary ticker within the submenu prior to using the functions. 

Screen Shot 2022-08-05 at 10 49 09 AM

DidierRLopes:
* Fixes #2257
* Remove `po` shortcut from `ca` controller

### DIFF
--- a/i18n/en.yml
+++ b/i18n/en.yml
@@ -149,7 +149,6 @@ en:
   stocks/options/screen/scr: screen data from this preset
   stocks/options/screen/_screened_tickers: Last screened tickers
   stocks/options/screen/ca: take these to comparison analysis menu
-  stocks/options/screen/po: take these to portoflio optimization menu
   stocks/options/pricing/_ticker: Ticker
   stocks/options/pricing/_expiry: Expiry
   stocks/options/pricing/add: add an expected price to the list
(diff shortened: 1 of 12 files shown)


JerBouma:
Some others places `po` is mentioned where it shouldn't be:

- options/screener/screener_controller.py
- stocks/screener/screener_controller.py

Background why we remove this integration:

Portfolio optimization is not a straightforward thing. All of the parameters you can define within each `po` command are massive considerations that have an incredibly large impact on the portfolio. E.g. we currently have defaults set but they are not based on any financial theory that it's the right approach. Simply calling `maxsharpe` has no meaning whatsoever unless you thought of it extensively.

We do not want to give our users the impression we believe that integrations with screeners and/or comparison analysis lead to effective and/or sound portfolios. This requires 10 times much more thought than this integration indicates. Furthermore, you don't tend to optimise a non-existing portfolio in the first place.

DidierRLopes:
@JerBouma done!

JerBouma (approved):
Looks good to me.

PR #2278, opened 2022-08-09
Scale text size with size of box in crypto/ov/hm

soggyomelette:
# Description
Attempts to scale text size of box in crypto/ov/hm to prevent crowding as seen below and in #2220 . Although we use an external package for the plotting, the creator recommends writing code yourself for text related issues (https://github.com/laserson/squarify/issues/19) 
Before:
image
After:
image


Chavithra (approved):
Nice !

PR #2325, opened 2022-08-16
com2ann: Convert type comments into Python type hints

cclauss:
# Description
https://pypi.org/project/com2ann

# How has this been tested?

# Checklist:

# Others


DidierRLopes (approved):

piiq:
What is the rationale to go away from using the typing module? The load time penalty is not that high but having 1 way to type hint the variables in the code improves readability through homogeneity by a lot. What am I missing @cclauss?

cclauss:
@piiq Type hints are moving towards **_PEP 604_** and `pyupgrade` automatically makes these changes.
* https://github.com/asottile/pyupgrade#pep-604-typing-rewrites

piiq:
Thanks for pointing to the PEP. Is my understanding correct that:
1. `Union[str,int]` changes to `str | int`
2. `Optional[str]` changes to `str | None` (and the `~str` proposal was not approved)
3. The `from __future__ import annotations` is going to stay there until the lowest supported version reaches 3.10?

cclauss:
Yes for 1. and 2. but I think the jury is still out on 3. ... See the footnote https://docs.python.org/3/library/__future__.html

I have found it useful to run `pyupgrade` on various test files to understand its transforms.

piiq:
Thanks. In that case I have another question. From your perspective is it possible to move away from `typing` module at all? Is there an `Any` annotation or an annotation for `Callable`?
I'm asking because I am _ironically_ trying to follow the `There should be one– and preferably only one –obvious way to do it.` guiding principle

piiq (approved):

cclauss:
PEP20:
> There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.

The Dutch part is significant here.  The solution does not need to be too simplistic.

The community is not saying that we should completely abandon the typing module.

Instead, it advocates using pre-existing built-in types and the `collections` module types where possible.  Having both `List` and `list` has confused new Python developers and it leads to unnecessary imports (both time complexity and code complexity).  Similarly, having both `typing.Callable` and `collections.abc.Callable` suffers from the same issues.  The proposed approach may take a bit of time to get used to but there is good logic behind this move.

piiq:
Thanks for sharing your thoughts.
Would you mind updating the [CONTRIBUTING.md](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md) to mention the change in how types should be used? And we can merge it

PR #2337, opened 2022-08-17
fix: typo in tradinghours_controller.py

pierreraii:
Fix minor typo in the message shown when ticker is not found in equities list


DidierRLopes:
Thanks @pierreraii !!

PR #2397, opened 2022-08-23
Add nasdaq as an options source

jmaslek:
This PR adds nasdaq as an option source for `chains`, `oi`,`vol`,`voi`.

This actually allows pulling the entire chain in one command, so calculations for delta neutral as example could be performed.

Nasdaq allows for some things not carried on yahoo, like NDX (youre welcome danglewood).

Also going through shows a bunch of refactorings that need to go through, but that is out of the scope here.


colin99d:
Screen Shot 2022-08-22 at 9 51 50 PM

Sad

jmaslek:
i uh didnt touch that one

colin99d:
Ok, while I will still approve this PR. I think it brings up an important refactor that needs to happen in the future. Currently, if I load with a non-default source, I then have to specify a source on every command I run. While this my not be hard on some, I have a full time job, a wife, and two kids at home. So I certainly don't have time to type `--source=nasdaq` after every command. I think a band aid fix for this in the near future would be having the controller remember your source, and a more long-term source would be to have all models send data in the same format to allow for standardization of views. Thanks for coming to my Ted Talk.

colin99d (approved):

DidierRLopes:
> Ok, while I will still approve this PR. I think it brings up an important refactor that needs to happen in the future. Currently, if I load with a non-default source, I then have to specify a source on every command I run. While this my not be hard on some, I have a full time job, a wife, and two kids at home. So I certainly don't have time to type `--source=nasdaq` after every command. I think a band aid fix for this in the near future would be having the controller remember your source, and a more long-term source would be to have all models send data in the same format to allow for standardization of views, thanks for coming to my Ted Talk.

I literally implemented this some weeks ago 😄  @colin99d 

Screenshot 2022-08-23 at 08 56 39

Issue #2412, opened 2022-08-24
[Bug] Economy/FRED - Flag for starting date is not working


deeleeramone:
The flag for setting the dates of the series being requested does not work.

![Screen Shot 2022-08-24 at 4 43 00 PM](https://user-images.githubusercontent.com/85772166/186542975-3aa4a746-5f9d-4543-a9c2-db4cc7caa16d.png)

![Screen Shot 2022-08-24 at 4 43 44 PM](https://user-images.githubusercontent.com/85772166/186543055-fa0bf1f6-e2ac-4fd1-b186-09273b6cd9fd.png)

This series goes back to 1986, but despite there being no default for the start/end dates it is stopping at two years.

![Screen Shot 2022-08-24 at 4 44 47 PM](https://user-images.githubusercontent.com/85772166/186543195-9409137f-049b-48b5-8a1a-51e6099a191d.png)

PR #2421, opened 2022-08-25
Fix economy/fred start date

jmaslek:
#2412 

Fix starting date issue.  Took out the if limit==-1 because limit has to be > 0


deeleeramone:
This is what I'm getting:
```
2022 Aug 25, 21:05 Danglewood /economy/ $ fred -p MORTGAGE30US
Error: get_aggregated_series_data() got an unexpected keyword argument 'limit'

Error: not enough values to unpack (expected 2, got 0)
```
Screen Shot 2022-08-25 at 9 06 31 PM

Screen Shot 2022-08-25 at 9 14 46 PM

Pretty sure I am checked out of the right branch because when I checkout back to origin/main, I get the old output with the start date being ignored.

Screen Shot 2022-08-25 at 9 18 35 PM

DidierRLopes (approved):

PR #2428, opened 2022-08-26
Added files outside OpenBB

### LINKED ISSUE #2453: [Bug] Stored API Keys Lost after Upgrade
FFS cant you check if users have pre-defined stored API keys before force overwriting the keys during upgrade installation?

colin99d:
# Description

Creates an OpenBB folder in the home directory and adds required files insider. This PR will also look to change `openbb_terminal/core/config/constants.py` to the single source of truth for referencing files. Fixes #2453 Still need to do:

- [x] Actually use the new custom_imports folder
- [x] Actually export to the new exports folder
- [ ] Actually read from new other folders

# How has this been tested?

# Checklist:

# Others


JerBouma:
A few checks:

- What happens when I do a fresh install (both Python and Installer), does it overwrite these files? E.g. will I still lose all my API keys?
- Can there be a confirmation window of some kind that says "The folder already exists, do you wish to overwrite?"
- Does this work for `importing` and `exporting` files?
- Does this also work for functionalities that look at certain folders (e.g. Econometrics `load`) and is this correctly redirected?

colin99d:
> * What happens when I do a fresh install (both Python and Installer), does it overwrite these files? E.g. will I still lose all my API keys?
It checks if folders already exist before creating them, so old files will never be overwritten
> * Can there be a confirmation window of some kind that says "The folder already exists, do you wish to overwrite?"
Same answer as aboce
> * Does this work for `importing` and `exporting` files?
Importing works, still need to testing exporting. I dont beleive style, routing, and ini files work yet
> * Does this also work for functionalities that look at certain folders (e.g. Econometrics `load`) and is this correctly redirected?
Econometrics load works with this, forecast will be modified once this is merged
> This also addresses the unfortunate loss of API keys mentioned here #2453

colin99d:
It is important to note that since some of the functions this uses are conditional on whether the given terminal is an installer, this will not be fully tested until we generate an installer with this code.

colin99d:
```
USER_DATA_DIR = Path(str(Path.home()) + "/OpenBBUserData")
if not os.path.exists(USER_DATA_DIR):
    os.mkdir(USER_DATA_DIR)

USER_HOME = Path(os.path.expanduser("~"))
ENV_FILE_DIR = Path(str(Path.home()) + "/.openbb_terminal")
if not os.path.exists(ENV_FILE_DIR):
    os.mkdir(ENV_FILE_DIR)

ENV_FILE = ENV_FILE_DIR.joinpath(".env")
if not ENV_FILE.is_file():
    with open(str(ENV_FILE), "w"):
        pass
```
Also, is there a reason we have a separate folder for just the env variable? Especially since this file is hidden by file browsers by default.

soggyomelette:
> ```
> USER_DATA_DIR = Path(str(Path.home()) + "/OpenBBUserData")
> if not os.path.exists(USER_DATA_DIR):
>     os.mkdir(USER_DATA_DIR)
> 
> USER_HOME = Path(os.path.expanduser("~"))
> ENV_FILE_DIR = Path(str(Path.home()) + "/.openbb_terminal")
> if not os.path.exists(ENV_FILE_DIR):
>     os.mkdir(ENV_FILE_DIR)
> 
> ENV_FILE = ENV_FILE_DIR.joinpath(".env")
> if not ENV_FILE.is_file():
>     with open(str(ENV_FILE), "w"):
>         pass
> ```
> 
> Also, is there a reason we have a separate folder for just the env variable? Especially since this file is hidden by file browsers by default.

This was something @piiq and @Chavithra suggested since we don't really want the user to have direct access to the .env file and instead set it using the keys or settings menu

Also, another thing that was suggested is instead of having an openbb timezone file, we could add it as a variable in the .env file

piiq:
> Also, is there a reason we have a separate folder for just the env variable?

@colin99d 
The user is able to define any folder to be a OpenBBUserData folder. This setting is saved as a record in the .env file.
When a new updated version is installed/cloned the new code doesn't know where to look for the location of the user data folder. Splitting user data and the settings file is to solve this issue. User level settings are saved in .* files by in user home by many programs. My user home includes a lot of stuff including
image

So as you can see by .gamestonk_terminal.hist` we're already utilizing that place for quite some time

PR #2435, opened 2022-08-27
allow to do 'macro TB' instead of 'macro -p TB'

DidierRLopes:
Minor enhancement.


jmaslek (approved):
sorry this one got past me

PR #2436, opened 2022-08-27
forex quote to use yahoo finance instead

### LINKED ISSUE #2403: [Bug] Forex Quote API blocked via premium account
**Describe the bug**
AlphaVantage seems to have converted the API used for quotes into a premium feature. Since this is a pretty simple ask, it may be better to use a different API that's free so users can access quote information.

**To Reproduce**
`forex/load GBPUSD/quote`

**Screenshots**
Screen Shot 2022-08-23 at 1 14 07 PM

- OS: Mac Big Sur 11.6
- Python version 3.8.13

**Additional context**

It might be possible to convert quote to a Yahoo Finance ticker using their currency quote tickers in their API. Remapping each forex pair to their format of `GBPUSD=X` could be an option. [See here for a reference page.](https://finance.yahoo.com/quote/GBPUSD=X?p=GBPUSD=X&.tsrc=fin-srch)

DidierRLopes:
* Fixes #2403


jmaslek (changes requested):
Should leave the alphavantage function and just add this as a new choice

DidierRLopes:
@jmaslek done ✅

jmaslek (approved):

PR #2439, opened 2022-08-27
Improve export folder to custom one

### LINKED ISSUE #2345: [Bug] Paths set from the Settings menu creates a weird mess.
Default path is not the default:
image

image
image

Setting path without quotes, "/path":
image

Setting path, to the actual exports folder that already exists because I have exported files there already:
image

image

DidierRLopes:
* Fix #2345 and allows to export custom folder


deeleeramone:
@DidierRLopes, the assignment of folders appears to work now. It would be a good idea to inform the user in the help dialogue that the path must be enclosed in quotation marks. There were problems with paths on PC/Linux, because the slash goes the other direction for those systems, so someone with a Windows and Linux OS should also test this for functionality. 

One issue I have discovered is that: `export default` sets the exports folder as `~/OpenBBTerminal/openbb_terminal/exports`, which is one folder too deep. The existing folder is at: `~/OpenBBTerminal/exports`

Screen Shot 2022-08-27 at 1 24 57 PM

DidierRLopes:
@deeleeramone done ✅

jmaslek (approved):

PR #2483, opened 2022-09-02
Improve ux experience through auto completer

DidierRLopes:
Give it a go


deeleeramone:
`/economy/index` after the first index is selected, autocomplete wants to add a second index to the chart when the `-` is entered after the first choice. 

Screen Shot 2022-09-02 at 9 31 42 AM

Screen Shot 2022-09-02 at 9 33 25 AM

Using the `-i` flag under `/economy/index` brings up choices that should be activated after `space` is entered in the command string as opposed to immediately being triggered by the single dash. 

Screen Shot 2022-09-02 at 9 34 08 AM

Screen Shot 2022-09-02 at 9 36 17 AM

DidierRLopes:
> `/economy/index` after the first index is selected, autocomplete wants to add a second index to the chart when the `-` is entered after the first choice.
> 
> Screen Shot 2022-09-02 at 9 31 42 AM
> 
> Screen Shot 2022-09-02 at 9 33 25 AM
> 
> Using the `-i` flag under `/economy/index` brings up choices that should be activated after `space` is entered in the command string as opposed to immediately being triggered by the single dash.
> 
> Screen Shot 2022-09-02 at 9 34 08 AM
> 
> Screen Shot 2022-09-02 at 9 36 17 AM
> 
> Also noteworthy is that double-dashes do not present any choices.

I only tested the `macro economy`. Will look into what's going on here. Thanks!

DidierRLopes:
> `/economy/index` after the first index is selected, autocomplete wants to add a second index to the chart when the `-` is entered after the first choice.
> 
> Screen Shot 2022-09-02 at 9 31 42 AM
> 
> Screen Shot 2022-09-02 at 9 33 25 AM
> 
> Using the `-i` flag under `/economy/index` brings up choices that should be activated after `space` is entered in the command string as opposed to immediately being triggered by the single dash.
> 
> Screen Shot 2022-09-02 at 9 34 08 AM
> 
> Screen Shot 2022-09-02 at 9 36 17 AM
> 
> Also noteworthy is that double-dashes do not present any choices.

Notes:

The `index` only had autocomplete for "index WORD" hence why there are no "--" and "-" args autocompleters - the behaviour is the same.

This auto-completer is designed to need a "--" flag to proceed to autocomplete, hence why results are sucky with "-". Having both would increase time by 2x, so we will only support an autocomplete for one.

deeleeramone:
```
This auto-completer is designed to need a "--" flag to proceed to autocomplete, hence why results are sucky with "-". Having both would increase time by 2x, so we will only support an autocomplete for one.
```
By this, do you mean that autocomplete will support only "--", and not "-"?

DidierRLopes:
> ```
> This auto-completer is designed to need a "--" flag to proceed to autocomplete, hence why results are sucky with "-". Having both would increase time by 2x, so we will only support an autocomplete for one.
> ```
> 
> 
>      
>   
> 
> By this, do you mean that autocomplete will support only "--", and not "-"?

Likely yes. The reason being for performance (speed), not because we can't make it happen :)

DidierRLopes:
@deeleeramone try it out now. I fixed the entire menu to have autocomplete for "--".

@jmaslek @piiq  check the implementation on the economy menu if u have some time

deeleeramone:
It seems to work, but with a caveat being after the first command is run, the functions in the menu no longer trigger autocomplete. so, if you run `overview` once, the menu has the functions autocomplete, but the second time it does not and is only triggered when `--` is deployed.

Screen Recording 2022-09-02 at 9 44 34 PM

This is more of a blocker than above, @DidierRLopes: `map --type` brings up the choices only for `overview --type`. 

Screen Shot 2022-09-02 at 10 02 45 PM

DidierRLopes:
> It seems to work, but with a caveat being after the first command is run, the functions in the menu no longer trigger autocomplete. so, if you run `overview` once, the menu has the functions autocomplete, but the second time it does not and is only triggered when `--` is deployed.
> 
> Screen Recording 2022-09-02 at 9 44 34 PM [ Screen Recording 2022-09-02 at 9 44 34 PM ](https://user-images.githubusercontent.com/85772166/188256548-fb3de299-b120-4df8-baf4-5bcf7d1e7b69.gif) [ ](https://user-images.githubusercontent.com/85772166/188256548-fb3de299-b120-4df8-baf4-5bcf7d1e7b69.gif)
> 
> This is more of a blocker than above, @DidierRLopes: `map --type` brings up the choices only for `overview --type`.
> 
> Screen Shot 2022-09-02 at 10 02 45 PM

Very nice catch. This is not easy to trigger, it happens when you delete the command that you had first input, I am going to fix this - thanks for spotting it.

DidierRLopes:
@deeleeramone both those cases where caused by the fact of you going back and changing first argument, I didn't account for that behaviour. 

It should now be taken care of. Let me know if it works well now!

jmaslek:
This is definitely a sweet improvement.

Is a little buggy. This behavior changed at some point.  Seems to be when i was in the middle of a completer than I do a backspace. (Confirms this happens when backspacing)

Screen Shot 2022-09-03 at 2 43 20 PM

Also if I start to type, the autocompleter doesnt know where I am in the list.

Screen Shot 2022-09-03 at 2 45 46 PM

deeleeramone:
`industry` shows as a choice for a function and not an argument. 

Screen Shot 2022-09-03 at 1 26 48 PM

`bigmac --country` doesn't populate with countries, and instead presents column names as choices. 

Screen Shot 2022-09-03 at 1 41 33 PM

`ycrv --country` is not able to handle the spaces where a country name is two words. 
Screen Shot 2022-09-03 at 1 43 55 PM

DidierRLopes:
> This is definitely a sweet improvement.
> 
> Is a little buggy. This behavior changed at some point. Seems to be when i was in the middle of a completer than I do a backspace. (Confirms this happens when backspacing)
> 
> Screen Shot 2022-09-03 at 2 43 20 PM
> 
> Also if I start to type, the autocompleter doesnt know where I am in the list.
> 
> Screen Shot 2022-09-03 at 2 45 46 PM

This is with my last change, messed up that end. Will look into this.

DidierRLopes:
@jmaslek and @deeleeramone . Both should be fixed now!

Issue #2478, opened 2022-08-31
[Bug] Stocks/CA/cashflow: Two flags exist for selecting quarterly data. `-t` flag should be for the year.


deeleeramone:
The `-t` flag here looks wrong. Default is last (default: None)? Flag for quarterly data already exists, so this second argument  is actually the year for the comparison. 

```
optional arguments:
  -q, --quarter         Quarter financial data flag. (default: False)
  -t S_TIMEFRAME, --timeframe S_TIMEFRAME
                        Specify yearly/quarterly timeframe. Default is last. (default: None)
```
<img width="1355" alt="Screen Shot 2022-08-31 at 2 08 51 PM" src="https://user-images.githubusercontent.com/85772166/187782902-a49a1f82-9e7e-47a6-9652-2f4dba13a519.png">

<img width="733" alt="Screen Shot 2022-08-31 at 2 04 05 PM" src="https://user-images.githubusercontent.com/85772166/187781741-7778694c-ec0f-4d99-9444-c2bd88be2272.png">

PR #2486, opened 2022-09-02
[Bug] Fixes the stocks ca cashflow command

northern-64bit:
# Description

Fixes issue https://github.com/OpenBB-finance/OpenBBTerminal/issues/2478

# How has this been tested?

Using the command with various tickers and inputs.

# Checklist:

# Others

### DIFF
--- a/openbb_terminal/stocks/comparison_analysis/ca_controller.py
+++ b/openbb_terminal/stocks/comparison_analysis/ca_controller.py
@@ -737,7 +737,8 @@ class ComparisonAnalysisController(BaseController):
             dest="s_timeframe",
             type=str,
             default=None,
-            help="Specify yearly/quarterly timeframe. Default is last.",
+            help="Specify year/quarter of the cashflow statement to be retrieved. The format for year is YYYY and for "
+            "quarter is DD-MMM-YYY (for example, 30-Sep-2021). Default is last year/quarter.",
         )
         ns_parser = self.parse_known_args_and_warn(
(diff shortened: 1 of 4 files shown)


openbb_terminal/stocks/comparison_analysis/ca_controller.py
```diff
@@ -737,7 +737,8 @@ def call_cashflow(self, other_args: List[str]):
             dest="s_timeframe",
             type=str,
             default=None,
-            help="Specify yearly/quarterly timeframe. Default is last.",
+            help="Specify year/quarter of the cashflow statement to be retrieved. The format for year is YYYY and for "
```
jmaslek, line 740: Technically its not quarter, since specifying year quarter would be something like 2021/Q3, but this requires the quarter end date.  Not sure if there is an easy workaround other than having a list of all possible dates ?

northern-64bit: The problem is that different firms publish their quarterly report on different dates, so you cannot really have autocomplete with this.
One option is to change the quarter flag to having int representing the quarter to retrieve, then we only need to enter the year in the timeframe flag. However this requires extensive refactoring and other commands use the same functions too, thus they would also have to be changed.

jmaslek: Yeah this will require some thinking and we should probably also move away from using this scraping source moving forward.  But I am good with how it is now.

jmaslek (approved):

Issue #2479, opened 2022-08-31
[Bug] test_econometrics_panel.openbb not working anymore


tinix84:
**Describe the bug**
test_econometrics_panel.openbb not working anymore

**To Reproduce**
exe ./scripts/test_econometrics_panel.openbb

**Screenshots**
![image](https://user-images.githubusercontent.com/16301528/187784131-45f383cf-6076-4231-aa09-1387144d95b7.png)

**Desktop (please complete the following information):**

- OS: wsl
- Python version 3.9

**Additional context**
Add any other information that you think could be useful for us.

PR #2489, opened 2022-09-03
add -f to load if not present

jmaslek:
Fix for #2479 

The load script had no "-f", so I added one if its not present


colin99d:
Great PR! Keep it up James.

colin99d (approved):

PR #2490, opened 2022-09-03
Big refactor to be explicit about data sources

DidierRLopes:
The goal of this PR is to move the sources from just being pasted on the controller to relying on `data_sources_default.json`.


DidierRLopes:
Screenshot 2022-09-04 at 02 17 21

@Chavithra do you know why it passes on my machine and not in remote?

DidierRLopes:
I am going to merge this because this is a substantial PR that will change the way we add data sources. This should only affect `controller` files and not `model` or `views`.

Will work on a PR about CONTRIBUTING guidelines, since the current one is rather outdated

PR #2492, opened 2022-09-04
added wiki command

hjoaquim:
# Description

Added wiki command -> type wiki and find a simple title/url/summary for any expression (from the wikipedia)

image

another one, using the "-e":

image

It's hard for average users/ users with no financial education to keep up with all the jargons we use on the terminal

# How has this been tested?

type "wiki -e EXPRESSIONS" or "wiki EXPRESSIONS" from anywhere

# Checklist:

- [X] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [X] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).

# Others
- [X] New and existing unit tests pass locally with my changes. You can test this locally using `pytest tests/...`.


hjoaquim:
@DidierRLopes thought it would be really cool to have some kind of glossary on the terminal - leveraging the Wikipedia API would allow users to search for virtually anything if they have doubts regarding the meaning of something while using the terminal without leaving (and even pursue more info through the URL if wanted)
What are your thoughts on this?

colin99d (approved):
Great work! I am proud of you.

terminal.py
```diff
@@ -147,6 +147,8 @@ def print_help(self):
         mt.add_menu("portfolio")
         mt.add_menu("dashboards")
         mt.add_menu("reports")
+        mt.add_raw("\n")
+        mt.add_cmd("wiki")
```
DidierRLopes, line 151: Let's add this just below update, ok?

openbb_terminal/helper_funcs.py
```diff
@@ -1703,3 +1703,40 @@ def get_ordered_list_sources(command_path: str):
...
+    Parameters
+    ----------
+    expression: str
+        Expression to search for
+    """
+
+    url = f"https://en.wikipedia.org/api/rest_v1/page/summary/{expression}"
+
+    response = requests.request("GET", url, headers={}, data={})
+    status = response.status_code
```
DidierRLopes, line 1720: These are minor but:
* No need to declare status variable since you are only checking it once on that if condition
* No need to declare response as response.text, just `json.loads(response.text)`

DidierRLopes (commented):
Couple of comments, but very nice addition!!

terminal.py
hjoaquim: Thank you for the feedback!! Just committed the adjustments

openbb_terminal/helper_funcs.py
hjoaquim: ^^^
same here!

Issue #2470, opened 2022-08-31
[Bug] Crypto/Disc/CMCTOP - Error: display_cmc_top_coins() got an unexpected keyword argument 'descend'


deeleeramone:
<img width="1136" alt="Screen Shot 2022-08-30 at 11 55 11 PM" src="https://user-images.githubusercontent.com/85772166/187613202-036d08cd-becb-4750-91db-591816ac8256.png">

jmaslek:
This was merged.  Yay

PR #2495, opened 2022-09-04
[bug] fixes cmc top coins issue #2470

et41:
# Description
Fixed argument mismatch in display_cmc_top_coins() and changed --descend default to false.


jmaslek (approved):
nice catch!  thanks!

PR #2500, opened 2022-09-05
Add End of Day Historical Data to FA : https://eodhistoricaldata.com/r/?ref=869U7F4J to 

pauljsymonds:
# Description

- Implementation of fa, Initial fundamentals added:

- Income, 
- Balance, 
- Cash 
- 
- with Ratios and Plots for eodhd  [premium source](https://eodhistoricaldata.com/r/?ref=869U7F4J)

* This must be tested with the key -   demo 
* Demo key only works with AAPL

@DidierRLopes @colin99d please check this out and provide feedback.

# Checklist:

# Others


colin99d:
Looks great Paul! Only change I need is that adding the limit argument does not change the amount of output.

pauljsymonds:
> Looks great Paul! Only change I need is that adding the limit argument does not change the amount of output.

That limit argument is specifically for ratios only as that was the slow part, the rest should return as normal all data.

Let me know how you would prefer it to work.  For certain though Ratios cannot work for much more than that.

colin99d (approved):

PR #2508, opened 2022-09-06
New path for .env

soggyomelette:
Adding a new path to .env in /.openbb_terminal


soggyomelette (commented):
@colin99d @Chavithra I have added some extra code in config/constants.py which adds the name of the user data folder to the .env file following what Theo said

Chavithra (approved):
Looks ok to me.

PR #2509, opened 2022-09-06
New path for styles and add timezone as environment variable

soggyomelette:
Add new path for styles in openbb user data folder and add timezone as an environment variable instead of having an timezone.openbb file


Chavithra (approved):
Looks good to me.

PR #2523, opened 2022-09-09
Fix CMD log to be a well-formed JSON string

hjoaquim:
With this small change, we ensure that the logs we're generating are a well-formed (and **escaped**) JSON string. Thus, avoiding a lot of problems on the analytics side, when we try to read the "CMD" as a JSON - which ultimately was leading to loss of data.


Chavithra (approved):
Looks good to me.

PR #2552, opened 2022-09-15
fix: Reports not working on installer 

soggyomelette:
# Description

Fixes the reports not working on installer bug by replacing papermill with nbconvert to convert ipynb to py files on the fly and executes .py instead of .ipynb


piiq (changes requested):
We should definitely not discard notebooks but rather figure out a way to fix/overcome the underlying issue

Chavithra:
> We should definitely not discard notebooks but rather figure out a way to fix/overcome the underlying issue

We can use that as a patch and fixing that on the next increment.

soggyomelette:
I've been making good progress with parametrising without papermill and using just nbconvert, should be able to tell more in a few hours

soggyomelette:
> We should definitely not discard notebooks but rather figure out a way to fix/overcome the underlying issue

@piiq the new changes should allow us to retain the notebooks without depending on papermill and also allowing parametrisation

openbb_terminal/reports/reports_controller.py
```diff
@@ -209,27 +213,66 @@ def switch(self, an_input: str):
...
+                    )
+
+                notebook_file_copy = notebook_file + "copy"
 
-            if not result["metadata"]["papermill"]["exception"]:
+                with open(notebook_file_copy + ".ipynb", "w") as file:
+                    file.write(notebook_content)
+
+            # use nbconvert to switch to .py
+            os.system(f"jupyter nbconvert --to script --no-prompt {notebook_file_copy}")
```
piiq, line 246: can you use the nbconvert python api insteas of an os.system call?
see https://nbconvert.readthedocs.io/en/latest/nbconvert_library.html

piiq (changes requested):

openbb_terminal/reports/reports_controller.py
soggyomelette: Thanks for this, I wasn't able to figure this out from their github

piiq: you can also look into nbconverts tooling to load and parse the notebook and get the cells

piiq:
handled here https://github.com/OpenBB-finance/OpenBBTerminal/pull/2570

Issue #2549, opened 2022-09-15
[FR] Arthur Hayes USD Liquidity Index - Public Data and Simple Subtraction of Columns


deeleeramone:
From Reddit (https://www.reddit.com/r/openBB/comments/x8euyf/new_request_arthur_hayes_usd_liquidity_index/):

He spun up this index for his Bloomberg terminal so I figured why not request it for my favorite superiorly customizable stocks platform!USD Liquidity Conditions Index = [The Fed’s Balance Sheet] — [NY Fed Total Amount of Accepted Reverse Repo Bids] — [US Treasury General Account Balance Held at NY Fed]

![image](https://user-images.githubusercontent.com/85772166/190319646-3fc7ca97-d764-457a-88c5-dac824683d4f.png)

This index is measured on a weekly resolution because the Federal Reserve Balance Sheet is published on Wednesdays.

The Wednesday Level of the Fed's Balance Sheet is: WALCL

The amount of accepted reverse repo bids is here: https://www.newyorkfed.org/markets/desk-operations/reverse-repo

The Wednesday Level of the US Treasury General Account Balance is: WDTGAL

jmaslek:
for my future reference, the second one is the fred series: RRPONTSYD

jmaslek:
Okay so with #2559 merged, this is kind of possible.

If we do 
```
fred RRPONTSYD WALCL CBBTCUSD WTREGEN -s 2020-01-01
```

```
eval liquid = WALCL*1000000000**-1-WTREGEN-RRPONTSYD
```

This is my attempt at it.  Not sure why the division is needed tp replicate this, but if you know your formula, you can do it now!

deeleeramone:
If I follow your snippets here (erasing CBBTCUSD, because what is that doing there?), this is my output:
![Screen Shot 2022-10-23 at 8 43 36 PM](https://user-images.githubusercontent.com/85772166/197444523-e3464152-31af-46c8-bb0a-5a300b4369ed.png)

![Screen Shot 2022-10-23 at 8 44 04 PM](https://user-images.githubusercontent.com/85772166/197444608-dd643c47-01b6-473e-a7df-a17c5d1940a7.png)


![Screen Shot 2022-10-23 at 8 41 17 PM](https://user-images.githubusercontent.com/85772166/197444375-26fa12bd-e0e7-4a78-8026-b632c3737de7.png)

jmaslek:
[...]

PR #2559, opened 2022-09-18
Enhance economy menu with custom math functions

jmaslek:
Based off of #2549.

One thing we don't support is the ability to do some simple operations on certain columns.  This is a basic idea to do so.

The workflow would be something along the lines of (this loads in the 5y and 2y treasury yields and then creates a new column that is the difference [yes i know this series exists, but for illustrations this works] ).  Note this uses pandas.DataFrame.eval() so anything available there is fair game (i.e log functions, not just addition/subtraction)

```
economy
fred dgs2 dgs5
eval two_minus_five = dgs2-dgs5
plot two_minus_five
```

Voila.  

A couple notes: would need to add some potential support for multiple series.  Currently just overwriting DATASETS["custom"] -- should be EZ. 

The slash issue will likely confuse users (but can be avoided by doing `series**-1`).

Obviously will have some issues if users load in datasets with different time indexes.  

Curious on your thoughts @deeleeramone @DidierRLopes @JerBouma


DidierRLopes:
This is a great idea! 👍🏽

JerBouma:
Think this is great, could even calculate the duration with this. Could this be integrated in more areas? E.g. it also makes sense for Forecasting and Econometrics. Like for Statistics you sometimes take the log to normalize the series.

jmaslek:
> Think this is great, could even calculate the duration with this. Could this be integrated in more areas? E.g. it also makes sense for Forecasting and Econometrics. Like for Statistics you sometimes take the log to normalize the series.

Should be straightforward as long as there are data frames we can use the evaluation method on.

deeleeramone:
It would be really useful to have a `--title` flag on the custom plot functions. A chart with no title is just "incorrect", and it would be impossible to know the user's intention for generating a title automatically.

deeleeramone:
As a user, I want to be able to *not* use, rely on, Sheets/Excel in order to accomplish arithmetic functions for OpenBB Terminal data. 

Screen Shot 2022-09-19 at 7 35 32 PM

DidierRLopes:
Why does the sum doesn't work?
Screenshot 2022-10-08 at 22 11 01

DidierRLopes:
@piiq is there a way to allow the user to do something like `eval newel = dgs2 / dgs5`.

Can you look into this?

colin99d:
Screen Shot 2022-10-09 at 9 12 27 AM
@jmaslek this command just needs a better help message so users know how to use it.

jmaslek:
> Why does the sum doesn't work? Screenshot 2022-10-08 at 22 11 01

Think I fixed this?   Issue was doing concat with axis=0 not axis=1.

Also note I coded this so that if a column exists, I appended a _duplicate to distinguish.

jmaslek:
> Screen Shot 2022-10-09 at 9 12 27 AM
> 
> @jmaslek this command just needs a better help message so users know how to use it.

Okay added error messages for invalid syntax and also if the user tries to use a column that does not exist.  Can add more edge cases.

colin99d (approved):

PR #2560, opened 2022-09-18
Change code to allow homebrew install

DidierRLopes:
I need us to release a new version so I can test the Homebrew formula.


terminal.py
```diff
@@ -876,8 +877,20 @@ def main(
...
     routines_args : List[str]
         One or multiple inputs to be replaced in the routine and separated by commas. E.g. GME,AMC,BTC-USD
+    install : bool
+        Install all packages from scratch for the first time
     """
 
+    if install:
+        import subprocess
+
+        subprocess.run(  # nosec
```
piiq, line 887: @DidierRLopes This is high risk of screwing up the end user system python

DidierRLopes: Why is this high risk of screwing up the end user system python?

This was a suggestion from @aia to make it easier to install OpenBB from homebrew. Adding the resources to the Formula is unmanageable. 

Let me know if you have a better idea to achieve this.

piiq: 1. subprocess with shell=True is risky by itself (that's why you have the nosec flag there)
2. when you launch poetry like this you don't know which poetry it's going to use. if the user has a new mac and installed brew just to get the terminal via brew this would fail as there will be no poetry around. If for some reason there is poetry in the system python, all the deps will be attempted to be installed into the system python. This is not good, unless this particular version of the terminal is going to be the only python application that is ever going to run on that computer
3. In addition to this not all packages in the dependency tree have binaries for all apple platforms which will lead to pip attempting to build the binaries locally which will fail as nothing for performing that task is going to be around (see "cmake not found" errors)

altogether this might lead to a half working system python environment that noone would know how to clean up

DidierRLopes: Regarding 2. the brew file would rely on homebrew poetry (https://formulae.brew.sh/formula/poetry) so there shouldn't be a concerned there. Similarly to that the python we would use would be installed from brew.

When creating the formula ruby file, you can put different conditions based on the platform you are running brew from.

Do you have any suggestions then? More than 1 person has asked for this, and I'd like for us to try to be able to support it

simmonsj330:
@DidierRLopes we can release at anytime this week! I believe it would be v1.9.0

DidierRLopes:
> @DidierRLopes we can release at anytime this week! I believe it would be v1.9.0

Let me know before you do so please, so we can discuss whether we go ahead with this or not

piiq:
Distribution via homebrew will be added in future. Closing this PR

PR #2567, opened 2022-09-20
Add path for custom_imports outside the terminal

soggyomelette:
# Description
Adds a path for custom_imports in the user data folder. Also copies all example/default files from custom_imports in the repo folder and then deletes the folder so the user isn't confused about which folder to use


Chavithra (approved):
Looks good to me.

PR #2569, opened 2022-09-20
Improving styles loading

Chavithra:
# Description

Currently version assume that we won't add / remove parameters from default style.
This PR is there to remove that assumption.


soggyomelette (approved):
Looks good

piiq (approved):

PR #2570, opened 2022-09-20
fresh reports fix

soggyomelette:
# Description

Fresh PR for reports installer fix as previous didn't merge correctly


piiq:
on my machine when I run the report all the charts pop up on screen and. stay there

piiq (approved):

soggyomelette:
Charts in terminal don't show up after a report is run - not able to identify the issue so far

soggyomelette:
At the moment, the issue is with the bt module switching to a non-interactive plotting backend but attempts to switch back don't seem to work. So, as a temporary fix, the bt module is not available to use from the api but whenever a fix is identified, it will be modified in a future PR

Chavithra (approved):
Looks good to me.

PR #2592, opened 2022-09-22
Be able to change user data folder and add path for exports

soggyomelette:
# Description

Be able to change user data folder (basically replaced being able to set exports to being able to set user data) and change location of exports


Chavithra (approved):
Implementation looks good to me.

PR #2611, opened 2022-09-23
New path for portfolio data

soggyomelette:
# Description

New path for portfolio data outside the terminal


Chavithra (approved):
Looks good.
Would just remove the copy_files.

minhhoang1023:
@JerBouma Can you give this a try to see if it all works as expected?

soggyomelette:
We could also look at deprecating the pa menu since it doesn't seem to be used by the portfolio base controller

PR #2646, opened 2022-09-27
New path for data_sources_default.json

soggyomelette:
# Description

Create a data_sources_default.json file in user data folder and set any changes to data_sources to that file by default


Chavithra (approved):
Looks good to me.

Issue #2612, opened 2022-09-23
[Bug] Integration tests failing


minhhoang1023:
Shared by @piiq 

```
Terminal/scripts/test_alt_oss.openbb: display_rossindex() got an unexpected keyword argument 'top'
Terminal/scripts/test_crypto_defi.openbb: display_defi_protocols() got an unexpected keyword argument 'descend'
Terminal/scripts/test_crypto_disc.openbb: display_coins() got an unexpected keyword argument 'top'
Terminal/scripts/test_econometrics_panel.openbb: 'NoneType' object has no attribute 'name'
Terminal/scripts/test_econometrics_ts.openbb: 'NoneType' object has no attribute 'name'
Terminal/scripts/test_economy.openbb: get_aggregated_macro_data() got an unexpected keyword argument 'currency'
Terminal/scripts/test_etf.openbb: list.remove(x): x not in list
Terminal/scripts/test_etf_ta.openbb: view_ma() got an unexpected keyword argument 'series'
Terminal/scripts/test_funds.openbb: ERR#0015: error 403, try again later.
Terminal/scripts/test_stocks_ba.openbb: 'AAPL'
Terminal/scripts/test_stocks_disc.openbb: list index out of range
Terminal/scripts/test_stocks_fa.openbb: 'NoneType' object has no attribute 'empty'
Terminal/scripts/test_stocks_options.openbb: 'Unrecognized kwarg="upexpiry_width_config"'
Terminal/scripts/test_stocks_ta.openbb: fibonacci_retracement() got an unexpected keyword argument 'period'
Terminal/scripts/test_reports_crypto.openbb: A cell timed out while it was being executed, after 4 seconds.
Terminal/scripts/test_reports_equity.openbb: Exception encountered at "In [13]": TypeError: 'NoneType' object is not subscriptable
Terminal/scripts/test_reports_etf.openbb: Exception encountered at "In [7]": AttributeError: 'Index' object has no attribute 'date'
Terminal/scripts/test_reports_forex.openbb: A cell timed out while it was being executed, after 4 seconds. The message was: Timeout waiting for IOPub output.
Terminal/scripts/test_reports_portfolio.openbb: Exception encountered at "In [7]": FileNotFoundError: [Errno 2] No such file or directory: 'portfolio/holdings/Public_Equity_Orderbook_test.xlsx'
```

Apart from that there are a few tests that weren’t maintained or they break automatic execution:

```
[...]

PR #2677, opened 2022-09-29
Fix google trends. Error: The request failed: Google returned a response with code 429.

montezdesousa:
This PR fix the error showing on integration tests "Error: The request failed: Google returned a response with code 429." -> too many requests : mentioned on #2612 

The PR also fixes an issue with ba controller, where `mentions` was prompting:
image

Instead of:
image

When user tries to run `mentions` without any ticker loaded.

The 429 error might occur with commands: `mentions`, `regions`, `queries` and `rise`.

Integration tests: `python terminal.py scripts\test_stocks_ba.openbb`

api
```python

from openbb_terminal.api import openbb

openbb.stocks.ba.mentions(symbol="AAPL")
openbb.stocks.ba.regions(symbol="AAPL")
openbb.stocks.ba.queries(symbol="AAPL")
openbb.stocks.ba.rise(symbol="AAPL")
```

Might me useful to overload `TrendReq` to get the desired error with:

```python
for i in range(10000):
       openbb.stocks.ba.mentions(symbol="AAPL")
```


hjoaquim (approved):
Nothing to add imo.
Test pass now:
image

PR #2683, opened 2022-09-30
fix: error #2676

### LINKED ISSUE #2676: [Bug] Terminal settings are not being read from the file they are written
I noticed this because the contents of the old `.env` file in my OpenBBTerminal folder had been updated to be a blank file. Settings for plotautoscaling , dpi , and monitor are being read from the old location .env file. If there are no entries in this file, it goes to "default". 

You can test this by removing entries from the `.env` in the old location, changing a setting in the Terminal settings ( you will see this be updated in `./openbb_terminal/.env`) - like dpi 300 - restarting and then looking at the setting menu. You should see that any changes made will be forgotten.

soggyomelette:
# Description
Fixes #2676 - although the variable was being written to the user_env_file, it didn't load the user_env_file while setting the variables in the terminal

### DIFF
--- a/openbb_terminal/config_plot.py
+++ b/openbb_terminal/config_plot.py
@@ -1,5 +1,12 @@
 import os
 
+import dotenv
+
+from openbb_terminal.core.config.paths import USER_ENV_FILE, REPOSITORY_ENV_FILE
+
+dotenv.load_dotenv(USER_ENV_FILE)
+dotenv.load_dotenv(REPOSITORY_ENV_FILE, override=True)
+
 PLOT_DPI = int(os.getenv("OPENBB_PLOT_DPI", "100"))
 
 # Backend to use for plotting


colin99d:
Om were you able to reproduce his issue?

soggyomelette:
> Om were you able to reproduce his issue?

Not exactly since the dpi it shows in the settings menu is the same as the value you enter but when it goes to actually plot something, it uses the value it loads from the user env value which wasn't being loaded correctly

colin99d:
I see. So I will know if this work based on whether the actual chart dpi changes?

soggyomelette:
> I see. So I will know if this work based on whether the actual chart dpi changes?

Yep

colin99d:
Lol, works perfectly

colin99d (approved):

deeleeramone:
#2676

PR #2702, opened 2022-10-03
Add record session into routine capabiltiies

DidierRLopes:
https://user-images.githubusercontent.com/25267873/193699845-a8225906-67ed-4391-b90f-8996333f5a84.mp4

### DIFF
--- a/i18n/en.yml
+++ b/i18n/en.yml
@@ -11,6 +11,8 @@ en:
   sources: select your preferred data sources
   keys: set API keys and check their validity
   wiki: search for an expression in Wikipedia (https://www.wikipedia.org/)
+  record: record session to be converted into a .openbb routine
+  stop: stop session recording
   _configure_: Configure your own terminal
   _main_menu_: Main menu
   keys/_keys_: This menu allows you to set your own API keys. Type 'about' for more information
--- a/openbb_terminal/parent_classes.py
+++ b/openbb_terminal/parent_classes.py
@@ -1,7 +1,7 @@
 """Parent Classes"""
 __docformat__ = "numpy"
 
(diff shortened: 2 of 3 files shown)


colin99d:
It would mean a lot to me if the command was q and you could name the macro with one letter.

colin99d:
https://www.redhat.com/sysadmin/use-vim-macros#:~:text=To%20record%20a%20macro%20and,again%20to%20stop%20the%20recording.

DidierRLopes:
> It would mean a lot to me if the command was q and you could name the macro with one letter.

"q" is already used for quit unfortunately.

I can have just the command `record` to start and stop. What do you think?

For the macro being a single letter, we prob need a mapping from routine to macro. I can work on that in the future.

colin99d:
> > It would mean a lot to me if the command was q and you could name the macro with one letter.
> 
> "q" is already used for quit unfortunately.
> 
> I can have just the command `record` to start and stop. What do you think?
> 
> For the macro being a single letter, we prob need a mapping from routine to macro. I can work on that in the future.

Im sorry it was just a vim joke

colin99d:
@DidierRLopes. Maybe this is out of scope for this pr but when running the `exe` command should we remove the .openbb and just show `test` instead of `test.openbb`?

colin99d (approved):
I bullied it, and it survived. Nice work!

DidierRLopes:
> @DidierRLopes. Maybe this is out of scope for this pr but when running the `exe` command should we remove the .openbb and just show `test` instead of `test.openbb`?

Probably I would leave the .openbb so that the user knows that the file needs to have that extension

PR #2710, opened 2022-10-04
replace ln and ld by search

### LINKED ISSUE #2542: [IMPROVE] ETF replacing ld and ln by search
We should just have a `search` command instead of `ld` and `ln`.

The search should have `-n/--name` and `--d/--desc` arguments.

DidierRLopes:
* This fixes #2542 2542


DidierRLopes:
@Chavithra need a hand with the tests here pleaaseeee

Chavithra:
The following code only work if you have space separated option/value, like `search --name oil --source FinanceDatabase`:
```python
            required="-h" not in other_args
            and "-n" not in other_args
            and "--name" not in other_args,
```

If you are using option/value like this ` search --name=oil --source=FinanceDatabase` it will cease to work as expected.

DidierRLopes:
> Hi Didier,
> 
> The following code only work if you have space separated option/value, like `search --name oil --source FinanceDatabase`:
> 
> ```python
>             required="-h" not in other_args
>             and "-n" not in other_args
>             and "--name" not in other_args,
> ```
> 
> If you are using option/value like this ` search --name=oil --source=FinanceDatabase` it will cease to work as expected.

Ohhh.. but that's how it was before. How was it working? 🤔

Chavithra:
Like if you have `"--name=oil  --source=FinanceDatabase", then:

```python
 other_args = [
    "--name=oil",
    "--source=FinanceDatabse",
]
```

So the `"--name" in other_args` statement won't work.

Chavithra (approved):
The `etf/search` command works on my side.

DidierRLopes:
> Like if you have `"--name=oil --source=FinanceDatabase", then:
> 
> ```python
>  other_args = [
>     "--name=oil",
>     "--source=FinanceDatabse",
> ]
> ```
> 
> So the `"--name" in other_args` statement won't work.

I see now. That's smart. Nice catch, wouldn't have catch it that easily. Thanks @Chavithra

PR #2757, opened 2022-10-09
Enable forecasting on m1 macs

piiq:
This PR add support for the forecasting menu on native M1 environments.

@deeleeramone @martinb-bb @colin99d @montezdesousa @simmonsj330 @DidierRLopes can you help me test it, please?

The setup here would require cmake to update some of the dependencies, but we can potentially lock a number of third level deps if it works and we'd like to avoid the build.


martinb-bb:
Working on Linux. To test M1.

deeleeramone:
Everything seems to be working with M1 and Python 3.9.6 or 3.10.8 - terminal and SDK. 

@piiq: The only issue I am running into, and it makes no difference which version of Python is used, is that it fails to start the kernel in the Jupyter app (lab and notebook). Oddly enough, it works fine with both iPython and the Jupyter integration within VS Code. With the x86 environment this does work in all scenarios.

piiq:
Most of the changes from this PR were implemented in #3432 
Closing

PR #2773, opened 2022-10-11
Add demark indicator to stocks/ta

### LINKED ISSUE #2249: [FR] Tom DeMark's Sequential Indicator, from pandas-ta
The proprietary version of this indicator (https://demark.com/sequential-indicator/) has been recreated in the pandas-ta library, used by the OpenBB Terminal. https://github.com/twopirllc/pandas-ta#new-indicators

> Tom DeMark's Sequential (td_seq) attempts to identify a price point where an uptrend or a downtrend exhausts itself and reverses. Currently exlcuded from df.ta.strategy() for performance reasons. See help(ta.td_seq)

This indicator is popular among professional traders, across many different asset classes and backtesting, for signalling market timing (exhaustion) to both the up and down side.

> A Buy Setup occurs when there are 9 consecutive closes less than the close four bars earlier. A Sell Setup occurs when there are 9 consecutive closes greater than the close four bars earlier.

> The Risk Level identifies the potential residual trend momentum once Countdown is completed. The Risk Level is often considered a secondary level of price exhaustion, not only defining the zone in which the market is primed for a reversal but also providing support/resistance to the chart.
> 
> When a Countdown series is completed a Risk Level is generated, identifying the zone within which there should be a reversal, as well as the amount of residual momentum the trend could experience and yet still preserve the buy or sell reading. The Risk Level associated with a completed Buy Countdown represents the price above which the 13 buy remains active, and serves as support on weakness; conversely, the Risk Level associated with a completed Sell Countdown represents the price below which the 13 sell remains active, and serves as resistance on strength. It is helpful to think of this level as the maximum threshold that the price exhaustion series can withstand before the momentum overwhelms the signal and renders it ineffective.
> 
> As with many other DeMARK Indicators, the qualified and disqualified Breakout Rules apply to help gauge whether or not the Sequential signal remains intact.

> The twelve price bar metric states that a reversal should occur within 12 bars of a completed 13. If the chart has not responded in that time the 13 is unlikely to have the desired response, and the existing trend is likely to continue.

Adding this component to our technical analysis functions will provide users with an open source alternative to a popular proprietary indicator. This would also be an excellent tool for the Forecast Menu (@martinb-bb) to include.

jmaslek:
This will fix #2249 and by fix, I mean this will implement the demark indicator.

This plot is pretty busy and I chose to use the ohlc instead of candle in mpf intentionally.  This results in a less busy plot as mentioned in the discussion in that thread.

Screen Shot 2022-10-10 at 9 02 06 PM

### DIFF
--- a/i18n/en.yml
+++ b/i18n/en.yml
@@ -430,6 +430,7 @@ en:
   stocks/ta/fisher: fisher transform
   stocks/ta/cg: centre of gravity
   stocks/ta/clenow: clenow volatility adjusted momentum
+  stocks/ta/demark: Tom Demark's sequential indicator
   stocks/ta/_trend_: Trend
   stocks/ta/adx: average directional movement index
   stocks/ta/aroon: aroon indicator
--- a/openbb_terminal/common/technical_analysis/momentum_model.py
+++ b/openbb_terminal/common/technical_analysis/momentum_model.py
@@ -184,7 +184,7 @@ def cg(values: pd.Series, window: int) -> pd.DataFrame:
         Length for indicator window
     Returns
     ----------
-    d.DataFrame
+    pd.DataFrame
         Dataframe of technical indicator
     """
     return pd.DataFrame(ta.cg(close=values, length=window).dropna())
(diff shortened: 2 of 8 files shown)


DidierRLopes (approved):
Noice!

PR #2785, opened 2022-10-11
Encode string

colin99d:
# Description

@martinb-bb right now this is a bad implementation because it uses increasing numbers, so the computer believe there are relationships that do not actually exist. Lets talk about a better implementation.

# How has this been tested?

- [ ] Ensure the api still works

# Checklist:

# Others


openbb_terminal/forecast/forecast_model.py
```diff
@@ -542,3 +542,24 @@ def corr_df(df: pd.DataFrame) -> pd.DataFrame:
...
+        The dataset you wish to modify
+    target_column : str
+        The column you wish to encode
+
+    Returns
+    -------
+    pd.DataFrame:
+        Dataframe with added encoded column
+    """
+    dataset[target_column] = dataset[target_column].astype("category").cat.codes
```
jmaslek, line 563: Why not the standard sklearn.preprossessing.LabelEncoder?

jmaslek:
This is in general a bad idea because as you say, it associates higher numbers.  Look into one hot encoding (which can possibly add a lot of variables) but is better for non-ordinal features

martinb-bb:
@colin99d Something to look into is to see if Darts supports non number variables. Might be a simple solution.

openbb_terminal/forecast/forecast_model.py
DidierRLopes: Yup, that's why Colin added that description. We already talked about this 👍🏽 

We should defo be using sklearn for this.

I would still allow to have an option for this type of encoding which makes sense in the context of quantitative variables, e.g. "How hot was the water": "cold", "not much". "hot", "boiling". In here 1,2,3,4 would make sense.

martinb-bb: For reference: 

```python
>>> from sklearn import preprocessing
>>> le = preprocessing.LabelEncoder()
>>> le.fit(["paris", "paris", "tokyo", "amsterdam"])
LabelEncoder()
>>> list(le.classes_)
['amsterdam', 'paris', 'tokyo']
>>> le.transform(["tokyo", "tokyo", "paris"])
array([2, 2, 1]...)
```
Then we can replace the column with this newly encoded array?

martinb-bb: Also, for later versions: https://stackoverflow.com/questions/24458645/label-encoding-across-multiple-columns-in-scikit-learn

martinb-bb:
Also, I think we may not want to make this is a menu function. Instead just have it as an automatic check and process when creating `TimeSeries()` and the user specifies a categorical column as a covariate or forecast target.

openbb_terminal/forecast/forecast_model.py
jmaslek: > Yup, that's why Colin added that description. We already talked about this 👍🏽
> 
> We should defo be using sklearn for this.
> 
> I would still allow to have an option for this type of encoding which makes sense in the context of quantitative variables, e.g. "How hot was the water": "cold", "not much". "hot", "boiling". In here 1,2,3,4 would make sense.

Yes, but if the encoder sees cold, boiling, hot, not much, in that order, it will encode it as 1,2,3,4 even though that is not the right ordinal ordering.  What kind of strings are being encoded?

DidierRLopes: Yup. But there are examples where preserving that relationship between variables can be important. We could allow the user to specify those :)  But prob would be better to have this as a request rather than just adding it without knowing if users would like that.

colin99d:
We are holding off on this for now. Darts static covariates might handle this already.

PR #2811, opened 2022-10-12
Add --export to load functions

jmaslek:
As the title says, this adds export to the various loads.  The reason is that this makes life easier to use the econometrics and forecasting menu.

Now we can have a script that would do

```
stocks
load aapl --export aapl.csv
load msft --export msft.csv
load amzn --export amzn.csv
econometrics
....
```


openbb_terminal/parent_classes.py
```diff
@@ -1181,3 +1192,9 @@ def call_load(self, other_args):
                     ns_parser.exchange,
                     self.current_interval,
                 )
+                export_data(
+                    ns_parser.export,
+                    os.path.dirname(os.path.abspath(__file__)),
+                    "load",
+                    self.current_df.copy(),
```
DidierRLopes, line 1199: Why `copy()` here?

jmaslek: i don't want to goof up the dataframe lol

DidierRLopes (approved):

openbb_terminal/parent_classes.py
DidierRLopes: What do you mean by goof up? At that stage we just save it, there's nothing happening with it after.

jmaslek: I had an edge case in my brain at the time but can't think of what it is now

martinb-bb (approved):

Issue #2732, opened 2022-10-07
[Bug] pred/conv1d is broken


husmen:
**Describe the bug**
Trying to run predictions using `conv1d` results in the following error:

```
Could not load library libcudnn_cnn_infer.so.8. Error: libcuda.so: cannot open shared object file: No such file or directory
```

All other models worked with no issue.

**To Reproduce**
- crypto
- load eth
- pred
- conv1d

**Screenshots**

**Desktop (please complete the following information):**

- OS: WSL2 (Ubuntu 22.04.1 LTS) on Windows 11
- Python version 3.9

**Additional context**
I tried tensorflow from both `conda` and `pip`. the former automatically installed  `cudatoolkit 11.7.0` and `cudnn 8.4.1.50`. And for the latter, I manually installed `cuda 11.7.0` and `cudnn 8.5.0` under `/usr/local/cuda/`. Both resulted in the same error.

colin99d:
I am going to close this because we removed prediction with our new menu! @husmen please tell me how you like the new forecast menu, it blows pred out of the water.

husmen:
The menu is nice but the issue persists. This time in `tcn`.
In the meantime, I submitted a PR #2829 to enable GPU support for some of the models.

PR #2829, opened 2022-10-13
GPU support in forecast

husmen:
# Description
All forecast models were using the CPU exclusively, so I added GPU support for the models that can benefit from that:
- brnn
- nbeats
- nhits
- rnn
- tft
- trans

CPU is still used by default. GPU is used only if the user sets `--use-gpu` flag and GPU availability check passes.

Dependency wise, no change is required, but the default torch installation only supports CUDA capabilities sm_37 sm_50 sm_60 sm_70. In order to run on newer architectures, the user will have to manually re-install the appropriate version as described [here](https://pytorch.org/get-started/locally/):
```bash
pip uninstall torch
pip install torch --extra-index-url https://download.pytorch.org/whl/cu116
```

**Note**: `tcn` model should be able to support GPU acceleration, but I had to omit it due to the same bug reported in #2732.

- [X] Summary of the change / bug fix.
- [X] Link # issue, if applicable.
- [X] Relevant motivation and context. 
- [X] List any dependencies that are required for this change.

# How has this been tested?
I tested it by running all forecast methods in the terminal with and without `--use-gpu` flag.

- [X] Make sure affected commands still run in terminal
- [X] Ensure the api still works

# Checklist:

- [X] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [X] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).

# Others


colin99d:
@husmen this is awesome!! @martinb-bb what do you think about this implementation?

martinb-bb:
@husmen Thanks for the contribution! This is a great start. 😄 

We cannot merge just yet, as we must wait for the current installer to be released before our 2.0 launch within the week. 🚀 

In the meantime, I will test and work diligently with you to have this merged in the next couple of weeks. We should also consider updating the installer to install GPU support if the user has one automatically. (something to think about)

piiq:
Just contributing to the discussion

@husmen, I few months ago I've been testing MPS (Apple's new Metal Performance Shaders) for both training and inference. My findings are that for the models that were in the terminal at that time the performance boost was negative (computation on MPS was over 5 times slower than on CPU) because the models and the batches are not big enough to actually make use of the parallelization that is offered by MPS. I assume it's the same with CUDA. I have to note that this was before the forecasting menu has been built and the prediction menu models used tensorflow, not pytorch.

Did you do some benchmark test that compares execution on CPU to execution on GPU? It would be great to see those numbers if you have them

husmen:
@piiq, Actually yes, I can confirm that running on CUDA was slower than on CPU. I will provide the numbers when I find the time in few days, but if I remember correctly the slowdown was not as dramatic as 5 times.
That being said, I still think it will be useful to offer the feature for anyone who wants to play with larger datasets and tune the model arguments accordingly. I might be able to test that as well.

Chavithra:
Hi @martinb-bb, is this PR still relevant with the current version of the forcast menu?
Thanks

martinb-bb:
@Chavithra it's relevant but we are not going to support GPU for now until we update install docs to add support for GPU enabled libraries.

piiq:
> until we update install docs to add support for GPU enabled libraries

The documentation update can be a part of this PR? Can you specify what is needed?

Issue #2833, opened 2022-10-14
[Bug] `reports/equity tsla` Error : 'NoneType' object is not iterable


colin99d:
**Describe the bug**
Equity command gives error.

**To Reproduce**
`python terminal.py reports/equity tsla`

**Screenshots**
<img width="755" alt="Screen Shot 2022-10-14 at 9 06 17 AM" src="https://user-images.githubusercontent.com/72827203/195795587-4266578a-83d0-4dd6-b507-8e2df98d9c54.png">

**Desktop (please complete the following information):**

- OS: [e.g. Mac Sierra]
- Python version [e.g. 3.6.8]

**Additional context**
Add any other information that you think could be useful for us.

kairosdojo:
Hi! I could reproduce this too. It seems to be an API keys missing (at least in my case). 

The problem is that in [L280 of reports_controller.py](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/openbb_terminal/reports/reports_controller.py#L280) the `stdout` of the copied notebook is captured, therefore hiding the error.

Disabling that context I got this:

![Screenshot_2022-10-14_16-06-53](https://user-images.githubusercontent.com/22766463/195867553-61efadf7-820f-41c9-9531-9b8bc5243ed2.png)

Probably the lack of API KEYS should be handled.

colin99d:
Ok. We should just need better handling of None. Do you want to take this one @kairosdojo?

kairosdojo:
Ok, I'll try :)

tinix84:
I have the same error but also sometimes other like 
![image](https://user-images.githubusercontent.com/16301528/196007615-3de75e47-1435-459b-a4e1-54cc72c28a71.png)

hjoaquim:
@tinix84 in the meantime, #2861 was merged. Would you update and try again pls??

tinix84:
> @tinix84 in the meantime, #2861 was merged. Would you update and try again pls??

I have tried yesterday evening after the update, and I wrote in discord.

kairosdojo:
Hello, Riccardo! Maybe it's related to https://github.com/OpenBB-finance/OpenBBTerminal/issues/2860 ? 

What's curious is that your error prints out the cell block number.  

Also, another thing: in my example I had already got an `AAPL.csv` file loaded as a custom dataset; you did just type AAPL (but you hadn't one already saved)?

colin99d:
[...]

PR #2834, opened 2022-10-14
Removed warning when importing API

colin99d:
# Description

Right now the API prints a warning on import. This warning is no longer necessary, so I removed the print statement.

# How has this been tested?
    - There was no specific command to test, so I ran `python terminal.py` to confirm that the terminal still works. I did not test more than this because I am confident that removing a print statement will not have effects on other places in the terminal.
- [x] Ensure the api still works
   - I opened up a jupyter notebook and ran `from openbb_terminal.api import openbb` to confirm no error message was shown. I then ran commands from the stocks and forecast menu to ensure that there were no unforeseen side effects. Because of the very remote chance that deleting a print statement would change more code, I stopped after this limited testing.
    - Ran all reports (except portfolio) and confirmed that they ran error free and the output seemed reasonable.
    - I did find an issue with equity, but it seemed unrelated to my PR. I reported the bug: #2833.

# Checklist:

# Others


minhhoang1023:
As discussed earlier with @colin99d , I'll approve this later this sprint. Right now it doesn’t hurt us to have it, and can be relevant for old users to have the pointer towards new documentation.

As long as we have this removed by the launch, this should be fine.

deeleeramone:
How about a warning for when the filename of `api.py` changes and breaks everyone's scripts/notebooks.

minhhoang1023:
@deeleeramone We will have a warning every time there's a breaking change.

piiq (approved):
Finally we're getting rid of this warning!

minhhoang1023 (approved):

PR #2850, opened 2022-10-14
Downgrade Protobuf for Forecast Usage

simmonsj330:
# Description
After investigating if forecasting is supported if a user downloads OpenBB SDK via pip (pypi), it is apparent that the user will need to create a conda environment, install `pytorch-lightning` from conda-forge, and install `u8darts-torch` from conda-forge. After doing those things and installing OpenBB SDK via pip, the follow error appeared:

Screen Shot 2022-10-13 at 3 47 50 PM

This error actually does not disallow usage of forecasting features, but mitigating this error would be still a good idea. As such, I have downgraded to 3.20.1.

# How has this been tested?
I tested this by creating a new conda environment, installing the two forecasting packages, and downloading the pip package locally  with the command `pip install .` After doing so, the error did not appear.

# Others

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2548,7 +2548,7 @@ python-versions = ">= 3.5"
 
 [[package]]
 name = "protobuf"
-version = "3.20.3"
+version = "3.20.1"
 description = "Protocol Buffers"
 category = "main"
 optional = false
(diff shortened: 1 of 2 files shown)


simmonsj330:
After further investigation, this error only occurs if the user downloads `pytorch-lightning` before `u8darts`. In order to have forecasting working in the first place, the user must download darts first, which in turn mitigates this error. Closing the PR.

PR #2851, opened 2022-10-14
Fix settings not updating right away

### LINKED ISSUE #2676: [Bug] Terminal settings are not being read from the file they are written

colin99d:
# Description

Fixes #2676 

Currently after using the settings menu a user needs to exit and then reopen the terminal to have the new settings apply. I fixed this by calling load_dotenv at the end of every function call. This automatically applies any new changes.

# How has this been tested?
While this PR does affect a lot of terminal operations, it does not have many possible negative consequences, since all it does is load the env file. To provide assurance that bugs were not introduced I performed the following operations:
    - To make sure the settings instantly change terminal wide, I switched between changing the dpi `settings/dpi [dpi]` and plotting candles `stocks/candle` the changes were applied immediately.
    - To make sure I did not forget a menu I navigated to `settings` ran each command, and then ran `h` after to ensure that the updated was applied.
- [ ] Ensure the api still works
    - Not needed, controllers do not affect the API
    - Not needed, controllers do not affect reports

# Checklist:

# Others


jmaslek:
I think this works, but couldn't you just do `set_key(.., export=True)`

colin99d:
Yeah I did this wrong. Turns out timezone is the only broken one. My local branch has a lot more progress in fixing.

colin99d:
> I think this works, but couldn't you just do `set_key(.., export=True)`

When I tried this time zone did not update right away.

jmaslek (approved):

PR #2858, opened 2022-10-15
Moving `custom_imports` into `openbbterminal` folder.

Chavithra:
# Description

Moving `custom_imports` into `openbbterminal` folder.

I was able to confirm the new file paths worked in the simplified way inside the terminal using:
`python terminal.py forecast/load [file you have added at ~/OpenBBUserData/custom_imports/]`
Tested api through jupyter-lab using the same file.
Screen Shot 2022-10-17 at 10 27 42 AM

Ran `python terminal.py reports/forecast` 

To confirm the that futures menu still works I ran `python terminal.py reports/futures -n`, and got the following result:
Screen Shot 2022-10-17 at 10 55 14 AM


colin99d:
@Chavithra wasn't the idea to delete `custom_imports` from being inside the repository, and only use it inside `OpenBBUserData`?

Chavithra:
> @Chavithra wasn't the idea to delete `custom_imports` from being inside the repository, and only use it inside `OpenBBUserData`?

I am all for but right now these two files seems to use this `custom_imports` folder:
- openbb_terminal/dashboards/voila/futures.ipynb
- openbb_terminal/forecast/forecast_model.py

colin99d:
Would it be better if this PR just moved where those functions point?

colin99d (approved):
I was able to successfully get the new file path to work in terminal, api, and report environments. The futures menu also worked flawlessly!

PR #2889, opened 2022-10-17
Eliminate `update` for packaged app

jmaslek:
Since the update does not do anything on installer, can we hide it?


DidierRLopes (approved):
I thought this was added a long time ago... smh

PR #2899, opened 2022-10-18
Add intro command + improve plot style + internationalisation consistency

DidierRLopes:
This is what this PR looks like:
image

Still working on getting new colors so that when we have more than 5 lines they are more distinguishable between themselves.

Also adds an intro command:
Screenshot 2022-10-19 at 03 55 48


DidierRLopes:
> Can you add a way to leave the `intro` instead of having to go through all 10 things

Done!

jmaslek (approved):

PR #2916, opened 2022-10-19
Make tests work with the pip package

colin99d:
# Description

When testing the pip package we ran into two issues:

- `docs/generate.py` could not be found. This error arises because the `docs` folder is not in the `openbb_terminal` folder. The fix for this was to skip the tests (they are still run on all merges to main, so there is no need to run them here)
- `import darts` was passing even though darts was not installed. This meant that the SDK was trying to use darts, even though it was not installed. I fixed this by also importing `darts.utils` to ensure that darts is actually installed.

To run these tests use the following steps:

- pip uninstall openbbterminal -y
- poetry build
- pip install dist/OpenBBTerminal-X.Y.Z-py3-none-any.whl
- cd ..
- rm -fr tests
- cp -r OpenBBTerminal/tests .
- pytest tests

# How has this been tested?
These changes only affecting tests and running the SDK. Tests of reports were skipped and the terminal was tested minimally.
    - `python terminal.py`
    - `from openbb_terminal.sdk import openbb`
    - Not affected, did not check

# Checklist:

# Others


Chavithra (approved):
Looks good to me.

PR #2926, opened 2022-10-20
Making Crypto/Defi/DPI --sort Pretty Again

simmonsj330:
# Description
Making changes to revert back to the use of "M" (millions), "B" (billions), and "%" (percent), which looks better than simply using floats.

Screen Shot 2022-10-19 at 10 50 55 PM

Screen Shot 2022-10-19 at 10 50 43 PM

# How has this been tested?
Re-ran pytest on the model and view.

# Others

### DIFF
--- a/openbb_terminal/cryptocurrency/defi/defipulse_model.py
+++ b/openbb_terminal/cryptocurrency/defi/defipulse_model.py
@@ -13,6 +13,23 @@ logger = logging.getLogger(__name__)
 
 
 @log_start_end(log=logger)
+def num_to_int(num):
+    if num[-1] == "M":
+        return int(float(num[1:-1])) * 1000000
+    if num[-1] == "B":
+        return int(float(num[1:-1])) * 1000000000
+    if num[-1] == "k":
+        return int(float(num[1:-1])) * 1000
+
+    return int(num)
+
+
+@log_start_end(log=logger)
+def percent_to_float(num):
+    return float(num[0:-1])
+
+
+@log_start_end(log=logger)
 def get_defipulse_index(sortby: str = "TVL_$", ascend: bool = False) -> pd.DataFrame:
     """Scrapes data from DeFi Pulse with all DeFi Pulse crypto protocols.
     [Source: https://defipulse.com/]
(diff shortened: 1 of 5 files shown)


openbb_terminal/cryptocurrency/defi/defipulse_model.py
```diff
@@ -12,6 +12,23 @@
...
 
 
+@log_start_end(log=logger)
+def num_to_int(num):
+    if num[-1] == "M":
+        return int(float(num[1:-1])) * 1000000
+    if num[-1] == "B":
+        return int(float(num[1:-1])) * 1000000000
+    if num[-1] == "k":
+        return int(float(num[1:-1])) * 1000
```
jmaslek, line 22: small change.  Probably want to make sure that we look at .upper() and match the "k" to "K".

openbb_terminal/cryptocurrency/defi/defipulse_model.py
```diff
@@ -12,6 +12,23 @@
 logger = logging.getLogger(__name__)
 
 
+@log_start_end(log=logger)
```
DidierRLopes, line 15: I don't think we want to log a helper func

openbb_terminal/cryptocurrency/defi/defipulse_model.py
```diff
@@ -12,6 +12,23 @@
...
+def num_to_int(num):
```
DidierRLopes, line 16: type hint missing and doctoring

openbb_terminal/cryptocurrency/defi/defipulse_model.py
```diff
@@ -12,6 +12,23 @@
...
+    if num[-1] == "M":
+        return int(float(num[1:-1])) * 1000000
```
DidierRLopes, line 18: For visibility, you can do: `1_000_000` which is quite cool.

Also, int(float(X)) is redundant, you can just do `int`

openbb_terminal/cryptocurrency/defi/defipulse_model.py
DidierRLopes: Seems like the "num" var is a bit confusing, as you are playing with a string.

openbb_terminal/cryptocurrency/defi/defipulse_model.py
```diff
@@ -12,6 +12,23 @@
...
+    if num[-1] == "B":
+        return int(float(num[1:-1])) * 1000000000
+    if num[-1] == "k":
+        return int(float(num[1:-1])) * 1000
+
+    return int(num)
+
+
+@log_start_end(log=logger)
```
DidierRLopes, line 27: Same comments as above apply.

I would also avoid creating a function with 1 line of code. I would look into lambda to do that one liner function: 

e.g. `df['age']=df.apply(lambda x: x['age']+3,axis=1)` See https://www.analyticsvidhya.com/blog/2020/03/what-are-lambda-functions-in-python/

DidierRLopes (changes requested):

openbb_terminal/cryptocurrency/defi/defipulse_model.py
simmonsj330: I thought thousand was normally lowercase?

simmonsj330:
Closing because command was removed

PR #2946, opened 2022-10-20
Adding more integration tests

jmaslek:
Going through some QA, trying to improve "coverage" if you will.  WIP as it relies on some bugs being fixed.


openbb_terminal/miscellaneous/scripts/test_stocks.openbb
```diff
@@ -1,9 +1,13 @@
 stocks
-search apple
+search -c United States -s Consumer Cyclical -i Specialty Retail -e USA -l 5
+load aaple -i 5
 load aapl -s 2022-10-01
+quote
 tob
 tob -e EDGX
-candle
+candle -t --ma 5,10
```
simmonsj330, line 8: Should we keep commands with default parameters AND commands with additional flags/arguements?

jmaslek: Probably depends. Running `candle` runs the same code as `candle -t --ma 5,10` just skipping those flags, so if this fails, you are correct that we would not know if its because default pars, the -t par or the --ma par.

colin99d:
I said this in slack, but if we are going to make integration tests really big should we add multithreading?

openbb_terminal/miscellaneous/scripts/test_stocks_bt.openbb
```diff
@@ -5,6 +5,14 @@ whatif
 whatif -n 2.34
 whatif 2021-01-01
 ema
+ema --spy
+ema --spy --no_bench
+ema -l 2
 ema_cross
+ema_cross -l 10 -s 3
```
DidierRLopes, line 12: This is going to break with the name change

DidierRLopes:
> I said this in slack, but if we are going to make integration tests really big should we add multithreading?

We cannot easily implement multithreading due to the way matplotlib is running on a thread already, at least afaik

jmaslek:
> > I said this in slack, but if we are going to make integration tests really big should we add multithreading?
> 
> We cannot easily implement multithreading due to the way matplotlib is running on a thread already, at least afaik

We can run on multiple processes for the tests:
```
import subprocess
import pathlib
import concurrent.futures


def run_test(path_to_file):
    subprocess.call(["python", "terminal.py", f"scripts/{path_to_file}", "-t"])


def get_all_it_tests():
    directory = pathlib.Path(__file__).parent.absolute()
    it_tests = directory / "openbb_terminal" / "miscellaneous" / "scripts"
    files = [f.name for f in it_tests.iterdir()]
    return files


if __name__ == "__main__":
    files = get_all_it_tests()
    executor = concurrent.futures.ThreadPoolExecutor(max_workers=10)
    futures = [executor.submit(run_test, file) for file in files]
    concurrent.futures.wait(futures)
```

But the processes are unreadable:
Screen Shot 2022-10-21 at 9 16 59 AM

DidierRLopes (approved):

Issue #2997, opened 2022-10-23
[Bug] Exception when calling balance in crypto / dd


0xDareDevil:
**Describe the bug**
Error raised when calling `balance` in crypto

**To Reproduce**
- Follow this: `crypto / load [coin] / dd / balance`

**coin** is any symbol that has some balance

**Screenshots**
![image](https://user-images.githubusercontent.com/114890815/197377364-967bd738-19b6-4bd9-a7c3-19c609fa8867.png)

**Desktop (please complete the following information):**

- OS: Ubuntu 22.04.1 LTS
- Python 3.10.6

PR #2998, opened 2022-10-23
Fixed bug giving exception when calling balance in crypto

0xDareDevil:
# Description

- Fixes: https://github.com/OpenBB-finance/OpenBBTerminal/issues/2997
- Checking dataframe using `not df` raises error. Replace it with condition `df is None`

image

# How has this been tested?

# Others


jmaslek:
I don't have a binance account to double check with, but I don't see why this wouldn't work (and the example shown works)

jmaslek (approved):

PR #3016, opened 2022-10-24
Docker + Pypi : update logs + builds

Chavithra:
# Description

For both `docker` and `pypi` packages here we are:
- updating the way we log the `commit hash`
- updating the build script


hjoaquim (approved):
Looks good to me. Went through it with @Chavithra and this accomplish the goal in a simple and easy to understand way, even improving readability.

PR #3022, opened 2022-10-24
Ta/atr function

jmaslek:
specific request @minhhoang1023 

added to sdk as well


DidierRLopes (approved):

minhhoang1023:
@jmaslek was about to share with ANC capital and i realized we don't yet have the documentation on Hugo for this. Can you add this? I'll create the ticket too.

PR #3133, opened 2022-10-27
Better UX when model cant find best theta

DidierRLopes:
From
image

To
Screenshot 2022-10-27 at 17 23 08


openbb_terminal/forecast/theta_model.py
```diff
@@ -96,12 +96,19 @@ def get_theta_data(
...
+            pred_theta = model.predict(len(val))
+            res = mape(val, pred_theta)
+            if res < best_mape:
+                best_mape = res
+                best_theta = theta
+        except Exception:  # noqa
+            continue
+
+    if best_theta == 0:
+        console.print("Theta Model failed to find best theta")
```
martinb-bb, line 110: Do you think we could tell the user why? In the previous it mentioned lack of data.

martinb-bb: If you don't mind, I'd like to add 1 quick sentence for that such as: 
"Please consider using a series with at least 50 elements."

martinb-bb:
On second thought, 

there is a fundamental UX problem if the user loads a super small dataset and tries to use any model. 
For instance, Expo: 

image

Instead of changing all models, I vote to just change the load requirements to check dataset size and make sure it hits a min # of rows. If it does not, we tell the user that "Data cannot be loaded due to min size requirements."

This will also stop users from uploading data sets that will provide no insights, since for any forecasting model to work properly, it needs an appropriate number of time stamps.

martinb-bb:
To my point above, how does this look: 

image

DidierRLopes:
> On second thought,
> 
> there is a fundamental UX problem if the user loads a super small dataset and tries to use any model. For instance, Expo:
> 
> image
> 
> Instead of changing all models, I vote to just change the load requirements to check dataset size and make sure it hits a min # of rows. If it does not, we tell the user that "Data cannot be loaded due to min size requirements."
> 
> This will also stop users from uploading data sets that will provide no insights, since for any forecasting model to work properly, it needs an appropriate number of time stamps.
> 
> what do you think @DidierRLopes ?

Sorry, I missed this message.

Yes, that idea seems much more robust! 

BUT because users will still be able to access the models through the SDK, we should probably add the messaging in there as well.

martinb-bb:
Added in expo model error handling and general purpose loading through terminal. 
Should be good with both API and normal terminal use now!

martinb-bb (approved):

DidierRLopes:
Amazing! Neat!

PR #3179, opened 2022-10-30
Improve UX when wrong command is selected

DidierRLopes:
From:

Screenshot 2022-10-30 at 13 00 52

To:

Screenshot 2022-10-30 at 12 55 11


jmaslek (approved):

DidierRLopes:
@Chavithra need your help to fix the tests.

The issue comes from: `exit_on_error=False,` line.

It works well, but all tests fail on `__init__`

openbb_terminal/parent_classes.py
```diff
@@ -134,9 +134,10 @@ def __init__(self, queue: List[str] = None) -> None:
         self.completer: Union[None, NestedCompleter] = None
 
         self.parser = argparse.ArgumentParser(
-            add_help=False, prog=self.path[-1] if self.PATH != "/" else "terminal"
+            add_help=False,
+            prog=self.path[-1] if self.PATH != "/" else "terminal",
+            exit_on_error=False,
```
Chavithra, line 139: Looks like inside our MacOS CI : `ArgumentParser` doesn't have `exit_on_error`.

A solution for the tests to pass would be to remove this argument from `ArgumentParser`:
```python
exit_on_error=False,
```

PR #3180, opened 2022-10-30
[SDK] Allow silencing verbose output in commands that use stocks/load

DidierRLopes:
As per @piiq request:
Screenshot 2022-10-30 at 15 24 06


piiq:
@DidierRLopes @jmaslek 

I've reverted the verbosity related changes introduced to the controllers. In the `stocks_helper.load` function (that's used across multiple menus of the terminal) the `verbose` argument is now types as `bool`, is optional and `True` by default. _This means that the default behaviour will not change neither on the CLI nor on the SDK._ 

The only thing this PR changes is introduces an ability to explicitly silence the printout produced by the helper load in stocks and forex menus where this functionality is linked directly.

image


`stocks_helper.load` is also mentioned in `etf` and some `qa` commands, but the commands themselves require refactoring and I will leave it outside of the scope of this PR

---

While editing the files, I've updated the docstrings to be compliant with `pydocstyle` linting rules as a part of our move towards better docstrings.
And fixed some typos

piiq:
@DidierRLopes @jmaslek @Chavithra This PR is now expected to fix the failing futures tests that block the CI for everyone. Please stamp and merge once tests pass

thanks to @Chavithra for proper debug

jmaslek (approved):

PR #3211, opened 2022-10-31
New format for SDK documentation

montezdesousa:
# Description

This PR changes SDK documentation formatting from .md to .rst to allow a bit more customization.

- [x] Clean up the `generate.py` code
- [x] Extend format to futures menu
- [x] Extend format to forecast menu
- [x] Clean buggy docstrings

Before:
image

After:
Screenshot 2022-11-07 at 19 07 52


JerBouma:
@montezdesousa Is it correct that the `export` argument is no longer an option for the SDK with this PR? Otherwise, it needs to be included in the Docstrings properly.

DidierRLopes:
@montezdesousa on the return can we remove the indentation?

Like:
```
alt.covid.slopes(
    arg,
    arg,
) -> pandas.core.Frame.DataFrame
```

piiq:
If I understand correctly the export arg of the CLI is meaningless in SDK

JerBouma:
> If I understand correctly the export arg of the CLI is meaningless in SDK

Yes, I discussed with @montezdesousa that we won't mention it in the documentation and that the functionality itself will be removed at some point for the SDK. My main point is that currently it is an actual argument you can use and pretty sure it will still export something but that for the SDK there are more intuitive ways as you are directly programming in Python.

montezdesousa:
> If I understand correctly the export arg of the CLI is meaningless in SDK

Yes, for now just removed it from the doc but later I've to actually pop it from the function args

montezdesousa:
> @montezdesousa on the return can we remove the indentation?
> 
> Like:
> 
> ```
> alt.covid.slopes(
>     arg,
>     arg,
> ) -> pandas.core.Frame.DataFrame
> ```

fixed

JerBouma:
Can you let me know when this should be ready? Then I can stamp it, put on my headphones and cram out those examples.

JerBouma:
> The new format looks much better! Well done.
> 
> I noticed that we only have 1 documentation for both the view & model. What happened if the list of input arguments differ between the two?

It is currently very confusing that we list the same command twice for every function that also has the ability to plot something. Sometimes, the only difference is `chart=True`. We've discussed to combine this in one function doc in which we explicitly mention that you can call upon the graph with `chart=True` (and other related commands).

I understand that **we** know there is a model and view behind this but the user of the SDK is unaware so I wouldn't want to confuse them with this by listing every function call twice. When there are more commands, either we should look at if it makes sense these are only available in the view (other than chart and external axes) and if not, integrate in the model.

E.g. this adds little value: https://openbb-finance.github.io/OpenBBTerminal/SDK/economy/macro/

montezdesousa:
> Can you let me know when this should be ready? Then I can stamp it, put on my headphones and cram out those examples.

Sure, I'll send you details with the process, but we should be able to work in parallel

montezdesousa:
> > The new format looks much better! Well done.
> > I noticed that we only have 1 documentation for both the view & model. What happened if the list of input arguments differ between the two?
> 
> It is currently very confusing that we list the same command twice for every function that also has the ability to plot something. Sometimes, the only difference is `chart=True`. We've discussed to combine this in one function doc in which we explicitly mention that you can call upon the graph with `chart=True` (and other related commands).
> 
> I understand that **we** know there is a model and view behind this but the user of the SDK is unaware so I wouldn't want to confuse them with this by listing every function call twice. When there are more commands, either we should look at if it makes sense these are only available in the view (other than chart and external axes) and if not, integrate in the model.
> 
> E.g. this adds little value: https://openbb-finance.github.io/OpenBBTerminal/SDK/economy/macro/

@JerBouma @minhhoang1023 actually there is a point here, because we might have functions where the view might have chart specific arguments that we will be hiding.

JerBouma:
> > > The new format looks much better! Well done.
> > > I noticed that we only have 1 documentation for both the view & model. What happened if the list of input arguments differ between the two?
> > 
> > It is currently very confusing that we list the same command twice for every function that also has the ability to plot something. Sometimes, the only difference is `chart=True`. We've discussed to combine this in one function doc in which we explicitly mention that you can call upon the graph with `chart=True` (and other related commands).
> > 
> > I understand that **we** know there is a model and view behind this but the user of the SDK is unaware so I wouldn't want to confuse them with this by listing every function call twice. When there are more commands, either we should look at if it makes sense these are only available in the view (other than chart and external axes) and if not, integrate in the model.
> > 
> > E.g. this adds little value: https://openbb-finance.github.io/OpenBBTerminal/SDK/economy/macro/
> 
> @JerBouma @minhhoang1023 actually there is a point here, because we might have functions where the view might have chart specific arguments that we will be hiding. 

We shouldn't be hiding _any_ argument, we just make it clear that you need _chart=True_ to use them.

minhhoang1023:
@JerBouma We had a lengthy discussion on this one. I'm with you on making it as simple as possible to use, and I prefer the current format at the moment.

However, `chart=True` is not the only difference between the models & views. Even if we ignore other parameters such as `limit`, and `raw`, which are only relevant for CLI, there are still many differences between models & views. 

If we were to combine them, we would have to take into great account this difference. Check [this one in specific](https://openbb-finance.github.io/OpenBBTerminal/SDK/common/qa/rolling/). There is one required argument, `target`, that exists in the view but not the model. If we only present one piece of documentation, either from model or view, it will throw an error. 

Therefore, I'd suggest first making sure all arguments are the same between views and models, except for things like `limit`, `raw`, and `external_axes`. Only then, we can present only the documentation page of the `model`. If we skip this step, users will follow the documentation page and still come across errors - which is something we should avoid from happening at all costs.

Chavithra (approved):
Looks ok to me.

PR #3263, opened 2022-11-02
Formatting, first image is not given a new line.

deeleeramone:
Attempting to resolve this:

Screenshot 2022-11-02 at 12 45 15 AM

Proposed solution is an additional line between content and embedded image.


JerBouma (approved):

JerBouma:
Interesting issue, only happens on a 1080p screen and not 4k. This should fix it though for both.

**1080p**
image

**4k**
image

Issue #3128, opened 2022-10-27
[Bug] Autocomplete


jmaslek:
@DidierRLopes why this no work?
in forex
type:
load E
-- nothing pops up from the list 

type load -t E and it pops up

jmaslek:
happens in `sia/clear m`

Chavithra:
historical --help doesn't work unless --ticker is provided.

DidierRLopes:
It's working for me. But when you start typing with the argument, the list of options disappears

<img width="437" alt="Screenshot 2022-11-01 at 09 17 44" src="https://user-images.githubusercontent.com/25267873/199202887-447a622f-c1fe-4f75-8628-b0aa317f734c.png">

<img width="427" alt="Screenshot 2022-11-01 at 09 17 38" src="https://user-images.githubusercontent.com/25267873/199202893-24b7c939-f45d-45ba-92aa-0da0c213cf3b.png">

<img width="420" alt="Screenshot 2022-11-01 at 09 17 31" src="https://user-images.githubusercontent.com/25267873/199202894-b7e3dd2c-b3bb-4d46-83ea-2d6b9a4ccfa3.png">

I get the same on curve, I think this may be a hard thing to fix. But I can look into it!

andrewkenreich:
go to crypto/ov

and do cr borrow

and select it the first time - it works fine then do cr again and it works fine

but now...do cr borrow - and dont select it and type our borrow fully

then do cr again

and nothing pops up


This happens across the full terminal I believe whenever you type out the auto complete instead of selecting

PR #3274, opened 2022-11-02
Reset the auto-complete options in custom prompt

jmaslek:
This is definitely not the hard work of @andrewkenreich spending hours debugging prompt toolkit :)

This seems to address some of the issues in #3128


DidierRLopes (commented):
The funny thing about this code is that this is a 4 line code change and it took probably as long as it took to do that PR of fixing a lot of bugs in crypto 🤣

andrewkenreich:
> The funny thing about this code is that this is a 4 line code change and it took probably as long as it took to do that PR of fixing a lot of bugs in crypto 🤣

Lol so true 🤣🤣

DidierRLopes (approved):
LGTM  🚀

Nice work! 🦋

PR #3275, opened 2022-11-02
fix(stocks): --show-all accessing out of range index on vis (#3249)

### LINKED ISSUE #3249: [Bug] Stocks/SIA - `vis --show-all`: Error: list index out of range
```
2022 Nov 01, 10:39 (🦋) /stocks/sia/ $ select MS

2022 Nov 01, 10:39 (🦋) /stocks/sia/ $ vis --show-all

The following args couldn't be interpreted: ['--show-all']
If it takes too long, you can use 'Ctrl + C' to cancel.

Error: list index out of range

Error: not enough values to unpack (expected 2, got 0)
```

Screenshot 2022-11-01 at 10 40 28 AM

gsegatti:
- Fixes #3249 by @deeleeramone.

**The Problem:**
- Having no valid arguments caused `ns_parser` to have no `finance_key`.
- This eventually caused `display_plots_financials()` to access [`used_statement[0]`](https://github.com/OpenBB-finance/OpenBBTerminal/blob/1f9a771101de0b5bd77e19919c11e0f2f7bbfb91/openbb_terminal/stocks/sector_industry_analysis/stockanalysis_view.py#L95).
- used_statement was empty since there's no `finance_key`.

**Possible Solution:**
- Like @jmaslek mentioned [here](https://github.com/OpenBB-finance/OpenBBTerminal/pull/3275#issuecomment-1302297401), removing the autocomplete option for `--show-all` and its check on `parser.add_argument` was enough.

Current output with this PR:
image


openbb_terminal/stocks/sector_industry_analysis/sia_controller.py
```diff
@@ -213,7 +213,6 @@ def __init__(
                 "--currency": None,
                 "-c": "--currency",
                 "--raw": {},
-                "--show-all": {},
```
gsegatti, line 216: I'm actually not sure whether this should be removed or not. Based on [this comment](https://github.com/OpenBB-finance/OpenBBTerminal/issues/3249#issuecomment-1299038630) it kinda seems like it should?

Appreciate any clarifications.

gsegatti:
Hm, seems like many other tests use the `parent_classes.py` so maybe this solution isn't feasible.

openbb_terminal/stocks/sector_industry_analysis/sia_controller.py
jmaslek: This should not be here. I added it then took it away from argparse but not autocomplete

jmaslek:
Does just removing the auto complete show all fix this bug?  Running the function with no args should show all possible values so adding a check may get rid of that?

gsegatti:
I have to double check, but if the autocomplete is removed and someone still types `--show-all` we'll have the same error.
Unless we're assuming people won't do that?

jmaslek:
--show-all isn't a valid flag so I should get the same behavior as if I type in --jejdjsjxjrjsjjahdiwjshdh.

gsegatti:
Yeah, checked it and we still get the same error message by just removing the autocomplete. 
`The following args couldn't be interpreted: ['--show-all']
If it takes too long, you can use 'Ctrl + C' to cancel.

Error: list index out of range

Error: not enough values to unpack (expected 2, got 0)`

gsegatti:
hmmm, interesting, I had seen this but didn't really think it could be the cause, will look into it.

jmaslek (approved):

PR #3289, opened 2022-11-03
SDK : faster and simpler architecture

Chavithra:
# Description

Currently the SDK contains different features:
- shortcuts
- selection between model and view : according to the chart=True/False parameter
- documentation of the module
- documentation of the functions
- logging

All these parts are entangled, which makes it hard to check/change one specific part.

**GOAL**
This PR is meant to make code more team work friendly.

**SIDE EFFECT**
The loading time of the SDK is reduced.


piiq:
Is this mergeable?

P.S. How do we import the sdk after this change?

hjoaquim:
> 

It would be like this:

```python
from openbb_terminal.core.library.sdk import openbb
```

jmaslek:
Why are we making this much harder to import for an end user?  From openbb.sdk is fine, but now needing openbb.core.library.sdk is not.

Chavithra:
> Why are we making this much harder to import for an end user? From openbb.sdk is fine, but now needing openbb.core.library.sdk is not.

Yes this was just for the drafting time.

Changed it to `openbb.sdk`

hjoaquim (approved):
Looks good to me!

hjoaquim:
@montezdesousa @jmaslek this is ready imo, but since I was also pushing some commits, would be cool if one of you took a look at this! Tks!

hjoaquim:
Also, @jmaslek we're keeping the `from openbb_terminal.sdk import openbb` so there are no changes from the user's perspective!

PR #3292, opened 2022-11-03
Get rid of funds menu....for now

jmaslek:
As the title says.  Since the investpy project is still waiting to hear back from investing.com, I want to deprecate this menu for the time being.

I believe most of the functionality can probably be replaced with yahoo finance, but that will require development.


Chavithra (changes requested):
We need to generate the:
- sdk yml files in the `website` folder
- html page in the 

We might want to automate these thing.

jmaslek:
> I thing you now need to generate the:
> 
> * sdk yml files in the `website` folder
> * html page in the
> 
> We might want to automate these thing.

Alrighty.  I got rid of all the html pages and regenerated all the sdk files.

Chavithra (approved):
Looks ok to me.

PR #3298, opened 2022-11-04
Adds Intro Guide for the Stock Options SDK Module

deeleeramone:


website/content/SDK/stocks/options/_index.md
```diff
@@ -1,6 +1,99 @@
...
+options_df.set_index(keys = 'symbol', inplace = True)
+
+options_df
+```
+
+Depending on the depth of options available, compiling the data may take upwards of thirty seconds.
+
+![DataFrame With all Expirations](https://user-images.githubusercontent.com/85772166/199893166-35ef062d-c16e-464d-a392-378ef4c6a1ee.png "DataFrame With all Expirations")
+
+The data can then be cleaned for better presentation, indexing, or analysis; for example:
```
JerBouma, line 84: Let's get rid of the section down below, we show SDK functionality and not how data can be cleaned. If this step was mandatory, then it should have been in the SDK in the first place.

JerBouma (changes requested):
@deeleeramone Please put your attention to the higher level menus first, e.g. Economy, Forex, Econometrics, Portfolio. We can think of these sub-menus at a later stage.

In any case, have some suggestions and then we can still merge this one.

deeleeramone:
I would consider the Options menu as a high-level menu.

JerBouma (changes requested):
Minor minor detail

JerBouma (approved):

PR #3299, opened 2022-11-04
Refactors Econometrics and Forecast load commands and controller

### LINKED ISSUE #3044: [Bug] Forecast & Econometrics menu - Columns with `/` in the header trigger a new command
This becomes apparent when attempting to rename a column. To recreate, in either menu, load any data set and then rename a column to include a `/` in it. For example, `rename -c SPY -o adj_close -n adj/close`

colin99d:
# Description

Fixes #3044

This PR refactors econometrics and forecast to use the same load_model. It also removes the separate definitions of file types inside each controller. This PR then improved the load function by making it no longer need a file_types argument, and adding error handling around file loading. Lastly, it automatically changes `/` to `_` in df column names. 

Also fixes a bug when trying to run `load --examples` in the `econometrics` menu.

Also changes some parameter names from df to data.

@martinb-bb for right now I am l leaving in Jeroen's example datasets but not explicitly letting users know they are there. I could add an --examples command like @JerBouma has to show users the examples, however a lot of the datasets are rather small (30-100) points. I see the following four ways to proceed, let me know which you prefer (all of these are very easy to do):
1. Remove sample datasets from load functionality
2. Keep sample datasets but give no users indication they exist (this is what we have now, if you do `load cancer`, you get a cancer dataset, but you have no way of knowing that)
3. Keep sample datasets, and only show ones with enough data (add n `--examples` argument that shows all datasets with >100 rows`) (not to bias you but I think this one makes the most sense)
4. Keep sample datasets, show all datasets 

Here are the sample datasets users would have access to:
Screenshot 2022-11-04 at 8 52 50 AM

# How has this been tested?
This refactor centers around the load function, so I tested it a lot.
    - Load csvs into forecast and econometrics with and without the `/` character in the header. If the `/` was in the header I ensured that the load function replaced it with a `_`.
    - `openbb.forecast.load("sample.csv")`
    - `openbb.forecast.econometrics("sample.csv")`
    - `python terminal.py reports/forecast`

# Checklist:

# Others


JerBouma:
These examples are datasets purposely designed for Econometrics. E.g. Longley's dataset is a mere example to show the accuracy of least squares (you can get an R-squared of 1.0). They serve little purpose for a Forecasting menu in my opinion but are relevant and useful for Econometrics.

colin99d:
Ok, I removed them.

colin99d:
Found a bug when trying to show outputs for timeseries without dates. Will remove do not merge once @martinb-bb and I decide how to fix it.

PR #3316, opened 2022-11-04
Adding some test coverage

jmaslek:
Not sure how far I will get before I head out but will update here

- Our formatting tests were failing because futures called `tickers` instead of `symbols` so I changed that
- Adding unit tests on economy macro -- which will help with making sure the csv countries are kept (note economy controller coverage was 48%). is now 54.
- Added unit tests on fred from controller
- Unit test on call index.  Up to 65%
- Unit test on call_treasury.  72% coverage on controller


andrewkenreich (approved):
wow what a coder - this guy is a UNIT

PR #3322, opened 2022-11-05
Removing autocomplete for all limit

### LINKED ISSUE #3253: [IMPROVE]. Get rid of auto complete for limit everywhere
Having 1,2,3,4,5 pop up is more distracting than helpful and every single limit function should have the same input

simmonsj330:
# Description
Fixes #3253 by removing limit autocompletion across the entire terminal.

# Others


openbb_terminal/alternative/covid/covid_controller.py
```diff
@@ -49,22 +49,22 @@ def __init__(self, queue: List[str] = None):
...
             choices["ov"] = {
                 "--raw": {},
-                "--limit": {str(c): {} for c in range(1, 1000)},
+                "--limit": None,
                 "-l": "--limit",
             }
             choices["rates"] = {
                 "--raw": {},
-                "--limit": {str(c): {} for c in range(1, 1000)},
+                "--limit": None,
```
jmaslek, line 57: I think these all need to be "--limit":{}

simmonsj330: @jmaslek whats the difference?

jmaslek: actually I think it does not need to be that?  I seethe line above we do {} instead of None so I wasnt completely sure.

jmaslek (approved):
Fix merge conflict then we good 👍

PR #3371, opened 2022-11-10
Fix SDK website documentation script

montezdesousa:
# Description

This PR fixes https://github.com/OpenBB-finance/OpenBBTerminal/issues/3376 -> `generate.py`. Also removed the italic format from `bool` in `chart` argument.

Disclaimer: there are functions with wrong indentation, that will be fixed by removing datetimes here: https://github.com/OpenBB-finance/OpenBBTerminal/pull/3354

Some signatures are being regenerated every time we run the script bc we are using time dependent defaults like `.now()`, no bug introduced though.


Chavithra (approved):
Looks good to me.

PR #3378, opened 2022-11-10
Remove mockings

hjoaquim:
This PR aims to remove the function `openbb_terminal/core/log/generation/formatter_with_exceptions.mock_break_lines`.

In the way logging is working at the moment, there is no real benefit on having this function, since the logs that depend on the user input are already addressed by:

`openbb_terminal/parent_classes.log_cmd_and_queue` and `openbb_terminal/core/library/operation.__log_method_info` --> using `json.dumps` ensure the string is in a valid json format.

Also, by changing the `\`s in an already valid json string we take the risk of making the json string invalid thus, eventually affecting the ETL pipeline or the queries - that expect a well-formed json string.


openbb_terminal/core/log/generation/formatter_with_exceptions.py
```diff
@@ -67,20 +67,9 @@ def mock_home_directory(text: str) -> str:
...
-        text_mocked = text_mocked.replace("\r", " MOCKING_CARRIAGE_RETURN ")
-        text_mocked = text_mocked.replace("\t", " MOCKING_TAB ")
-        text_mocked = text_mocked.replace("\\n", " MOCKING_BREAKLINE ")
-        text_mocked = text_mocked.replace("\\r", " MOCKING_CARRIAGE_RETURN ")
-
-        return text_mocked
-
     @classmethod
     def filter_special_tags(cls, text: str) -> str:
-        text_filtered = cls.mock_break_lines(text=text)
```
Chavithra, line 82: I think we need to replace that line by:
```python
text_mocked = text.replace("\n", " MOCKING_BREAKLINE "
```

Chavithra (approved):
Looks ok except for the line that should be added.

PR #3379, opened 2022-11-10
Refactoring econometrics for SDK usability

### LINKED ISSUE #3323: [Bug] SDK - `openbb.econometrics.granger` - Error: get_granger_causality() missing 1 required positional argument: 'lags'
According to docstring, this value is set by default to 3.

Screenshot 2022-11-05 at 3 19 21 PM

### LINKED ISSUE #3366: [Bug] SDK - `openbb.econometrics.norm` should produce a chart
The following example produces a table while it should be producing a chart.

```python
data = openbb.econometrics.load("longley")
openbb.econometrics.norm(data['TOTEMP'], chart=True)
```

Returns:
image

### LINKED ISSUE #3367: [Bug] SDK - `openbb.econometrics.ols` requires dictionary datasets instead of DataFrames
Datasets need to be loaded directly into a dictionary to work with `ols`. Otherwise it returns an error, e.g.

```python
datasets = {"longley": openbb.econometrics.load("longley"), "macrodata": openbb.econometrics.load("macrodata")}
```

It doesn't work if I do the following:

````python
data = openbb.econometrics.load("longley")
openbb.econometrics.ols(["TOTEMP", "GNPDEFL"], data=data)
````

image

Whereas this does:

image

### LINKED ISSUE #3368: [Bug] SDK - `openbb.econometrics.ols` displays more objects than feasible
When I wish to show regression results, it shows me much more than only the regression table.

image

It also still shows the table if I assign variables:

image

### LINKED ISSUE #3369: [IMPROVE] SDK - `openbb.econometrics.ols` input can be much nicer
How you enter an ols regression is not intuitive. Ideally what you would want to do is the same pattern as Statsmodels uses which is `EXECUTIONS ~ SOUTH + INCOME`. What we currently have to do is the following:

image

Instead of Statmodels version:

image

### LINKED ISSUE #3370: [Bug] SDK - `openbb.econometrics.granger` returns a messy output
This leaves the user completely confused what he is looking at while it could be a nice table (or better, a DataFrame) like the OpenBB Terminal has.

**OpenBB SDK**
image

**OpenBB Terminal**
image

### LINKED ISSUE #3372: [Bug] SDK - `openbb.econometrics.dwat`, `openbb.econometrics.bgod` and `openbb.econometrics.bpag` can't be used
These functionalities require the residual of an OLS regression. That residual is hidden within the `ols` regression functionality and it is tough to make this work.

image

Tricky for the user to understand this needs to be done:

image

And for `bpag` and `bgod` it still doesn't work:

image

Also, the tables for all of these functionalities are missing. See:

image

### LINKED ISSUE #3373: [Bug] SDK - `openbb.econometrics.coint` is messy and you can't plot anything
This is what is being returned:

image

As opposed to the OpenBB Terminal:

image

There is also nothing happening when I try to plot.

image

### LINKED ISSUE #3374: [Bug] SDK - `openbb.econometrics.panel` shows the same thing twice
The output of the Panel regression is messy and the same result (the table with regression results) is outputted twice.

image

### PR DESCRIPTION
**Pull request by jmaslek**
The econometrics menu is nice, but the SDK usage is not friendly because how we do it in the terminal.  This menu needs some refactoring so that the SDK is usable.  Still love you @JerBouma.

Things done:

- Refactor OLS so that the SDK can take in dataframes instead of a dictionary of dataframes:  This will do fixes #3368 fixes #3367 fixes #3369
Screen Shot 2022-11-10 at 2 59 54 PM

- Fixes #3366 by making chart=True by default
Screen Shot 2022-11-10 at 4 05 41 PM

- Fixes #3373 by refactoring the coint function to return a df only.  I also changed it to allow a variable number of inputs in the sdk and allowing for the plot to show up when doing chart=True

Screen Shot 2022-11-10 at 5 19 52 PM

- Fixes #3370 and fixes #3323 for the granger SDK function
Screen Shot 2022-11-11 at 9 20 37 AM

- Refactor dwat function to take in an OLS model instead of the residuals.  I think this cleans up.  Made a note about the view in the model docstring
```
    Returns
    -------
    float
        Test statistic of the Durbin Watson test.

    Notes
    ------
    When using chart = True, the dependent variable in the regression must be passed to view the residuals
````
Screen Shot 2022-11-11 at 11 11 45 AM

- Complete the fixes #3372 -- sdk functions now input models
Screen Shot 2022-11-11 at 11 46 00 AM

- Refactor panel.  Fixes #3374 
Screen Shot 2022-11-13 at 1 36 03 PM


jmaslek:
There are a couple functions that need some cleaning up for the sdk, but if anyone wants to review this overnight feel free :)

JerBouma:
> The econometrics menu is nice, **bla bla bla** love you @JerBouma.

Thank you <3

jmaslek:
I just have `panel` left

JerBouma:
Make sure by the way that if you do `chart=True` (or anything else) that it doesn't print the richt table. Those should always be DataFrames.

jmaslek:
> Make sure by the way that if you do `chart=True` (or anything else) that it doesn't print the richt table. Those should always be DataFrames.

Ummm that's the desired behavior to print the nice charts when chart = True.  Doing otherwise would be a whole refactor.

JerBouma:
> > Make sure by the way that if you do `chart=True` (or anything else) that it doesn't print the richt table. Those should always be DataFrames.
> 
> Ummm that's the desired behavior to print the nice charts when chart = True. Doing otherwise would be a whole refactor.

Just making sure we don't mean different things, I mean that if you use `chart=True` I don't want to see the table too, see:

image

Whereas this screenshot also shows the Rich table. That looks wrong.

image

JerBouma:
Ok ok

JerBouma:
> Ideally we would have the `raw=True` when we want to output the table. Is this not possible? @jmaslek @JerBouma
> 
> Spitting both is very suboptimal, imo. Would be better to just spit the chart than both

The rich table shouldn't be an option in the first place, you can rely on the default output (without `chart=True`) which is a DataFrame object. But if this requires too big of an architectural change it isn't too big of a deal for now.

jmaslek:
> Ideally we would have the `raw=True` when we want to output the table. Is this not possible? @jmaslek @JerBouma
> 
> Spitting both is very suboptimal, imo. Would be better to just spit the chart than both

I think this gets into the question of the default output.  In this menu its primarily tables so we want raw=True by default.  You can actually pass plot=False in these cases.  So I think its a matter of what the default output is

jmaslek:
> The rich table shouldn't be an option in the first place, you can rely on the default output (without `chart=True`) which is a DataFrame object. But if this requires too big of an architectural change it isn't too big of a deal for now.

If you are in a script, you wouldnt get the default output and could instead see a table with view. So if you want the default output, dont use the _view.

PR #3419, opened 2022-11-14
Change common trails to no trails

jmaslek:
Instead of the confusing notation of `openbb.common.ta.cmd`, I removed the leading trail so that we can do `openbb.ta.cmd`


Chavithra (approved):
Looks ok to me.

PR #3421, opened 2022-11-15
Fixing `exit` command on forecast menu

### LINKED ISSUE #3415: [Bug]`Exit` does not exit from the terminal while in forecast menu
**Describe the bug**
When the user is in the `forecast` menu, they have to use the `exit` command twice before being able to exit properly out of the terminal itself.

Screen Shot 2022-11-14 at 10 48 22 AM

**To Reproduce**
`stocks/load aapl/forecast/exit/exit`

- OS: Mac Monterey 12.3
- Python version: 3.9.6

simmonsj330:
# Description

Fixes #3415.

The forecast menu exhibits a bug where if the user uses the `exit` command, they are directed back to the base menu. This occurs only when loading in data from another menu like stocks with the following commands `stocks/load aapl/forecast`. However, this error does not occur if they user navigates to the forecast menu first to load in their own dataset. Additionally, this error does not occur if the user inputs a ticker from another menu, navigates to the base menu, and then navigates to the forecast menu.

## Source of the bug
This error occurs due to the number of times `quit` is called relative to the menu path. The `exit` command is used to leave the entire application and the backend of the exit command is simply X number of quits where X represents how "deep" in the terminal the user is, i.e what menu or sub-menu. For example, if the user is at the `crypto/dd` menu, then they are 3 deep: base->crypto->dd. 
When applying this to forecasting where the user loads the ticker `aapl` and *then* calls forecasting, it is synonymous to being 3 menus deep because the user is going from base->stocks->forecasting. This seems to be true because if the user uses the `quit` command, they will be directed back to the `stocks` menu, and then if they use it twice more, they will be directed back to the base menu and then out of the terminal.
However, the `exit` command directs the user back only to the base menu. This means that there is one missing `quit` call. This occurs because forecasting is a primary menu option off of the base menu and its path is `forecast/`, (base->forecast). The `exit` function utilizes this path and thus only calls `quit` twice. 

## Solution
The solution is to simply increase the total number of quits used when exiting. This will cover the forecast edge case when there are not enough quits. However, this means that all other `exit` calls will have one additional, unneeded `quit`. This does not seem to cause any errors or lead to adverse functionality.

Screen Shot 2022-11-14 at 7 42 16 PM

# Others


jmaslek (approved):
works for me both from /forecast and /stocks/forecast

martinb-bb (approved):
works for me too

PR #3432, opened 2022-11-15
Move portfolio optimization to extras

piiq:
This PR creates an optional extra out of the po menu. This will enable the terminal and the SDK to be pip-installable on most of the platforms out-of-the-box (no reliance on the presence of CMake and specific conda packages)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ openbb = 'openbb_terminal.terminal_controller:parse_args_and_run'
 
 [tool.poetry.dependencies]
 python = "^3.8,<3.11, !=3.9.7"
-scipy = "1.8.1"
+scipy = "^1.8.1"
```
simmonsj330, line 20: Looks good to me! I've made an [installer](https://github.com/OpenBB-finance/OpenBBTerminal/actions/runs/3500392516) that runs successfully. This particular line installs `scipy 1.9.3` which is needed to subvert another `libgfortran.5.dylib` error.

simmonsj330 (approved):

PR #3444, opened 2022-11-17
Fix reports after SDK changes

jmaslek:
Title.  

- Equity Report
- Forex Report
- Crypto Report
- ETF Report


Chavithra (approved):
Looks to to me.

PR #3445, opened 2022-11-17
Filtering duplicate logs

hjoaquim:
This PR aims to avoid duplicate logs on the SDK.

A good point raised by @jose-donato is that a user can actually use the SDK endpoints inside a `while True` generating way to many entries on the logs.

Solution:
   - Adding a comparison on "current operation" with "last operation" (arguments included). If that's the case, the logging should be suppressed for that call.

This PR also:
 - Removes logging from `pyth_model.get_price` as it is used inside a `while True`
 - Forces the maximum log file size to 2MB in order to be sent to S3.


Chavithra (approved):
Looks ok to me.

Let's just move this `trail_stack` into `OperationLogger`.

jose-donato:
Wouldn't we still be interested in tracking if users are calling this command though?

Chavithra:
> 

If they are using the command we will know through the controller call log.

And on the SDK we have the `trail`.

PR #3451, opened 2022-11-18
Removing `libgfortran.5.dylib` by upgrading `scipy`

simmonsj330:
# Description
There has been thorn in the build system's side for ages and this PR finally fixes it. Scipy 1.8.1 packaged `libgfortran.5.dylib` within it's wheels and this lead to errors when bundling/packaging the terminal because it was in ARM64 architecture. A variety of binaries reference this dynamic library for some reason. We previously 'fixed' the issue by upgrading certain libraries, such as `numpy` that have a binary which references `libgfortran.5.dylib`. After upgrading `numpy` last time, we were able to build installers off of the automated build system. However, we recently discovered that issue was persistant when building installers *locally*.

Screen Shot 2022-11-17 at 4 35 52 PM

In order to address this issue fully, we upgrading `scipy`, which was the source of `libgfortran.5.dylib`, because `scipy 1.9.3` does not package this file within their wheel, as shown below:
Screen Shot 2022-11-17 at 7 34 18 PM

# Others

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -77,7 +77,7 @@ pandas = ">=0.18"
 toolz = "*"
 
 [package.extras]
-dev = ["black", "docutils", "flake8", "ipython", "m2r", "mistune (<2.0.0)", "pytest", "recommonmark", "sphinx", "vega-datasets"]
+dev = ["black", "docutils", "ipython", "flake8", "pytest", "sphinx", "mistune (<2.0.0)", "m2r", "vega-datasets", "recommonmark"]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


piiq:
Thanks @simmonsj330 
This should be covered by the larger dependency related PR #3432 
The dependencies changed to provide a range for scipy so that the tree can be solved
image

Can you test the build from that branch and tell me if it works? If it does we'll close this PR in favour of the other one

simmonsj330:
Closing because this is addressed in #3432.

PR #3489, opened 2022-11-19
Add a single unit test to check sdk import

jmaslek:
This just adds a quick check to make sure the import works.  Will avoid issues in trail map with commas.


simmonsj330 (approved):
lgtm!

PR #3522, opened 2022-11-22
Suppress logs for dashboards

hjoaquim:
This PR aims to add log suppression to the dashboards, similarly to what happens in the reports.

Currently, there is no suppression on the `dashboards_controller` what leads to having the impression of a sdk "sub_app" running while using e.g. an installer version of the terminal, example:

image

How to replicate/test:

1. On main: 
   - Open the terminal and go to `dashboards`
   - Run `crypto -n`
   - Check your logging directory and see that 4 files were generated (wrongly - as it was supposed to be only 2). Open the `.log` files and see `"LOGGING_SUB_APP": "sdk"` in the configs.

2. Fixed branch:
   - Repeat the process - now you should only have 2 files and see `"LOGGING_SUB_APP": "terminal"`


Chavithra (approved):
Looks ok to me.

PR #3538, opened 2022-11-23
Fix linter issue

jmaslek:
We missed a lint rip

Gonna look into the deploy issue real quick

- Looks like it was not doing poetry -E optimization


Chavithra (approved):
Looks ok.

PR #3564, opened 2022-11-24
Forecast docstring fixes and other things

martinb-bb:
# Description
- FIxes docstrings


deeleeramone:
Models featuring past-covariates are missing the shortcut for `all-past-covariates`. It is simpler and faster to use when the dataframe is constructed via scripting or routines.

deeleeramone:
The raw output for models would ideally be served as a dataframe instead of a console print, eliminating the steps of creating a file via exporting, then re-importing 

Screenshot 2022-11-24 at 1 40 05 PM

colin99d (approved):

PR #3579, opened 2022-11-25
Enforce logging to use `str`

hjoaquim:
As reported by @DidierRLopes :

image

Despite, I was not able to reproduce this behavior, I can see why this could be an issue.

In this PR I suggest we explicitly enforce the `last_method` to be a `str` (with a max of 100 chars).


Chavithra (approved):
Looks ok to me.

PR #3589, opened 2022-11-25
Etf bug fix

jmaslek:
Changed to get the hardcoded string from their etf list, which was breaking.

Fixes #3851


jmaslek:
> stockanalysis.com changed their site.......again
> 
> Managed to get the hardcoded string from their etf list, which was breaking.
> 
> This is a bandaid while we replace the source

Fixes #3851

Also adds some docstring examples

DidierRLopes (approved):

PR #3600, opened 2022-11-26
Remove colors from dataframes when saved

DidierRLopes:
Before | After

Screenshot 2022-11-26 at 20 15 19

Also solves the issue to save in json, since we need to reset_index to avoid issues.


jmaslek (approved):
Do we need more colors though?

DidierRLopes:
> Do we need more colors though?

Are there more that we use on tables? I think we may actually use magenta too in insider table, but that should be about it

PR #3603, opened 2022-11-26
Add parameter to automatically color a table based on positive and negative numbers

DidierRLopes:
Before:
image

After:
image

ALSO ALLOWS TO SELECT COLUMNS AND ROWS TO LOOK AT SPECIFICALLY:
image

EXTRA:
* Also fixed a bug in command: `stocks/ins/act` - which has been ported already!


ruimaia:
Maybe you could use regex to check for numbers in a string (see [`link`](https://pythonexamples.org/python-regex-extract-find-all-the-numbers-in-string/#:~:text=Python%20Regex%20%E2%80%93%20Get%20List%20of,single%20digit%20in%20the%20string.)). I guess this would require way less lines of code.

Additionally, you could even have only one function that would do everything at once by simply returning the desired color (i.e. green, red, yellow or grey) based on three simple regex conditions to check for positive/negative/zero numbers.

Let me know if this makes sense to you and/or if you need any help.

DidierRLopes:
> Maybe you could use regex to check for numbers in a string (see [`link`](https://pythonexamples.org/python-regex-extract-find-all-the-numbers-in-string/#:~:text=Python%20Regex%20%E2%80%93%20Get%20List%20of,single%20digit%20in%20the%20string.)). I guess this would require way less lines of code.
> 
> Additionally, you could even have only one function that would do everything at once by simply returning the desired color (i.e. green, red, yellow or grey) based on three simple regex conditions to check for positive/negative/zero numbers.
> 
> Let me know if this makes sense to you and/or if you need any help.



> Maybe you could use regex to check for numbers in a string (see [`link`](https://pythonexamples.org/python-regex-extract-find-all-the-numbers-in-string/#:~:text=Python%20Regex%20%E2%80%93%20Get%20List%20of,single%20digit%20in%20the%20string.)). I guess this would require way less lines of code.
> 
> Additionally, you could even have only one function that would do everything at once by simply returning the desired color (i.e. green, red, yellow or grey) based on three simple regex conditions to check for positive/negative/zero numbers.
> 
> Let me know if this makes sense to you and/or if you need any help.

That is a good idea re regex!! Would definitely avoid the if/else and the try/catch I'd say.

I didn't understand the 2nd part regarding returning the colors. How would you put that into applymap?

ruimaia:
> I didn't understand the 2nd part regarding returning the colors. How would you put that into applymap?

Pseudo-code:

```python
def return_colored_value(value: str):
    if re.findall(positive_number_regex, value):
        return  f"[green]{value}[/green]"

    elif re.findall(negative_number_regex, value):
        return  f"[red]{value}[/red]"

    elif re.findall(zero_number_regex, value):
        return  f"[yellow]{value}[/yellow]"

    else:
        return  f"[gray]{value}[/gray]"
```

I didn't bother with the definition of the regular expressions but I guess you can get the gist of it.

Then you could simply have:
```python
 if obbff.USE_COLOR and automatic_coloring:
            df = df.applymap(lambda x:  return_colored_value(x))
```

DidierRLopes:
> > I didn't understand the 2nd part regarding returning the colors. How would you put that into applymap?
> 
> Something along the lines of:
> 
> ```python
> def return_colored_value(value: str):
>     if re.findall(positive_number_regex, value):
>         return  f"[green]{value}[/green]"
> 
>     elif re.findall(negative_number_regex, value):
>         return  f"[red]{value}[/red]"
> 
>     elif re.findall(zero_number_regex, value):
>         return  f"[yellow]{value}[/yellow]"
> 
>     else:
>         return  f"[gray]{value}[/gray]"
> ```
> 
> I didn't bother with the definition of the regular expressions but I guess you can get the gist of it.
> 
> Then you could simply have:
> 
> ```python
>  if obbff.USE_COLOR and automatic_coloring:
>      df = df.applymap(lambda x:  return_colored_value(x))
> ```

Love it. Yes, that should work fine.

Am on it 🙏🏽

DidierRLopes:
> > > I didn't understand the 2nd part regarding returning the colors. How would you put that into applymap?
> > 
> > 
> > Something along the lines of:
> > ```python
> > def return_colored_value(value: str):
> >     if re.findall(positive_number_regex, value):
> >         return  f"[green]{value}[/green]"
> > 
> >     elif re.findall(negative_number_regex, value):
> >         return  f"[red]{value}[/red]"
> > 
> >     elif re.findall(zero_number_regex, value):
> >         return  f"[yellow]{value}[/yellow]"
> > 
> >     else:
> >         return  f"[gray]{value}[/gray]"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I didn't bother with the definition of the regular expressions but I guess you can get the gist of it.
> > Then you could simply have:
> > ```python
> >  if obbff.USE_COLOR and automatic_coloring:
> >      df = df.applymap(lambda x:  return_colored_value(x))
> > ```
> 
> Love it. Yes, that should work fine.
> 
> Am on it 🙏🏽

Much better, thank you @ruimaia 🫡

Screenshot 2022-11-27 at 09 35 37

jmaslek (changes requested):
There are no new unit tests that make sure the coloring works and prints out the expected results.

DidierRLopes:
> There are no new unit tests that make sure the coloring works and prints out the expected results.

That is normal, because there are no flags set to True.

This PR adds the functionality, but it doesn't set it on. Please have the team discuss in which commands this would make sense. 

I can set it to true for a specific command, for the sake of creating a unit test. Just let me know how to proceed

DidierRLopes:
@Chavithra @jmaslek need help with the tests, for some reason the coloring doesn't show up. Am wondering if am doing the mocking correctly

jmaslek:
> @Chavithra @jmaslek need help with the tests, for some reason the coloring doesn't show up. Am wondering if am doing the mocking correctly

Something weird is happening.  The print_rich_table isnt actually being called.  If I just put 1/0 in the first line of the function, it still passes when it should raise an error.  I switched the mocker to a monkeypatch for good measure but I am confuzzled.

jmaslek:
@Chavithra I remember.  We mock the console so that the print is just a normal print because of something related with the table.  So how can we test this does what we expect?

DidierRLopes:
@jmaslek removed the tests as per your request!

hjoaquim:
@jmaslek I saw you just added the `# type:ignore` a couple hours ago, but It was alright without it locally - if it breaks on the CI I'll add it again!

jmaslek:
> @jmaslek I saw you just added the `# type:ignore` a couple of hours ago, but It was alright without it locally - if it breaks on the CI I'll add it again!

Yup.  No idea what was causing it, since locally was fine.  This was the only solution I could find.

jmaslek (approved):

hjoaquim:
@jmaslek @DidierRLopes looks like `Legacy.Version` is being deprecated in packaging 22.0 --> https://github.com/pypa/packaging/issues/631

On a fresh conda environment I have this:
```
(obb) ➜  OpenBBTerminal git:(colortable) pip list|grep packaging
packaging                     21.3
```
As long as we're using packaging 21.3 this should not be an issue.

Should we add this?:
poetry add packaging="^21.0"

Issue #3521, opened 2022-11-22
[Bug] Exit needed twice for `stocks/scr/technical/ca`


simmonsj330:
**Describe the bug**
Running the commands `stocks/scr/technical/ca` in that order directs the user to the `ca` menu, but the user needs to use the `exit` command twice to exit the terminal.
<img width="895" alt="Screen Shot 2022-11-22 at 2 28 10 AM" src="https://user-images.githubusercontent.com/53658028/203263514-60831bf5-b8f3-414c-8543-04de1ef50dd5.png">

PR #3607, opened 2022-11-27
Double exit bug fix on `ca` menu

simmonsj330:
# Description
Temporary fix to #3521. This is the same error that occurs on the forecast menu which was addressed with #3421. This PR implements the same fix on the ca controller by overwriting the exit function. This is only a temporary fix until a better way to handle such issues can be implemented after 2.0. A potential solution would be to track the path of menus taken.

Screen Shot 2022-11-27 at 1 09 08 AM

# Others


jmaslek (approved):
Probably need to adjust the ca controller tests

simmonsj330:
> Probably need to adjust the ca controller tests
Fixed with latest commit.
Screen Shot 2022-11-27 at 2 22 02 PM

PR #3619, opened 2022-11-28
Fix portfolio optimization dates parameters bug

montezdesousa:
# Description
Fixes bug when wrong date format was injected through `file --file`.

2 cases were failing:
* when user filled `start_period` or `end_period` in the parameters excel template, excel was converting the date to datetime instead of string, crashing the terminal afterwards
* when user filled `start_period` or `end_period` in `.ini` or `.xlsx` files with random values not with `YYYY-MM-DD` format, crashed the terminal

Fix: validate `start_period` or `end_period` format and pop parameters if random bad input or cast `datetime` to `str` if valid date provided

Test:
* Attribute different values to `start_period` and `end_period` in the parameters files
* Load the file changed with `file --file`
* Load some tickers with `load --file`
* Perform some optimization like `maxutil`


jmaslek (approved):

deeleeramone:
Still crashing. Prompt Toolkit seems to be the culprit this time.


https://user-images.githubusercontent.com/85772166/204317429-817273e0-caa8-470a-8792-eb1fdd27460d.mov

deeleeramone:
In the `params` submenu, you are not able to load XLSX files.

Screenshot 2022-11-28 at 7 41 37 AM

deeleeramone:
Removing `start_period` created a new error: `Error: 'numpy.float64' object has no attribute 'to_dict'`

With  start_period = 1995-01-01

Screenshot 2022-11-28 at 7 49 14 AM

With no start_period defined:

Screenshot 2022-11-28 at 7 48 07 AM

deeleeramone:
It's somewhat annoying that Excel wants to convert input values to a %, and then round the decimals out so that you are not sure if the values entered were accepted as intended.

PR #3629, opened 2022-11-28
Fix Docs Auto Generation

tehcoderer:
#3610 fix for terminal docs generation


Chavithra:
@tehcoderer is it possible to explain the issue and the fix?

PR #3653, opened 2022-11-29
Change ins/lip (and others) to show full 100 results on the page

### LINKED ISSUE #3628: [Enhancement] /stocks/ins Return more than 100 results
**Describe the bug**
regardless of limit parameter the command returns max ~50 items

**To Reproduce**
OpenBB Terminal v1.9.0
/stocks/ins/lip -l 500 - returns max ~50 items. 
OpenInsider web page returns full 500 rows table

- OS: Windows
- Python version 3.11.0

jmaslek:
As the title says:
- Fixes #3628 
Fix was to get rid of the dropna as it was dropping things we dont want to drop.  Now will return up to 100 results.


jmaslek:
> I would rather keep the `limit` arg on the `model` and have the `df.head(limit)` also on the model - this way, people can have the same arguments using view/model on the sdk. Wdyt?

I lean towards having it in the view. The reasoning is if I am using the SDK and I call the model, I can get as much data as I want without having to worry about adding the limit flag.

hjoaquim (approved):

hjoaquim:
> > I would rather keep the `limit` arg on the `model` and have the `df.head(limit)` also on the model - this way, people can have the same arguments using view/model on the sdk. Wdyt?
> 
> I lean towards having it in the view. The reasoning is if I am using the SDK and I call the model, I can get as much data as I want without having to worry about adding the limit flag.

Fair enough!

Issue #3646, opened 2022-11-29
[IMPROVE] SDK - method to print/get SDK version 


jczernia:
**What's the feature that should be improved?**
Taking into consideration more releases are on the roadmap sanity check which get the SDK version used would be helpful for SDK users. Something like openbb.get_version() which will return the string would be perfect

jmaslek:
You can now do `openbb.__version__` to get the version.  If you run pip, you can also check from `pip list|grep openbb`

PR #3656, opened 2022-11-29
add version to sdk

jmaslek:
As requested in #3646 this just gets the verison from the feature flag
Screenshot 2022-11-29 at 1 23 58 PM


Chavithra (approved):
Looks ok to me.

PR #3676, opened 2022-12-01
Redirect users to correct installation tab based on OS

jose-donato:


andrewkenreich:
Need a catch all here for random user agent types -

image

This is shown if it isnt found

andrewkenreich (changes requested):
Other should show Windows

jose-donato:
@andrewkenreich should be good now. nice catch

andrewkenreich:
looks good to me - @jmaslek you want to review as well?

andrewkenreich (approved):

PR #3705, opened 2022-12-05
Core: move parse_simple_args

Chavithra:
# Description
Move the function `parse_simple_args`:
- from: `openbb_terminal.helper_funcs`
- into: `openbb_terminal.parent_classes`

This function seems specific to controllers.

Plus that will ease automatic completion into some classes.


openbb_terminal/parent_classes.py
```diff
@@ -690,13 +689,57 @@ def call_screenshot(self, other_args: List[str]) -> None:
...
+        if ns_parser.help:
+            txt_help = parser.format_help()
+            console.print(f"[help]{txt_help}[/help]")
+            return None
+
+        if l_unknown_args:
+            console.print(
+                f"The following args couldn't be interpreted: {l_unknown_args}"
+            )
+            console.print("\n")
```
jmaslek, line 736: Too many newlines.

jmaslek (approved):

PR #3718, opened 2022-12-06
Creates Crypto/Discovery SDK Guide.

deeleeramone:
Creates new folder in the directory tree for Crypto. Next commit will move `crypto.md` to the folder and rename as `index.md`


website/content/sdk/guides/intros/crypto/discovery.md
```diff
@@ -0,0 +1,131 @@
...
+```python
+categories[10]
+```
+
+```console
+'big-data'
+```
+
+```python
+openbb.crypto.disc.coins(category = categories[10], sortby = 'market_cap').head(5)
```
JerBouma, line 64: There something off here, this shouldn't all be different code blocks, I'd group them all together. E.g. I can not run these individually anyways.

JerBouma (changes requested):
Just one small note.

website/content/sdk/guides/intros/crypto/discovery.md
deeleeramone: I removed the part showing the category name; now they are just two separate commands. Better?

PR #3724, opened 2022-12-06
SDK - Small enhancement to `openbb.stocks.options.chains`, when source is Tradier.

deeleeramone:
# Description

- [X] Summary of the change / bug fix.

The main purpose of this PR is to restore the missing greeks on the SDK layer, where `source = 'Tradier'`. 

Besides restoring greeks, this accomplishes two minor presentation items:

- Cleans the column names.  
- Changes the order of the columns.

- [X] Screenshot of the feature or the bug before/after fix, if applicable.

Before:

before

After:

Screenshot 2022-12-06 at 1 17 47 PM

- [X] Relevant motivation and context.

Greeks were missing and there is currently no way to view or work with them in the SDK layer. The table presented is also now easier to read.

# How has this been tested?

Tested using `python`, `ipython`, and `Jupyter`.

For example:

```python
from openbb_terminal.sdk import openbb
openbb.stocks.options.chains('OXY', source = 'Tradier')
openbb.stocks.options.chains('VIX', source = 'Tradier')
openbb.stocks.options.chains('SPX', source = 'Tradier')
openbb.stocks.options.chains('NDX', source = 'Tradier')
```

Works with, or without, the `expiration` argument.

- [X] Make sure affected commands still run in terminal

These changes do impact Terminal functions. There are no downstream SDK functions or Terminal operations which can break because they are relying on this specific function to pass it data.

Screenshot 2022-12-06 at 1 32 59 PM

- [X] Ensure the SDK still works

The SDK functions as expected, no changes to the `def_get` structure, I/O variables, or docstrings were made.

# Checklist:

- [] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).

No changes required.

No changes were made that would require updating the tests.

Not adding a feature.

# Others

        Happy to add comments if required, but it's nothing complicated.


jmaslek:
So I prefer having the calls and puts in the same row to allow for plotting to be simplified.

That said, currently all our options functions return different things, so this suffices for now and is a solid addition until we improve the sdk portion for option.

openbb_terminal/stocks/options/tradier_model.py
```diff
@@ -142,21 +142,78 @@ def get_full_option_chain(symbol: str) -> pd.DataFrame:
...
-        temp = get_option_chains(symbol, date)
-        calls = temp[temp.option_type == "call"][option_cols].rename(
-            columns=option_col_map
-        )
-        puts = temp[temp.option_type == "put"][option_cols].rename(
-            columns=option_col_map
-        )
-        temp = pd.merge(calls, puts, how="outer", on="strike", suffixes=("_c", "_p"))
-        temp["expiration"] = date
+    options_df = pd.concat(options_dfs)
```
jmaslek, line 150: Note you are saying options_dfs is a dataframe then defining a list.

You can simply define options_df = pd.DataFrame() and then concat in the loop

df = pd.DataFrame()
for expiry in expirations:
   df = pd.concat([df,get_option_chains(symbol,expiry)])

jmaslek (approved):
Not too shabby for your first coding PR!

Some things we could do slightly differently, but this gets the job done!

deeleeramone:
> So I prefer having the calls and puts in the same row to allow for plotting to be simplified.
> 
> That said, currently all our options functions return different things, so this suffices for now and is a solid addition until we improve the sdk portion for option.

I had a couple of lines that I had left out, that separated calls and puts

> So I prefer having the calls and puts in the same row to allow for plotting to be simplified.
> 
> That said, currently all our options functions return different things, so this suffices for now and is a solid addition until we improve the sdk portion for option.

I had this part, but I left it out so not to muck with the `return`. It could easily be modified to return a Tuple of [calls,puts] if that was the desired. 

```
calls = options_df.filter(like = 'Call', axis = 0)
puts = options_df.filter(like = 'Put', axis = 0)
```

jmaslek:
> ```
> calls = options_df.filter(like = 'Call', axis = 0)
> puts = options_df.filter(like = 'Put', axis = 0)
> ```

Also not a fan of dealing with multiple datatsets.  For now, this is perfect while a proper design is being figured out for the menu.

PR #3734, opened 2022-12-08
Tensorboard Integration + model export location

martinb-bb:
# Description

No additional install dependancies needed. 

Example: 

```
load aapl.csv
trans aapl
```

Any models that use tensorboard to track gradients, now automatically saves logs for viewing. The user is then presented with instructions on how to run it and where to get files:
image

When running it will look like this in a new terminal: 
```
(obb) martinbufi@Martins-MacBook % tensorboard --logdir=/Users/martinbufi/OpenBBUserData/exports/forecast_models/trans_model/logs
TensorFlow installation not found - running with reduced feature set.

NOTE: Using experimental fast data loading logic. To disable, pass
    "--load_fast=false" and report issues on GitHub. More details:
    https://github.com/tensorflow/tensorboard/issues/4784

Serving TensorBoard on localhost; to expose to the network, use a proxy or pass --bind_all
TensorBoard 2.11.0 at http://localhost:6006/ (Press CTRL+C to quit)
```
image
image

Lastly, all models are now saved in `OpenBBUserData`
image


jmaslek:
This is pretty nifty.  My only question is should we put this behind a --tensorboard (or other) flag?

Reasoning being some of these files `can` get large, so saving them could unknowingly make the forecast models folder pretty large, so only doing that intentionally seems like the desired behavior?

martinb-bb:
> This is pretty nifty. My only question is should we put this behind a --tensorboard (or other) flag?
> 
> Reasoning being some of these files `can` get large, so saving them could unknowingly make the forecast models folder pretty large, so only doing that intentionally seems like the desired behavior?

@jmaslek I thought about that, but then figured it wouldn't be nessary to have it behind a flag since only a small portion of the models support it and the file size is so small.

image

`logs` will get re-written on each run, only keeping the most recent training. I figured this would be convenient for people to have done automatically, since the terminal will notify them that the specific model is supported by tensorboard. (else, they would have to search through documentation or `-h`)

Would you still like me to have this as a flag?

jmaslek:
> > This is pretty nifty. My only question is should we put this behind a --tensorboard (or other) flag?
> > Reasoning being some of these files `can` get large, so saving them could unknowingly make the forecast models folder pretty large, so only doing that intentionally seems like the desired behavior?
> 
> @jmaslek I thought about that, but then figured it wouldn't be nessary to have it behind a flag since only a small portion of the models support it and the file size is so small.
> 
> image
> 
> `logs` will get re-written on each run, only keeping the most recent training. I figured this would be convenient for people to have done automatically, since the terminal will notify them that the specific model is supported by tensorboard. (else, they would have to search through documentation or `-h`)
> 
> Would you still like me to have this as a flag?

I see. I was being silly and it was the model checkpoints that were saving by default.
Screenshot 2022-12-12 at 3 17 47 PM

In this case, its fine without the flag.  May want the saving checkpoints False by default, but thats out of scope.

martinb-bb:
@jmaslek Yup sounds good. Ready to merge whenever. 

Also the PR includes a quick URL change towards the website docs now - added in the last commits. (just FYI)

jmaslek:
> @jmaslek Yup sounds good. Ready to merge whenever.
> 
> Also the PR includes a quick URL change towards the website docs now - added in the last commits. (just FYI)

Sounds good.  We are going to get #3764 in first then this can be merged.

jmaslek (approved):

Issue #3702, opened 2022-12-04
[Bug] stocks/dd/est error handling causes equity report to fail [for "TM" (toyota) ticker]


DidierRLopes:
```
Executing:  17%|███████████▌                                                         | 11/66 [00:10<00:51,  1.07cell/s]Traceback (most recent call last):
  File "[http://terminal.py](https://t.co/NVZJO9Vvpj)", line 10, in <module>
  File "openbb_terminal\terminal_controller.py", line 1210, in parse_args_and_run
    main(
  File "openbb_terminal\terminal_controller.py", line 1129, in main
    terminal()
  File "openbb_terminal\terminal_controller.py", line 796, in terminal
    log_all_settings()
  File "openbb_terminal\core\log\generation\settings_logger.py", line 41, in log_all_settings
    do_rollover()
  File "openbb_terminal\core\log\generation\settings_logger.py", line 127, in do_rollover
    handler.doRollover()
  File "openbb_terminal\core\log\generation\path_tracking_file_handler.py", line 170, in doRollover
    super().doRollover()
  File "logging\[http://handlers.py](https://t.co/n65E0mvoEz)", line 403, in doRollover
  File "logging\[http://handlers.py](https://t.co/n65E0mvoEz)", line 115, in rotate
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\Jordan\\OpenBBUserData\\logs\\616d5c58-5a79-4a9a-aef3-3b63b87cb40c\\gst_packaged_1670125169' -> 'C:\\Users\\Jordan\\OpenBBUserData\\logs\\616d5c58-5a79-4a9a-aef3-3b63b87cb40c\\gst_packaged_1670125169.2022-12-03_19.log'
[2236] Failed to execute script 'terminal' due to unhandled exception!
Executing:  17%|███████████▌                                                         | 11/66 [00:13<01:05,  1.20s/cell]

An error was encountered in cell [8], check the notebook:
C:\Users\Jordan\OpenBBUserData\reports\20221203_193927_equity_tm.ipynb
```

montezdesousa:
Seems to be a bug with yf. They are adding "k" to the EPS rows and it breaks our casting to float.

<img width="779" alt="image" src="https://user-images.githubusercontent.com/79287829/206687069-e9ec3820-4b38-4fd4-b66e-7d0954427bd0.png">

PR #3739, opened 2022-12-09
Fix empty equity report

montezdesousa:
# Description

Fixes https://github.com/OpenBB-finance/OpenBBTerminal/issues/3702

# How has this been tested?
`openbb reports`
`equity --symbol TM`
`equity --symbol AAPL`


jmaslek:
The terminal function still works:

Screenshot 2022-12-09 at 1 49 18 PM

However on equity TM, I get stuck at In[16] which has teh following error

```
KeyError                                  Traceback (most recent call last)
Cell In [16], line 10
      6             df_analyst["target"] = df_analyst["target_to"].combine_first(
      7                 df_analyst["target"]
      8             )
      9     else:
...
   3808     #  InvalidIndexError. Otherwise we fall through and re-raise
   3809     #  the TypeError.
   3810     self._check_indexing_error(key)

KeyError: 'target_to'
```

I also note that in this notebook, there is `print(df_analyst.columns)` -- which was probably left over from a debug statement?

montezdesousa:
> The terminal function now works:
> 
> Screenshot 2022-12-09 at 1 49 18 PM
> 
> However on equity TM, I get stuck at In[16] which has teh following error
> 
> ```
> KeyError                                  Traceback (most recent call last)
> Cell In [16], line 10
>       6             df_analyst["target"] = df_analyst["target_to"].combine_first(
>       7                 df_analyst["target"]
>       8             )
>       9     else:
> ...
>    3808     #  InvalidIndexError. Otherwise we fall through and re-raise
>    3809     #  the TypeError.
>    3810     self._check_indexing_error(key)
> 
> KeyError: 'target_to'
> ```
> 
> I also note that in this notebook, there is `print(df_analyst.columns)` -- which was probably left over from a debug statement?

Forgot to indent that block, can you see if work now? Seems to be good on my side with `TM`, `AAPL` or default

jmaslek (approved):
All good on my end now

Issue #3707, opened 2022-12-05
[Bug] Stocks/TH/exchange: bursa_model.py:39: FutureWarning: The frame.append method is deprecated. Use pandas.concat instead.


deeleeramone:
Non critical but will eventually stop working. Use any exchange for: `/stocks/th/exchange`

```console
OpenBBTerminal/openbb_terminal/stocks/tradinghours/bursa_model.py:39: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
```

deepgohil:
I would like to take this up. As this will be my first contribution to this repo. if you can guide me it will be very helpful

deepgohil:
@deeleeramone I would like to take this up. As this will be my first contribution to this repo. if you can guide me it will be very helpful

deeleeramone:
Happy to help where I can, which aspect are you looking for guidance on?

PR #3753, opened 2022-12-10
replaced deprecated df.append method with pd.concat

skhan59bbk:
# Description

append method deprecated. pandas suggest replacing with concat. 
Fixes issue #3707 

# How has this been tested?

load any exchange in the following menu
`/stocks/th/exchange`

future warning no longer shows after method replaced.


montezdesousa (approved):
Looks good to me!

Issue #3750, opened 2022-12-10
[Bug] new release of pypa/packaging might break openbb 


koioannis:
**Describe the bug**
I think new [release](https://github.com/pypa/packaging/releases/tag/22.0) of `pypa/packaging` breaks openbb

**To Reproduce**
```
conda create -n obb python=3.9.6 -y
conda activate obb
pip install openbb
touch main.py
```
In `main.py`:
```
from openbb_terminal.sdk import openbb
```
then:
```
python main.py
```
results to:

```
Traceback (most recent call last):
  File "/Users/koioannis/dev/work/gdx/new-openbb/main.py", line 1, in <module>
    from openbb_terminal.sdk import openbb
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/sdk.py", line 19, in <module>
    openbb = Breadcrumb(
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/core/library/breadcrumb.py", line 105, in __init__
    BreadcrumbLogger()
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/core/library/breadcrumb.py", line 151, in __init__
    self.__check_initialize_logging()
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/core/library/breadcrumb.py", line 155, in __check_initialize_logging
    self.__initialize_logging()
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/core/library/breadcrumb.py", line 160, in __initialize_logging
    from openbb_terminal.core.log.generation.settings_logger import (  # pylint: disable=C0415
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/core/log/generation/settings_logger.py", line 13, in <module>
    from openbb_terminal.terminal_helper import is_packaged_application
  File "/Users/koioannis/opt/miniconda3/envs/obb/lib/python3.9/site-packages/openbb_terminal/terminal_helper.py", line 283, in <module>
    latest_version: Union[version.LegacyVersion, version.Version],
AttributeError: module 'packaging.version' has no attribute 'LegacyVersion'
```

**Desktop:**

- OS: MacOS Monterey 12.6
- Python version 3.9.6

[...]

PR #3764, opened 2022-12-12
Changing packaging version

hjoaquim:
This PR patch the deprectation of `LegacyVersion` under `packaging` - also discussed here: https://github.com/pypa/packaging/issues/631

- Removes `LegacyVersion`
- Specifies that `packaging=">=22.0"`

Steps taken: 

```bash
poetry add packaging=">=22.0"
poetry export -f requirements.txt  -o requirements.txt --without-hashes --dev
poetry export -f requirements.txt  -o requirements-full.txt --extras prediction --extras optimization --without-hashes --dev
```

Fix:
- #3750


jmaslek:
I think the better solution will be to change:
```
current_version: Union[version.LegacyVersion, version.Version]
```
to 
```
current_version: version.Version
```

And then add packaging^22.

Thoughts?

hjoaquim:
Does the Tests / General Linting create a fresh environment each time? The error does not make sense since `version.parse()` returns only `Version` with packaging 22

jmaslek:
> Does the Tests / General Linting create a fresh environment each time? The error does not make sense since `version.parse()` returns only `Version` with packaging 22

It uses the actions/cache@v3 to restore dependencies.  I am not sure how to change those, so you may need to just add `packaging==22` to the yml

jmaslek (approved):

jmaslek:
> > Does the Tests / General Linting create a fresh environment each time? The error does not make sense since `version.parse()` returns only `Version` with packaging 22
> 
> It creates a fresh env when the dep tree is re-solved

The tests create a new env, but the linting uses the cache from gh actions.  That has not been updated in 9 months.  

I used this doc to open a new PR that will edit the cache when the poetry lock changes: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

hjoaquim:
@jmaslek @piiq thanks for the clarification!

Issue #3758, opened 2022-12-11
[Bug] Reports/Economy


deeleeramone:
`/reports/economy`:

```console
AttributeError                            Traceback (most recent call last)
Cell In [9], line 17
     14 fig.savefig(f, format="svg")
     15 ycrv_chart_1 = f.getvalue().decode("utf-8")
---> 17 if not ycrv_data_1.empty:
     18     ycrv_data_1.drop(["Maturity"], axis=1, inplace=True)
     19     ycrv_data_1.index = [
     20         "1M",
     21         "3M",
   (...)
     30         "30Y",
     31     ]

AttributeError: 'str' object has no attribute 'empty'
```

PR #3766, opened 2022-12-12
Fixes Bug #3758 Reports/Economy yield curve notebook

skhan59bbk:
# Description

Bug #3758 
.ycrv() function returns one object but prev code was assigning to two variables.
image

# How has this been tested?

Notebook cell runs without error and shows yield curve figure.

image

Separately... might want to change x-axis to show rate tenors (i.e. 1M, 3M, ... etc) rather than years to maturity. But that's an EN outside the scope of this bug fix.


jmaslek:
Hey thanks for taking a look into this.  Is there a reason that the diff is 80k lines?

DidierRLopes:
> Hey thanks for taking a look into this. Is there a reason that the diff is 80k lines?

I believe because @skhan59bbk commit with the notebook with images/output in cells which gets stored.

Can you reset and clean report and commit? That way we only get the code changes

skhan59bbk:
Please ignore for now (or close), while I figure this out! Thought I had definitely removed the output cells this time but evidently not.

PR #3792, opened 2022-12-16
Small options fix + docstring

jmaslek:
1) Fix the greek calculation in `chains` where we were using .seconds instead of .total_seconds
2) Change the time to expiration calculation in options to be 4PM instead of midnight am (hence the + 16 hours)
3) Fix the typo I reverted from @hjoaquim


hjoaquim (approved):

hjoaquim:
No worries :laughing:

PR #3799, opened 2022-12-19
Out of the box forex report

### LINKED ISSUE #3570: [Bug] Reports/Forex Candlestick charts fail to draw
`/reports/forex NZDCAD`

Screenshot_20221124_070201

The Box and Whisker plot is also stylistically-confused.

Screenshot_20221124_070518

hjoaquim:
1. Fixes #3570
2.  Removes the "Box and Whisker plot" as it end up grabbing the terminal theme (from the config file) and gives the report a weird look. - should this be reconsidered?


jmaslek:
Works for me!  I think it is fine to leave that out for our default out of the box one.

jmaslek (approved):

PR #3804, opened 2022-12-20
Fixes ValueError in Cell 21.

joshuabuildsthings:
# Description
- [ x] Summary of the change / bug fix.
 - In the recently updated economy report there was a ValueError in Cell 21 caused by a truthiness check on a pandas data frame. This caused economy reports to fail.
 - This was a minor change to one line and shouldn't require changes to documentation or existing tests.

# How has this been tested?

 - Unable to run economy report prior to the fix. Able to run it now.

# Checklist:

- [ x] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [ x] Update our tests following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/tests).
- [ x] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).
- [x ] If a feature was added make sure to add it to the corresponding [scripts file](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/scripts).

# Others
- [ x] I have performed a self-review of my own code.
- [x ] I have commented my code, particularly in hard-to-understand areas.


joshuabuildsthings:
Hi @hjoaquim & @jmaslek!

Thank you for the work on the custom economy report. I tried to run it and received an error; this patch should correct the issue.

hjoaquim:
Hey @joshuabuildsthings, thanks for this nice detail :rocket: 

I think this will not work in case the API key is defined for FRED. Reason is because of this line returning a tuple (if API key defined):

`fred = openbb.economy.fred(series_ids=["T10Y3M"], start_date="1980-01-01") # this returns a tuple`

IMO, the fix should be something like this instead:

```python
fred = openbb.economy.fred(series_ids=["T10Y3M"], start_date="1980-01-01") # this returns a tuple

if fred:
    df, df_dict = fred

    (... code ...)
else:
    fred = None # we can ommit this if we want
```

And then we don't need to touch the final condition.

What do you think?

PS - I've tried to add those directly myself but got permission denied :disappointed:

hjoaquim:
Also, don't forget to "Clear Output of All Cells" before commit :smile:

joshuabuildsthings:
> Hey @joshuabuildsthings, thanks for this nice detail 🚀
> 
> I think this will not work in case the API key is defined for FRED. Reason is because of this line returning a tuple (if API key defined):
> 
> `fred = openbb.economy.fred(series_ids=["T10Y3M"], start_date="1980-01-01") # this returns a tuple`
> 
> IMO, the fix should be something like this instead:
> 
> ```python
> fred = openbb.economy.fred(series_ids=["T10Y3M"], start_date="1980-01-01") # this returns a tuple
> 
> if fred:
>     df, df_dict = fred
> 
>     (... code ...)
> else:
>     fred = None # we can ommit this if we want
> ```
> 
> And then we don't need to touch the final condition.
> 
> What do you think?
> 
> PS - I've tried to add those directly myself but got permission denied 😞

@hjoaquim Thank you for the feedback! I made the recommended changes & re-tested & it looks like everything worked as expected. 

Hopefully I got the Clear Output of All Cells right this time 😄

openbb_terminal/reports/templates/economy.ipynb
```diff
@@ -10,10 +10,22 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "driven-billion",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
hjoaquim, line 16: Would it be possible to get rid of these?

hjoaquim (changes requested):
Other than the small comment, is looking alright!
Works perfectly on my end, both with and without FRED API key.

joshuabuildsthings:
@hjoaquim - I think I cleared the cells as you requested this time; also brought the feature branch up to date with the 2.1.0 from today.

hjoaquim (approved):

hjoaquim:
@joshuabuildsthings not sure why (prob your fork settings) but can't update your branch :cry: 
We need you to update it before we're able to merge! :rocket:

joshuabuildsthings:
@hjoaquim - Sorry about that. Adjusted the settings & merged the recent changes from main back in.

PR #3812, opened 2022-12-21
Fix the greeks calculations of the chain command

### LINKED ISSUE #3779: [Bug] Large Difference in Options Greeks Values Between Sources.
The two values are not even in the ballpark, for the same option.

Screenshot 2022-12-14 at 8 32 32 PM

northern-64bit:
# Description
This makes the greeks calculations in the chains command (with YahooFinance) accurate. Closes #3779. There will still exist a difference between the sources due to different IV values, current asset prices (when the market is open) and so on ..., namely due to the data given.

# How has this been tested?

Running the command and comparing results with Nasdaq, Tradier and the greeks command.

# Checklist:

# Others

### DIFF
--- a/openbb_terminal/stocks/options/yfinance_model.py
+++ b/openbb_terminal/stocks/options/yfinance_model.py
@@ -128,13 +128,15 @@ def get_option_chain_expiry(
     # Add in greeks to each df
     # Time to expiration:
     dt = (
-        datetime.strptime(expiry, "%Y-%m-%d") + timedelta(hours=16) - datetime.now()
-    ).total_seconds() / (60 * 60 * 24)
+        (datetime.strptime(expiry, "%Y-%m-%d") - datetime.now()).total_seconds()
+        + 16 * 60 * 60
+    ) / (60 * 60 * 24)
+    rf = get_rf()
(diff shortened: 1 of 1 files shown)


openbb_terminal/stocks/options/yfinance_model.py
```diff
@@ -128,27 +128,29 @@ def get_option_chain_expiry(
     # Add in greeks to each df
     # Time to expiration:
     dt = (
-        datetime.strptime(expiry, "%Y-%m-%d") + timedelta(hours=16) - datetime.now()
-    ).total_seconds() / (60 * 60 * 24)
+        (datetime.strptime(expiry, "%Y-%m-%d") - datetime.now()).total_seconds()
+        + 16 * 60 * 60
+    ) / (60 * 60 * 24)
```
jmaslek, line 133: Just curious why you decided to change from the time delta (which I added in #3792).  Technically not wrong, but using time delta is more readable imo.

northern-64bit: Oh, I worked on an old version and then just overwrote the changes when I fetched. I'll switch back 👍

jmaslek (approved):

PR #3850, opened 2023-01-01
Display daily returns as bar chart with color=red for negative return…

LarsKarbach:
…s and color=green for positive return

# Description
The displayed daily return as given by the `dret_chart` command is implemented as a connected line chart. I think a bar chart is more visually appealing. I've gone further and added red/green color for returns below/above zero. See example:

Before:
Daily_return_old

After:
daily_return_new


jmaslek:
This definitely looks a lot nicer!

One question,  instead of using red/green, can we use the custom theme files?

They are:

```
from openbb_terminal.config_terminal import theme
```

Then it is theme.up_color or theme.down_color instead of red/green

LarsKarbach:
Of course, fixed it.

jmaslek (approved):

jmaslek:
Nice!  The linter is now yelling.  Running `black ./` should fix the file and make it happy.

Issue #3822, opened 2022-12-22
[Bug] openbb.stocks.fa.analysis for ticker that doesn't exist


DidierRLopes:
<img width="429" alt="Screenshot 2022-12-22 at 19 36 16" src="https://user-images.githubusercontent.com/25267873/209212928-54527cd5-3db1-447b-81bc-d2c461224884.png">

northern-64bit:
Has been fixed by #3858.

PR #3858, opened 2023-01-02
stocks/fa/analysis returns empty dataframe instead of error when no data

skhan59bbk:
# Description
Quick and unsophisticated fix for Bug #3822 

Now just checks if resulting list is empty and returns empty DataFrame. In the terminal, this will trigger printing a message (filings not found) and in SDK this will just return a blank DataFrame. 

No error message appears in either case but if you'd prefer some sort of message in the notebook then please say!

# How has this been tested?
Tested various US and non-US tickers (and some made up tickers). 

`stocks -> load coin -> fa -> analysis `

In terminal:
image

In notebook:
image

As expected, COIN should not return results. 
Apple, however, should run just fine...

`stocks -> load aapl -> fa -> analysis `

In terminal:
image

In notebook:
image


jmaslek (approved):
Nice!  Just need to fix the linter.  Running `black ./` should auto fix it!

PR #3860, opened 2023-01-03
Fix issue on loading stock

hjoaquim:
Fix index out of bonds - issue for nonexistent YTD dataframe when loading a stock.

image


montezdesousa (approved):
Works fine on my end

Issue #3835, opened 2022-12-23
[FR] Download price history from the SDK for multiple symbols at once


jasonstrimpel:
**What's the problem of not having this feature?**
I'd like to download price history through the SDK for multiple symbols at one time, similar to yfinance.

**Describe the solution you would like**

symbols = ["MSFT", "GME"]
prices = openbb.stocks.load(symbols, start_date=start, end_date=end)['Adj Close']

**Describe alternatives you've considered**
Loop through the list and append columns to DataFrames.

**Additional information**
Add any other information or screenshots about the feature request here.

deeleeramone:
This functionality exists already:

```
openbb.economy.index(['MSFT','AAPL'])

```

![Screenshot 2022-12-27 at 9 56 00 AM](https://user-images.githubusercontent.com/85772166/209704483-976cdba4-da84-4711-ac51-ef056c5d378b.png)


That said, it would be nice to get groups from other sources too.

jmaslek:
The sdk function I would recommend would be to use `openbb.stocks.ca.hist(ticker_list)`

jasonstrimpel:
`openbb.stocks.ca.hist(ticker_list)` doesn't allow an end date so does not fit my requirements.

`openbb.economy.index(['MSFT','AAPL'])` better fits.

That said, both solutions may be better in a different namespace.

Per the documentation, `openbb.economy.index(['MSFT','AAPL'])` is for comparables and `openbb.economy.index(['MSFT','AAPL'])` is for indexes.

jmaslek:
This was actually very easy to add: so I did #3867.

I think stocks.load( tickers ) may be more intuitive down the line, but I did not want to break that load that is used EVERYWHERE, and for loading in multiple symbols, comparison (ca) made sense to me.

That pr will let you do
```
start = "2012-08-22"
end = "2015-02-02
stocks.ca.hist( list_of_tickers , start, end)
```
(Also applied into the terminal functions!)

jmaslek:
Hey @jasonstrimpel, any chance you took a look and would consider your requirements met?

jasonstrimpel:
This is cool thanks @jmaslek closing

PR #3867, opened 2023-01-03
Adds historical end date on the ca menu

jmaslek:
This was asked for in #3835 and has been applied tp the terminal and the sdk.  Note sdk usage:

Tests updated and this was added to all the yf calls in the menu.

Screenshot 2023-01-03 at 4 40 17 PM


montezdesousa:
End date inputs before or equal to start date yield an error in the terminal for `historical` and `volume` commands. Seem ok for the sdk
 
Screenshot 2023-01-04 at 14 36 17

jmaslek:
> End date inputs before or equal to start date yield an error in the terminal for `historical` and `volume` commands. Seem ok for the sdk
> 
> Screenshot 2023-01-04 at 14 36 17

Adding a helper to check this:
```
2023 Jan 04, 10:29 (🦋) /stocks/ca/ $ volume --start 2022-01-01 --end 2022-01-01

Start date and end date cannot be the same.

2023 Jan 04, 10:29 (🦋) /stocks/ca/ $ volume --start 2022-01-01 --end 2021-01-01

Start date cannot be greater than end date.
```

openbb_terminal/stocks/comparison_analysis/ca_controller.py
```diff
@@ -476,7 +477,15 @@ def call_historical(self, other_args: List[str]):
...
             default=(datetime.now() - timedelta(days=366)).strftime("%Y-%m-%d"),
             dest="start",
-            help="The starting date (format YYYY-MM-DD) of the stock",
+            help="The starting date (format YYYY-MM-DD) of the stocks",
+        )
+        parser.add_argument(
+            "-e",
+            "--end",
+            type=valid_date,
+            default=(datetime.now() - timedelta(days=366)).strftime("%Y-%m-%d"),
```
hjoaquim, line 486: Suggestion: we can have the end date as today (instead of on it being the same as start date) so we have the `historical` command work out of the box, instead of this:
image

hjoaquim (changes requested):

openbb_terminal/stocks/comparison_analysis/ca_controller.py
jmaslek: good call.  silly me copy pasting the start date

Issue #3838, opened 2022-12-27
[Bug] Economy/Eval - Using `plot` for the results from `eval` is messed up, but the exported file is fine.


deeleeramone:
![Screenshot 2022-12-27 at 2 38 00 PM](https://user-images.githubusercontent.com/85772166/209731527-cf4bdcd2-f4dd-4821-9c5b-45acdc575757.png)

Recreate with:

```
/economy/fred AMERIBOR,SOFR
eval spread = AMERIBOR - SOFR
plot spread
```

Exporting the results from `eval` shows that the data is fine. 

![Screenshot 2022-12-27 at 2 47 32 PM](https://user-images.githubusercontent.com/85772166/209731774-231e6289-6d85-4518-9a25-c6de2a71a79e.png)

deeleeramone:
Another notable quirk is that the `eval` results do not appear in the choices for the `plot` function.

PR #3881, opened 2023-01-06
Fix plot look when using eval and add query to choices

montezdesousa:
# Description

Fixes https://github.com/OpenBB-finance/OpenBBTerminal/issues/3838

Screenshot 2023-01-06 at 15 45 32
nance/OpenBBTerminal/issues/3838

Test:
`economy`
`fred SOFR,AMERIBOR`
`eval spread=SOFR-AMERIBOR`
`plot spread`

# How has this been tested?

- [ ] Ensure the SDK still works

# Checklist:

# Others

### DIFF
--- a/openbb_terminal/economy/economy_controller.py
+++ b/openbb_terminal/economy/economy_controller.py
@@ -965,7 +965,8 @@ class EconomyController(BaseController):
                         [
                             self.DATASETS["treasury"],
                             df,
-                        ]
+                        ],
+                        axis=1,
                     )
 
                     cols = []
--- a/openbb_terminal/economy/economy_controller.py
+++ b/openbb_terminal/economy/economy_controller.py
@@ -1562,7 +1562,7 @@ class EconomyController(BaseController):
             self.stored_datasets = economy_helpers.update_stored_datasets_string(
                 self.DATASETS
             )
-        console.print()
+            self.update_runtime_choices()
 
     @log_start_end(log=logger)
     def call_qa(self, _):
--- a/openbb_terminal/economy/economy_controller.py
+++ b/openbb_terminal/economy/economy_controller.py
@@ -587,12 +587,19 @@ class EconomyController(BaseController):
                 )
 
                 if not df.empty:
+
                     df.columns = ["_".join(column) for column in df.columns]
 
                     if ns_parser.transform:
                         df.columns = [df.columns[0] + f"_{ns_parser.transform}"]
 
-                    self.DATASETS["macro"] = pd.concat([self.DATASETS["macro"], df])
+                    for column in df.columns:
+                        if column in self.DATASETS["macro"].columns:
+                            self.DATASETS["macro"].drop(column, axis=1, inplace=True)
+
+                    self.DATASETS["macro"] = pd.concat(
+                        [self.DATASETS["macro"], df], axis=1
+                    )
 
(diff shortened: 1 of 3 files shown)


jmaslek:
kind of out of scope, but `eval spread=sofr-ameribor` doesn't work because it expects a match on capital SOFR.  Should we make the query upper case?

jmaslek:
> kind of out of scope, but `eval spread=sofr-ameribor` doesn't work because it expects a match on capital SOFR. Should we make the query upper case?

actually this won't work because macro has mixed cases.  Should have a uniform way of doing these,

jmaslek (approved):
This does fix the bug.  The feature itself has a lot more that could be improved (*looks at who wrote feature* - jk its perfect)

montezdesousa:
> > kind of out of scope, but `eval spread=sofr-ameribor` doesn't work because it expects a match on capital SOFR. Should we make the query upper case?
> 
> actually this won't work because macro has mixed cases. Should have a uniform way of doing these,

what's the case where it does not work?

jmaslek:
> > > kind of out of scope, but `eval spread=sofr-ameribor` doesn't work because it expects a match on capital SOFR. Should we make the query upper case?
> > 
> > 
> > actually this won't work because macro has mixed cases. Should have a uniform way of doing these,
> 
> what's the case where it does not work?

So if you just do macro gdp, the macro will have column: `united_states_GDP`.  (note doing `eval newCol=united_states_GDP*SOFR` wont work because of mismatched indices), but we should be case insensitive

montezdesousa:
> This does fix the bug. The feature itself has a lot more that could be improved (_looks at who wrote feature_ - jk its perfect)

I actually found another bug while running the integration test script. So if you load the same series twice it was concatenating both instead of replacing it. That was blocking `qa` menu afterwards

montezdesousa:
> > > > kind of out of scope, but `eval spread=sofr-ameribor` doesn't work because it expects a match on capital SOFR. Should we make the query upper case?
> > > 
> > > 
> > > actually this won't work because macro has mixed cases. Should have a uniform way of doing these,
> > 
> > 
> > what's the case where it does not work?
> 
> So if you just do macro gdp, the macro will have column: `united_states_GDP`. (note doing `eval newCol=united_states_GDP*SOFR` wont work because of mismatched indices), but we should be case insensitive

got it, so we either block the user if not perfect match between indices, or we can merge the indices and apply the eval transform on the overlap of both series only

jmaslek:
> > > > > kind of out of scope, but `eval spread=sofr-ameribor` doesn't work because it expects a match on capital SOFR. Should we make the query upper case?
> > > > 
> > > > 
> > > > actually this won't work because macro has mixed cases. Should have a uniform way of doing these,
> > > 
> > > 
> > > what's the case where it does not work?
> > 
> > 
> > So if you just do macro gdp, the macro will have column: `united_states_GDP`. (note doing `eval newCol=united_states_GDP*SOFR` wont work because of mismatched indices), but we should be case insensitive
> 
> got it, so we either block the user if not perfect match between indices, or we can merge (inner) the indices and apply the eval transform on the overlap of both series only

Yeah.  Right now I think I just loop through each dictionary and concat dfs to create a giant df, so it probably breaks at that stage. So improving that could be a future step.  But also outside the scope of this bug/pr.

PR #3888, opened 2023-01-07
Add crypto dd functions to sdk

### LINKED ISSUE #3884: [Bug] Can't access certain crypto.dd functions from SDK
Apologies if I'm just missing something, but I cannot seem to access crypto dd `score` and `dev` from the SDK. They do not appear in `dir(openbb.crypto.dd)`. I'm having a hard time tracing this back with the Breadcrumbs() structure. Given they appear in `due_diligence_api.py` I would have expected them to be accessible.

Thanks!

jmaslek:
This will fix #3884.

Thie issue is that coingecko functions did not have the correct model definitions, so I created an sdk helper to call the methods from the coingecko coin object.


hjoaquim (changes requested):
Tiny things (didn't change it myself to don't mess with your work)
It's pretty much done!

hjoaquim (approved):
Sweet!

PR #3964, opened 2023-01-14
Fix portfolio/summary command (issue #3960)

### LINKED ISSUE #3960: [Bug] Portfolio/summary - Error: Found input variables with inconsistent numbers of samples
Recreate with:

`/portfolio/load -f Public_Equity_Orderbook.xlsx/summary`

```
Error: Found input variables with inconsistent numbers of samples: [3209, 3198]
```

Doesn't matter what the portfolio file is.

```
Portfolio: megacap.csv
Risk Free Rate: 0.00%
Benchmark: SPDR S&P 500 ETF Trust (SPY)

2023 Jan 13, 19:27 (🦋) /portfolio/ $ summary

Error: Found input variables with inconsistent numbers of samples: [3991, 3990]
```

northern-64bit:
# Description
Fixes the summary command by removing data points, which are not in booth the portfolio and the benchmark data frame, from the data for the r2 calculations. It was these points that caused the error.

Closes #3960

# How has this been tested?

Running the command with multiple different portfolios. Checking that there are no data leaks by comparing results with and without data modifications.

# Checklist:

# Others


jmaslek (commented):
So while this will fix the bug, why are we dropping data?  We should try to figure out why the index lengths mismatch first before just hiding it.

northern-64bit:
@jmaslek The issue can be from different holdings being in different countries and thus having different holidays and vacation, thus there's a difference in trading days. I checked some different portfolios and it is nearly always just one or two dates a year which is missing. Here's one example of the days missing in the benchmark:
```
Date
2010-05-03   -1.000000
2011-08-06   -0.405741
2012-08-04    0.071733
2013-02-03   -0.065389
2015-03-08    0.014973
2015-12-05   -0.021659
2016-02-06   -0.007925
2018-09-03    0.000046
2019-05-04   -0.024390
2020-03-08    0.001432
2020-12-12    0.015081
Name: Total, dtype: float64
```
I haven't been able to see something systematic in the index lengths mismatch.

jmaslek (approved):
approving since it is just for r2.  if you dropped dates for returns calculations, that would be a different story

PR #3973, opened 2023-01-15
Wrong default key values

sergio-abu:
# Description
When using the sdk it becomes clear that the default string for the keys is wrong due to not being capitalized

Example documentation:
https://docs.openbb.co/sdk/reference/forecast/rsi

### Does not work:
```
STOCK = openbb.stocks.load('BABA')
print(STOCK)
RSI = openbb.forecast.rsi(STOCK)
RSI.plot()
```

### Does work:
```
STOCK = openbb.stocks.load('BABA')
print(STOCK)
RSI = openbb.forecast.rsi(STOCK, target_column="Close")
RSI.plot()
```

### DIFF
--- a/openbb_terminal/forecast/forecast_model.py
+++ b/openbb_terminal/forecast/forecast_model.py
@@ -157,7 +157,7 @@ def clean(
 
 @log_start_end(log=logger)
 def add_ema(
-    dataset: pd.DataFrame, target_column: str = "close", period: int = 10
+    dataset: pd.DataFrame, target_column: str = "Close", period: int = 10
 ) -> pd.DataFrame:
     """A moving average provides an indication of the trend of the price movement
     by cut down the amount of "noise" on a price chart.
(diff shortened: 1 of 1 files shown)


deeleeramone:
You are not using the module correctly. It is for adding a column that is the RSI of the `target_column` and to use as a past-covariant in Forecast functions.  

`RSI.plot()` is not how you would go about plotting the results.

https://docs.openbb.co/sdk/guides/intros/forecast#how-to-use
```
One important distinction between the Terminal and SDK is that the target_column must be explicitly declared when using the SDK, if it is not labeled as "close". It is case-sensitive.
```


Your changes would break functionality within the Terminal application.

Screenshot 2023-01-15 at 10 22 34 AM

sergio-abu:
I feared that it would break the terminal, since i did not checked how the terminal works.
But that also means the documentation here is wrong for the sdk: if you pass only the necessary parameters (dataset: pd.DataFrame) it causes an error due to invalid key, since the default value for target_column is not capitalized ("close"). 
Error: 'close'

But due the documentation the forecast is meant for the terminal:
https://docs.openbb.co/sdk/guides/intros/forecast#how-to-use

PR #3980, opened 2023-01-16
Quantile Anomaly Detection on Timeseries

martinb-bb:
# Description
Flags anomalies that are beyond some quantiles of historical data.
image

image

# Others


martinb-bb:
Got it. Will rename

jmaslek:
> image

I think you may want to twinx the chart.  Its also not clear what is 1 and what is 0, since on the chart it shows -100 and says 1 = True

openbb_terminal/forecast/qanom_model.py
```diff
@@ -0,0 +1,62 @@
+# pylint: disable=too-many-arguments
+"""Linear Regression Model"""
```
jmaslek, line 2: whoops

martinb-bb: whoops! all fixed.

jmaslek:
So what exactly does this detect/whats the general idea?  Does it take the initial split as training, find the 99th quantile (you use (high_quantile=0.99)) of that train data and then flag what values in the test set that are above that?

Or does it use a local window (in your KMeansScorer) to find what sequences are the most "different"?

martinb-bb:
> So what exactly does this detect/whats the general idea? Does it take the initial split as training, find the 99th quantile (you use (high_quantile=0.99)) of that train data and then flag what values in the test set that are above that?
> 
> Or does it use a local window (in your KMeansScorer) to find what sequences are the most "different"?

Great question. Basically the latter. 

This detector flags anomalies that are beyond some quantiles of historical data. 

The KMeansScorer is a model that uses the k-means algorithm to detect anomalies in a time series. When the "fit" function is called, the model applies a moving window to the input series, creating a set of vectors of size "W", where "W" is the window size. The k-means model is then trained on these vectors. The "score" function applies the same moving window to the input series and returns the distance to the closest of the "k" centroids for each vector of size "W".

jmaslek (approved):

martinb-bb:
Made it one plot on a single ax. easier to follow for now including removing anomaly score since it may not be obvious for the user.

image

jmaslek:
> Made it one plot on a single ax. easier to follow for now including removing anomaly score since it may not be obvious for the user.
> 
> image

this looks much nicer

PR #3997, opened 2023-01-17
Add ticker to the load export file

### LINKED ISSUE #3994: [Bug] Export of `stocks/load` leaves you with unrecognizable Excel file names
See the following, this is an export of multiple different tickers but it is not possible to recognize which is which:

image

The command `stocks/candle` does this better but downloading data via `load` is more efficient.

jmaslek:
This is a super super small thing for @JerBouma that will just make the exports noticeable.

(kinda) Fixes #3994


JerBouma (approved):
Yup, this works!

PR #4009, opened 2023-01-18
Fix Win build

hjoaquim:
Fix the windows build by explicitly pull `scipy.libs` into the build if it exists.


hjoaquim:
Maybe this issue is now gone?
Looks like the @luqmanbello build is not running into the `scipy.libs` issue: https://github.com/OpenBB-finance/OpenBBTerminal/actions/runs/3950639416
@luqmanbello is this build from `develop` branch?

hjoaquim:
^
We'll need to wait until the integration tests, if I'm not wrong.

tehcoderer:
Change to `"scipy.libs/"` with the forward slash at the end, it's how I got it to work on my Plots POC PR
```    
added_files.append(
        (os.path.join(f"{os.path.dirname(scipy.__file__)}.libs"), "scipy.libs/"),
    )
```

build/pyinstaller/terminal.spec
```diff
@@ -75,6 +75,9 @@ added_files = [
     (os.path.join(pathex, "blib2to3", "PatternGrammar.txt"), "blib2to3"),
 ]
 
+if is_win and os.path.exists(os.path.join(pathex, "scipy.libs")):
+    added_files.append((os.path.join(pathex, "scipy.libs"), "scipy.libs"))
```
hjoaquim, line 79: should I change this line for that? @tehcoderer

tehcoderer:
```
if is_win:
    print("Adding scipy.libs to bundle")
    added_files.append(
        (os.path.join(f"{os.path.dirname(scipy.__file__)}.libs"), "scipy.libs/"),
    )
```

PR #4010, opened 2023-01-18
Fix M1 build

hjoaquim:
^


tehcoderer:
Add this file to hooks 
https://github.com/OpenBB-finance/OpenBBTerminal/blob/6e6464dae545134a898e3848065f9d1794b3ebca/build/pyinstaller/hooks/hook-xgboost.py

build/pyinstaller/terminal.spec
```diff
@@ -75,6 +75,10 @@ added_files = [
     (os.path.join(pathex, "blib2to3", "PatternGrammar.txt"), "blib2to3"),
 ]
 
+
+if is_darwin and os.path.exists(os.path.join(pathex, "xgboost")):
+    added_files.append((os.path.join(pathex, "xgboost"), "xgboost"))
```
hjoaquim, line 80: should the hook replace this? @tehcoderer

tehcoderer: yes sir you can remove that cause it' run with this
image

jmaslek:
So I think this works?

jmaslek:
So I think this works?

jmaslek (approved):

PR #4017, opened 2023-01-19
Refactor `stocks/options/voi,vol,oi,chains`

### LINKED ISSUE #3403: [Bug] stocks/options/voi - Chart weirdness + `--raw` flag does not work.
`stocks/load qqq/options/exp 0/voi`

The chart displays `volume` and `oi`. However, `oi` is added on to the end of the `volume` graph, instead of being an overlapping value that begins at 0. The chart title also states that open interest is only displayed during market hours, it appears to be inaccurate. 

When the strike range is defined as `-m 250 -M 290`, the chart does not display the full-width of the values. One needs to zoom out before seeing the complete bar.

Screenshot 2022-11-12 at 2 26 29 PM

`voi --raw` does not show the raw table.

hjoaquim:
# Description

This PR changes the way the `vol,voi,oi,chains` functions are used under the `stocks/options` menu.

Instead of having it working differently for each source, those functions were centralized in the new `options_view.py` file.

TODO before merge:
   - [x] Add unit tests for the new `options_view.py` file.

Bonus:
   - Fix #3403


jmaslek:
I am seeing that `voi` and `oi` do not have auto complete on the -e flag, but `vol` does

jmaslek:
```
2023 Jan 19, 11:48 (🦋) /stocks/options/ $ voi

No expiry loaded. First use `exp {expiry date}`
```

This message should append saying something like "or specify an expiration with the `-e` flag"

openbb_terminal/stocks/options/options_controller.py
```diff
@@ -814,20 +830,17 @@ def call_chains(self, other_args: List[str]):
...
-                                to_display=ns_parser.to_display,
-                                min_sp=ns_parser.min_sp,
-                                max_sp=ns_parser.max_sp,
-                                calls_only=ns_parser.calls,
-                                puts_only=ns_parser.puts,
-                                export=ns_parser.export,
-                            )
-                        else:
-                            console.print("TRADIER TOKEN not supplied. \n")
+                    if ns_parser.source == "Tradier":
```
jmaslek, line 833: Since we already have the option chain loaded, shouldn't chains be source independent and similar to the vol/oi/voi commands that use the loaded df and allow to specify exp ?

hjoaquim:
Autocompletion fixed :heavy_check_mark: 
Adjusted message :heavy_check_mark: 

Regarding `chains` cmd, yes, agreed - but, should I address that in another PR so this one doesn't keep growing? (wip already)

jmaslek:
> Autocompletion fixed ✔️ Adjusted message ✔️
> 
> Regarding `chains` cmd, yes, agreed - but, should I address that in another PR so this one doesn't keep growing? (wip already)

Lets make that the last addition here.

The only reason is so that we have the same -e usage across the primary commands used.

hjoaquim:
^
on it !

hjoaquim:
wip: making the piece of code on `yfinance_model.get_option_chain_expiry` that calculates the greeks general to apply on `display_chains`

hjoaquim:
`chains` cmd refactored :heavy_check_mark:

jmaslek (approved):
Perfect.  Ignoring the new yf errors :D

PR #4019, opened 2023-01-19
Feature - Added Integration Test Automation

luqmanbello:
# Description

# How has this been tested?

- [ ] Ensure the SDK still works

# Checklist:

# Others


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,51 @@
+name: Integration Tests
+
+on:
+  push:
+    branches:
+      - release/*
+  pull_request:
+    branches:
+      - develop
+      - main
```
jmaslek, line 10: should we also run these on a cron from develop?  We do the nightly build (which needs to be updated to run the int tests correctly)

luqmanbello: I can invoke the workflow from the nightly build workflow. But since we would be running the build and there is an integration test on each build, won't that be an overlap?

jmaslek: I guess it would (assuming the build gets to that step nightly -- which it should), so not necessary.

PR #4023, opened 2023-01-19
Feature/excel sheets

### LINKED ISSUE #3995: [IMPROVE] Have the ability to specify sheet name
When collecting a lot of data, it is cumbersome if all pieces of data are inside separate Excel sheets especially if you have no background in programming and just need the data for whatever type of research you are doing. As an example, if you make use of the `routines` functionality to download e.g. fundamental data it can be neat that if you are doing so for multiple companies you can specify the sheet name.

**This is mostly important to work directly in Excel with the data as that works neat when its in the same Excel file and very poorly when its in multiple.** It isn't too hard to apply some Excel magic to combine them or to do any type of analysis on them as I so desire (see example below). The purpose this serves is to make it more accessible to people that are looking for data and know their way around Excel.

image

So for example as depicted with this piece of a routine script (.openbb) it would store every single in a seperate Excel sheet.
```
# Enter stocks
stocks

# Download DATA!
load MMM -s 2022-01-01 -e 2023-01-01 --export xlsx
load AOS -s 2022-01-01 -e 2023-01-01 --export xlsx
load ABT -s 2022-01-01 -e 2023-01-01 --export xlsx
...
```

See result:
image

Instead, you would want one file that has all of these combined:

image

Something like `load AAPL -s 2022-01-01 -e 2023-01-01 --export Market Data.xlsx --sheet_name AAPL` could work whatever is tidy and possible. Having everything on one tab is also an option but this requires more work than the proposed solution especially if you want to use it for fundamental data too (which results in 3D DataFrames).

jmaslek:
Fixes #3995 

As asked, apparently excel folks like to have 500 sheets instead of 500 files.  So this adds the ability to specify a sheet name.  If you run:

```
2023 Jan 19, 16:09 (🦋) /stocks/ $ load aapl --export stonk.xlsx --sheet-name aapl
2023 Jan 19, 16:09 (🦋) /stocks/ $ load msft --export stonk.xlsx --sheet-name msft
2023 Jan 19, 16:09 (🦋) /stocks/ $ load wm --export stonk.xlsx --sheet-name wm
```

You will now get a single xlsx file with 3 sheets. woohoo.

But I wasnt done.  What happens if I do this for 2000 tickers then get lazy and do
```
load something --export stonk.xlsx
```

My stonk.xlsx is gone!  The horror.  So much time for nothing.  If only..... oh wait.  Whats this?

```
2023 Jan 19, 16:09 (🦋) /stocks/ $ load aapl --export stonk.xlsx

Loading Daily data for AAPL with starting period 2020-01-15.

File already exists. Overwrite? [y/n]: 
```

Is the terminal asking me if I want to overwrite the file?  Whew.  All that work would have been for nothing.  (I will still overwrite the sheet name).

So - where does one find this?  Well, since we have base controllers, this will be in anything that inherits from the base controller!

Can I do this in menu xxx ?  Not yet.  But, wIth one easy step, you just need to add "--sheet-name" as an argument and pass it to export data!  

EDIT:  Can I do this in menu xxx ?  ALMOST.  If you go to ANY function with an xlsx export option, you can see that --sheet-name is in the list of options.  -- But it doesn't work--  

LUCKY US:  @JerBouma has volunteered his time to go through and add the ns_parser.sheet_name to EVERY SINGLE export function.  What a guy.

MORE:  If you want to play openbb on hard mode, you can enable the feature flag: FILE_OVERWRITE to ignore this extra check.


JerBouma:
I've never seen such a beautiful addition to the OpenBB Terminal. 🦋

montezdesousa (approved):

PR #4032, opened 2023-01-22
Add back pull request to linting action

jmaslek:
We had edited the action, so that it would run on any pull_request.  That was removed and I am not sure why.

Currently, we will not run the required action on any PR from a fork, even if the naming is correct.  

EDIT: Also getting rid of the charset line in conda yaml.  Messed up a rebase sometime and accidentilly merged that line in a previous PR


luqmanbello:
We can remove the `pull_request` trigger since linting should happen when commiting to the branches directly.

I will commit a change to that effect now.

luqmanbello (approved):

PR #4033, opened 2023-01-22
Refactor requests throughout

jmaslek:
Abstracting our requests calls so that we can 1) not duplicate so much code and 2) get ruff up to date with them yelling at our non timeouts.

This will also let us investigate some of this SSL certificate stuff


andrewkenreich (approved):
Looks B-E-A-U-tiful

PR #4035, opened 2023-01-23
Copy paste all the sheet names

JerBouma:
# Description

This PR adds `sheet_name` to all the functions that have the ability to export. This is directly related to #4023 that added the base implementation of being able to specify to what sheet you wish to export. This functionality is relevant because when a user would like to continue from within Excel, it is very cumbersome all data you need gets exported to (e.g.) 5-10 Excel files instead of one.

Example of output: [msft.xlsx](https://github.com/OpenBB-finance/OpenBBTerminal/files/10480036/msft.xlsx)

I've also updated the documentation to include explanation on the `--sheet-name` parameter.

# How has this been tested?

I've tested this by running the functionalities at random. Also, I've made sure all linting ran flawlessly going with pylint through each individual "menu" and fixing any issues. Furthermore, looked at all tests and fixed them were neccesary. 

# Checklist:

- [**Irrelevant**] Update [our Hugo documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).
- [**Irrelevant** ] Update our tests following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/tests).
- [ **Irrelevant** ] If a feature was added make sure to add it to the corresponding [scripts file](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/scripts).

# Others


JerBouma:
@jmaslek Locally tests seem to pass but here it gives all kinds of decoding errors. Any clue?

jmaslek:
> @jmaslek Locally tests seem to pass but here it gives all kinds of decoding errors. Any clue?

Not completely sure, but if you are rewriting stuff that involves yfinance -- you need to delete the local yf cache before running.  You can find it in ` /Users/YOURDIRECTORY/Library/Application Support/py-yfinance`

JerBouma:
Got it! That can very well be it. 👍🏼

JerBouma:
I've fixed all the integration tests however some now fail due to #4044.

jmaslek:
Hate to break it to you, but in export data:

```
" ".join(sheet_name) if sheet_name else None
```
Is not the right implementation.  The _view functions should accept a string (or None) and that string should be generated from the parser/controller  as `" ".join(ns_parser.sheet_name)`.  

I also am noticing that I have a bug in the initial implementation. IF only add the sheet-name if we use the `EXPORT_ONLY_RAW_DATA_ALLOWED` when in fact it should be that OR `EXPORT_BOTH_RAW_DATA_AND_FIGURES`

JerBouma:
> 

Uff, oh well at least from here on it's mostly Find & Replace.

JerBouma:
Hi @jmaslek,

Can you see if you can merge this PR in today, it touches so many files that it will cause merge conflicts when I make a bunch of changes to the Terminal regarding Yahoo Finance? I've made the following changes:

- Changed everything to be what you suggested.
- Also made sure `sheet_name` is `None` by default since `None` just means "Sheet1" whereas `" "` results in an error.
- Updated the documentation so people understand this exists.
- Updated the `parse_known_args_and_warn` by adding `EXPORT_BOTH_RAW_DATA_AND_FIGURES` too

See:

```
# If excel is an option, add the sheet name
if export_allowed in [
    EXPORT_ONLY_RAW_DATA_ALLOWED,
    EXPORT_BOTH_RAW_DATA_AND_FIGURES,
]:
    parser.add_argument(
        "--sheet-name",
        dest="sheet_name",
        default=None,
        nargs="+",
        help="Name of excel sheet to save data to. Only valid for .xlsx files.",
    )
```

jmaslek (approved):

PR #4057, opened 2023-01-27
Removes the "mutual funds" from the code base

### LINKED ISSUE #3975: [Bug] Docs - Mutual Fund in the Terminal Reference Section Shouldn't Exist
It's being automatically generated because it still exists as a module in the code base. 

https://docs.openbb.co/terminal/reference#funds

Noticing another small quirk where you cannot navigate back to the #menu.

Screenshot 2023-01-15 at 9 46 23 AM

hjoaquim:
^ and fix #3975

> Disclaimer: in the future we can always bring it back.


jmaslek:
So I think some of these deletions may be undone in #4038.  May want to cross check just to avoid a ton of merge conflicts.

hjoaquim:
My bad, was not aware of that PR - which is very cool btw!
Will keep this as draft in the meantime and when it's merged I'll do the necessary adjustments.

JerBouma:
I would suggest that we rename `etfs` to `funds` and integrate any still working functionality in there. The reason for that is that an ETF is a fund too and shares the same characteristics on Yahoo Finance making combining the two relatively easy to do.

I am 100% fine if all functionalities need to be deleted it's just that the whole ETF menu also works for Funds and the benefit this also has is that we aren't dropping an asset class.

Thoughts? @jmaslek @hjoaquim

hjoaquim:
I think if we can get #4068 merged, we can keep the `funds` menu a separate one

hjoaquim:
Closing this PR in favor of #4068 and #4113

PR #4058, opened 2023-01-27
`openbb.economy.fred_ids` to always return dataframe

### LINKED ISSUE #3621: [Bug] openbb.economy.fred_ids returns None instead of pd.DataFrame()
These models should return pd.DataFrame so that subsequent commands do not give AttriuteErrors.

hjoaquim:
Fix #3621 and adds tests to that function.


jmaslek (approved):
good one

Issue #4154, opened 2023-02-06
[Bug] Stocks/FA/enterprise - Error: 'Namespace' object has no attribute 'source'


deeleeramone:
Just trying out the latest merge to the develop branch, the ticker loaded is AAPL.

```
2023 Feb 06, 18:27 (🦋) /stocks/fa/ $ enterprise

Error: 'Namespace' object has no attribute 'source'
```

JerBouma:
?

<img width="1209" alt="image" src="https://user-images.githubusercontent.com/46355364/217271311-f0a15080-397c-4878-bb09-e0a1235deef7.png">

deeleeramone:
It works on the main branch:
![Screenshot 2023-02-07 at 6 32 07 AM](https://user-images.githubusercontent.com/85772166/217273680-33aaf62c-09b2-4f84-8f05-0af4f2cfa5c4.png)


On Develop branch, I get the changes to the menu, but `enterprise`  does not work. 
```
(obb-m1) Danglewoods-MacBook-Pro:OpenBBTerminal danglewood$ git branch
* (HEAD detached at origin/develop)
```

![Screenshot 2023-02-07 at 6 35 39 AM](https://user-images.githubusercontent.com/85772166/217274411-d00dbcd2-58e4-4b7f-bc09-973351736af2.png)

JerBouma:
What commit are you on, I am on the develop branch.

JerBouma:
Seems your `data_sources_default.json` is not updated. Please pull the latest develop commit.

deeleeramone:
`HEAD is now at ef12295f33 Remove code of legacy jupyter lab extensions from the repo (#4146)`

JerBouma:
See this file: https://github.com/OpenBB-finance/OpenBBTerminal/blob/develop/openbb_terminal/miscellaneous/data_sources_default.json. Yours is old.

deeleeramone:
I think you're the one with old code, your branch is missing a function that was added in 2.3.1:
 
![Screenshot 2023-02-07 at 6 46 18 AM](https://user-images.githubusercontent.com/85772166/217277385-435e8ed5-76e3-44de-88c7-8635ccbc835b.png)

This is yours:
![Screenshot 2023-02-07 at 6 44 29 AM](https://user-images.githubusercontent.com/85772166/217276780-dfb7b530-1627-44c6-8e62-07cc0ef5aa0b.png)

[...]

Issue #4022, opened 2023-01-19
[Bug] Error when charting multiple country indices


jerryduluk:
# Bug #
Trying command: `/economy/ $ index --indices sp500,ftse100,jp_n225,hk_hsi -s 2022-01-01`
Only the sp500 is coming up. However, when individually requested, each index is able to load

Add “-r” and I get “Error: local variable “plot_data’ referenced before assignment

![Screenshot 2023-01-19 at 10 57 56 AM](https://user-images.githubusercontent.com/123030228/213535286-03a02a2a-be17-4be9-ae72-d608fe0349d1.png)

JerBouma:
Also related to #4020

PR #4155, opened 2023-02-07
Fixing smaller issues and moving a few things around

### LINKED ISSUE #4020: [Bug] Economy Menu - Plotting function not plotting two things
One works:

`/economy/index sp500`

Screenshot 2023-01-19 at 9 41 27 AM

`/economy/index hk_hsi`

Screenshot 2023-01-19 at 9 42 24 AM

Two together, no dice:

`/economy/index sp500,hk_hsi`

Screenshot 2023-01-19 at 9 43 27 AM

Plotting using the `plot` function also produces the wrong scale for the second y-axis:

`plot --y1 sp500 --y2 hk_hsi`

Screenshot 2023-01-19 at 9 45 17 AM

### LINKED ISSUE #4044: [Bug] stocks/fa/info - Error: Yahoo has again changed data format, yfinance now unsure which key(s) is for decryption
Recreate with:

`/stocks/load aapl/fa/info`

image

Related to this: https://github.com/ranaroussi/yfinance/issues/1329

### LINKED ISSUE #4082: [IMPROVE] Treasury rates from `economy/treasury` are wrongly exported to Excel
If you export to Excel the number is off (`/economy/treasury --export xlsx`). The data is meant to be depicted as a percentage in Excel while it should be Excel doing the formatting there. So instead of the value being 4.71 it should be 0.0471. For example, if I would calculate the Sharpe Ratio and use this number to deduct from a stock's annual return, I would get very odd results.

So pretty simple divided by 100 in the code before exporting for all series 😉 

image

JerBouma:
This changes the following things:
- Moved `arktrades` to `screener`. It has no business being in `stocks/fa`.
- Made fixes to the Equity report. This tended to refer to old data, now runs smoothly.
- Updated the docs to remove any mention of `dd` as this menu is integrated in `stocks/fa`
- Updated SDK paths where they were pointing to the wrong location.
- Moved the `filings` command back in but moved it to `stocks/disc` since it has no business being on the home page of `stocks` fixing #4162.
- I saw a small mistake with sheet_name again, saying `heet_name` in `call_shrs`. Fixed that too.
- Fixed treasury rates not displaying properly when exported. It was a simple `/ 100`. See this source: https://www.federalreserve.gov/releases/h15/ (same as EconDB which we get data from) for the 10 year and then try `economy/treasury --export xlsx`. Previously this gave you 353%, a bit high don't you think? Fixes #4082
- Indices couldn't be plotted due to missing `ignore_tz=True`. Now fixed which fixes #4020 and #4022.
- Adding fast_info to `overview --source YahooFinance` which fixes  #4044.


jmaslek:
plz fix #4154 as well

JerBouma:
?

image

jmaslek:
> ?
> 
> image

You fixed it.  Magic.

JerBouma:
Impressed by myself sometimes that running the command is enough

jmaslek:
Note that on the reports, we run integration test on those, so if a key is not provided (which they aren't in the CI), they should not fail. Can you confirm this by changing your api key to something invalid or deleting from the env file?

JerBouma:
@jmaslek I did it mom, deleted .env and this is the result.

image

jmaslek (approved):

JerBouma:
🥳

PR #4157, opened 2023-02-07
SDK login with OpenBB token

hjoaquim:
In this PR:

- Save and check status of OpenBB key directly on the terminal
- Login to the SDK using just the OpenBB key 
   - Session can be kept if `keep_session=True` - the key will not be saved
   -  To save the OpenBB key, one must use `openbb.keys.openbb("key")` - just like the other keys


hjoaquim:
@montezdesousa thought that could be cool to use a stored OpenBB key to log in when there is no local session and any arg is provided, but that kinda conflicts with the fact that if there is no session in place, no keys are available, right?

montezdesousa:
> @montezdesousa thought that could be cool to use a stored OpenBB key to log in when there is no local session and any arg is provided, but that kinda conflicts with the fact that if there is no session in place, no keys are available, right?

Sounds good, but where do we store the OpenBB key?

hjoaquim:
@montezdesousa on the `.env`. But that is kind of what me want to avoid, right?

hjoaquim:
@montezdesousa we could have a special flag like "write to local storage", just for that one tho

montezdesousa:
> @montezdesousa on the `.env`. But that is kind of what me want to avoid, right?

that might work

PR #4178, opened 2023-02-08
[FEAT] Additional metrics for evaluating forecasting (RMSE+ MSE)

martinb-bb:
Adding in `RMSE` and `MSE` metrics for accuracy calculations when performing backtesting on models.

[Twitter](https://twitter.com/adam_cjohnson/status/1623154046331924480?s=46&t=Mrg0MsCFae0uDE2RCvW7kA)
image

In terminal: 
image
image

SDK:
image


martinb-bb:
@DidierRLopes ☝️

jmaslek:
Any reason we can't add the others from darts.metrics?

jmaslek:
Thats fair.  Only reason I ask is because I was going to try using the sdk for a kaggle comp that used SMAPE as the scoring metric, but I noticed it wasn't an option, so I didnt use it.  Having as many is both helpful and overkill.

martinb-bb:
Added smape for you :) 
image

jmaslek (approved):

PR #4183, opened 2023-02-09
Restores numeric values

### LINKED ISSUE #3872: [IMPROVE] Ensure all exports to Excel are always numbers and not strings
Whenever you export something to Excel, e.g. `stocks/load MSFT/income --source Polygon --export xlsx` you will receive output that includes Bs, Ms and similar which will always require data adjustments if you wish to do anything with the data. This is very undesirable because it requires additional work from the user side in which it can question why not just get this from somewhere else if it's such a burden to edit?

There needs to be a helper functionality that automatically recognizes these type of numbers and converts it accordingly.

image

hjoaquim:
This PR aims to remove the formatting that sometimes is done in order to beautifully present tables in the terminal - which latter causes numeric values to be presented as alphanumeric (specially in Excel sheets).

Fix #3872

> Also took to opportunity to do a minor cleaning on `helper_funcs.py`


JerBouma:
I'll look at this in a bit thank you.

JerBouma (approved):
Thank you for this, this is really great and helpful.

Issue #4182, opened 2023-02-09
[Bug] Stock splits are not correctly accounted for within `portfolio`


JerBouma:
As noted here: https://twitter.com/jrp980/status/1623386325142347784?s=46&t=_PF5iVHF-dn4emakfIM6ag

If I purchase a share of Tesla at 890 on July 2022 and the stock split occurs on Aug 2022 (3:1), my initial order size needs to be adjusted (3 shares instead of 1). Currently, this isn't happening while the stock price data **is** adjusted. So when we display a table like below (or any other performance metric) it falsely tells me I made a massive loss on some tickers.

This can be relatively easy acquired by collecting split data which we already have within `stocks/fa/splits` and multiplying volumes when orders are before the split. Do let the user know this is happening. A flag  ("adjust for splits") shouldn't be required since it is entirely wrong without the adjustment.

```
(🦋) /portfolio/ $ perf --show_trades

                              Portfolio vs. Benchmark - Individual Trades in period
┏━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┓
┃ Date       ┃ Ticker ┃ Portfolio Value ┃ Portfolio % Return ┃ Benchmark Value ┃ Benchmark % Return ┃ Alpha     ┃
┡━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━┩
│ 2010-05-03 │ GOOGL  │ 303.84          │ 863.35%            │ 137.32          │ 335.38%            │ 527.97%   │
├────────────┼────────┼─────────────────┼────────────────────┼─────────────────┼────────────────────┼───────────┤
│ 2010-07-06 │ AMZN   │ 3313.95         │ 10765.41%          │ 154.62          │ 406.95%            │ 10358.46% │
├────────────┼────────┼─────────────────┼────────────────────┼─────────────────┼────────────────────┼───────────┤
│ 2011-08-06 │ AAPL   │ 298.11          │ 667.73%            │ 196.85          │ 406.95%            │ 260.78%   │
[...]

PR #4195, opened 2023-02-09
Adjust for stock splits in portfolio engine

rajanprabu:
@JerBouma  @jmaslek  

As we discussed over twitter (  https://twitter.com/jrp980/status/1623386325142347784?s=46&t=_PF5iVHF-dn4emakfIM6ag) about the stock split bug in portfolio_engine. Here is the fix to adjust for stock splits. 

https://github.com/OpenBB-finance/OpenBBTerminal/issues/4182

here is the PR for that. Its done in step 5 of __preprocess_transactions.


openbb_terminal/portfolio/portfolio_engine.py
```diff
@@ -284,6 +285,19 @@ def __preprocess_transactions(self):
...
             )
+            
+            # Adjust quantity and price for splits 
+            for ticker in self.__transactions["Ticker"].unique() :
+                try :
+                    splits_df = get_splits(ticker)
+                    if not splits_df.empty :
+                        splits_df = splits_df.tz_localize(tz=None)
+                        for split_date in splits_df.index :
+                            self.__transactions['Quantity'] = np.where((self.__transactions['Ticker'] == ticker) & (self.__transactions['Date'] < split_date), self.__transactions['Quantity'] * splits_df.loc[split_date].values , self.__transactions['Quantity'])
```
jmaslek, line 296: I think you have the condition reversed.  If the date is > split_date, then we multiply.  For np.where it yields the first value if condition is True.

We also need to be careful and make sure that get_splits is sliced so that anything that happened before our initial purchase doesn't count )i.e, `splits_df = splits_df[splits_df.index > self.__transactions['Date'][0]]`

openbb_terminal/portfolio/portfolio_engine.py
```diff
@@ -284,6 +285,19 @@ def __preprocess_transactions(self):
...
+                            self.__transactions['Price'] = np.where((self.__transactions['Ticker'] == ticker) & (self.__transactions['Date'] < split_date), self.__transactions['Price'] / splits_df.loc[split_date].values , self.__transactions['Price'])
```
jmaslek, line 297: Also isn't this not needed because the price is already split adjusted (which is why we need to adjust quantity in the first place)?  If the value is Price * Quantity and its incorrect, then Price / Split * Quantity * Split is the same and will still be incorrect.

rajanprabu: yes Im aware that we use Adj. Close. But if you don't adjust the price here in get_transactions one will get quantity that is after split but price that is before split. Hence its introduced here.

openbb_terminal/portfolio/portfolio_engine.py
rajanprabu: If we are holding the scrip before the event (split)  then quantity will be multiplied. I think that the condition is right. I request you to run with an example and check. AAPL for example 1 share bought before 1987-06-16 should be 224 shares now. 

Slicing is also taken cared off. AAPL shares bought during 2010 should be 28 shared today.

jmaslek: So I have an example.  If I bought 1 share of appl at the beginning of 2005, it underwent splits of 2,7,4, so 1 share then = 56 now.  Using your code says we initially bought 56:
Screenshot 2023-02-10 at 8 30 14 AM

If I switch the < to > but dont slice split, it accounts for everything historical, saying I started at 4 shares
Screenshot 2023-02-10 at 8 29 13 AM

If I slice splits to not include before we bought, we get the desired behavior:

Screenshot 2023-02-10 at 8 31 49 AM

jmaslek: I think I see where I am looking differently.  Your method is correct if I enter into the template how many shares I bought from adjusted price.  So if  I bought 400$ of appl a few years ago, I only got 1 share.  If you enter that as 4 shares at 100$ into your excel, then your split adjustment is correct.  If you enter into the excel as 1 share at 400, then you need to do the reverse way I have here.

rajanprabu: In the PR I have inserted the code in the portfolio_engine.py under the preprocess_transactions..  I modify the quantity immediately after reading the transactions so that rest is taken cared of. I checked with GOOG and AAPL and it worked correctly.

rajanprabu: I see whats conflicting here. You are applying to the data frame that contains historical price data. Im applying the logic to transactions itself ( before it gets the historical price). I found it easier to adjust price and quantity in transactions itself.

JerBouma:
I haven't looked into the code but basically what my expectation is, is that the volume is corrected after the split in which the user is informed this happend. So the following (values are random). Please note the remark regarding Yahoo Finance. If this source is used, the split is factored in within **all historical data**.

image

[Book1.xlsx](https://github.com/OpenBB-finance/OpenBBTerminal/files/10720911/Book1.xlsx)

rajanprabu:
@JerBouma  for the example here

https://github.com/OpenBB-finance/OpenBBTerminal/pull/4195#discussion_r1102819521
<!DOCTYPE html>

Date | Ticker | Type | Price | Quantity | Side | Currency | Fees | yf_Ticker
-- | -- | -- | -- | -- | -- | -- | -- | --
02/11/22 | GOOGL | Stock | 89.25 | 5 | Buy | USD | 1 | GOOGL
27/09/22 | GOOGL | Stock | 99.75 | 5 | Buy | USD | 1 | GOOGL
26/04/22 | GOOGL | Stock | 2400 | 1 | Buy | USD | 1 | GOOGL
22/04/22 | GOOGL | Stock | 2400 | 1 | Buy | USD | 1 | GOOGL


with the modified code.. I got the following  as portfolio returns 

Screenshot 2023-02-20 at 21 30 03

JerBouma:
> @JerBouma for the example here
> 
> [#4195 (comment)](https://github.com/OpenBB-finance/OpenBBTerminal/pull/4195#discussion_r1102819521)
> 
> Date	Ticker	Type	Price	Quantity	Side	Currency	Fees	yf_Ticker
> 02/11/22	GOOGL	Stock	89.25	5	Buy	USD	1	GOOGL
> 27/09/22	GOOGL	Stock	99.75	5	Buy	USD	1	GOOGL
> 26/04/22	GOOGL	Stock	2400	1	Buy	USD	1	GOOGL
> 22/04/22	GOOGL	Stock	2400	1	Buy	USD	1	GOOGL
> with the modified code.. I got the following as portfolio returns
> 
> Screenshot 2023-02-20 at 21 30 03

Yes, this is correct. @jmaslek Can you take it from here? This is how it should be. Note again that Yahoo Finance adjusts **all** prices also those before the split.

jmaslek:
can merge once the linters pass

jmaslek (approved):

PR #4206, opened 2023-02-10
Login on the SDK using the OpenBB Personal Access Token

hjoaquim:
In this PR:

- Save and check status of OpenBB key directly on the terminal
- Login to the SDK using just the OpenBB key 
   - Session can be kept if `keep_session=True` - the key will not be saved
   -  To save the OpenBB key, one must use `openbb.keys.openbb("key")` - just like the other keys


montezdesousa (approved):
Looks good to me

PR #4213, opened 2023-02-10
Feature/general improvements

jmaslek:
This PR mainly allows users to enter any stocks submenu without doing load first (except for research).

Within the fa menu, this adds ticker arguments to all functions, ie (cash -t aapl) instead of load appl/cash.

### DIFF
--- a/openbb_terminal/stocks/stocks_controller.py
+++ b/openbb_terminal/stocks/stocks_controller.py
@@ -25,6 +25,7 @@ from openbb_terminal.menu import session
 from openbb_terminal.parent_classes import StockBaseController
 from openbb_terminal.rich_config import MenuText, console, translate
 from openbb_terminal.stocks import cboe_view, stocks_helper, stocks_view
+from openbb_terminal.terminal_helper import suppress_stdout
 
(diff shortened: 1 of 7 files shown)


openbb_terminal/stocks/stocks_controller.py
```diff
@@ -106,7 +107,6 @@ def print_help(self):
         mt.add_cmd("load")
         mt.add_raw("\n")
         mt.add_param("_ticker", stock_text)
-        mt.add_raw(self.add_info)
         mt.add_raw("\n")
         mt.add_cmd("quote", self.ticker)
```
hjoaquim, line 111: Regarding `quote, tob, candle, codes, news`, if those are usable upon `--ticker` arg, shouldn't they be blueish? Also, why exclude `candle` and `news?`

hjoaquim (approved):
Looks great! I think this will improve a lot the UX

openbb_terminal/stocks/stocks_controller.py
jmaslek: Yes and I forgot them :(

PR #4215, opened 2023-02-10
Sunset some dependencies

jmaslek:
This PR removes the following deps from our tree:

- sentiment-investor : unused
- investpy : api has not recovered from months ago
- investiny : same as above
- pyEx: unmaintained, unused and iex is deprecating the api it hits


jmaslek:
> Looks good to me.
> 
> Side question: the `.md` files for the website will be auto-regenerated, right?

The reference docs will be, yes.

PR #4217, opened 2023-02-11
Display correct risk free rate in the portfolio menu help message

### LINKED ISSUE #4200: [Bug] Portfolio/load - RFR displayed is *0.01 what is entered. Don't know what the actual value for calculations is.
Do I enter it as 300% to get the correct calculations? Unclear. 

image

Screenshot 2023-02-09 at 8 58 20 PM

northern-64bit:
# Description

Closes #4200 which was caused by saving incorrect value of the rfr.

# How has this been tested?

Running the command multiple time with different risk free rates and checking which rfr value is used for the calculations.

# Checklist:

# Others


openbb_terminal/portfolio/portfolio_controller.py
```diff
@@ -356,9 +356,9 @@ def call_load(self, other_args: List[str]):
                 f"\n[bold][param]Portfolio:[/param][/bold] {self.portfolio_name}"
             )
 
-            self.risk_free_rate = ns_parser.risk_free_rate / 100
+            self.risk_free_rate = ns_parser.risk_free_rate
             console.print(
-                f"[bold][param]Risk Free Rate:[/param][/bold] {self.risk_free_rate:.2%}"
+                f"[bold][param]Risk Free Rate:[/param][/bold] {self.risk_free_rate}%"
```
jmaslek, line 361: What is the desired behaviour?  If I enter 3.  Is that 3% -> 0.03?

If so then we want to keep the /100 and print 100*rate% ?

JerBouma: The original intention was to put it in %. So if you put in 5 it would convert it to 5%.  The proxy for the risk-free rate tends to be pretty low (0% - 5%) meaning that you'd need to enter a bunch of zeros anyways. In 2020 at some point it was 0.65% which would be 0.0065 in the Terminal.

If we do not have this behaviour anywhere else in the Terminal perhaps 0.0065 would still be better though.

northern-64bit: > If so then we want to keep the /100 and print 100*rate% ?

The problem is that every other command also divides the risk free rate input by hundred, thus it would divide the default risk free rate twice:
Input: 5  ---(Saves to)---> 0.05 ---(Gets executed as)---> 0.0005; thus resulting in the incorrect result

jmaslek (approved):

PR #4234, opened 2023-02-14
Minor fixes on login

hjoaquim:
Aims to fix:

- banner not showing if credentials expire  --> Fixed :heavy_check_mark: 
- `openbb --login` not working  --> Fixed :heavy_check_mark:


Chavithra:
Let's leave the conditions like this for now.

jmaslek:
> @Chavithra with the conditions as it's now on the `develop` branch, I can't launch the terminal without an account. Is that the default behavior?

This should not be the case.  There should be nothing about accounts on develop yet by default

hjoaquim:
^I misunderstood the purpose of the condition and reverted to the original state

hjoaquim:
> The dependency tree has been changed. Please update requirements.txt and requirements-full.txt

why?

montezdesousa (commented):
`openbb --login` does not yield the prompt if there is no session save, it logs the user as guest. Is this the desired behavior?

jmaslek:
> > The dependency tree has been changed. Please update requirements.txt and requirements-full.txt
> 
> why?

ignore.  bad bot

montezdesousa:
> `openbb --login` does not yield the prompt if there is no session save, it logs the user as guest. Is this the desired behavior?

`poetry install` solved it

montezdesousa (approved):

PR #4236, opened 2023-02-14
Code Coverage

luqmanbello:
- Implement code coverage


Chavithra:
Hi @luqmanbello,

fixed the tests.

I think we should combine "coverage" CI and tests CI.

Or something like this to avoid running the tests twice.

Chavithra (approved):
Looks ok to me.

luqmanbello:
> Hi @luqmanbello,
> 
> fixed the tests.
> 
> I think we should combine "coverage" CI and tests CI.
> 
> Or something like this to avoid running the tests twice.
> 
> Thanks

Since we are not enforcing it yet, we can combine it with the unit test

jmaslek:
Should we have the code cov depend on the check for changes so that the tests arent running when no .py files are changed?

luqmanbello:
> Should we have the code cov depend on the check for changes so that the tests arent running when no .py files are changed?

That's what I just implemented. it will also only run on the `base-test`

PR #4245, opened 2023-02-15
Update documentation

DidierRLopes:
* Simplifies and improves main README of the project.


website/content/sdk/quickstart/pypi.md
```diff
@@ -0,0 +1,50 @@
...
+More information on this product can be found [here](https://openbb.co/products/sdk)
+
+
+## Installation
+
+The command below provides access to the core functionalities behind the [OpenBB Terminal](https://openbb.co/products/terminal).
+
+```python
+pip install openbb
+```
```
JerBouma, line 23: This goes against what we mention in the docs since we mention to use Miniconda and install an environment specifically. This could result in weird bugs when people do not expect our package to bring all these kind of packages. Generally, when you install something like Pandas this is all good, for us it isn't. See: https://docs.openbb.co/sdk/quickstart/installation

website/content/sdk/quickstart/pypi.md
```diff
@@ -0,0 +1,50 @@
...
+Everything you need to use the OpenBB SDK can be found on our [Official Documentation](https://docs.openbb.co/sdk).
+
+Main uses cases can be seen below.
+___
+
+#### RAW FINANCIAL DATA AT YOUR FINGERTIPS
+
+Access raw financial data from any data source that you are interested. The open source nature of this SDK makes it so that this is an ever-growing project and that everyone can contribute to.
+
+![image-20221025-092439 2](https://user-images.githubusercontent.com/25267873/218906336-cebd1fc8-7e7a-45bc-a5fc-641eb19c3e8c.png)
```
JerBouma, line 44: So I get that you want to show examples but both are not impressive and do not really show the true capability of the SDK. Collecting historical data I can also use `yfinance` for, why do I need this package?

website/content/sdk/quickstart/pypi.md
```diff
@@ -0,0 +1,50 @@
...
+
+#### GENERATE INSIGHTS 10X FASTER
+
+For most of the functionalities, adding `_chart` to the function will allow you to visualize the output directly from a Jupyter Notebook. Not only speeding up the time it takes to create a plot for the data you are interested in, but making building custom reports much faster.
+
+<img width="992" alt="Group 550" src="https://user-images.githubusercontent.com/25267873/218906112-b2272d43-11fc-4ec1-9a8f-b2d8e2ed7dc1.png">
```
JerBouma, line 50: Why is one of the key use cases having the ability to create charts?

JerBouma (commented):
Up to you but I am not a fan of this. I believe we need to push people to the docs to result in confusion since your installation instructions are different than we have in the Docs. Next to that, the use-cases can be much better explained (and most importantly, updated) from within the docs. This will otherwise be yet another document we need to keep up-to-date.

jmaslek:
You will also need to change this line in pyproject.toml :

```
readme = "website/content/sdk/quickstart/installation.md"
```

website/content/sdk/quickstart/pypi.md
DidierRLopes: Because allows you to retrieve data much faster, if you want to have a quick POC. Used the same that we have on the marketing website.

DidierRLopes:
> please rename your branch to start with feature/ or hotfix/

Will do!

Issue #2764, opened 2022-10-10
[FR] Improve Shroom 


hjoaquim:
In the crypto menu there are already some commands related to Shroom:

Menu:  `crypto/onchain`
Commands: `tvl`, `ds`, `dt`

There is some space for improvements/ adding more shroom features, such as:

- add a command where users can provide their own queries and use the output (will be really cool for reports)
- add more commands that consume shroom sdk

Resources:
- https://sdk.flipsidecrypto.xyz/shroomdk
- https://docs.flipsidecrypto.com/our-data/tutorials

PR #4246, opened 2023-02-15
Adds shroom command and removes old ones

### LINKED ISSUE #3943: [Bug] `crypto/onchain/dt` - Error: Error creating query, got response: {"errors":"SQL compilation error: Schema 'FLIPSIDE_PROD_DB.ETHEREUM' does not exist or not authorized."}
All three functions using the ShroomDK need to be updated. 

`FLIPSIDE_PROD_DB.ETHEREUM` is no longer supported. Endpoints to the SQL query need to be updated.

https://docs.flipsidecrypto.com/our-data/data-updates#may-2022-ethereum-schema

jose-donato:
Addresses both #2764 and #3943 

- removes old commands (flipside remove/renamed the tables we were using for those commands)
- adds new command that allows to perform any flipsidecrypto query

Fixes #3943

much more work can be done here:
- prettify tables (detect if number and format)
- allow users to display in chart (maybe wait until plotly PR is merged)
- add templates of sql queries that users can use


openbb_terminal/cryptocurrency/onchain/shroom_view.py
```diff
@@ -18,6 +19,45 @@
...
+        Export dataframe data to csv,json,xlsx file
+    """
+    df = get_query_data(query)
+    if df.empty:
+        console.print("[red]No data found.[/red]")
+    elif not df.empty:
+        if raw:
+            console.print(df)
+        else:
+            print_rich_table(df.head(limit))
```
jmaslek, line 51: dont think this needs a raw input.  raw differentiates between showing a chart and a table

jmaslek, line 51: also limit is an argument.  with the interactive, cropping here will only show a small amount

openbb_terminal/cryptocurrency/onchain/shroom_view.py
```diff
@@ -18,6 +19,45 @@
...
+
+        export_data(
+            export,
+            os.path.dirname(os.path.abspath(__file__)),
+            "query",
+            df,
+        )
```
jmaslek, line 58: No sheet name argument

jmaslek (commented):
query -h does not print help text

```
2023 Apr 27, 16:05 (🐛) /crypto/onchain/ $ query -h

usage: query -q QUERY [QUERY ...] [-h] [--export EXPORT] [--sheet-name SHEET_NAME [SHEET_NAME ...]] [-l LIMIT]
query: error: the following arguments are required: -q/--query
```

jmaslek (approved):
llstm

PR #4260, opened 2023-02-15
Feature/improve docs

DidierRLopes:
Improve main README and also PyPI page


DidierRLopes:
@jmaslek is there anything else you want me to add here?

JerBouma:
I just want to say again that the installation instructions could result in issue as they are put there. We ask people to create an environment in here (https://docs.openbb.co/sdk/quickstart/installation) for a reason. If I'd think "this is just another package" and I do `pip install openbb` in my personalised environment that results into many conflicts most likely.

DidierRLopes:
> I just want to say again that the installation instructions could result in issue as they are put there. We ask people to create an environment in here (https://docs.openbb.co/sdk/quickstart/installation) for a reason. If I'd think "this is just another package" and I do `pip install openbb` in my personalised environment that results into many conflicts most likely.

Is this still required even without portfolio optimization and forecasting out of scope?

JerBouma:
You are still blasting the environment with https://github.com/OpenBB-finance/OpenBBTerminal/blob/develop/requirements.txt with specific Pandas, NumPy etc. For some packages we stay on older version on purpose. If `pip` doesn't entirely recognize that stuff, it could be that they can't launch the Terminal or SDK because of dependency errors. I believe this was the main reason we ask people to make an environment but please @jmaslek or @piiq confirm this.

piiq:
`pip` looks at the dependency range specified in `pyproject.toml` and installs the dependencies accordingly. 
When you manually `pip install openbb` into an existing environment that has same packages that openbb depends on, if the version of an existing package does not match the range specified in pyproject.toml - the existing package will be overwritten.
This will happen with any package that you install with pip not openbb specifically. does that answer your question @JerBouma 

Let's outline what should we improve in the docs

JerBouma:
So to me this is a problem if you would install the package within an environment that is already filled with different packages. Is it not? Otherwise, why do we mention in the Docs to create an environment?

piiq:
Making a reproducible environment with a dependency tree that works nicely is a problem for everyone, not just you. This problem triggered an inception of a universe of tools. Examples of such tools include general purpose containerization tools like docker and podman and tools from the python ecosystem like conda, pyenv and poetry

My experience for creating a vanilla pip environment is to have openbb as the primary dependency and only installing on top whatever is not available in openbb. I haven't yet encountered a situation when I would specifically need a different version of some package different than the one defined by openbb

JerBouma:
Alright, thank you for the explanation. My question then is to you and @DidierRLopes:

Does it make sense that on the PyPi page we say just do `pip install openbb` and in the Docs we say "make environment etc etc". If not, let's at least say the same thing on both then.

piiq (approved):
Re-read pypi.md
I think this is just fine @JerBouma

JerBouma:
Ok all good then :)

PR #4307, opened 2023-02-20
Speed up pytest with xdist plugin

jmaslek:
Make unit tests speedy


tehcoderer:
Meh... Too slow IMO

image

Chavithra (approved):
The tests passes on my side.

Really handy for local tests run too.

PR #4378, opened 2023-03-02
Update helper_classes.py bug on line 126

colin99d:
I have not tested this fix, but the current code crashes when I make an installer from the SDK.


jmaslek:
branch name

jmaslek (approved):

PR #4386, opened 2023-03-02
Interactive tables using react

jose-donato:
[demo](https://twitter.com/josedonato__/status/1631434312095617024)

This PR adds the functionality to spin up a pywry window containing react (or any frontend library code). 

* sets foundation for adding more web components
* adds interactive tables

todo:
* add docs on how to change and how it is structured
* move all formatting code inside print_rich_table
* do the formatting inside (after sending the raw df to the pywry window)
* style with our components


JerBouma:
Another must have is being able to filter by dates. I think this change will actually allow the `limit` flag to be removed. Due to how easy it is to search, we don't need that flag anymore.  For a lot of functionality the limit flag is just there to reduce clutter while with this table functionality it no longer is clutter since you can filter so well. 

In my opinion just nice to have not necessarily must-have:

**1.** It would be nice if we can have clickable URLs as seen in the screenshot below. Here it is getting cut-off (and thus not usable). We got that complaint a couple of times, also happens in `stocks/fa/mgmt`.

image

**2.** What's powerful is that you can **plot** a series within this table.  So for example if I obtain the income statement as you see here, if I could click a row and press plot that's powerful. Then you can see revenue growth over time with ease etcetera. Probably a lot more use cases outside of this one.

image

JerBouma:
Do limit even higher if possible, some data series are based on dates that can go all the way back to e.g. 1980 and offer daily data. Unless this would slow everything down a lot of course!

JerBouma:
I'd say go with a thousand for now. Little bit worried we might cut-off some data in due time so there needs to be a more dynamic solution at some point in time (basically no limit).

image

jose-donato:
> I'd say go with a thousand for now. Little bit worried we might cut-off some data in due time so there needs to be a more dynamic solution at some point in time (basically no limit).
> 
> image

I like the no limit but have to talk with @jmaslek for the best way to do this without breaking anything

jmaslek (approved):

PR #4388, opened 2023-03-02
Feature/fmp endpoints

jmaslek:
As per @minhhoang1023's request I have added the desired endpoints to the sdk.  They are not standardized in any way, just wrapping the endpoints.

Next step will be to integrate these into the terminal, but that would blow up the size of this and we decided not to do that

To test:
```
python --m doctest openbb_terminal/core/models/fmp.py
```


openbb_terminal/core/models/fmp.py
```diff
@@ -0,0 +1,840 @@
...
+        -------
+        dict
+            Dictionary containing the revenue by segment
+
+        Examples
+        --------
+        >>> from openbb_terminal.sdk import openbb
+        >>> revseg = openbb.fmp.revseg("AAPL")
+        """
+        # TODO: make dataframe since this is not a fixed shape
```
minhhoang1023, line 347: was about to ask this as a DF :P We should also add a way to query the year / or quarter. This is because when going back more than 5 years, they completely changed their segment names. 
image

minhhoang1023 (commented):
Can you add this endpoint too? It provides the list of holders' names 🙌 https://site.financialmodelingprep.com/developer/docs#Stock-Ownership-by-Holders

openbb_terminal/core/models/fmp.py
```diff
@@ -0,0 +1,840 @@
...
+        >>> from openbb_terminal.sdk import openbb
+        >>> ratios = openbb.fmp.ratios("AAPL")
+        """
+        url = (
+            f"{cls._BASE_URL()}ratios/{symbol}?period=quarter&"
+            if quarter
+            else f"{cls._BASE_URL()}ratios/{symbol}?"
+        )
+        r = request(url + f"apikey={cls._KEY()}&limit={limit}")
+        return pd.DataFrame(cls.validate_request(r)).T
```
minhhoang1023, line 249: This should not be transposed. 
image

openbb_terminal/core/models/fmp.py
```diff
@@ -0,0 +1,840 @@
...
+            Historical income statement growth
+
+        Examples
+        --------
+        >>> from openbb_terminal.sdk import openbb
+        >>> income_growth = openbb.fmp.income_growth("AAPL")
+        """
+        url = f"{cls._BASE_URL()}income-statement-growth/{symbol}?"
+        r = request(url + f"apikey={cls._KEY()}&limit={limit}")
+        return pd.DataFrame(cls.validate_request(r)).T
```
minhhoang1023, line 274: This too.

openbb_terminal/core/models/fmp.py
minhhoang1023, line 249: can we remove the symbol/period rows?
image

minhhoang1023 (commented):
.

jmaslek:
> Can you add this endpoint too? It provides the list of holders' names 🙌 https://site.financialmodelingprep.com/developer/docs#Stock-Ownership-by-Holders

How do you want it to be?  The api requires a `date` argument and I can't figure out exactly what dates it wants?  I have gone back by hand and there is currently nothing from the current year.

openbb_terminal/core/models/fmp.py
minhhoang1023: yeah, that's all good. Mainly because now it became a row and not column after the transpose, it'll be easier now to remove it.

minhhoang1023:
> 

hm that's strange. i dont need it from my side. Will DM you

deeleeramone:
> How do you want it to be? The api requires a date argument and I can't figure out exactly what dates it wants? I have gone back by hand and there is currently nothing from the current year.

There example date is September 30, which would be end of calendar Q3. Maybe date needs to be one of the four? 

While were at it, let's also include the RSS feed which is the most current data:
https://financialmodelingprep.com/api/v4/institutional-ownership/rss_feed?page=0&apikey=YOUR_API_KEY

https://site.financialmodelingprep.com/developer/docs/institutional-holder-rss-feed-api/

minhhoang1023:
@jmaslek These are the endpoints I need

- **[Institutional Stock Ownership per Security]**(https://site.financialmodelingprep.com/developer/docs/institutional-stock-ownership-api) - This provides a summary of aggregated institutional ownership
- **[Stock Ownership by Holders per Security]**(https://site.financialmodelingprep.com/developer/docs/stock-ownership-by-holders-api) - This provides a breakdown in terms of the largest holders and their names. Once a quarter, these institutional investment managers are obligated to report their equity securities holding on Form 13. Because of that, we will need to provide the end-of-quarter date. They don't have the latest endpoint, but we can set the default to be the last quarter.

minhhoang1023:
> > How do you want it to be? The api requires a date argument and I can't figure out exactly what dates it wants? I have gone back by hand and there is currently nothing from the current year.
> 
> Their example date is September 30, which would be end of calendar Q3. Maybe date needs to be one of the four?
> 
> While were at it, let's also include the RSS feed which is the most current data: https://financialmodelingprep.com/api/v4/institutional-ownership/rss_feed?page=0&apikey=YOUR_API_KEY
> 
> https://site.financialmodelingprep.com/developer/docs/institutional-holder-rss-feed-api/

How does this fit into the Equity Analyst use case?

minhhoang1023 (approved):
Legoo!

jmaslek:
soooo.  are we shipping this?  this changes the sdk structure (using fmp. instead of stocks. ...) and I dont know if that gets more confusing?  @minhhoang1023

PR #4390, opened 2023-03-03
hotfix whisper folder check

martinb-bb:
Need to create folder before checking cache in case user has never run the command.


martinb-bb:
@jmaslek please merge asap when you can :)

jmaslek (approved):

Issue #4399, opened 2023-03-06
[Bug] Provide correct unavailable commands


DidierRLopes:
A while ago the "unavailable" items on the stocks menu was modified

<img width="1119" alt="Screenshot 2023-03-05 at 4 30 41 PM" src="https://user-images.githubusercontent.com/25267873/222995274-6ae89e40-f985-4d4c-af2f-15b83b564321.png">

This means that now we don't need to load a ticker to go into these menus. However, that introduced a a change of pattern that wasn't fixed in advance.

When going into "stocks/fa" without any ticker this is what I get
<img width="1175" alt="Screenshot 2023-03-05 at 4 32 33 PM" src="https://user-images.githubusercontent.com/25267873/222995344-1b6459a6-e3dc-44ee-829c-3add2ab0364d.png">

This is wrong, the coloring below "load" should all be unavailable to reflect that a ticker is needed, and protections need to be in place to ensure that.

Another bug is that if I go into "stocks/ta" without a ticker loaded, this is what I get:

<img width="728" alt="Screenshot 2023-03-05 at 4 33 42 PM" src="https://user-images.githubusercontent.com/25267873/222995402-19146a39-0834-47ee-a2e9-14b9f75ac151.png">

PR #4405, opened 2023-03-06
Make the UI a little cleaner

jmaslek:
Addresses #4399


openbb_terminal/stocks/fundamental_analysis/fa_controller.py
```diff
@@ -44,6 +44,10 @@
 
 logger = logging.getLogger(__name__)
 
+no_ticker_message = (
```
hjoaquim, line 47: shouldn't the cmds under this menu also be grey if ticker is not loaded?

openbb_terminal/stocks/technical_analysis/ta_controller.py
```diff
@@ -129,46 +129,48 @@ def print_help(self):
             stock_str = f"{s_intraday} {self.ticker}"
 
         mt = MenuText("stocks/ta/", 90)
-        mt.add_param("_ticker", stock_str)
+        mt.add_param(
```
hjoaquim, line 132: should we also have the `load` cmd here? the functionality is there tho

openbb_terminal/stocks/fundamental_analysis/fa_controller.py
jmaslek: So the reason they aren't is because they can all be run with cmd -t

openbb_terminal/stocks/technical_analysis/ta_controller.py
jmaslek: yes i should

hjoaquim (approved):

PR #4424, opened 2023-03-07
Improve `stocks/fa/enterprise`, now `stocks/fa/mktcap`, functionality

JerBouma:
# Description

Main issue was that the command was a bit strange to work with. It gave insights but only for a relative short period. This makes it so you get all the periods by default and includes a start and end date. Given that enterprise value shares a lot of similarities with that of stock prices, it made sense it followed a similar logic.

Mostly did this change since I am using it in the Academia presentations.

image

image

# How has this been tested?

# Checklist:

# Others


openbb_terminal/stocks/fundamental_analysis/fa_controller.py
```diff
@@ -385,24 +385,23 @@ def call_enterprise(self, other_args: List[str]):
...
-            default=5,
-            help="Limit of latest years/quarters, relevant for the source FinancialModelingPrep.",
-        )
 
         parser.add_argument(
             "-s",
             "--start",
             type=valid_date,
-            default=(datetime.now() - timedelta(days=3 * 366)).strftime("%Y-%m-%d"),
+            default="1900-01-01",
```
Raghavendrapara, line 393: Hey, I had a query, does this change impact the response time or structure of the charts prepared(namely the axis markings getting skewed) based on the default start date?

openbb_terminal/core/sdk/trail_map.csv
```diff
@@ -449,7 +449,7 @@ stocks.fa.info,stocks_fa_yahoo_finance_model.get_info,
 stocks.fa.key,stocks_fa_av_model.get_key_metrics,
 stocks.fa.metrics,stocks_fa_fmp_model.get_key_metrics,
 stocks.fa.mgmt,stocks_fa_business_insider_model.get_management,
-stocks.fa.mktcap,stocks_fa_yahoo_finance_model.get_mktcap,stocks_fa_yahoo_finance_view.display_mktcap
+stocks.fa.mktcap_yf,stocks_fa_yahoo_finance_model.get_mktcap,stocks_fa_yahoo_finance_view.display_mktcap
```
jmaslek, line 452: We no longer use the convention cmd_source.

Either get rid of yf (wouldn't be mad) or write a function that accepts a source and use it here.

JerBouma: Alright, I'll make a new function. In what kind of module should that be included? E.g. it wouldn't fit in fmp_model or yahoofinance_model right?

jmaslek: yeah I would throw it in with the others in sdk_helpers.py

JerBouma: I mean what about for the Terminal itself where we have model and views?

jmaslek: Is fine how it is.  Just the sdk needs a single function

JerBouma: Got it

openbb_terminal/stocks/fundamental_analysis/fa_controller.py
JerBouma: Within the model it gets converted to datetime actually. This is mostly for convenient sake. The default set here is essentially "give me all data available if no start or end date are set"il

It won't change anything how data is plotted or how data is collected since the functions itself work with strings like these.

jmaslek: Think the question was asking about the chart performance with lots of data loaded.  For candles, there is a lot of data shown, so starting at 1900 (or however far back we can go) would make the window react slower and take up more memory.

For line charts like this, there will be a non-noticable difference between a couple thousand extra data points.

JerBouma: Ah yes and note that its annual data.

openbb_terminal/core/sdk/trail_map.csv
JerBouma: I've made `fa.enterprise` and `fa.mktcap` instead.

jmaslek (approved):

deeleeramone:
The value for market cap should not be limited to annual or quarterly, and there is a need for this number to be calculated on a daily level.

Screenshot 2023-03-08 at 9 48 52 AM

jmaslek:
I dont know why the merge queue is randomly failing the linting when it passes here.

jmaslek:
I found it.  Codespell updated 2 hours ago.  Guess they changed their dictionaries

PR #4428, opened 2023-03-08
Patch reports

### LINKED ISSUE #4353: [Bug] Reports/equity - An error was encountered in cell [63] - 'list' object has no attribute 'to_html'
```
reports/equity --symbol UPS
```

Screenshot 2023-02-28 at 7 55 05 AM

hjoaquim:
^

Fixing the reports to take into consideration the new changes on user preferences and fixed income menu.

Fix #4353

### DIFF
--- a/openbb_terminal/core/session/session_model.py
+++ b/openbb_terminal/core/session/session_model.py
@@ -88,7 +88,7 @@ def login(session: dict) -> LoginStatus:
         if response.status_code == 200:
             configs = json.loads(response.content)
             email = configs.get("email", "")
-            feature_settings = configs.get("features_settings", {})
+            feature_settings = configs.get("features_settings", {}) or {}
             hub_user.profile.load_user_info(session, email)
             set_current_user(hub_user)
             Local.apply_configs(configs=configs)
--- a/generate_sdk.py
+++ b/generate_sdk.py
@@ -60,7 +60,7 @@ class SDKLogger:
 
     @staticmethod
     def _try_to_login(sdk: "OpenBBSDK"):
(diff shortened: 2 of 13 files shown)


openbb_terminal/core/session/session_model.py
```diff
@@ -88,7 +88,7 @@ def login(session: dict) -> LoginStatus:
         if response.status_code == 200:
             configs = json.loads(response.content)
             email = configs.get("email", "")
-            feature_settings = configs.get("features_settings", {})
+            feature_settings = configs.get("features_settings", {}) or {}
```
jmaslek, line 91: ?  is this just in case `features_settings` is empty and not an empty dict?

hjoaquim: Yes. I was getting an error because somehow the key "features_settings" is there but with the value None (maybe because I had values stored in the past but not anymore?) - causing the terminal to yield an error when later trying `feature_settings.get("FLAIR", None)`

hjoaquim:
@jmaslek for the SDK, do I need to add something else?

jmaslek:
> @jmaslek for the SDK, do I need to add something else?

As long as `python generate_sdk.py` runs, you should be good

jmaslek (approved):

hjoaquim:
@jmaslek fixed it now! Should've changed just the trail map and not the files directly!
Also, did some fixes to the `generate_sdk.py` so it's now up-to-date (taking into consideration the new changes on the user preferences). The `sdk.py` should not be changed directly (at least not without also changing `generate_sdk.py`).

jmaslek (approved):

PR #4434, opened 2023-03-08
Update FinanceDatabase to v2.0.9

### LINKED ISSUE #4422: [IMPROVE] Security vulnerability caused by Pickle usage
Not sure if to flag this as a bug or improvement. Either way, I think it's fairly dangerous to have terminals download on-demand Pickle files from random places on the Internet.

Pickle files allow for remote code execution and Python offers no sandbox mechanism. This is not so different from loading a DLL. See the warning at: https://docs.python.org/3/library/pickle.html or https://pandas.pydata.org/docs/reference/api/pandas.read_pickle.html

I will open a similar issue on https://github.com/JerBouma/FinanceDatabase which OpenBB depends on and which presents the same issue.

---

Yahoo Finance model should move away from Pickle formats, e.g.:
https://github.com/OpenBB-finance/OpenBBTerminal/blob/fdcc6b0252dbaed3b24296a590ade74ba1a56617/openbb_terminal/portfolio/portfolio_optimization/yahoo_finance_model.py#L194

Same functionality using a safe serialization format.

**Possibly describe the ideal way to improve this**

E.g. [msgpack](https://msgpack.org/), BSON or compressed JSON instead of Pickle

**Additional information**

N/A

JerBouma:
Very small change, fixes #4422. Now loads different files, other than that nothing changes. Old files are still there so nothing breaks for earlier versions. Files are now CSV compressed with BZ2 instead of Pickle. I don't fully agree with the security risk (as it is just me maintaining, no PRs in the last 3 years and you would need to add malicious code to a CSV file) but this just gets rid of the risk altogether.


jmaslek (approved):

JerBouma:
Need to rewrite a whole bunch of tests just because the file extension changed from `.pkl` to `.bz2`, smh.

jmaslek:
> Need to rewrite a whole bunch of tests just because the file extension changed from `.pkl` to `.bz2`, smh.

plz just not infinite numbers of tests

JerBouma:
> > Need to rewrite a whole bunch of tests just because the file extension changed from `.pkl` to `.bz2`, smh.
> 
> plz just not infinite numbers of tests

Testa passed so just 800k lines. I don't even get why it updates so much lol.

PR #4440, opened 2023-03-08
Update dev deps

jmaslek:
Codespell updated, so I just updated it and then also did pylint while at it


colin99d:
Might as well make ruff 0.0.256?

colin99d (approved):

jmaslek:
> Might as well make ruff 0.0.256?

Was already on it :)

Issue #3647, opened 2022-11-29
[Bug] Charts not showing on docker


hjoaquim:
**Describe the bug**

**To Reproduce**
Steps(from the start) and commands to reproduce the behavior

**Screenshots**

![image](https://user-images.githubusercontent.com/48914296/204537366-28e43a2b-a4c3-4bb4-afd1-60acc9bb09e8.png)

On running the instructions on the **[X-Server on Linux Desktop](https://docs.openbb.co/terminal/quickstart/installation?tab=docker)** section:

![image](https://user-images.githubusercontent.com/48914296/204537657-9a76d87d-6224-43bc-81f1-a8c84c8e794b.png)

**Desktop (please complete the following information):**

- OS: Ubuntu
- Docker image

Chavithra:
> @Chavithra tested for Win and same thing happens.

We merge a fix that should work on windows.

@Lukazcreations testing on M1 Mac.

Chavithra:
Charts are now working on macOS and Windows.
Ubuntu still need to to fix

colin99d:
@henriqueio did Juan's new PR fix this?

jiiiijiij:
did it fix now ? on linux ?

andrewkenreich:
Closing this as we have fixed issue here on the new version - This is related to an older version of docker now.

@jiiiijiij if you have issues still let us know.

https://github.com/OpenBB-finance/OpenBBTerminal/issues/4872

PR #4447, opened 2023-03-09
Add X11 support for docker to display charts

carsanbu:
# Add X11 support for docker to display charts

After running the docker image in my Debian Linux distribution I realized it was not possible to display charts so I added support mounting a volume in the X11 socket and fixing the DISPLAY variable in environment.

I tested this with Debian Linux but I can't test with Windows or macOS and I don't know if displaying charts is already supported in these OS, I don't want to break anything so, can anyone test this?

How I tested:

```bash
docker compose run openbb
stocks/candle -t LIN
```
The chart is now displayed:
image


piiq:
Hey @carsanbu, thanks for investing time into better support for linux 👏💪❤️ 

Sorry, this is most likely not work on windows because a lack of `/tmp`. 

Regarding fixing the DISPLAY variable. If it's not set on the host system what would be the behaviour? Do you think we can improve the docs and explain how to override the default settings and mount additional files into the container rather than change the compose file?

carsanbu:
Hi @piiq,
I understand, I think it's possible to create an [extension](https://docs.docker.com/compose/extends/) of the actual compose file and only use it in case you want to use X11. I think it would be the best solution.
I'm going to try this and maybe I can update with a new solution.

piiq:
That would be great! Post here if you need any help

jiiiijiij:
I think there should be a seperate docker-compose.yaml for linux. Thank your for the help. Now my million dollar dream can come true, lol

carsanbu:
Updated with the backwards compatible solution, Now in an extension file. It is possible to run in any X11 system using this command:
`docker compose -f build/docker/docker-compose.yaml -f build/docker/docker-compose.x11.yaml run openbb`

I would like to document this option somewhere but I don´t see the correct place.

hjoaquim:
> I would like to document this option somewhere but I don´t see the correct place.

@carsanbu `website/content/terminal/installation/docker.md` under `<TabItem value="Linux" label="Linux">` might be the place!

hjoaquim:
This fixes the issues I was experiencing with charts and tables on Ubuntu.
Fixes https://github.com/OpenBB-finance/OpenBBTerminal/issues/3647

piiq:
great job @carsanbu 
thanks a lot!
waiting for the docs and let's merge it 🚀

piiq (approved):

PR #4485, opened 2023-03-14
New table design

jose-donato:
TODO:
* show toast when columns are hidden --- **fixed**
* `stocks/fa/income` - improve formatting so doesnt catch these numbers --- **fixed**
* index shouldn't have date filter --- **fixed**
image

* `stocks/fa/income -l 30` too many columns shown by default --- **fixed**
image


andrewkenreich:
image


On stocks/dps/pos and a few others we can only sort on the last 2 columns and index, first 4 dont let you

jmaslek:
Reminders for me to check:

- [ ] urls
- [ ] long entries (like fred -q {} )

deeleeramone:
There is a need to auto-fit column width for content. 

`/econ/performance`

Screenshot 2023-03-21 at 11 49 04 PM

deeleeramone:
The space at the top, IMO, takes up too much of my precious screen space.

Screenshot 2023-03-22 at 12 00 12 AM

Font size adjustments has unintended consequences. 

Screenshot 2023-03-22 at 12 04 33 AM

deeleeramone:
The feature flag to turn off the interactive tables makes the terminal fail.


Screenshot 2023-03-22 at 12 20 26 AM

Enabling it returns results.

deeleeramone:
Columns are missing, and it should not be considered "advanced" to see all the data in a table. The table needs to dynamically sized for the content and column headers should respect the boundaries of the column width.

Screenshot 2023-03-22 at 12 29 18 AM

Before:
Screenshot 2023-03-22 at 12 22 57 AM

After:
Screenshot 2023-03-22 at 12 25 18 AM

jmaslek (approved):

jmaslek:
For you @jose-donato chart height:
Screenshot 2023-03-22 at 4 19 48 PM

deeleeramone:
> For you @jose-donato chart height: Screenshot 2023-03-22 at 4 19 48 PM

Can we do away with the `watermark` now?

jmaslek:
> Can we do away with the `watermark` now?

patience

deeleeramone:
This doesn't actually solve the problem. The solution is to be able to scroll to the right on the PyWry window.  

Screenshot 2023-03-22 at 2 11 40 PM


> Columns are missing, and it should not be considered "advanced" to see all the data in a table. The table needs to dynamically sized for the content and column headers should respect the boundaries of the column width.
> 
> Screenshot 2023-03-22 at 12 29 18 AM
> 
> Before: Screenshot 2023-03-22 at 12 22 57 AM
>

PR #4546, opened 2023-03-21
Hub | Add colors, sources sync

montezdesousa:
# Description
* Adds `colors` sync (should we spin off `colors` to a menu to match the hub?)
Screenshot 2023-03-21 at 17 00 07
Screenshot 2023-03-21 at 17 11 11

* Displays `keys` source
Screenshot 2023-03-24 at 17 17 02

Screenshot 2023-03-24 at 17 17 45

* Adds `sources` sync and displays source
Screenshot 2023-03-24 at 17 18 40

Screenshot 2023-03-24 at 17 18 01


DidierRLopes:
> # Description
> Screenshot 2023-03-21 at 17 00 07
> 
> Screenshot 2023-03-21 at 17 11 11
> 
> * Users can change between `dark` and `custom` options in `colors` command inside `settings` menu
> 
> 1. `dark` is a builtin style (this is the default)
> 2. `custom` is a user style
> 
> 
> We have 2 types of users. This is how they can change the terminal style and use `custom` option:
> 
> 1. Hub users:
> 
> * Set theme in https://my.openbb.dev/app/terminal/theme
> * Login and see their hub style applied if `custom` option is set in `settings`
> * Logout and terminal will revert style to local version
> 
> 3. Local users:
> 
> * Pre define custom theme in the website https://openbb.co/customize, follow instructions and set `custom` option in `settings`

@montezdesousa do we need a "dark" mode for default?

If the user wants to change locally they will edit the local file, so I think it makes more sense to have "local" and "hub" color settings. Where "local" is the one that the user sets manually and defaults to the "dark theme" and then the hub is the one associated with their account and defaults to "dark theme" too.

Also when accessing this menu would be good to have a preview of what those colors are directly there?

We could have something below like:
```
Colors: local
    [Menu, Commands, Help, Params, ..]
```

Where `Menu` would appear in the corresponding color, `Commands` in the corresponding color and so on.

Thoughts?

montezdesousa:
> 

We don't need, but I believe it's important to have an easy way to go back to a default or known state if for some reason user messes up with the colors.

montezdesousa:
> > # Description
> > Screenshot 2023-03-21 at 17 00 07
> > Screenshot 2023-03-21 at 17 11 11
> > 
> > * Users can change between `dark` and `custom` options in `colors` command inside `settings` menu
> > 
> > 
> > 1. `dark` is a builtin style (this is the default)
> > 2. `custom` is a user style
> > 
> > We have 2 types of users. This is how they can change the terminal style and use `custom` option:
> > 
> > 1. Hub users:
> > 
> > 
> > * Set theme in https://my.openbb.dev/app/terminal/theme
> > * Login and see their hub style applied if `custom` option is set in `settings`
> > * Logout and terminal will revert style to local version
> > 
> > 
> > 3. Local users:
> > 
> > 
> > * Pre define custom theme in the website https://openbb.co/customize, follow instructions and set `custom` option in `settings`
> 
> @montezdesousa do we need a "dark" mode for default?
> 
> If the user wants to change locally they will edit the local file, so I think it makes more sense to have "local" and "hub" color settings. Where "local" is the one that the user sets manually and defaults to the "dark theme" and then the hub is the one associated with their account and defaults to "dark theme" too.
> 
> Also when accessing this menu would be good to have a preview of what those colors are directly there?
> 
> We could have something below like:
> 
> ```
> Colors: local
>     [Menu, Commands, Help, Params, ..]
> ```
> 
> Where `Menu` would appear in the corresponding color, `Commands` in the corresponding color and so on.
> 
> Thoughts?

We don't need, but I believe it's important to have an easy way to go back to a default or known state if for some reason user messes up with the colors.

montezdesousa:
Like this for the colors preview @DidierRLopes ?

Screenshot 2023-03-22 at 10 19 01

DidierRLopes:
> Like this for the colors preview @DidierRLopes ?
> 
> Screenshot 2023-03-22 at 10 19 01

That is perfect, yup!  @montezdesousa 

I still don't think we should call it "dark" though. We can call it "default" so users know they are selecting the pre-built style. Since they may want to create their own dark theme.

Also, I think the main part about custom is the fact that if I select custom, I'm not sure if it comes from Hub or Terminal local settings

montezdesousa:
> > Like this for the colors preview @DidierRLopes ?
> > Screenshot 2023-03-22 at 10 19 01
> 
> That is perfect, yup! @montezdesousa
> 
> I still don't think we should call it "dark" though. We can call it "default" so users know they are selecting the pre-built style. Since they may want to create their own dark theme.
> 
> Also, I think the main part about custom is the fact that if I select custom, I'm not sure if it comes from Hub or Terminal local settings

Default looks good to me.

Oh I see it's not explicit. Anyway if your are logged in it comes from the hub, otherwise it comes from the .json inside OpenBBUserData. 3 options work `default`, `local` and `hub`?

DidierRLopes:
> > > Like this for the colors preview @DidierRLopes ?
> > > Screenshot 2023-03-22 at 10 19 01
> > 
> > 
> > That is perfect, yup! @montezdesousa
> > I still don't think we should call it "dark" though. We can call it "default" so users know they are selecting the pre-built style. Since they may want to create their own dark theme.
> > Also, I think the main part about custom is the fact that if I select custom, I'm not sure if it comes from Hub or Terminal local settings
> 
> Default looks good to me.
> 
> Oh I see it's not explicit. Anyway if your are logged in it comes from the hub, otherwise it comes from the .json inside OpenBBUserData. 3 options work `default`, `local` and `hub`?

That would be great. You can keep same mechanism, just change the local/hub based on wheather they are logged in or not :)

jmaslek (approved):

Issue #4572, opened 2023-03-23
[Bug] `Stocks/load`  inconsistent behaviour from each source. `end_date` not respected. Sometimes, `start_date` is ignored.


deeleeramone:
Inconsistent and buggy behaviour from the `load` function in general. 

For daily data:
- YahooFinance is not inclusive of the `end_date`.
- EODHD does not respect the start or end date.

For intraday:
- YahooFinance ignores the `end_date`. 
- Time is 4 hours ahead.
- AlphaVantage doesn't work when `end_date` is present.
- AlphaVantage returns only 100 results because the endpoint has an argument, `outputsize=compact`, which needs to be `full`.

```
In [13]: openbb.stocks.load("GME", start_date = "2023-03-01", end_date = "2023-03-02", source = 'Intrinio')
Loading Daily data for GME with starting period 2023-03-01.
Out[13]: 
             Open   High    Low  Close  Adj Close     Volume  dividend
date                                                                  
2023-03-01  18.92  19.09  18.14  18.15      18.15  4210100.0       0.0
2023-03-02  18.15  18.40  17.62  18.19      18.19  2627304.0       0.0

In [14]: openbb.stocks.load("GME", start_date = "2023-03-01", end_date = "2023-03-02", source = 'YahooFinance')
Loading Daily data for GME with starting period 2023-03-01.
Out[14]: 
             Open   High        Low  Close  Adj Close   Volume  Dividends  Stock Splits
date                                                                                   
2023-03-01  18.92  19.09  18.139999  18.15      18.15  4210100        0.0           0.0

In [15]: openbb.stocks.load("GME", start_date = "2023-03-01", end_date = "2023-03-02", source = 'AlphaVantage')
Loading Daily data for GME with starting period 2023-03-01.
Out[15]: 
             Open   High    Low  Close  Adj Close     Volume  Dividend amount  Split coefficient
date                                                                                            
2023-03-01  18.92  19.09  18.14  18.15      18.15  4210100.0              0.0                1.0
2023-03-02  18.15  18.40  17.62  18.19      18.19  2627304.0              0.0                1.0

In [16]: openbb.stocks.load("GME", start_date = "2023-03-01", end_date = "2023-03-02", source = 'EODHD')
Loading Daily data for GME with starting period 2022-03-23.
Out[16]: 
               Open      High      Low   Close  Adj Close    Volume
date                                                               
[...]

Issue #4562, opened 2023-03-22
[Bug] `stocks/load SPY --interval 1 --prepost --source YahooFinance` - The time is 4 hours ahead. Pre-market starts at 4 AM not 8.


deeleeramone:
```
openbb.stocks.load(symbol = 'spy', start_date = '2023-03-22', interval = 1, prepost = True)
```

![Screenshot 2023-03-22 at 8 50 00 AM](https://user-images.githubusercontent.com/85772166/226962009-d8875a16-f246-410a-93f8-b921feb9568b.png)

![Screenshot 2023-03-22 at 8 50 33 AM](https://user-images.githubusercontent.com/85772166/226962181-40efc62c-c8ea-4dc1-9a8f-0234a7726fcb.png)

The argument for `end_date` is not working:

```
openbb.stocks.load(symbol = 'spy', start_date = '2023-03-19', end_date = '2023-03-20', prepost = True, interval = 1)
```

![Screenshot 2023-03-22 at 8 52 28 AM](https://user-images.githubusercontent.com/85772166/226962783-a7dbce8d-3a01-4ff9-8f12-3573c8f28c83.png)

deeleeramone:
#4572

PR #4579, opened 2023-03-24
Fix Stocks Load

IgorWounds:
Fixed the issue of the market being 4h ahead of EST and not respecting the end date (#4562).

Fixed the issues of EODH not respecting the start and end dates and AlphaVantage not working when the end date was present and not showing the full output (#4572).

The thing to keep in mind is that Yahoo Finance and AlphaVantage return data up to the end date while other providers include that date too. AlphaVantage seems to do this only when intraday data is requested. 

Do we want to standardize the end-date behavior between different providers? This might make it easier for the internal flow but counter-intuitive for people that have experience with using these providers.


deeleeramone:
> The thing to keep in mind is that Yahoo Finance and AlphaVantage return data up to the end date while other providers include that date too. AlphaVantage seems to do this only when intraday data is requested.

According to the AlphaVantage documentation, there is no start and end date as variables to the API call. This would mean that the inclusiveness is being handled by a Pandas preference. 

https://www.alphavantage.co/documentation/#daily

deeleeramone:
When `--source EODHD`, and parameters fall outside of the subscription level, ideally, the error message will be the status code from the API response stating the reason for the call being rejected. Currently returns a Python error.  People with the appropriate paid plans for EODHD will have intraday and pre-post as available data being returned.

Screenshot 2023-03-27 at 8 18 53 AM

When the range of dates falls outside of the free account, I get this error:

```
(🦋) /stocks/ $ load aapl --start 2020-01-01 --end 2020-01-05 --source EODHD

Error: "None of [Index(['date', 'open', 'high', 'low', 'close', 'adjusted_close', 'volume'], dtype='object')] are in the "
```

The error code returned here is: 'Forbidden'.  I suggest this be made a little more friendly, to say, "Access forbidden. This is likely because the account is not subscribed to this data."


https://eodhistoricaldata.com/financial-apis/intraday-historical-data-api/
https://eodhistoricaldata.com/financial-apis/api-for-historical-data-and-volumes/

deeleeramone (changes requested):
This does resolve the issues raised. 
 
Screenshot 2023-03-27 at 3 07 28 PM

One exception is that the behaviour of the `load` function has now changed, where this used to be a valid input, and it would get the max history available. Is it possible to restore this functionality while keeping the fixes?

```
(🦋) /stocks/ $ load aapl -i 1


1 Failed download:
- AAPL: 1m data not available for startTime=1584849600 and endTime=1679954918. Only 7 days worth of 1m granularity data are allowed to be fetched per request.
```

deeleeramone:
I could have sworn I added this as a comment already, but I don't see it anywhere now, so it must have not completed. The functionality works, it's just the message being printed which will lead to user confusion.  Start date != start date, according to the console print.

Screenshot 2023-03-28 at 9 39 19 PM

deeleeramone:
YahooFinance is fixed, AlphaVantage still shows the same issue:

```
(🦋) /stocks/ $ load aapl -i 1 --source AlphaVantage

Loading Intraday 1min data for AAPL with starting period 2020-03-24.
```

PR #4583, opened 2023-03-24
fix: trying new bg color

jose-donato:
before
image

after
image


DidierRLopes (approved):
We need this merged so we can point users to installation page

Issue #4261, opened 2023-02-15
[Bug] Forex/Load - "Source Oanda not supported"


deeleeramone:
Not sure what this error message is about, `Oanda` is a choice for the `--source`. 

![Screenshot 2023-02-15 at 1 24 02 PM](https://user-images.githubusercontent.com/85772166/219170424-99d4942b-5f1d-4072-a994-08205368e6e9.png)

It also does not check the API key or notify you that there is no key entered.

![Screenshot 2023-02-15 at 1 24 51 PM](https://user-images.githubusercontent.com/85772166/219170733-a28eab0d-7614-4d77-a173-4d1eeb409f17.png)

PR #4596, opened 2023-03-27
Remove Oanda from forex sources

IgorWounds:
Removed Oanda from Forex sources according to #4261  and discussion with @jmaslek


openbb_terminal/forex/forex_controller.py
```diff
@@ -75,6 +75,7 @@ def __init__(self, queue: Optional[List[str]] = None):
         if session and get_current_user().preferences.USE_PROMPT_TOOLKIT:
             choices: dict = self.choices_default
             choices["load"].update({c: {} for c in FX_TICKERS})
+            del choices["load"]["--source"]["Oanda"]
```
jmaslek, line 78: Since this is autogenerated, you should be able to take Oanda from the sources instead of deleting it here.

IgorWounds: Thanks for spotting this!

jmaslek (approved):

PR #4625, opened 2023-03-29
Handling log supression

hjoaquim:
Goal: logs should only be generated by the SDK in standalone mode (i.e., if launched from within the terminal, the SDK should not generate new log files).

In this PR:

1. Remove the `system.json` solution previously in use
   a. Adds unnecessary complexity --> not easy to remove the file as soon as the terminal is closed (the `atexit` won't work if the terminal is closed by a signal and in turn Windows devices deal with signals differently than posix devices)
   b. Potentially problematic if different instances are in use at the same time (e.g. SDK + Terminal)

2. Move the responsibility to suppress logging to the reports and dashboards
   a. As the SDK can be used either from reports/dashboards and standalone, it can be tricky to check which is the case during runtime. By making each report and dashboard updating the current system to suppress the logs, we avoid getting into sketchy solutions.


jmaslek:
Do they get "un-suppressed" after running the suppress command?  I assume that the cfg.suppress only suppresses for that file, and not globally?

hjoaquim:
> Do they get "un-suppressed" after running the suppress command? I assume that the cfg.suppress only suppresses for that file, and not globally?

That's correct. Both reports and dashboards are their own instances, meaning that, we'll "return" to the terminal with no side effect.

PR #4633, opened 2023-03-30
Remove references to environment variables

hjoaquim:
### As the title suggest, this PR intends to remove as many references to the environment variables as possible. This allows us:

1. Greater control over how things change
2. Centralized approach leveraging `SystemModel`
3. Tied with the above: predictability

### Functionality check list - from the user standpoint, everything should be the same:

1. `DEBUG_MODE`
   a. w/ `python terminal.py --debug`
   b. `openbb`
2. `TEST_MODE`
   a. Run `python terminal.py -t -l` (check possible int tests) and then `python terminal.py -t 2 --subproc 0 --verbose`.
3. `LOG_COLLECTION`
   a. if you're a dev and have this flag set to `False` on the `env` file, you should see no files on your `archives` dir on your logs folder.
4. `ENABLE_AUTHENTICATION`
   a. if you've it set to `True`on you `.env` file and use `openbb --login` you should still be able to use the authenticated functionalities.

> All of this can be setup on the `.env` file and will be initialzed to `SystemModel` accordingly during the terminal initialization.
> `DEBUG_MODE` and `TEST_MODE` can obviouly, also be used throught the command line.

### Not addressed environment variables that can still be found (use `CTRL+shift+F` to check it) that we probably want to keep:

1. PYDEVD_DISABLE_FILE_VALIDATION --> pyinstaller hook
2. REQUESTS_CA_BUNDLE --> Necessary for installer so that it can locate the correct certificates for API calls and https
3. SSL_CERT_FILE --> Necessary for installer so that it can locate the correct certificates for API calls and https 
4. SERVER_SOFTWARE --> used for voila dashboards
5. PYTHONPATH --> used for dashboards


openbb_terminal/config_terminal.py
```diff
@@ -33,6 +33,27 @@ def change_logging_suppress(new_value: bool):
     set_current_system(updated_system)
 
 
+def change_test_mode(new_value: bool):
```
montezdesousa, line 36: Here wdyt having variable names as arguments instead and have it on `openbb_terminal/core/session/current_system.py`? Think it might scale better as we add new variables to the dataclass.

`set_system_variable("TEST_MODE", True)`


```python3
def set_system_variable(
    name: str,
    value: Any,
):
    """Set system variable

    Parameters
    ----------
    name : str
        Variable name
    value : Any
        Variable value
    """
    current_system = get_current_system()
    updated_system = dataclasses.replace(current_system, **{name: value})  # type: ignore
    set_current_system(updated_system)
```

montezdesousa:
At which point do we load the `SystemModel` with the .env?  The `--debug` flag seems to be off

Screenshot 2023-03-30 at 16 19 58

montezdesousa:
There are a few `load_env_vars` references which are loading environment variables as well. Should this go into the `SystemModel` as well?

hjoaquim:
@montezdesousa would you give it another try (maybe on a fresh env)?

Everything seems alright on my end:
1.
image
image
2.
image
image
4.
image

montezdesousa:
> @montezdesousa would you give it another try (maybe on a fresh env)?
> 
> Everything seems alright on my end: 1. image image 2. image image 4. image

You are right, don't know what was happening there!

openbb_terminal/config_terminal.py
hjoaquim: This is very good suggestion, thank you. Just addressed it on the most recent commits!

hjoaquim:
> `load_env_vars`

Good one. Was able to remove the function completely!

hjoaquim:
Related with:  #4665
@montezdesousa after this PR gets merged, we should use the auth flag on the SystemModel to control the `is_auth_enabled` function

hjoaquim:
> Related with: #4665 @montezdesousa after this PR gets merged, we should use the auth flag on the SystemModel to control the `is_auth_enabled` function

@montezdesousa did it here: https://github.com/OpenBB-finance/OpenBBTerminal/pull/4633/commits/0f007f6684498b7c590a8dae8f1c4d598b1d5f1a

hjoaquim:
> > Related with: #4665 @montezdesousa after this PR gets merged, we should use the auth flag on the SystemModel to control the `is_auth_enabled` function
> 
> What's left to do before merging?

everything's finished now, I think!

montezdesousa (approved):

PR #4640, opened 2023-03-30
Hub | sync, pull, sources and other updates

montezdesousa:
# Description

This PR should make the authenticated user experience more consistent and predictable

* Removes runtime configuration fetching from hub (`settings colors`, `sources get`, ...). Restarting the terminal is ALWAYS required to apply changes made on the hub during terminal runtime
* Following the point above, `pull` command is removed
* `sync` command is removed, so authenticated user api keys are always stored in the hub. Less freedom, but less complexity
* Changes `reset` so that it does reset as a guest regardless user being authenticated or not
* Warns user to delete session files if `logout` fails (not likely but...)

Additional bug fixes:
* Fix `sources` menu not updating sources on commands usage
* Fix `sources` menu not reading most updated version of json
* Saves terminal crashing if sources file is broken, falling back to openbb_defaults.json
* Fixes `test_sdk_session.py` tests with `@pytest.mark.skip("Not working?")`


jmaslek:
Hmm.  My first test was changing the theme, and after changing + saving + reset the changes werent reflected.

montezdesousa:
> Hmm.  My first test was changing the theme, and after changing + saving + reset the changes werent reflected.

Theme you refer to light/dark charts/tables in the hub? That's only implement in #4623

jmaslek:
> > Hmm.  My first test was changing the theme, and after changing + saving + reset the changes werent reflected.
> 
> Theme you refer to light/dark charts/tables in the hub? That's only implement in #4623

no, just the terminal theme (like menu, cmd etc)

montezdesousa:
> > > Hmm.  My first test was changing the theme, and after changing + saving + reset the changes werent reflected.
> > 
> > 
> > Theme you refer to light/dark charts/tables in the hub? That's only implement in #4623
> 
> no, just the terminal theme (like menu, cmd etc)

Issue was a "~/OpenBBUserData/styles/user/hub.richstyle.json" and that prevails over "openbb_terminal/miscellaneous/styles/user/hub.richstyle.json".

After looking into this, I think this is the expected behavior since "OpenBBUserData" has priority over the installation folder files. For your particular case, you were not aware it was there bc it was a leftover from a previous version I pushed during the week. I was downloading the "hub.richstyle.json" to "~/OpenBBUserData/styles/user" which is not the case anymore.

In a further iteration it would be good to refactor the whole `TerminalStyle` class and this kind of edge cases should be handled by design.

IgorWounds (approved):
Looks fine to me.

PR #4658, opened 2023-04-02
hotfix/ Avoid starting plots_backend twice on boot

tehcoderer:
- avoid starting plots_backend twice on boot
- declutter `call_intro` with `intro.json`
- bump pywry


jmaslek (approved):

openbb_terminal/config_terminal.py
```diff
@@ -96,11 +96,13 @@ def setup_logging_sub_app(sub_app: str):
     set_current_system(updated_system)
 
 
-def setup_config_terminal():
+def setup_config_terminal(app_name: str = "gst"):
```
hjoaquim, line 99: I feel that this can be misleading if the app_name is not actually "gst". Can we have something like `is_sdk: bool = False` instead?

openbb_terminal/terminal_helper.py
```diff
@@ -355,6 +356,7 @@ def reset(queue: Optional[List[str]] = None):
     logger.info("resetting")
     plt.close("all")
     plots_backend().close(reset=True)
+    load_env_files()
```
hjoaquim, line 359: Conflicts with https://github.com/OpenBB-finance/OpenBBTerminal/pull/4633, but can be easily fixed on whichever gets merged last.

openbb_terminal/config_terminal.py
tehcoderer: EZPZ good idea

PR #4662, opened 2023-04-03
Moves logging credentials under `CredentialsModel`

hjoaquim:
Avoiding having credentials on the `SystemModel`, otherwise, those will be logged.


hjoaquim:
After discussing this solution w/ @montezdesousa we decided to go into a different direction.

PR #4667, opened 2023-04-03
Change returning dict to Dict (from typing)

hjoaquim:
Change functions that use the pattern `-> dict[something, something]` to use `-> Dict[something, something]` (using typing).
Goal: make it compatible w/ Python 3.8

Example: 

image

image


tehcoderer (approved):
Perfect Phone Call!!

PR #4673, opened 2023-04-03
Removes `exchange` option from `stocks/load`

colin99d:
# Description

It looks like the yfinance does not work here, so there is no reason to provide a broken option.


jmaslek (approved):
aight

jmaslek (approved):

PR #4693, opened 2023-04-05
Fix issue with holder

### LINKED ISSUE #4691: [Bug] `stocks/fa/shrs` requires `--holder` when using `--export`
Bug only applies when using `--export` otherwise it works fine.

image

JerBouma:
This fixes #4691 

image


deeleeramone:
`shrs --holder major` returns a duplicate of the index instead of the values.

Screenshot 2023-04-07 at 8 57 09 PM

montezdesousa (approved):

PR #4697, opened 2023-04-05
Revert UltimaInsights from default data source on stocks/news

DidierRLopes:
This PR does 2 things:
* Reverts UltimaInsights from default data provider for `stocks/news`, as this would break current user workflows.
* Remove UltimaInsights from `news` commands which is meant to display news for anything and not necessarily `stock` related

CC: @AdiSai @jmaslek


DidierRLopes:
#4703 implements this.

PR #4698, opened 2023-04-06
Allow special characters in autocompletion and improve sources controller

montezdesousa:
# Description

* This PR fixes sources menu autocompletion, which was lost after typing "/"
* It also fixes `tz` command in the `settings` menu, which was having the same issue

Changing `WORD` to `True` inside "openbb_terminal/custom_prompt_toolkit.py" means that prompt_toolkit will consider the regex `_FIND_BIG_WORD_RE = re.compile(r"([^\s]+)")` to get the word before the cursor, instead of `_FIND_WORD_RE = re.compile(r"([a-zA-Z0-9_]+|[^a-zA-Z0-9_\s]+)")`. The latter is ignoring "/".

* Before:
image
image

* After:
Screenshot 2023-04-06 at 10 51 02
Screenshot 2023-04-06 at 11 02 47

* The changes in the sources controller allow for multiple white spaces in-between commands/flags, or to avoid typing the flags `--cmd`, `--sources`
Screenshot 2023-04-06 at 11 00 15


hjoaquim (approved):
Looks good to me. Functionality is also there, nice improvement.

PR #4705, opened 2023-04-06
Add a note on GC and Kaggle

IgorWounds:


hjoaquim:
@IgorWounds moved it, so it's near the installation material, wdyt?

hjoaquim (approved):

PR #4706, opened 2023-04-07
Hotfix/ TA multi handle no params + with params

tehcoderer:
## Before
`/ta/multi --indicators macd,obv,sma[50,150]`
image

## After
image


andrewkenreich (approved):
Good fix

PR #4711, opened 2023-04-07
Keep hub sources updated and allow only available sources in the terminal

montezdesousa:
# Description

* Uses `openbb_defaults.json` as base structure to update with source list
* Appends new and removes deprecated sources from sources dictionary on user login. Then send it to backend
* Allows only sources specified in `openbb_defaults.json` to ensure the user does not set a deprecated source for e.g.

Test:
* Add/remove sources from `openbb_defaults.json` and check that user can/cannot select them in `sources` menu
* Add/remove a command from `openbb_defaults.json` and check that user can/cannot select it in `sources` menu
* Add/remove sources from `openbb_defaults.json`, login and check that sources in the hub (website) are in sync with that change
* Change default sources in the terminal, check they are applied in commands, `sources` menu and persist after logout
* Change default sources in the hub and check if they are applied in the following terminal session


IgorWounds (approved):
Works for me.

PR #4716, opened 2023-04-07
Ensure consistency across terminal and improve intro

DidierRLopes:
Says view doesn't exist, even though it appears on menu
Screenshot 2023-04-07 at 2 52 11 PM

The ordering is wrong here, and those should be dimmed
Screenshot 2023-04-07 at 2 53 10 PM

I made fundamentals menu dim as well for consistency - it's like that in most of the terminal (e.g. ta menu)

Also improved intro


openbb_terminal/stocks/fundamental_analysis/fa_controller.py
```diff
@@ -128,41 +128,41 @@ def print_help(self):
...
+        mt.add_cmd("income", not self.stock.empty)
+        mt.add_cmd("balance", not self.stock.empty)
+        mt.add_cmd("cash", not self.stock.empty)
+        mt.add_cmd("growth", not self.stock.empty)
+        mt.add_cmd("metrics", not self.stock.empty)
+        mt.add_cmd("ratios", not self.stock.empty)
+        mt.add_cmd("dupont", not self.stock.empty)
+        mt.add_cmd("fraud", not self.stock.empty)
+        mt.add_cmd("sec", not self.stock.empty)
+        mt.add_cmd("analysis", not self.stock.empty)
```
jmaslek, line 156: these are all intentionally not dimmed because they can all be accessed using a ticker arg (income -t aapl).  Having them dimmed implied that something needs to be loaded first (which is why they are not dimmed in the ta menu)

PR #4719, opened 2023-04-08
hotfix/fix the lack of commit hash in pypi logs

### LINKED ISSUE #4453: [Bug] Pypi version of the terminal with no commit hash on the logs
image

hjoaquim:
Fix #4453

As mentioned in [poetry docs](https://python-poetry.org/docs/pyproject/#include-and-exclude) files that are on the `.gitignore` will not make it into the package, so we have to explicitly add it to `pyproject.toml` in order to use the build shell script as intended.

Changes were made to the `build/pypi/build.sh` so that we ensure that at the moment of the creation of the wheel, the content on the `.env` is known and, after it, the content is restored.


Chavithra:
Hi @hjoaquim,

I think we can dish the backup/restore of previous `.env`, since people are not suppose to add their `,env` here.

Thanks

Chavithra (approved):
Hi @hjoaquim,

looks good to me.

I think we can dish the backup/restore of previous `.env`, since people are not suppose to add their `.env` here.

Thanks

PR #4720, opened 2023-04-09
Improve plotting UX

DidierRLopes:
* Reorganized items
* Improved key bindings
* Did NOT added default inclination

Things to do - @tehcoderer:
* Key binding for auto-size is not working
* Limit charts to not go below a certain size (e.g. modbar should never rotate vertically)
* Start xaxis tickangle of -20 earlier as threshold is a bit late
* Edit color popup goes over date on top right, can we improve that location?

### DIFF
--- a/openbb_terminal/core/plots/web/main.js
+++ b/openbb_terminal/core/plots/web/main.js
@@ -78,7 +78,7 @@ function OpenBBMain(plotly_figure, chartdiv, csvdiv, textdiv, titlediv) {
     modeBarButtons: [
       [
         {
-          name: "Download CSV (Ctrl+S)",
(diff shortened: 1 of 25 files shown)


openbb_terminal/core/plots/web/main.js
```diff
@@ -220,6 +228,9 @@ function OpenBBMain(plotly_figure, chartdiv, csvdiv, textdiv, titlediv) {
     };
   }
 
+  // RIP: Juan hated it so had to remove it
```
tehcoderer, line 231: Approved

jmaslek:
+0
-0

noice

PR #4739, opened 2023-04-11
Updated support url

jose-donato:


jmaslek (approved):
this is awful. We are all worse coders for having looked at this pr

PR #4740, opened 2023-04-12
gimme -> askobb

DidierRLopes:
* Rename from gimme to askobb (AskOpenBB)
* Allow to run in one line
* The suggestion [y/n] mechanism could be used to iterate on the model, but most of the times we won't know if the output corresponds to what we expect. So if we are to implement it we may need to have it happening after chart/table is displayed.


minhhoang1023 (approved):
good points!

Issue #4713, opened 2023-04-07
[Bug] `econometrics\load macrodata` Cannot format dates properly to set the index.


deeleeramone:
If you load one of the sample data sets in the Econometrics menu, for this example it is `macrodata`, the column for `year` is formatted as an abbreviated integer in thousands. 

![Screenshot 2023-04-07 at 12 35 09 PM](https://user-images.githubusercontent.com/85772166/230667732-b0ea7e80-a865-442c-8029-4db919731311.png)

`type macrodata.year --format date` then converts them all to `0.000`.
![Screenshot 2023-04-07 at 12 36 22 PM](https://user-images.githubusercontent.com/85772166/230667989-c67eff3d-527c-4d16-81a9-69991dad94fd.png)

It would be a reasonable argument to make that `type --format date` is not sufficient for converting these values. This should convert to a `DatetimeIndex` and allow for the desired formatting of the date and time.

With the inability to correct this, you can't set the index and it essentially becomes impossible to do any meaningful analysis or calculations.  Furthermore, upon loading any data set, it should automatically find the column titled, `date` or `year`/`month` and format the DataFrame appropriately.

the-praxs:
@tehcoderer changing `year` to `str` will show `1960.0` instead of our desired output `1960`.

the-praxs:
Issue extends to the datetime type index

![image](https://user-images.githubusercontent.com/88149434/231344421-4d01ae36-5d24-44b7-8e85-60ed8a4d858c.png)

the-praxs:
> > @tehcoderer changing `year` to `str` will show `1960.0` instead of our desired output `1960`.
> 
> Convert the date to a string. An integer does not have a decimal, so you must have converted to a float first.

The dataset was loaded with `float64` dtype. I tried this approach but found a more direct and suitable way to fix it. Please feel free to review the commit (#4742). It works as intended on my end.

If you find any more issues, please ping me and I will fix them as soon as I can.

tehcoderer:
Fixed in #4848 👍

tehcoderer:
> @tehcoderer changing `year` to `str` will show `1960.0` instead of our desired output `1960`.

yeah had to convert to int then str for it to work when I tested today lol.

deeleeramone:
Fixed in #4848

PR #4742, opened 2023-04-12
Fixed `type` date formatting in `econometrics` menu

the-praxs:
# Description

- [x] Added code to detect column date type and assign a date format in the `call_load` function in `econometrics_controller.py`.
- [x] #4713 
- [x] Fixed date type
image
- [x] To add stability to the command.
- [x] No dependency change.

# How has this been tested?
Tested by running the command with different datasets in the terminal.

# Checklist:

# Others

### DIFF
--- a/openbb_terminal/econometrics/econometrics_controller.py
+++ b/openbb_terminal/econometrics/econometrics_controller.py
@@ -850,11 +850,19 @@ class EconometricsController(BaseController):
             if ns_parser.name:
                 if "." in ns_parser.name:
                     dataset, column = ns_parser.name.split(".")
+                    # Determine date type and accordingly assign date format
+                    if column.lower() == "year":
+                        date_format = "%Y"
+                    elif column.lower() == "month":
+                        date_format = "%m"
+                    elif column.lower() == ["date", "period"]:
+                        date_format = "%Y-%m-%d"
+
                     if ns_parser.format:
                         if ns_parser.format == "date":
                             self.datasets[dataset][column] = pd.to_datetime(
-                                self.datasets[dataset][column].values,
-                            )
+                                self.datasets[dataset][column],
+                            ).dt.strftime(date_format)
                         else:
                             self.datasets[dataset][column] = self.datasets[dataset][
                                 column


the-praxs:
@deeleeramone please review and let me know your thoughts on this.

deeleeramone (changes requested):
This solution does not work.

Using the `type` command to change the Year column now converts it to an object type, when type should be `DatetimeIndex` stored as a string value.

Screenshot 2023-04-12 at 9 17 47 AM

Using the `show` command, all values are now 1970.

Screenshot 2023-04-12 at 9 19 53 AM

the-praxs:
@deeleeramone thank you for your review. I will find a solution and fix it by today.

the-praxs:
While working on the issue today, I found out an issue with the `to_json()` function in the `pandas` API. 

It seems that the `date_format` parameter  has `epoch` type by default so it causes the date column to be dumped as UNIX timestamp in `json` file.

The second type is `iso`(**ISO8601** format). However this means the date column will be formatted as `%Y-%m-%dT%H:%M:%S`.

I will try to find a workaround to this problem.

deeleeramone (changes requested):
The solution to modify the backend.py file needs to be reversed, this impacts every part of the terminal instead of being a targeted repair within the menu. This file does not need to be touched.  The dataframes can be modified within the individual files of the econometrics folder, like `econometrics_model` and `econometrics_view`.

When I attempt to set a multi-index, it appears to "set", but then `show` gives an error:

```
(🦋) /econometrics/ $ index ccard -i age,income

Successfully updated 'ccard' index to be 'age, income'


(🦋) /econometrics/ $ show

Error: cannot insert income, already exists


(🦋) /econometrics/ $ show ccard

Error: cannot insert income, already exists


(🦋) /econometrics/ $
```

the-praxs:
> The solution to modify the backend.py file needs to be reversed, this impacts every part of the terminal instead of being a targeted repair within the menu. This file does not need to be touched. The dataframes can be modified within the individual files of the econometrics folder, like `econometrics_model` and `econometrics_view`.
> 
> When I attempt to set a multi-index, it appears to "set", but then `show` gives an error:
> 
> ```
> (🦋) /econometrics/ $ index ccard -i age,income
> 
> Successfully updated 'ccard' index to be 'age, income'
> 
> 
> (🦋) /econometrics/ $ show
> 
> Error: cannot insert income, already exists
> 
> 
> (🦋) /econometrics/ $ show ccard
> 
> Error: cannot insert income, already exists
> 
> 
> (🦋) /econometrics/ $
> ```

This is a bug I believe. I switched to the `develop` branch and used the same commands, and got the same error.

the-praxs:
@deeleeramone can you please confirm if the problem arises without the modified code?

deeleeramone:
This was fixed in #4848, so closing this PR. Thanks, @the-praxs!

PR #4744, opened 2023-04-12
Changing default data source for `crypto.load`

### LINKED ISSUE #4367: [Bug] [SDK] Crypto/Load - Default source (CoinGecko) is returning bad data.
Screenshot 2023-03-01 at 10 13 38 AM

hjoaquim:
Fix #4367

TL;DR:
   CoinGecko was the default source, but the data returned is from every four days.
   Changing it to `yf` to improve the user experience.

> I think we shouldn't drop the CG as a whole bc it might be helpful for some use cases. Also, they might change the way they provide the data in the future or users with their Pro API key have access to the full range of dates.


jmaslek:
Does this need to get changed in the Default Sources json as well?

hjoaquim:
> Does this need to get changed in the Default Sources json as well?

On the default data sources file (`openbb_terminal/miscellaneous/sources/openbb_default.json`) it's `yf` already for `crypto/load` but I don't think this impact the SDK usage.

jmaslek (approved):

PR #4754, opened 2023-04-12
stocks/fa unit tests

IgorWounds:
Added unit tests for the stocks/fa submenu. It is essentially up to 80% coverage but there is a `dcf_view.py` file that wasn't written to be easily unit-testable due to the size of requests it makes (recorded requests are cumulatively over 100MB). Thus, I marked them to be skipped for the time being.


hjoaquim (approved):
This looks very good to me! Awesome to see the pytest recorder in action. Looking forward to expand it trought our codebase 🚀

PR #4798, opened 2023-04-17
Fixed default routines message and evident display

### LINKED ISSUE #4794: [Bug] `account\list` displays `No routines found.` but opens a windows with 3 default routines
**Describe the bug**
I have not set any routine in my account. So executing `accounts\list` command displays the proper output -

image

However, it opens a window that displays 3 routines configured - 

openbb_Default_routines_20230415_154016

the-praxs:
# Description
Fixes #4794.
Modified code to display default routines only when personal routines are not found. 
Added message if personal routines are found and displayed.

# Screenshots
image


montezdesousa:
Hi, thanks for contributing.

IIUC your solution, after the first personal routine list command w’d never display default routines again? 
A better solution might be to add `--type` argument where user can specify `personal` or `default`, so it mimics the hub experience where users have two tabs to check both at any time.

Screenshot 2023-04-17 at 12 45 54
Screenshot 2023-04-17 at 12 48 48

the-praxs:
Thanks for the review. It sounds like a good idea. We can set the `--type` to `personal` by default. So it will work as intended. 

I do have a question: what if the user wants to disable displaying the default routines? How do we achieve that?

montezdesousa:
> Thanks for the review. It sounds like a good idea. We can set the `--type` to `personal` by default. So it will work as intended.
> 
> I do have a question: what if the user wants to disable displaying the default routines? How do we achieve that?

Taking https://my.openbb.dev/app/terminal/routines as reference.

`--type personal` is the default and we display a message if no routines
`--type defaults` can be accessed explicitly (maybe defaults is not the best naming, we can check that later and make the website consistent)

the-praxs:
I see. I will modify the code later to reflect this suggestion. Thanks for the information!

openbb_terminal/account/account_controller.py
```diff
@@ -259,15 +268,21 @@ def call_list(self, other_args: List[str]):
...
-                if not df.empty:
-                    self.REMOTE_CHOICES += list(df["name"])
-                    self.update_runtime_choices()
-                    display_personal_routines(df, page, pages)
-                else:
-                    console.print("[red]No routines found.[/red]")
+
+                if ns_parser.type == "personal":
+                    if not df.empty:
+                        console.print("[info]Displaying personal routines.[/info]")
```
montezdesousa, line 274: I'd remove this print here and the similar one for default routines, either the table pops up or is printed in the console if `USE_INTERACTIVE_DF='False'`

openbb_terminal/account/account_controller.py
```diff
@@ -259,15 +268,21 @@ def call_list(self, other_args: List[str]):
...
+                        self.REMOTE_CHOICES += list(df["name"])
+                        self.update_runtime_choices()
+                        display_personal_routines(df, page, pages)
+                    else:
+                        console.print("[red]No personal routines found.[/red]")
+                elif ns_parser.type == "default":
+                    console.print("[info]Displaying default routines.[/info]")
```
montezdesousa, line 281: ^ Related to comment above

montezdesousa (commented):
Looks good to me, just added two comments on review

openbb_terminal/account/account_controller.py
the-praxs: I removed the print to let the table pop up and show the results.

openbb_terminal/account/account_controller.py
the-praxs: Same as above!

montezdesousa:
Thanks for the contribution, this also fixes https://github.com/OpenBB-finance/OpenBBTerminal/issues/4803

montezdesousa (approved):

PR #4799, opened 2023-04-17
Fix `exe --file` choices on login

montezdesousa:
* Makes sure that `exe --file` choices are updated right after login

Test:
* login with `openbb --login` and `account/login` and check that `exe --file` command on `home` menu displays the hub routines (default and personal)


deeleeramone:
Can you merge this with the other PR, #4802? Hard to test the other without this in place. :)

montezdesousa:
> Can you merge this with the other PR, #4802? Hard to test the other without this in place. :)

done

PR #4813, opened 2023-04-17
Keep session on reset

montezdesousa:
* Keeps user authenticated on `reset`
* Removes unnecessary new lines printed when `reset`

### DIFF
--- a/openbb_terminal/common/behavioural_analysis/finbrain_view.py
+++ b/openbb_terminal/common/behavioural_analysis/finbrain_view.py
@@ -137,4 +137,4 @@ def display_sentiment_analysis(
         fig,
     )
 
-    return fig.show(external=external_axes)
+    return fig.show(external=raw or external_axes)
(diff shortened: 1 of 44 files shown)


jmaslek:
Okay I have a question that has been bugging me forever.

Is it possible to get rid of all the newlines when we do reset?  I tracked down a bunch a while ago, but that deletes newlines between commands.

This is not a requireement, just a question that has bugged me for a while

montezdesousa:
> Okay I have a question that has been bugging me forever.
> 
> Is it possible to get rid of all the newlines when we do reset? I tracked down a bunch a while ago, but that deletes newlines between commands.
> 
> This is not a requireement, just a question that has bugged me for a while

Removed some now, can you confirm pls?

jmaslek (approved):
noice
```
2023 Apr 17, 15:35 [jmaslek11] 🦋 / $ reset
resetting...

╭────────────────────────────────────────────────────────────────────────────── Home ───────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                                   │
│ Information, guides and support for the OpenBB Termina
```

PR #4815, opened 2023-04-17
Adding GARCH

northern-64bit:
# Description

Adding GARCH, its model fit and forecasts. It also reorganizes the econometrics menu.

# How has this been tested?

Comparing to sample GARCH results from Hull.

# Checklist:

# Others


openbb_terminal/econometrics/econometrics_model.py
```diff
@@ -326,6 +328,74 @@ def get_coint_df(
...
+        Lag order of the asymmetric innovation
+    q: int
+        Lag order of lagged volatility or equivalent
+    mean: str
+        The name of the mean model
+    horizon: int
+        The horizon of the forecast
+    """
+    model = arch_model(
+        100 * data.pct_change().dropna(), vol="GARCH", p=p, o=o, q=q, mean=mean
```
jmaslek, line 391: So this is what is tripping me up.  I was trying to load in returns and I was getting something whacky.  This should take in a selected data set.  We should maybe have an exploration function that calculates returns?

JerBouma: Yes, it makes sense we have an exploration command that calculates the returns for you. Also if you export data from `load` or `candle` you also won't get a returns column so to prevent needing to do this manually in Excel, we can offer this.

openbb_terminal/econometrics/econometrics_controller.py
```diff
@@ -1890,6 +1940,122 @@ def call_bpag(self, other_args):
...
+            - u_t; the innovation/discovery through current market price changes
+            - simgma_t; the last estimate
+
+            GARCH can be understood as a model, which allows to optimize these 3 variance components to the time series.
+            This is done assigning weights to variance components: (1 - alpha - beta) for sigma_l, alpha for u_t and
+            beta for sigma_t. [2]
+
+            The weights can be estimated by iterating over different values of (1 - alpha - beta) sigma_l which we
+            will call omega, alpha and beta, while maximizing: SUM(-ln(v_i) - (u_i ^ 2) / v_i). With the constraints:
+            - alpha > 0
```
jmaslek, line 1968: So I think we can support latex here, which will look nicer on the reference pages :)

openbb_terminal/econometrics/econometrics_model.py
```diff
@@ -326,6 +329,86 @@ def get_coint_df(
...
+    mean: str
+        The name of the mean model
+    horizon: int
+        The horizon of the forecast
+    """
+    model = arch_model(data.dropna(), vol="GARCH", p=p, o=o, q=q, mean=mean)
+    model_fit = model.fit(disp="off")
+    pred = model_fit.forecast(horizon=horizon, reindex=False)
+
+    return pred.variance.values[-1, :] * np.sqrt(252), model_fit
```
jmaslek, line 408: I believe this should return sqrt of the variance.? And then annualizing only makes sense if we annualize the input?

jmaslek:
Curious what @JerBouma thinks about this:  I am curious if we should have some historical comp?
Screenshot 2023-04-28 at 10 00 39 AM

JerBouma:
> Curious what @JerBouma thinks about this: I am curious if we should have some historical comp? Screenshot 2023-04-28 at 10 00 39 AM

What am I looking at here?

jmaslek:
> > Curious what @JerBouma thinks about this: I am curious if we should have some historical comp? Screenshot 2023-04-28 at 10 00 39 AM
> 
> What am I looking at here?

smh. essentially the 'backtested' prediction to compare actual with garch.  Instead of just spitting out a few steps forecasted.

openbb_terminal/econometrics/econometrics_controller.py
northern-64bit: Will do! 👍

openbb_terminal/econometrics/econometrics_model.py
northern-64bit: Good catch, it should take the square root! Regarding the annualizing, I'll remove it since users may want to adjust it themselves.

JerBouma:
> > > Curious what @JerBouma thinks about this: I am curious if we should have some historical comp? Screenshot 2023-04-28 at 10 00 39 AM
> > 
> > 
> > What am I looking at here?
> 
> smh. essentially the 'backtested' prediction to compare actual with garch. Instead of just spitting out a few steps forecasted.

Makes sense to me to include this. Especially to demonstrate whether what GARCH produces makes any logical sense. Isn't it also a way to test the underlying assumptions? If it is very off consistently, I'd assume that its future prediction is also not really that useful.

openbb_terminal/econometrics/econometrics_model.py
```diff
@@ -326,6 +329,89 @@ def get_coint_df(
...
+
+@log_start_end(log=logger)
+def get_garch(
+    data: pd.Series,
+    p: int = 1,
+    o: int = 0,
+    q: int = 1,
+    mean: str = "constant",
+    horizon: int = 100,
+):
```
jmaslek, line 352: This should include an example of how to use in the sdk.

Also technically not annualized since we get rid of the sqrt

JerBouma:
So I think we should add the historical backtested GARCH in a different PR, yes it definitely is helpful to have but the PR otherwise would get top big in my opinion.

jmaslek:
> So I think we should add the historical backtested GARCH in a different PR, yes it definitely is helpful to have but the PR otherwise would get top big in my opinion.

Yup. Just waiting on additional documentation in the model for how to use with sdk

jmaslek:
Note that running doctest on the example in the test fails:

```
Failed example:
    openbb.econometrics.garch(openbb.stocks.load("AAPL").pct_change()*100)
Exception raised:
      File "/Users/james/miniconda3/envs/obb/lib/python3.9/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest econometrics_model.get_garch[1]>", line 1, in <module>
        openbb.econometrics.garch(openbb.stocks.load("AAPL").pct_change()*100)
      File "/Users/james/OpenBBTerminal/OpenBBTerminal/openbb_terminal/core/sdk/sdk_helpers.py", line 106, in __call__
        method_result = method(*args, **kwargs)
      File "/Users/james/OpenBBTerminal/OpenBBTerminal/openbb_terminal/econometrics/econometrics_model.py", line 401, in get_garch
        model = arch_model(data.dropna(), vol="GARCH", p=p, o=o, q=q, mean=mean)
      File "/Users/james/miniconda3/envs/obb/lib/python3.9/site-packages/arch/univariate/mean.py", line 1965, in arch_model
        am = ConstantMean(y, hold_back=hold_back, rescale=rescale)
      File "/Users/james/miniconda3/envs/obb/lib/python3.9/site-packages/arch/univariate/mean.py", line 1103, in __init__
        super().__init__(
      File "/Users/james/miniconda3/envs/obb/lib/python3.9/site-packages/arch/univariate/mean.py", line 269, in __init__
        super().__init__(
      File "/Users/james/miniconda3/envs/obb/lib/python3.9/site-packages/arch/univariate/base.py", line 191, in __init__
        self._y_series = cast(pd.Series, ensure1d(y, "y", series=True))
      File "/Users/james/miniconda3/envs/obb/lib/python3.9/site-packages/arch/utility/array.py", line 49, in ensure1d
        raise ValueError(f"{name} must be squeezable to 1 dimension")
    ValueError: y must be squeezable to 1 dimension
```

northern-64bit:
> Note that running doctest on the example in the test fails:

I've now fixed it :)

jmaslek:
> > Note that running doctest on the example in the test fails:
> 
> I've now fixed it :)

The docstring does not represent a functional example

```
    51         return np.asarray(x.iloc[:, 0])

ValueError: y must be squeezable to 1 dimension
```

You need to specify a column from the load.  Also linters are failing

jmaslek (approved):

PR #4827, opened 2023-04-19
Fix pt command when no splits are found

jmaslek:
As per the title, this PR fixes an issue in `fa/pt` for when there are no splits found.


serkannpolatt:
.

PR #4864, opened 2023-04-22
First introduction of news_biztoc — A new partnership with BizToc.com

marban:
# Description

Introduces the first set of features for news_biztoc, following a new partnership with [BizToc](https://biztoc.com). 
Available at launch:
- News / latest
- News / per source
- News / search
- News / tag
- Tags / list
- Sources / list

TBD:
- News / hot
- News / page

See [https://api.biztoc.com](https://api.biztoc.com) for details.

# Others


openbb_terminal/terminal_controller.py
```diff
@@ -93,6 +94,7 @@ class TerminalController(BaseController):
         "exe",
         "guess",
         "news",
+        "news_biztoc",
```
minhhoang1023, line 97: Can you call biztoc inside the `news` command and not a separate command? Then users can specify `news --sources Biztoc`

openbb_terminal/terminal_controller.py
```diff
@@ -251,6 +254,81 @@ def call_news(self, other_args: List[str]) -> None:
...
+        parse.add_argument(
+            "-t",
+            "--term",
+            dest="term",
+            default="",
+            type=str,
+            nargs="+",
+            help="search for a term on the news",
+        )
+        parse.add_argument(
```
minhhoang1023, line 273: can you add these available options to the auto-completion? So it will look something like this:
and users don't need to run a separate command to see all the available tags.

Screenshot 2023-04-26 at 15 19 22

jmaslek: I can do the tags for auto complete, but the sources for auto complete won't work because then it overwrites the other source.

minhhoang1023: @jmaslek we should change the `source` to something else. `source` here should refer to `provider`, not news source

jmaslek: so source does make sense here, ie find bbg or yahoo etc.  we also have the sources flag, which may be kinda similar, but the autocomplete differentiates between them.

minhhoang1023 (approved):

openbb_terminal/terminal_helper.py
```diff
@@ -203,6 +203,7 @@ def open_openbb_documentation(
             "update",
             "wiki",
             "news",
+            "news_biztoc",
```
minhhoang1023, line 206: should be removed

minhhoang1023 (commented):
.

PR #4876, opened 2023-04-24
Update installation docs for release

piiq:
- Point download links for environment and docker artifacts to main instead of a blob
- Clean-up debug echo in macos launcher script

### DIFF
--- a/build/pyinstaller/macOS_package_assets/OpenBB Terminal/OpenBB Terminal
+++ b/build/pyinstaller/macOS_package_assets/OpenBB Terminal/OpenBB Terminal
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-echo "1"
+clear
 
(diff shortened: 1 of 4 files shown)


jmaslek:
I cant build locally:

```
Module build failed (from ./node_modules/@docusaurus/mdx-loader/lib/index.js):
SyntaxError: /Users/james/OpenBBTerminal/OpenBBTerminal/website/content/terminal/installation/macos.md: Expected corresponding JSX closing tag for <img>. (48:0)
```

deeleeramone:
Also cannot build locally. 

Screenshot 2023-04-24 at 9 49 39 AM

deeleeramone:
Do we update the download links now? Is there a way to make this more robust so that it doesn't need to be updated manually every release?

piiq:
> Do we update the download links now?

No, James does this as a part of the release routine in the release branch



> Is there a way to make this more robust so that it doesn't need to be updated manually every release?

This question is outside of the scope of this PR

deeleeramone (approved):
Looks good.

PR #4885, opened 2023-04-24
Last minutes enhancements

jmaslek:
- Remove featflag/tab
- Fix featflag/overwrite not overwriting
- Change fundamentals exporting to do transpose


tehcoderer (approved):
LGTM

Issue #4886, opened 2023-04-24
[IMPROVE] Futures Curve Tenor Display


skhan59bbk:
Two suggested improvements to be made to the futures curve function...

1. The contract shows as a datetime in the --raw table output, which is meaningless in this context. Also the header "Futures" is an odd choice. Perhaps "Last Price" would be better.

`futures/curve -t ES --raw`

![image](https://user-images.githubusercontent.com/57355784/234116209-7619993d-860d-4cba-b308-aa74933243bd.png)

2. The chart visual has a continuous date x-axis. In the context of futures curves, the convention is to show more of a discrete x-axis, corresponding to the actual tenors of the curve. 

`futures/curve -t ES`

![image](https://user-images.githubusercontent.com/57355784/234116021-e0166d47-0949-40b9-a461-22bff73782fe.png)

skhan59bbk:
Also when you export the table to excel (with featureflag 'interactive' enabled), the contract identifiers come out like so: 

![image](https://user-images.githubusercontent.com/57355784/234376954-064129ca-35ec-4b80-b4ce-f0291f03153a.png)

PR #4901, opened 2023-04-25
Futures Curve: minor cosmetic improvements 

skhan59bbk:
Catering for #4886 

1. Raw table column name changed from "Futures" to "Last Price" - in both interactive and basic view.
2. Contract names "prettified" more in-line with convention (mmm-yyyy).  Also amends excel output format.

image

image

image

3. Curve graph x-axis lines up actual contract months now, rather than continuous dates. 

image


deeleeramone:
Hi @skhan59bbk, thanks for the PR and the improvement!  The code linting is failing because it has not been formatted according to Black's wishes.  Run it from the command line at the root of the OpenBBTerminal folder:

```
black openbb_terminal/futures
```

However, because the expected outputs have been changed, you will also have to re-record the unit test cassettes so that the changes to the results are captured.

```
pytest tests/openbb_terminal/futures --rewrite-expected 
```

This will update a few files in the `/tests` folder. Add those, along with the changes made by the code linting to the `yfinance_model` and `yfinance_view` files, in a commit.

skhan59bbk:
Ah my bad. Linting fixed but pytest is asking me to uninstall brotli - is this expected? 

image

jmaslek (approved):

PR #4937, opened 2023-04-30
Adds options helper function for DTE, updates SDK trail map to not use yFinance.

deeleeramone:
Adds a new helper function in `openbb_terminal/stocks/options/op_helpers.py` for appending the chain DataFrame with a column for DTE.

This PR updates the SDK so that:

- `openbb_terminal.stocks.options.op_helpers` is a valid SDK path.
- the mapping for `openbb.stocks.options.DTE` to come from the `op_helpers`.

The docstring has been updated:

```
In [1]: from openbb_terminal.sdk import openbb

In [2]: openbb.stocks.options.dte?
Signature:      openbb.stocks.options.dte(chain: pandas.core.frame.DataFrame) -> pandas.core.frame.DataFrame
Call signature: openbb.stocks.options.dte(*args: Any, **kwargs: Any) -> Any
Type:           get_dte
String form:    <openbb_terminal.stocks.options.op_helpers.Operation object at 0x104e871c0>
File:           ~/GitHub/OpenBBTerminal/openbb_terminal/stocks/options/op_helpers.py
Docstring:     
Returns a new column containing the DTE as an integer, including 0.
Requires the chain to have the column labeled as, expiration.
```

It works with all sources, relying on there being a called called: `expiration`.

Screenshot 2023-04-30 at 2 40 56 PM


openbb_terminal/stocks/options/op_helpers.py
```diff
@@ -594,3 +594,20 @@ def Vomma(self, change):
...
+    """
+    Returns a new column containing the DTE as an integer, including 0.
+    Requires the chain to have the column labeled as, expiration.
+    """
+    if "expiration" not in chain.columns:
+        raise ValueError("No column labeled 'expiration' was found.")
+
+    now = datetime.now()
+    temp = pd.DatetimeIndex(chain.expiration)
+    temp_ = (temp - now).days + 1
```
jmaslek, line 610: If expirarion = now, then this would return 1 instead of 0.

I would like to see this done how I do elsewhere, where you look at the total_seconds between now the expiration (need to add 16 hours) and then convert seconds to days.  This will also avoid the divergence for 0dte calculates.

jmaslek (approved):

PR #4951, opened 2023-05-02
Feature/improve posthog logs

### LINKED ISSUE #4303: [Bug] logs should only be collected on `main` branch
Logs should only be collected on the `main` branch, otherwise we might be polluting our infrastructure with debug / development logs.

This (probably) can be achieved by looking at the commit hash, i.e.,
**if** `commit hash > main's commit hash` **then** `we don't want to collect `

> this is just an initial idea

hjoaquim:
This PR proposes the following:

- [improving handling of logging handlers](https://github.com/OpenBB-finance/OpenBBTerminal/commit/afa05d01bebcabbd5047a0e60c1c93e1b50fc253) --> changing the responsibility to the `SystemModel` instead

- [adding the commithash to the system so it gets logged](https://github.com/OpenBB-finance/OpenBBTerminal/pull/4951/commits/ea941b3c8096bc90f46730d093550dc50ea3561b)

- [don't send logs to S3 if posthog is an active handler](https://github.com/OpenBB-finance/OpenBBTerminal/pull/4951/commits/cdb064f135a0fcfab4dbce64b31bcc0c2d9ca804)

- [branch name on the posthog logs](https://github.com/OpenBB-finance/OpenBBTerminal/pull/4951/commits/e330fb29a47ecf3889b8d75c8468d1c2a0e4cad0)

Fixes #4303 by clearly identify the branch name for each event (posthog) and each session (file/s3).


andrewkenreich:
image

We can update this so it shows something

andrewkenreich:
also if i follow this route 


Load up terminal in guest 

Run a few commands 

Login

Run a few more commands


- I dont get alliased anymore

hjoaquim:
Nice catch! I think last commits solves both comments thank you, @andrewkenreich

jmaslek (approved):

PR #4959, opened 2023-05-03
Removing wrong transposition

hjoaquim:
Removing the `transposed()` so that years are displayed as columns instead of rows, for the following commands:

- income
- balance
- cash
- growth
- metrics
- ratios


jmaslek:
These transpose in the exports are intentional. Without them, you cannot overlay them onto the pywry charts.

hjoaquim:
> These transpose in the exports are intentional. Without them, you cannot overlay them onto the pywry charts.

I see!
@jmaslek @JerBouma should we fix this by having an argument `--trasponse` that controlls this behaviour?

JerBouma:
Perhaps this is an isolated case but for Financial Statements, you should be able to export them as we had originally.

I don't want to need to add a `--transpose` flag just to get a Financial Statement how it is supposed to be presented.

jmaslek:
lets get rid and @tehcoderer will figure out how to have a transpose option in PyWru

jmaslek (approved):

PR #4986, opened 2023-05-07
Create KEYCHRON_MACROS.md

DidierRLopes:
In progress

### DIFF
--- a/KEYCHRON_MACROS.md
+++ b/KEYCHRON_MACROS.md
@@ -1,0 +1,5 @@
+# Keychron Macros
+
+The OpenBB Terminal is a powerful open source investment research platform.
+
+tbd


the-praxs:
Should this be implemented as a general feature, allowing anyone using VIA to create their own macros?

DidierRLopes:
> Should this be implemented as a general feature, allowing anyone using VIA to create their own macros?

Everyone should be able to use VIA to create their own macros, all needed is a mechanical keyboard that VIA supports. We can probably change the title away from "Keychron" as VIA supports other types of keyboards. Thoughts?

DidierRLopes:
> Should this be implemented as a general feature, allowing anyone using VIA to create their own macros?

I have refactored this to allow support to more configurators like VIA and not just highlight Keychron. @the-praxs @hjoaquim @colin99d

the-praxs:
Reviewed it, looks good to me.

hjoaquim (approved):

PR #4991, opened 2023-05-08
Fixed equity report

the-praxs:
Changed `symbol` from `FRC` to `MSFT`.


the-praxs:
What is the error in the test indicting?

jmaslek:
> What is the error in the test indicting?

not a clue.  Just re ran it with changing nothing and it worked.

jmaslek (approved):

the-praxs:
Looks like a temp fault. yFinance acts weird sometimes.

PR #5018, opened 2023-05-13
Adds test for collinearity to the `econometrics` menu

northern-64bit:
# Description

Adds `vif` command, which tests collinearity by calculating the variance inflation factor. It also rearranges the commands in the menu by adding a `Assumption Testing` section.

# How has this been tested?

By using it in the SDK and in the  terminal. 

Testing it on data with known collinearity. Use for instance the following test case:
```
df = pd.DataFrame(
    {'a': [1, 1, 2, 3, 4],
     'b': [2, 2, 3, 2, 1],
     'c': [4, 6, 7, 8, 9],
     'd': [4, 3, 4, 5, 4]}
)
```
The result should be and are:
```
   Values
a   22.95
b    3.00
c   12.95
d    3.00
```

# Checklist:

# Others


JerBouma:
Just calling `vif` returns error.

```
(🦋) /econometrics/ $ vif

Error: 'NoneType' object is not iterable


(🦋) /econometrics/ $
```

Autocomplete is missing for columns, shouldn't be just when I type `vif`.

```
(🦋) /econometrics/ $ vif --columns

usage: vif [-c COLUMNS] [-h] [--export EXPORT] [--sheet-name SHEET_NAME [SHEET_NAME ...]]
vif: error: argument -c/--columns: expected one argument

(🦋) /econometrics/ $ vif --columns
```

More importantly, I think the default should be to show all columns. The purpose of VIF is to see if other variables seem to be able to estimate the value of the variable. If you only include a subset, the results will definitely vary where, in most cases, you would like to use your entire dataset. Then, with `--columns` you can specify which ones you don't and do want if you do wish to mix this up.

image

This result (from `load --file wage_panel -a wp`) indicates that most variables are not highly correlated (close to 1). However experience and experience squared do seem to be somewhat explained by other variables. As you look into the dataset, this actually makes perfect sense given that this is panel data, meaning it tracks multiple individuals over time, and we have a year column. So you'd see that if year goes up by 1, experience goes up by 1 too meaning from the year column you could have inferred experience too.

jmaslek:
Issue with error handling when only one dataset input:
```
2023 May 23, 18:42 (🐛) /econometrics/ $ vif k0.high

['k0.high']
```

jmaslek:
Looks like a rogue print:
```
2023 May 23, 18:44 (🐛) /econometrics/ $ vif -d k0.high,k0.low

['k0.high', 'k0.low']
```

jmaslek:
The output is not clear. 

Screenshot 2023-05-23 at 6 45 18 PM

If we are looking for multicollinarity in a regression, we should be 1) setting a dependent/indep variable and 2) getting a multi dimensional output when there are more than 2 outputs.

jmaslek (approved):
Nice!  Thanks for making all the changes!

PR #5024, opened 2023-05-15
Hotfix/docs-sdk-basics: Updates SDK basics page.

deeleeramone:
This PR updates the SDK basics page to be current.


deeleeramone:
@jmaslek - I removed three pages, incorporating them into the basics page.  Do you agree with this choice?

website/content/sdk/usage/basics/index.md
```diff
@@ -4,196 +4,390 @@ sidebar_position: 1
...
 import HeadTitle from '@site/src/components/General/HeadTitle.tsx';
 
-<HeadTitle title="Basics - SDK | OpenBB Docs" />
+`<HeadTitle title="Basics - SDK | OpenBB Docs" />`
 
-### Importing the SDK
+## Overview
 
-Now that you have the SDK installed, the first step is to import the OpenBB SDK in your preferred code editor. Nearly everything required to interact with any function from the OpenBB Terminal gets imported in one line. Begin a Python script or Notebook file with:
+The OpenBB SDK provides programmatic access to nearly all Terminal commands, allowing custom scripts and functions to power creations built on top of the core platform.  It is easy to use and as flexible as one's imagination.  The sections below will outline how to get started using it.  If it isn't already installed, follow the instructions to install via [PyPi (pip)](https://docs.openbb.co/terminal/installation/pypi) or from the [source code](https://docs.openbb.co/terminal/installation/source).
```
jmaslek, line 13: So we should probably differentiate that the sdk itself (without all) can be installed pretty much anywhere?  I can just fire up a venv, run pip install openbb and I am good to go.

deeleeramone: mmmm...that would mean we are telling people to install the Terminal this way, and we still want people to install by following the instructions so they don't make up their own methods and complain when "pip install doesn't work".  

I think I have some install steps for Google CoLab that are now stable.

deeleeramone:
> What doesn't make sense is having the same api-keys.md page twice (except for how to set).

Ya, that was a discussion with @DidierRLopes during the Hub push, and we'll have to think of a better solution for this instead of maintaining two separate pages.

website/content/sdk/usage/basics/index.md
deeleeramone: Not sure if this is the ideal spot for these instructions... thoughs?

jmaslek:
One of the things that I am not seeing anywhere is the fact that all the base functionalities can be run using :

```
python -m venv test
source test/bin/activate
pip install openbb
```

deeleeramone:
> One of the things that I am not seeing anywhere is the fact that all the base functionalities can be run using :
> 
> ```
> python -m venv test
> source test/bin/activate
> pip install openbb
> ```

Shouldn't these things be on the installation page?

website/content/sdk/usage/basics/index.md
```diff
@@ -4,196 +4,411 @@ sidebar_position: 1
...
+## Cloud Installation
+
+:::note
+**Cloud installations are experimental and cannot be guaranteed to work.  Results may vary by host.**
+
+The steps below assume that the developer container is Ubuntu-based.  If using Google Colab, remove `sudo` from the commands.
+
+**Step 1**: `sudo apt update && sudo apt upgrade`
+**Step 2**: `sudo apt install libwebkit2gtk-4.0-dev`
+**Step 3**: `pip install openbb`, or,  `pip install openbb[optimization]`
```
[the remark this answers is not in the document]
deeleeramone: If you want to launch it without the warning messages, then you need to install `libwebkit`.

jmaslek:
> > One of the things that I am not seeing anywhere is the fact that all the base functionalities can be run using :
> > ```
> > python -m venv test
> > source test/bin/activate
> > pip install openbb
> > ```
> 
> Shouldn't these things be on the installation page?

Yeah

jmaslek (approved):

PR #5056, opened 2023-05-22
Preserve stock ticker suffix 

### LINKED ISSUE #4965: [Bug] Tickers are stripped of their suffix, then sources downstream fail or get the wrong symbol.
This behaviour can be easily observed from the Stocks menu and a dual-listed symbol.

```
/stocks/load RY.TO/quote
```

VS.

```
/stocks/quote RY.TO
```

Screenshot 2023-05-03 at 3 28 33 PM

Non dual-listed stocks is problematic because it totally fails.

Screenshot 2023-05-03 at 3 29 40 PM

The ticker symbol must be preserved in order for the things to work.

IgorWounds:
This PR closes #4965 and adds a way to preserve the stock ticker suffix. Please double confirm that I didn't break other things :D

### DIFF
--- a/openbb_terminal/parent_classes.py
+++ b/openbb_terminal/parent_classes.py
@@ -1209,7 +1209,8 @@ class StockBaseController(BaseController, metaclass=ABCMeta):
                     stocks_helper.show_quick_performance(self.stock, ns_parser.ticker)
                 if "." in ns_parser.ticker:
                     self.ticker, self.suffix = ns_parser.ticker.upper().split(".")
-                    self.ticker = ns_parser.ticker.upper()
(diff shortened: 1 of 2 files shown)


deeleeramone (changes requested):
Funny quirk here where I enter the FA menu and the suffix is doubled.

Screenshot 2023-05-22 at 7 31 59 AM

```
(🦋) /stocks/fa/ $ income

Could not get data


(🦋) /stocks/fa/ $ income --ticker RY.TO
```


Screenshot 2023-05-22 at 7 34 32 AM

deeleeramone (approved):
Works as advertised.

Issue #5050, opened 2023-05-19
[Bug] `/stocks/fa/earnings` --limit argument appears to have no impact.


deeleeramone:
Recreate with:

```
/stocks/load aapl/fa/earnings --source YahooFinance --limit 2

earnings --source AlphaVantage --limit 2
```

Is there even a need to have the `limit` argument here?

PR #5058, opened 2023-05-22
hotfix/print_rich_table_limit

Ltwo3five:
# Description

### Summary of the change / bug fix.

> Limit was not being applied to the interactive table. I added the logic that was already present in print console table to print interactive table

### Link # issue, if applicable.
[   #5050](https://github.com/OpenBB-finance/OpenBBTerminal/issues/5050)

> rich_table

# How has this been tested?

### * Please describe the tests that you ran to verify your changes.

> End to End Test

### * Provide instructions so we can reproduce.
/stocks/load aapl/fa/earnings --source YahooFinance --limit 5
# Checklist:

# Others


jmaslek:
Hey!  Thanks for the PR.  

With the interactive tables, we don't need the limit.  To address that specific issue, the better solution would be to remove the limit argument from the command.

deeleeramone:
> With the interactive tables, we don't need the limit. To address that specific issue, the better solution would be to remove the limit argument from the command.

This makes me think, it's probably not a bug in this instance because it still works when the `interactive` flag is off.  This behaviour is also not consistent across functions.  Some will respect the `limit` argument when `interactive`, while others do not.  For example, `/arkord --limit 5`, works as both `interactive` and non.

LOne2three:
> Hey! Thanks for the PR.
> 
> With the interactive tables, we don't need the limit. To address that specific issue, the better solution would be to remove the limit argument from the command.

I have added logic to call_earnings to disable the limit argument when interactive is enabled because as far as I understand the limit argument should still be present when interactive is disabled. And according  @deeleeramone the limit command is inconsistent across functions, I have checked call_arkord in disc controller and the reason this is so is because the dataframe has the limit applied before being used as an argument in print_rich_table whereas for call_earnings the dataframe has the limit applied within print_rich_function.

deeleeramone:
FYI, PRs must be merged to the `/develop` branch.  They cannot be merged to `/main`.

PR #5092, opened 2023-05-31
fixes stocks/fa integration test file

deeleeramone:
This fixes a bunch of bad syntax that was in the `stocks/fa` integration test file.  Oddly enough, when I execute the test locally, it returns an error for an item that is not actually in the test. 

image

Removing any reference to `balance` and `YahooFinance` creates a new error, which again is never referenced in the test file. 

image

Every command in the `.openbb` file works and produces an output, so this is half bug report and half PR. :P


IgorWounds (approved):
Thanks for spotting this!

PR #5146, opened 2023-06-16
Small enhancements

### LINKED ISSUE #4834: [Improve] `stocks/ca` - `ticker` and `set` should do the same thing, having both is confusing.
this is `/stocks/ca/ticker tsla/add F,GM,VW,RIVN`

Screenshot 2023-04-19 at 6 39 50 PM

There should be a "base" ticker, and then everything else.  If you want to add a list of tickers, use `add`.

### LINKED ISSUE #5085: [Bug] stocks/qa/bw. only when loaded from qa menu
Wild sequence I found from the logs.
```
stocks/qa/load -t nvda/bw 
Error: ''
```

But if we load in from stocks, we are fine:

Aha.  I think the issue is because we need to have a `pick` after.  So the bug fix would be to set self.target after loading

jmaslek:
- Add a feedback option on askobb
- Couple small cosmetic changes from user feedback
- Fix an unreported bug in toplobbying function
- Fix #5085
- Cleans up ca ticker lists, i.e fix #4834

### DIFF
--- a/openbb_terminal/economy/finviz_view.py
+++ b/openbb_terminal/economy/finviz_view.py
@@ -51,12 +51,12 @@ def display_valuation(
 
     if df_group.empty:
         return
-
+    df_group = df_group.rename(columns={"Name": ""})
     print_rich_table(
         df_group,
         show_index=False,
         headers=list(df_group.columns),
-        title="Group Valuation Data",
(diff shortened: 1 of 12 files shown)


tehcoderer (approved):
LGTM

PR #5202, opened 2023-07-06
[Update] bump langchain and llama

martinb-bb:
Bump versions:
llama-index: 0.7.2
langchain: 0.0.225


deeleeramone:
Scipy is breaking things.

Needs to be kept at `1.10.1`

deeleeramone:
> > Scipy is breaking things.
> > Needs to be kept at `1.10.1`
> 
> Screenshot 2023-07-06 at 1 01 13 PM

Following the PyPI instructions leaves you with this:

Screenshot 2023-07-06 104315

jmaslek:
> > > Scipy is breaking things.
> > > Needs to be kept at `1.10.1`
> > 
> > 
> > Screenshot 2023-07-06 at 1 01 13 PM
> 
> Following the PyPI instructions leaves you with this:
> 
> Screenshot 2023-07-06 104315
> 
> The SDK then fails to import:
> 
> Screenshot 2023-07-06 104530

So thats an issue with the pypi build, not something addressed here

deeleeramone:
> So thats an issue with the pypi build, not something addressed here

This PR is addressing the failing builds though, `pip install openbb` will continue to fail, despite these changes, if Scipy isn't addressed.

jmaslek:
Even though the scope was to fix the langchain issue, I pinned scipy here to only be 1.10.1.

I also updated some dependencies by just running `poetry lock`.  Integration tests seem to run locally without any issues.  Running build now

@piiq can you approve that nothing should break.

jmaslek (approved):

PR #5205, opened 2023-07-07
Make some improvements to askobb

jmaslek:
Get rid of all the nested load functions (why we always see stocks/bt/load).   Change the ca files.  Update from .3 to .4


jmaslek:
@martinb-bb needs to be approved to merge from the queue :D

martinb-bb:
I'm just playing with it before approving. ☺️

jmaslek:
😭

martinb-bb (approved):

PR #5212, opened 2023-07-11
Update folder structure for the SDK v4

piiq:
This updates the folder structure and adds further decoupling on the extensions and providers

The package can be installed by running

`poetry install -C openbb_sdk` from the repo root.

to install ALL built-in extensions in editable mode you can use

```
poetry install -C openbb_sdk/core/sdk && \
poetry install -C openbb_sdk/core/provider && \
poetry install -C openbb_sdk/providers/fmp && \
poetry install -C openbb_sdk/providers/polygon && \
poetry install -C openbb_sdk/providers/benzinga && \
poetry install -C openbb_sdk/providers/fred && \
poetry install -C openbb_sdk/extensions/stocks && \
poetry install -C openbb_sdk/extensions/crypto && \
poetry install -C openbb_sdk/extensions/economy && \
poetry install -C openbb_sdk/extensions/news && \
poetry install -C openbb_sdk/extensions/forex && \
poetry install -C openbb_sdk/extensions/fixedincome && \
poetry install -C openbb_sdk/extensions/providers && \
poetry install -C openbb_sdk
```

with this monster all the extensions will appear as symlinks in the site-packages folder


the-praxs:
> 

Is there a way to configure a lock file to install all built-in extensions in editable mode rather than using this list of commands?

piiq:
> Is there a way to configure a lock file to install all built-in extensions in editable mode rather than using this list of commands?

likely yes, but it's going to be a hack that will be overwritten by poetry every time you do anything with the pyproject toml or the lock file

openbb_sdk/core/provider/openbb_provider/__init__.py
montezdesousa, line 1: Could we rename this `openbb_provider_core` and the package `openbb-provider-core` so that it follows the same convention as the sdk core?

piiq: Ok, will do

piiq: Maybe we should go with openbb_core and openbb_provider?

montezdesousa: and we rename the parent folder?
```
/openbb_sdk
           /sdk
               /provider
                        /openbb_provider
               /core
                        /openbb_core
           /extensions
           /openbb
 ```

piiq: like this?
image

montezdesousa: yes

PR #5218, opened 2023-07-14
Restructure folders and files

IgorWounds:
This PR restructures the SDK v4 file structure for easier usage and readability.


piiq:
I'm having trouble building the static in python3.9

piiq:
it works on 3.10 and fails to build on 3.9

the 3.9 bug is unrelated to the restructuring of the files

i like the proposed structure very much, thank you

piiq (approved):

PR #5238, opened 2023-07-21
Feature/improving to plotly

hjoaquim:
The ability to use the `to_plotly_json()` method on the `CommandOutput` level without relying on the `chart` argument. Also, allows for further customization. 

Example:

```python
from openbb import obb
stock_data = obb.stocks.load(symbol="AAPL", start_date="2023-01-01", provider="fmp")
indicators = dict(
    sma=dict(length=[20, 50, 100]),
    adx=dict(length=14),
    macd=dict(fast=12, slow=26, signal=9),
    rsi=dict(length=14),
    ema=dict(length=[20, 50]),
)
my_plotly_json = stock_data.to_plotly_json(**{"indicators":indicators})
stock_data.show()
```


hjoaquim:
To be fixed: chart indicators appear ploted backwards.

PR #5244, opened 2023-07-24
Setup unit tests folder structure

IgorWounds:
This PR will implement the unit testing framework for SDK v4.


piiq:
@luqmanbello this PR adds SDK v4 tests to GHA. Can you have a look at the workflow files to check if we can make the tests launch after all the required linters have passed? Now it seems to start right away, although I'm only making this assumption based on the image this I see

image

luqmanbello:
> @luqmanbello this PR adds SDK v4 tests to GHA. Can you have a look at the workflow files to check if we can make the tests launch after all the required linters have passed? Now it seems to start right away, although I'm only making this assumption based on the image this I see
> 
> image

@piiq You are correct it currently runs right away. We can do it in 2 ways
- Trigger the test from the linting test workflow once it is complete
-  Add it directly in the same workflow file as the linting test then use conditional statement.

the-praxs (approved):
Igor's tests work perfectly, nice implementation!

PR #5256, opened 2023-07-27
Leverage the OpenBB SDK v4 on the OpenBBTerminal

hjoaquim:
Using the `sdk-core` on the OpenBBTerminal.

Introducing `ArgparseTranslator` as a way to use `sdk-core` functionality out-of-the-box.

Ideas:
   - since all the arguments for the commands will be known apriori, we might find a better way to process the autocompletion.

TODOs:
   - [ ] Update controller tests
   - [ ] Refactor `argparse_translator` module
   - [ ] Add tests for the argparse translation (after refactor)

Open questions:
   - How to handle logs? Should the terminal have its own logs, plus the logs that the `sdk` will generate? Are the logs generated by the `sdk` enough, although there will be no info Terminal specific (like crashes on controllers, terminal style, etc.)?
   - Can we reach a state where the Terminal doesn't have to handle its own charting and rely only on the sdk's charting extension?


hjoaquim:
Ideas: 
 - since all the arguments for the commands will be known apriori, we might find a better way to process the autocompletion.

hjoaquim:
closing in detriment of #5302

PR #5276, opened 2023-08-02
Feature/sdk v4 updates

tehcoderer:
- added fmp financial ratios, stock quote
- correctly set return data type, rather then strictly make all return types `List[Data]`

```python
class FMPAvailableIndicesFetcher(
    Fetcher[
        FMPAvailableIndicesQueryParams,
        List[FMPAvailableIndicesData],  # We set the correct return type in fetcher generic class arg
    ]
):
```

- add missing  aliases and fields to base/provider models
- fix multiple validation error issues with datetime/dateTypes
- add intraday logic in StockEOD fetchers
- implement ability to pass multiple symbols on some Fetchers to avoid multiple sdk api calls
- create an `operation_id` for each router endpoint

- improve fastapi response schema's (below)
## Before
image

## After
image


openbb_sdk/extensions/stocks/openbb_stocks/fa/fa_router.py
```diff
@@ -191,6 +213,17 @@ def income(
     return CommandOutput(results=Query(**locals()).execute())
 
 
+@router.command(model="IncomeStatementGrowth")
+def income_growth(
```
piiq, line 217: @jmaslek how do you think we should handle this? commands with underscores have perfect sense in python but we tried to avoid them on the terminal. if this command will be exposed to the terminal as is it will break the convention. not sure if urls with underscores on the API are a good thing too.

jmaslek: I would prefer the following:
- terminal functions have no underscores
- terminal fnc = sdk fnc

by the coimmutative proprty - sdk has no underscore.  @DidierRLopes your thoughts?

IgorWounds (approved):
EZ PZ

PR #5280, opened 2023-08-03
Custom request for all providers

jmaslek:
Adds a custom request object and the ability to pass through additional user settings.

Cleans .to_dataframe() by dropping columns that are all None.


montezdesousa:
Code lgtm. Any objection to having `make_request` inside `helpers.py` instead of having a new file `utils.py`? Alternatively we could create utils/ folder and put all those utility files there (helpers, descriptions, etc.). It would mimic the provider extensions structure.

jmaslek:
> Code lgtm. Any objection to having `make_request` inside `helpers.py` instead of having a new file `utils.py`? Alternatively we could create utils/ folder and put all those utility files there (helpers, descriptions, etc.). It would mimic the provider extensions structure.

No objections here.

montezdesousa (approved):

PR #5283, opened 2023-08-04
[SDK v4] General improvements related with the charting extension

hjoaquim:
On this PR:
 - [x] Introduces `fig` on the chart model as way of accessing the `OpenBBFigure`
 - [x] Drop `to_plotly_json()` method in detriment of `to_chart()` to create or update the chart.
 - [x] Custom way to drop fields from the API response only.
 - [x] Support for tables on the charting extension


the-praxs (approved):
Awesome @hjoaquim!

IgorWounds (approved):
Nice!

PR #5291, opened 2023-08-07
Add pydocstyle linter

piiq:
This PR adds pydocstyle linter to pre-commit hooks and to the linting github action


IgorWounds (approved):

piiq:
I'll close this for now. The linter was added to pre-comit hook here https://github.com/OpenBB-finance/OpenBBTerminal/pull/5322 and we'll use that to highlight missing docstrings as it does a better job at checking the diff

PR #5302, opened 2023-08-09
[Feature] OpenBB Platform CLI

hjoaquim:

    - The OpenBB Terminal was unmaintainable.

    - Wrapping the OpenBB Platform in a CLI that resembles the former OpenBB Terminal.

    - NA to the Platform
    - ∞ to the Terminal/CLI

4. **Testing Done**:

    - Manually exploring the CLI.


the-praxs:
Commenting here to keep track of the discussions regarding the Terminal V4.

deeleeramone:
I'm still not able to launch it:

```
  File "/Users/danglewood/GitHub/OpenBBTerminal/terminal.py", line 17, in <module>
    main()
  File "/Users/danglewood/GitHub/OpenBBTerminal/terminal.py", line 13, in main
    session_controller.launch_terminal(dev=dev, debug=debug)
  File "/Users/danglewood/GitHub/OpenBBTerminal/openbb_terminal/core/session/session_controller.py", line 15, in launch_terminal
    terminal_controller.parse_args_and_run()
  File "/Users/danglewood/GitHub/OpenBBTerminal/openbb_terminal/terminal_controller.py", line 873, in parse_args_and_run
    main(
  File "/Users/danglewood/GitHub/OpenBBTerminal/openbb_terminal/terminal_controller.py", line 785, in main
    terminal()
  File "/Users/danglewood/GitHub/OpenBBTerminal/openbb_terminal/terminal_controller.py", line 476, in terminal
    t_controller = TerminalController(jobs_cmds)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/danglewood/GitHub/OpenBBTerminal/openbb_terminal/terminal_controller.py", line 125, in __init__
    self._generate_platform_commands()
  File "/Users/danglewood/GitHub/OpenBBTerminal/openbb_terminal/terminal_controller.py", line 148, in _generate_platform_commands
    target, reference=obb.coverage.reference
                      ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Coverage' object has no attribute 'reference'
```

hjoaquim:
@deeleeramone  itt works on a fresh environment following the highlighted steps.
I'll update the `pyproject.toml` as soon as there is a new Platform release so that the only step needed it `poetry install`.

hjoaquim:
@deeleeramone you're not using the latest dev:
image

deeleeramone:
> @deeleeramone you're not using the latest dev: ![image](https://private-user-images.githubusercontent.com/

Not sure how that would even be a choice. `git checkout feature/argparse_translator` -> new environment -> install -> `terminal.py` -> error

deeleeramone:
Getting somewhere, at least with new types of errors. :)

@hjoaquim, new way to break it, 

```
  File "/Users/danglewood/GitHub/OpenBBTerminal/argparse_translator/argparse_class_processor.py", line 81, in _custom_groups_from_reference
    return rp.custom_groups[route]
           ~~~~~~~~~~~~~~~~^^^^^^^

KeyError: '/econometrics/autocorrelation'

KeyError: '/quantitative/capm'
```

deeleeramone:
Now that I'm actually inside the operations, where do you want me to list bugs? 

All of the custom `regex` from the V3 Terminal will still be useful because `/` as a valid input breaks the CLI. For example, try setting a timezone.

deeleeramone:
Parameters that are `Literals` do not appear to even be getting passed through because otherwise I would be generating errors with incorrect choices.

```
/equity/screener --exchange asx
```

```
/equity/screener --exchange some_bad_exchange
```

Both produce the same output.


Another consideration is that if you are arbitrarily deciding which columns and rows get dropped, then there needs to be a way to sort/filter as part of the CLI command. Ideally you could do this without making a new request for the exact same data.

hjoaquim:
> Getting somewhere, at least with new types of errors. :)
> 
> @hjoaquim, new way to break it,
> 
> ```
>   File "/Users/danglewood/GitHub/OpenBBTerminal/argparse_translator/argparse_class_processor.py", line 81, in _custom_groups_from_reference
>     return rp.custom_groups[route]
>            ~~~~~~~~~~~~~~~~^^^^^^^
> 
> KeyError: '/econometrics/autocorrelation'
> 
> KeyError: '/quantitative/capm'
> 
> KeyError: '/technical/ad'
> ```
> 
> IIUC, toolkits and/or POST requests break it?

It shouldn't break! So I'll consider it a bug

hjoaquim:
> Now that I'm actually inside the operations, where do you want me to list bugs?
> 
> All of the custom `regex` from the V3 Terminal will still be useful because `/` as a valid input breaks the CLI. For example, try setting a timezone.

This is a good point - we might need to bring that back!

hjoaquim:
> Parameters that are `Literals` do not appear as choices no matter what. Additionally, they do not even get passed through to the Platform request, because otherwise I would be generating errors with incorrect choices.
> 
> A basic FAQ needs to be used as a low bar check, "how do I find stocks from India with OpenBB?"
> 
> Both of these arguments produce the same output.
> 
> ```
> /equity/screener --exchange asx
> ```
> 
> ```
> /equity/screener --exchange some_bad_exchange
> ```
> 
>     * 2 decimal places is not adequate for displaying a percent, especially when 1% == 0.01.
> 
> 
> The solution that does not involve guessing will be to read `json_schema_extra` from the data model fields, looking for `{"x-unit_measurement": "percent", "x-frontend_multiply": 100}
> 
> Another consideration is that if you are arbitrarily deciding which columns and rows get dropped, then there must be a way to sort/filter/limit/reduce as part of the CLI command. Without any controls what is displayed will be out of context.
> 
> Ideally you could do this without making a new request for the exact same data otherwise it is just a wasted API call and bandwidth. At least with the legacy interactive tables, there was the ability to filter/sort/refine the contents of the table.

Added these to the backlog and tagged you as reporter, thanks!

hjoaquim:
In order to successfully run this branch, you'll need to do a dev install, please follow:

    create new env: conda create --name platform-cli python=3.9
    install poetry and toml: pip install poetry toml
    install the cli: poetry install
    install latest platform: python openbb_platform/dev_install.py
    install latest charting extension: pip install -e openbb_platform/obbject_extensions/charting/

argparse_translator/argparse_class_processor.py
```diff
@@ -0,0 +1,148 @@
...
+            The target class whose methods will be processed.
+        add_help : Optional[bool]
+            Whether to add help to the ArgparseTranslators.
+        """
+        self._target_class: Type = target_class
+        self._add_help: bool = add_help
+        self._translators: Dict[str, ArgparseTranslator] = {}
+        self._paths: Dict[str, str] = {}
+
+        ArgparseClassProcessor._reference = reference if reference else {}
```
montezdesousa, line 42:
```suggestion
        ArgparseClassProcessor._reference = reference or {}
```

deeleeramone:
I'm not able to switch timezones:

```
timezone -t America/New_York

(🦋) /settings/ $ timezone -t America/New_York


(🦋) /settings/ $ timezone -t America

Invalid timezone. Please enter a valid timezone.
```

deeleeramone:
I can't see any differences when applying "`console_style`" settings.  What is supposed to happen between "light" and "dark"? The charts and tables each have their own preferences for light and dark, what is "console_style"?

IgorWounds:
> timezone -t America/New_York

Fixed @deeleeramone

hjoaquim:
> I can't see any differences when applying "`console_style`" settings. What is supposed to happen between "light" and "dark"? The charts and tables each have their own preferences for light and dark, what is "console_style"?

As the name says, it's the console "rich" style, example:
image

The difference it's subtle with the default style, but becomes more evident if you use a custom theme:
image

hjoaquim:
@deeleeramone added the other which I didn't answer directly to the Jira ticket so they can be separately addressed.

PR #5304, opened 2023-08-09
hotfix/options-hist: Fixes `options/hist` to display date column when raw is True

deeleeramone:
This PR fixes the `raw` output from `/stocks/options/hist` so that the date column is included in the table. 

Screenshot 2023-08-09 at 1 49 01 PM

Additionally:

- chartexchange_view.py now doesn't show the chart when `raw` or `export`.


openbb_terminal/stocks/options/chartexchange_view.py
```diff
@@ -109,22 +110,27 @@ def display_raw(
...
-    print_rich_table(
-        df,
-        headers=list(df.columns),
-        show_index=True,
-        index_name="Date",
-        title=f"{symbol.upper()} raw data",
-        export=bool(export),
-        limit=limit,
-    )
+    if export != "":
```
jmaslek, line 113: you can just do `if export`

"" evaluates to False

openbb_terminal/stocks/options/chartexchange_view.py
```diff
@@ -109,22 +110,27 @@ def display_raw(
...
+            export,
+            os.path.dirname(os.path.abspath(__file__)),
+            "hist",
+            df,
+            sheet_name,
+            fig,
+        )
+        return None
+
+    if raw is True:
```
jmaslek, line 124: same here. ` if raw`

jmaslek (approved):

PR #5307, opened 2023-08-11
Convert all return fields to snake_case

IgorWounds:
This PR converts all fields into snake_case for inner consistency.


hjoaquim (approved):
Consistency >>>
what do we need to also do it for `QueryParams`?

the-praxs (approved):

PR #5308, opened 2023-08-11
Dependency work to use the OpenBB SDK v4 on the OpenBBTerminal 

hjoaquim:
Exploring on how to have the SDK v4 integrated with the Terminal dependecy wise.

Known issues: 
- `reportlab` - is currently pinned on the Terminal (build error on MacOs)
  - Added the same restriction to the SDK to fix

- `statsmodel` - was using an older version on the Terminal
    - Bumped the version on the Terminal

- `pandas` - uses pandas = "^1.5.0 on the Terminal - degiro connector dependency is blocking us from updating it.
    - Downgraded the SDK to use the same as the Terminal


deeleeramone:
> pandas - uses pandas = "^1.5.0 on the Terminal - degiro connector dependency is blocking us from updating it.

This is going to be an ongoing theme.  Not using Pandas 2 will be more problematic into the future.

Screenshot 2023-08-11 at 1 15 10 PM

piiq:
@hjoaquim do you know who's pulling reportlab and why we need it?

deeleeramone:
Seems the Plotly backend uses it for PDF rendering, and the Portfolio menu uses it for something. 

> @hjoaquim do you know who's pulling reportlab and why we need it?

Screenshot 2023-08-15 at 10 58 37 AM

piiq:
> Seems the Plotly backend uses it for PDF rendering, and the Portfolio menu uses it for something.

@tehcoderer @andrewkenreich can we get rid of the render to pdf feature on our plotly backend? it seems we're not using it (at least i don't see it in the export tools), but in brings in some somewhat irritating dependencies like reportlab and svglib

image

andrewkenreich:
@piiq 
image
when you hit the far right button you have 3 options on export - We can just remove the pdf if it creates an issue

jmaslek:
> I vote for ditching svg and pdf from the core @jmaslek ?

kill it

deeleeramone:
> I vote for ditching svg and pdf from the core @jmaslek ?

Does this impact the ability to publish a report as a PDF?

deeleeramone:
@hjoaquim, got a new issue we can document, Numba:

- Numpy SDK-V4: 1.25.2
- Numpy Terminal: 1.23.4

What happens when I run the Terminal after installing SDK-V4:
 
Screenshot 2023-08-17 at 7 03 04 PM

tehcoderer:
@piiq we use to export images as pdf like `candle --export pdf`

image

piiq:
> @piiq we use to export images as pdf like `candle --export pdf`

@hjoaquim @tehcoderer if we really need it i would ask if it's possible to make it into a extension and remove from the core

jmaslek:
If it's on the charting end, isn't it in the extension already and not the core?

openbb_sdk/extensions/charting/pyproject.toml
```diff
@@ -12,12 +12,12 @@ openbb-core = { path = "../../sdk/core" }
 scipy = "^1.10.0"
 plotly = "^5.15.0"
 statsmodels = "^0.14.0"
-reportlab = "^4.0.4"
+reportlab = "<4" #Pinning due to build error on MacOs
 pywry = "^0.5.13"
 svglib = "^1.5.1"
 aiohttp = "^3.8.4"
 nbformat = "^5.9.0"
-pandas = ">1.5.3"
```
piiq, line 20: the caret ^ here would pin pandas to 1.X.X, is it what you actually want

piiq:
> If it's on the charting end, isn't it in the extension already and not the core?

you are right

openbb_sdk/extensions/charting/pyproject.toml
deeleeramone: Isn't this going to get extremely complex and messy when each extension has its own dependency list that will override the already installed versions defined in `openbb_core`?

piiq: exactly @deeleeramone . cleaning this up is what we need to do

the-praxs:
What's the status of this?

tehcoderer:
> What's the status of this?

out of scope

the-praxs:
> > What's the status of this?
> 
> out of scope

allegedly

piiq:
We'll re-address this given the changes to the core that have been made since this PR was created. Thanks for the initial exploration @hjoaquim

PR #5312, opened 2023-08-13
Fix routines creation from the Terminal to Hub

DidierRLopes:
This should include:
- [x] Check that title only has characters, digits or white spaces
- [x] Title needs to be provided
- [x] Default description to generic sentence
- [x] Allow to run exe and record only from main menu. But allow stop from anywhere. - this guarantees the order of stuff and makes it easier for us.
- [x] Add auto-completer for tags in record command
- [x] Check that tags selected by user are valid
- [x] Replace white spaces by underscores to save file locally
- [x] Create a flag that allows files to be saved locally, otherwise default to hub
- [x] Improve overriding of local routines logic
- [x] Add description as comment in local file
- [x] Add tags as comment in local file
- [x] Check that the tags selected are valid. Do it at record time and not "stop"
- [x] Ensure that there’s at least 4 commands being run before stop
- [x] Check that tags are validated before uploading to Hub
- [x] Check that the sharing link works with the Hub
- [x] Check that title, description and tags are correctly set in the Hub

@jmaslek please check PR and if you can fix any linting and others, I'd appreciate it 🙏🏽


powru9-gobVok-nutguj (commented):
[]()

powru9-gobVok-nutguj (commented):
[]()

PR #5314, opened 2023-08-13
Feature/sdk-v4-futures-term-structure - CBOE Provider Improvements

deeleeramone:
Refactors the helpers to get rid of loads on run time, unnecessary code, and also caches symbol directories for better performance (from #5338).

- updates `pyproject.toml` with required dependencies for `openbb_cboe`.

Adds CBOE as a provider for:

`obb.futures.curve()`

- Use one of these tickers to test the futures curve: ['VX', 'VXM', 'IBHY', 'IBIG', 'AMB1', 'AMT1', 'AMT3']
- The `date` parameter can be used to get the settlement price on that day for active contracts only, so the farther back you go, the less results return.

`obb.economy.available_indices()`
`obb.economy.index()`
`obb.economy.european_index()`
`obb.economy.european_index_constituents()`
`obb.economy.available_indices(europe=True, provider="cboe")`
`obb.economy.search_index("sector")`
`obb.economy.search_index("sector", europe=True)`
`obb.economy.search_index("six", ticker=True)`

Other general items:

- remove VWAP from provider standard model of `major_indices_eod`
- enforce snake_case columns for FMP, `obb.economy.index()`
- allow `^` in symbol to not break FMP, `obb.economy.index()`


deeleeramone:
Closing because this has been folded into #5339.

PR #5319, opened 2023-08-14
Creates a "--dev" flag that points to dev backend for testing with hub

DidierRLopes:
Allows to use --dev flag to point to .dev backend


tehcoderer:
fixedd 

lol your class wasn't a singleton so the urls weren't getting changed globally

DidierRLopes:
> fixedd
> 
> lol your class wasn't a singleton so the urls weren't getting changed globally

Ups.

Thank you!

PR #5329, opened 2023-08-15
hotfix/fix-stocks-search-sdk3: Fixes the stocks search so that it doesn't `print_rich_table()` for SDK.

deeleeramone:
Currently, if you use the SDK and `stocks.search()`, it will open a PyWry window while simultaneously returning a DataFrame.  This takes care of that.

### DIFF
--- a/openbb_terminal/stocks/stocks_helper.py
+++ b/openbb_terminal/stocks/stocks_helper.py
@@ -228,43 +228,11 @@ def search(
 
     df = df[["name", "country", "sector", "industry_group", "industry", "exchange"]]
     # To automate renaming columns
-    headers = [col.replace("_", " ") for col in df.columns.tolist()]
-
-    title = "Companies found"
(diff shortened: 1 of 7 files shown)


openbb_terminal/stocks/stocks_helper.py
```diff
@@ -1108,3 +1073,45 @@ def heikin_ashi(data: pd.DataFrame) -> pd.DataFrame:
     ]
 
     return pd.concat([data, ha], axis=1)
+
+
+def calculate_adjusted_prices(df: pd.DataFrame, column: str, dividends: bool = False):
```
deeleeramone, line 1078: This helper was committed unintentionally and is part of fixing something unrelated.  It can stay or go, but can be used to help fix some issues with the AlphaVantage adjusted prices data.

jmaslek (approved):

PR #5338, opened 2023-08-16
Small v4 cboe refactor

jmaslek:
Adding a 7 day cache on the import requests so that subseuent imports dont need to make the call.  The time can be adjusted if the files change more often, but I imagine they are pretty static.

Saves about 5 s on subsequent imports for me (also defaults to using sqlite in the default cache location, but could be changed if needed or put behind a flag of some sort)


the-praxs (approved):
Superb work Doc!

PR #5339, opened 2023-08-16
feature/quandl-provider: Add Quandl provider to SDK-V4

deeleeramone:
This PR adds Quandl (Nasdaq Data Link) to the data providers. Functions currently implemented: 

- `obb.economy.sp500_multiples()`
  - 36 different time series.  The default series is set as, "PE Ratio by Month".

  "Real Price by Month" offers the longest continuous series at ~150 years, on a monthly interval.

Screenshot 2023-08-16 at 4 36 55 PM

- `obb.economy.cot()`
  - CFTC Commitment of Traders Reports.  All Quandl parameters implemented.
  - Get reports by code ("`03565B`"), symbol ("`NG=F`"), or full name ("`Henry Hub Natural Gas Futures (NYM)`")
- `obb.economy.cot_search("energy")`
  - Curated list of assets to assist with finding the codes for the most common reports.

Similar to company financial statements, the number of columns returned, and their names, are dynamic; based on the combination of parameters and the underlying asset.  Standardization, or declaring all fields in the response data is not particularly helpful here, IMO.

This PR also refactors the CBOE provider, cleans up the code, and introduces data for over 1000 indices from the US and Europe:

Adds CBOE as a provider for:

- `obb.futures.curve()`

    Use one of these tickers to test the futures curve: ['VX', 'VXM', 'IBHY', 'IBIG', 'AMB1', 'AMT1', 'AMT3']
    The date parameter can be used to get the settlement price on that day for active contracts only, so the farther back you go, the less results return.

- `obb.economy.available_indices()`
  - `obb.economy.available_indices(europe=True, provider="cboe")`
- `obb.economy.index("VIX")`
  - `obb.economy.index("VIX", interval="1m")` 
- `obb.economy.european_index("BUK350P")`
  - `obb.economy.european_index("BUK350P", interval="1m")` 
- `obb.economy.european_index_constituents("BUK350P")`
- `obb.economy.index_search("sector")`
  - `obb.economy.index_search("sector", europe=True)`
  - `obb.economy.index_search("six", ticker=True)`
- `obb.economy.index_snapshots()`
  - `obb.economy.index_snapshots(region="EU")`

Adds Intrinio as a provider for:

- `obb.stocks.options.chains()`

Other general items:

- allow ^ in symbol to not break FMP, obb.economy.index()
- remove VWAP from `MajorIndicesEOD` and `StockEOD` standard models.
- allow for both `dateType` and `datetime` to be accepted by `MajorIndicesEOD` and `StockEOD` standard models.
- fix date validation in FMP intraday stripping time.


IgorWounds:
Hi,

The definition of a standard field is: "A field shared by two or more providers". Thus, I suggest you revert your changes regarding it. 

Also, I suggest you revert the snake_case enforcement as another PR I pushed a couple of days ago does this dynamically and the provider-specific models should be provider-specific and stick to their CamelCase.

deeleeramone:
> The definition of a standard field is: "A field shared by two or more providers". Thus, I suggest you revert your changes regarding it.

I disagree with this definition.  If two providers have field "VWAP" and three do not, does this make "VWAP" a standard field?

IgorWounds:
> > The definition of a standard field is: "A field shared by two or more providers". Thus, I suggest you revert your changes regarding it.
> 
> I disagree with this definition. If two providers have field "VWAP" and three do not, does this make "VWAP" a standard field? In the case of `MajorIndicesEOD` and FMP, it's not even clear why this column exists at all because, there is no "volume" to calculate a VWAP.

Hi,

This is the rule that all standardized models follow. We have also made it clear that if a third provider doesn't have such a field ->  that  becomes an Optional standard field. Moreover, all completely empty columns are dropped when using our helper methods.

"If two providers have field "VWAP" and three do not, does this make "VWAP" a standard field? " - Yes, it makes it an Optional standard field. I do agree that this threshold might change when we have 20+ providers but for now, it is 2 or more.

As for the MajorIndices, it could be a mistake that leaked into the standard model. Can @the-praxs validate this?

the-praxs:
> > > The definition of a standard field is: "A field shared by two or more providers". Thus, I suggest you revert your changes regarding it.
> > 
> > 
> > I disagree with this definition. If two providers have field "VWAP" and three do not, does this make "VWAP" a standard field? In the case of `MajorIndicesEOD` and FMP, it's not even clear why this column exists at all because, there is no "volume" to calculate a VWAP.
> 
> Hi,
> 
> This is the rule that all standardized models follow. We have also made it clear that if a third provider doesn't have such a field -> that becomes an Optional standard field. Moreover, all completely empty columns are dropped when using our helper methods.
> 
> "If two providers have field "VWAP" and three do not, does this make "VWAP" a standard field? " - Yes, it makes it an Optional standard field. I do agree that this threshold might change when we have 20+ providers but for now, it is 2 or more.
> 
> As for the MajorIndices, it could be a mistake that leaked into the standard model. Can @the-praxs validate this?

The standardization that @IgorWounds mentioned is our current definition. We have to abide by it in all circumstances until further revision of it.

The MajorIndices one was a bug, it was copied over from other EOD (now Historical) endpoints as they share the same code structure. Everything is fixed.

@deeleeramone I am merging this PR because it is big and there are many items that need to be modified to abide by our current guidelines, whether standardization, contribution, or the code structure.

PR #5358, opened 2023-08-21
Hotfix/squeeze import

montezdesousa:
Reduces import time by removing requests to github and lazy loading some modules like `statsmodels`, `scipy`, etc.

Once `cboe` requests are solved import time is reduced to around 2/3s after cold start.

without cboe requests
Screenshot 2023-08-21 at 17 21 26


openbb_sdk/sdk/core/openbb_core/app/logs/formatters/formatter_with_exceptions.py
```diff
@@ -14,9 +14,7 @@ class FormatterWithExceptions(logging.Formatter):
 
     DATEFORMAT = "%Y-%m-%dT%H:%M:%S%z"
     LOGFORMAT = "%(asctime)s|%(name)s|%(funcName)s|%(lineno)s|%(message)s"
-    LOGPREFIXFORMAT = (
-        "%(levelname)s|%(appName)s|%(commitHash)s|%(appId)s|%(sessionId)s|%(userId)s|"
-    )
+    LOGPREFIXFORMAT = "%(levelname)s|%(appName)s|%(appId)s|%(sessionId)s|%(userId)s|"
```
hjoaquim, line 17: This might not be a good idea. Although we're not using the log files anymore but Posthog instead, this would break the CSV structure we have in place.
I would advocate for us keeping the structure but simply hardcore `unknown-commit` as the "commitHash" value.

Wdyt @montezdesousa ?

montezdesousa: sure, thanks for the review

PR #5363, opened 2023-08-22
Feature/code cleanup

IgorWounds:
Code cleanup.


openbb_sdk/sdk/core/openbb_core/app/static/package/crypto.py
```diff
@@ -1,9 +1,8 @@
 ### THIS FILE IS AUTO-GENERATED. DO NOT EDIT. ###
 
 import datetime
-from typing import List, Literal, Union
+from typing import Annotated, List, Literal, Optional, Union
```
hjoaquim, line 4: this is not 3.8 compatible, we need to generate the static assets on a 3.8 environment - that's probably why tests are failing

hjoaquim (changes requested):

openbb_sdk/sdk/core/openbb_core/app/static/package/crypto.py
IgorWounds: Do you know how this change was introduced?

the-praxs (approved):
@IgorWounds lord of dicts

PR #5365, opened 2023-08-23
Adding AV and StockEOD

hjoaquim:
^

### DIFF
--- a/openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/__init__.py
+++ b/openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/__init__.py
@@ -1,0 +1,20 @@
+"""Alpha Vantage Provider module."""
+from openbb_provider.abstract.provider import Provider
+
+from openbb_alpha_vantage.models.stock_eod import AlphaVantageStockEODFetcher
+
+alpha_vantage_provider = Provider(
+    name="alpha_vantage",
+    website="https://www.alphavantage.co/documentation/",
+    description="""Alpha Vantage provides realtime and historical
+    financial market data through a set of powerful and developer-friendly data APIs
+    and spreadsheets. From traditional asset classes (e.g., stocks, ETFs, mutual funds)
+    to economic indicators, from foreign exchange rates to commodities,
+    from fundamental data to technical indicators, Alpha Vantage
+    is your one-stop-shop for enterprise-grade global market data delivered through
+    cloud-based APIs, Excel, and Google Sheets. """,
+    required_credentials=["api_key"],
+    fetcher_dict={
+        "StockEOD": AlphaVantageStockEODFetcher,
+    },
+)
--- a/openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
+++ b/openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
@@ -1,0 +1,208 @@
+"""Alpha Vantage Stock End of Day fetcher."""
+
+
+from datetime import datetime
+from typing import Any, Dict, List, Literal, Optional, get_args
+
+from dateutil.relativedelta import relativedelta
+from openbb_provider.abstract.fetcher import Fetcher
+from openbb_provider.standard_models.stock_eod import StockEODData, StockEODQueryParams
+from openbb_provider.utils.helpers import get_querystring
+from pydantic import Field, root_validator, validator, PositiveFloat, NonNegativeFloat
+
+from openbb_provider.utils.descriptions import DATA_DESCRIPTIONS
+
+import pandas as pd
+
+
+class AlphaVantageStockEODQueryParams(StockEODQueryParams):
+    """Alpha Vantage Stock End of Day Query.
+
+    Source: https://www.alphavantage.co/documentation/
+    """
+
+    function: Literal[
+        "TIME_SERIES_INTRADAY",
+        "TIME_SERIES_DAILY",
+        "TIME_SERIES_DAILY_ADJUSTED",
+        "TIME_SERIES_WEEKLY",
+        "TIME_SERIES_WEEKLY_ADJUSTED",
+        "TIME_SERIES_MONTHLY",
+        "TIME_SERIES_MONTHLY_ADJUSTED",
+    ] = Field(
+        description="The time series of your choice. ",
+        default="TIME_SERIES_DAILY_ADJUSTED",
+    )
+    interval: Optional[Literal["1min", "5min", "15min", "30min", "60min"]] = Field(
+        description="The interval between two consecutive data points in the time series.",
+        default="60min",
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+        required_on_functions=["TIME_SERIES_INTRADAY"],
+    )
+    adjusted: Optional[bool] = Field(
+        description="Output time series is adjusted by historical split and dividend events.",
+        default=True,
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+    )
+    extended_hours: Optional[bool] = Field(
+        description="Extended trading hours during pre-market and after-hours.",
+        default=False,
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+    )
+    month: Optional[str] = Field(
+        description="Query a specific month in history (in YYYY-MM format).",
+        default=None,
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+    )
+    outputsize: Optional[Literal["compact", "full"]] = Field(
+        description="Compact returns only the latest 100 data points in the intraday "
+        "time series; full returns trailing 30 days of the most recent intraday data "
+        "if the month parameter (see above) is not specified, or the full intraday "
+        "data for a specific month in history if the month parameter is specified.",
+        default="compact",
+        available_on_functions=[
+            "TIME_SERIES_INTRADAY",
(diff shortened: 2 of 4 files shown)


openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,217 @@
...
+    interval: Optional[Literal["1min", "5min", "15min", "30min", "60min"]] = Field(
+        description="The interval between two consecutive data points in the time series.",
+        default="60min",
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+        required_on_functions=["TIME_SERIES_INTRADAY"],
+    )
+    adjusted: Optional[bool] = Field(
+        description="Output time series is adjusted by historical split and dividend events.",
+        default=True,
+        available_on_functions=["TIME_SERIES_INTRADAY"],
```
deeleeramone, line 40: Adjusted is funny, this param is actually only valid for non-intraday endpoints and it doesn't really adjust the time series.  Instead, it adds another column "Adj Close" and all other columns remain unadjusted.  In the current Terminal, this breaks the candle charts because one column is not aligned with the others.

In a dream world, `adjusted=True` would only split-adjust prices and volume, leaving dividend adjustments out of the equation entirely.  Dividend adjustments aren't a reflection of the real price, just an assumption for valuation models.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,217 @@
...
+        description="Query a specific month in history (in YYYY-MM format).",
+        default=None,
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+    )
+    outputsize: Optional[Literal["compact", "full"]] = Field(
+        description="Compact returns only the latest 100 data points in the intraday "
+        "time series; full returns trailing 30 days of the most recent intraday data "
+        "if the month parameter (see above) is not specified, or the full intraday "
+        "data for a specific month in history if the month parameter is specified.",
+        default="compact",
```
deeleeramone, line 57: Probably want to have this set as "full" otherwise it will override any `start_date` and `end_date` settings.  Not much point in returning only the last 100 minutes (if one-minute).

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,217 @@
...
+from pydantic import Field, NonNegativeFloat, PositiveFloat, root_validator, validator
+
+
+class AlphaVantageStockEODQueryParams(StockEODQueryParams):
+    """Alpha Vantage Stock End of Day Query.
+
+    Source: https://www.alphavantage.co/documentation/
+    """
+
+    function: Literal[
```
deeleeramone, line 22: We should consider consolidating all interval choices into one, standardized parameter. Currently (Terminal and SDK V3), `monthly=True`.  Presently in SDK V4, there is:

- `interval="1mo"`
- `timespan="month"`
- `function="TIME_SERIES_MONTHLY"`
- `period="1mo"`

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""Alpha Vantage Provider module."""
+from openbb_provider.abstract.provider import Provider
+
+from openbb_alpha_vantage.models.stock_eod import AlphaVantageStockEODFetcher
```
the-praxs, line 4: You can rename AlphaVantage to AV to make the class names shorter :)

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,217 @@
...
+        "TIME_SERIES_MONTHLY",
+    ] = Field(
+        description="The time series of your choice. ",
+        default="TIME_SERIES_DAILY",
+    )
+    interval: Optional[Literal["1min", "5min", "15min", "30min", "60min"]] = Field(
+        description="The interval between two consecutive data points in the time series.",
+        default="60min",
+        available_on_functions=["TIME_SERIES_INTRADAY"],
+        required_on_functions=["TIME_SERIES_INTRADAY"],
```
the-praxs, line 35: Not needed if the above code is implemented.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
the-praxs: This can be worked on in the `transform_data` function. Not an issue atm.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
the-praxs: Change the default to `full`. That's it.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,217 @@
...
+    @staticmethod
+    def extract_data(
+        query: AlphaVantageStockEODQueryParams,
+        credentials: Optional[Dict[str, str]],
+        **kwargs: Any,
+    ) -> dict:
+        """Return the raw data from the Alpha Vantage endpoint."""
+
+        api_key = credentials.get("alpha_vantage_api_key") if credentials else ""
+
```
the-praxs, line 196: Here you can convert the `period` to upper case, attach it to `TIME_SERIES` string.

You can also check if the functions and the parameters are correct and raise the Exception. This is optional because for the wrong parameters, there will be an exception raised by the AV API.

the-praxs (changes requested):
@hjoaquim there are ways to optimize the code so that it follows the structure of the other data providers and is human-readable. Otherwise, LGTM 😄. Fantastic work! 🚀

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
hjoaquim: @the-praxs I propose we keep the `function` attribute so we can still have the validators that check `available_on_functions` and `required_on_functions`. IMO, it also makes clear that not all functions require the same query args and them being required vary from function to function.

Something we can/should do is to hide the `function` field with `exclude=True` and set it up on a validator based on the `period`.

Wdyt

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
hjoaquim: Would go for rely only on the validators and pydantic's own data validation and let the htpp response we'll get from AV do the rest.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
the-praxs: I believe that having the `function` attribute reduces user experience because they need to provide the whole parameter. Also, this attribute shares the same look as the `period` which we can use to make it simpler for the end user, in line with the other providers. The necessary details can be mentioned in the descriptions.

If there is a way to make the period versatile and prevent erroneous parameter compatibility, I will be intrigued on how we can do this. I haven't tried the `exclude=True` option so if you can make it work it would be good.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
the-praxs: Good idea!

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
hjoaquim: @the-praxs latest commit achieves this by having the function field as private. So the user does not have to worry about poor ux but we can keep validation for function specific fields!

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,244 @@
...
+        query_dict = query.dict()
+        query_dict["function"] = query_dict.pop("_function")
+
+        query_str = get_querystring(query_dict, ["start_date", "end_date"])
+
+        url = f"https://www.alphavantage.co/query?{query_str}&datatype=csv&apikey={api_key}"
```
the-praxs, line 226: Just an idea for data like this: we can store it as a binary file for a user.

openbb_sdk/providers/alpha_vantage/openbb_alpha_vantage/models/stock_eod.py
```diff
@@ -0,0 +1,244 @@
...
+
+        data = pd.read_csv(url)
```
the-praxs, line 228: if we can avoid using pandas, its much better.

the-praxs (approved):
@hjoaquim well done!

PR #5402, opened 2023-09-03
Router cleanup

montezdesousa:
Same functionality, just splits some code into functions and fixes a small bug when return type hint is `OBBject[List]` .


tehcoderer (commented):
LGTM! 🚀🚢

PR #5419, opened 2023-09-08
Fix/provider-bugs-intrinio: fix bad API key response for options symbols

deeleeramone:
Two small fixes:

- Bad API key response for querying symbol directory
- Correct DTE calculation because `now` is within the `date` field, +1 is for when `now` == datetime.now()

Then removes the functions in `helpers` that were moved from `helpers` to `options_chains`

### DIFF
--- a/openbb_sdk/providers/intrinio/openbb_intrinio/models/options_chains.py
+++ b/openbb_sdk/providers/intrinio/openbb_intrinio/models/options_chains.py
@@ -37,213 +37,13 @@ intrinio_session = requests_cache.CachedSession(
 )
 
 
-def get_options_tickers(api_key: str) -> List[str]:
-    """Returns all tickers that have existing options contracts.
-
-    Parameters
-    ----------
-    api_key: str
-        Intrinio API key.
-
-    Returns
-    -------
-    List[str]
-        List of tickers
-    """
-
-    url = f"https://api-v2.intrinio.com/options/tickers?api_key={api_key}"
-
-    r = intrinio_session.get(url, timeout=10)
-
-    if r.status_code != 200:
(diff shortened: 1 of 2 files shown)


openbb_sdk/providers/intrinio/openbb_intrinio/models/options_chains.py
```diff
@@ -302,6 +108,219 @@ class IntrinioOptionsChainsFetcher(
...
+                urls.append(
+                    IntrinioOptionsChainsFetcher.generate_url(
+                        symbol, expiration, date, api_key
+                    )
+                )
+            results = IntrinioOptionsChainsFetcher.download_json_files(urls)
+
+        df = pd.DataFrame()
+
+        for i in range(0, len(results)):
```
the-praxs, line 229: This function has n^2 complexity and can be optimized. Hardcoding can be avoided.

openbb_sdk/providers/intrinio/openbb_intrinio/models/options_chains.py
```diff
@@ -302,6 +108,219 @@ class IntrinioOptionsChainsFetcher(
 ):
     """Transform the query, extract and transform the data from the Intrinio endpoints"""
 
+    @staticmethod
+    def get_options_tickers(api_key: str) -> List[str]:
```
the-praxs, line 112: Functions like these use a common functionality so they must be in helpers. Setting them as `staticmethod` does not solve the issue.

the-praxs (changes requested):

the-praxs (approved):

the-praxs:
LGTM.

PR #5424, opened 2023-09-09
Allow user to set what do they want the data format to be by default

DidierRLopes:
As a user, if I'm doing a research on a jupyter notebook is annoying to have to do `to_dataframe()` every single time.

There's a high chance that whichever data output a user is looking for will remain the same within that same notebook / app, hence, this PR introduces a setting that allows the user to set what data type they are interested in having as default.

I just did DataFrame and Dict as an example. 

If everyone is happy with this solution we can then integrate the others: polars, numpy, ...

CC: @the-praxs thanks for the instructions on how to point to the local lib 👊

Screenshot 2023-09-09 at 2 24 12 AM

### DIFF
--- a/openbb_sdk/sdk/core/openbb_core/app/command_runner.py
+++ b/openbb_sdk/sdk/core/openbb_core/app/command_runner.py
@@ -2,6 +2,7 @@ import warnings
 from contextlib import nullcontext
 from copy import deepcopy
 from datetime import datetime
+import pandas as pd
 from inspect import Parameter, signature
 from sys import exc_info
 from time import perf_counter_ns
@@ -345,7 +346,7 @@ class StaticCommandRunner:
         /,
         *args,
         **kwargs,
-    ) -> OBBject:
+    ) -> Union[OBBject, Dict, pd.DataFrame]:
         timestamp = datetime.now()
         start_ns = perf_counter_ns()
 
@@ -372,6 +373,13 @@ class StaticCommandRunner:
                 "route": route,
                 "timestamp": timestamp,
             }
+
+        # Based on user configuration for data output
+        # we return the data in the desired format
+        if execution_context.user_settings.preferences.data_output == "df":
+            return obbject.to_dataframe()
+        if execution_context.user_settings.preferences.data_output == "dict":
+            return obbject.to_dict()
 
         return obbject
 
--- a/openbb_sdk/sdk/core/openbb_core/app/model/preferences.py
+++ b/openbb_sdk/sdk/core/openbb_core/app/model/preferences.py
@@ -19,6 +19,7 @@ class Preferences(BaseModel):
     table_style: Literal["dark", "light"] = "dark"
     request_timeout: PositiveInt = 15
     metadata: bool = True
+    data_output: Literal["OBBject", "df", "dict"] = "df"
 
     class Config:
         validate_assignment = True


openbb_sdk/sdk/core/openbb_core/app/command_runner.py
```diff
@@ -373,6 +374,13 @@ def run(
                 "timestamp": timestamp,
             }
 
+        # Based on user configuration for data output
+        # we return the data in the desired format
+        if execution_context.user_settings.preferences.data_output == "df":
+            return obbject.to_dataframe()
```
montezdesousa, line 380: I think doing the conversion here will break the API since it relies on `OBBject` as output and we avoid returning DataFrames there.

I'd rather do this in directly in the python UI. https://github.com/OpenBB-finance/OpenBBTerminal/tree/feature/openbb-sdk-v4/openbb_sdk/openbb/package

hjoaquim (commented):
I totally get the idea, and I agree that can be helpful and can save some typing time.

But I'm not sure if I'm a fan, here is why:
 - dataframes break the API (@montezdesousa suggestion is a workaround for this - although it's yet another detail that differs from the python interface to the api - ~~will potentially need to be an env variable instead of a user preference (during the execution of the `PackageBuilder` there isn't a user yet)) ~~ edit: @montezdesousa comment above - we can actually compute it in runtime, no need to do it during the static assets generation
 - we lose `OBBject` methods (specially the charting capabilities) + metadata available with it
 - what happens if you're using an endpoint that doesn't return a dataframe? We'll probably get an error, understand that we need to change the user preferences, use the endpoint and change the preferences again to return dataframe only? The time we saved initially was now lost in this process.

This is reasonable to me and in the end if the user wants this behavior needs to accept the tradeoffs!

openbb_sdk/sdk/core/openbb_core/app/command_runner.py
montezdesousa: Screenshot 2023-09-11 at 11 33 01

piiq:
I'm a fan of Didier's PRs but I I'm not sure if I'm a fan of this one 🤣😂

I'm still thinking if the trade-offs outweigh the benefits in this particular case. I understand the intent to save time on typing but should it be on the core. Casting the object to dataframe by default should work for a specific use case but apart from a broken api and the things added by henrique (no warnings, errors, helpers) I would add that not everything is nicely dataframable (see `obb.news.globalnews().to_dataframe()`) and with that in place there would be almost no way to get that nested dict that was returned by the provider.

The biggest risk I see is having it as a system setting. The system settings are system wide and not project wide (if you don't explicitly set it at runtime) and i easily envision a person following a notebook that they downloaded from the internet and messing up the system wide openbb configuration

On the other hand, can this be an extension? The BaseApp can be instantiated with a custom CommandRunner, so technically yes

montezdesousa:
> I'm a fan of Didier's PRs but I I'm not sure if I'm a fan of this one 🤣😂
> 
> I'm still thinking if the trade-offs outweigh the benefits in this particular case. I understand the intent to save time on typing but should it be on the core. Casting the object to dataframe by default should work for a specific use case but apart from a broken api and the things added by henrique (no warnings, errors, helpers) I would add that not everything is nicely dataframable (see `obb.news.globalnews().to_dataframe()`) and with that in place there would be almost no way to get that nested dict that was returned by the provider.
> 
> The biggest risk I see is having it as a system setting. The system settings are system wide and not project wide (if you don't explicitly set it at runtime) and i easily envision a person following a notebook that they downloaded from the internet and messing up the system wide openbb configuration
> 
> On the other hand, can this be an extension? The BaseApp can be instantiated with a custom CommandRunner, so technically yes

@piiq I wonder if we could just create an ipython magic method to improve the ux while not touching the core or anything else in the code base.

piiq:
i mean we can have a setting that would be used similarly how we sometimes

```python
import warnings
warnings.filterwarnings("ignore")
```

but i would not make it system wide

jmaslek:
@montezdesousa 

```
from IPython.core.magic import register_cell_magic
from openbb import obb
@register_cell_magic
def todf(line, cell):
    result = eval(cell)
    if isinstance(result, OBBject):
        result = result.to_dataframe()
    return result
```

```
%%todf

obb.stocks.load("AAPL")
```

Can we just include this somewhere?

montezdesousa:
> @montezdesousa
> 
> ```
> from IPython.core.magic import register_cell_magic
> from openbb_core.app.model.obbject import OBBject
> from openbb import obb
> @register_cell_magic
> def todf(line, cell):
>     result = eval(cell)
>     if isinstance(result, OBBject):
>         result = result.to_dataframe()
>     return result
> ```
> 
> To get a df:
> 
> ```
> %%todf
> 
> obb.stocks.load("AAPL")
> ```
> 
> Can we just include a set of these somewhere?

Yes here "openbb_sdk/openbb/__init__.py". Is there a way to make it persistent on several cells?

jmaslek:
> Yes here "openbb_sdk/openbb/**init**.py". Is there a way to make it persistent on subsequent cells?

Looks like: `IPython.core.getipython.get_ipython().events.register` can do it.

Since this requires IPython, would it make sense to make this an extension, like openbb-magics ?

DidierRLopes:
@piiq @jmaslek @montezdesousa @hjoaquim Thank you for chiming in.

I agree with all those points. Particularly, that it should only work on a project basis and remain visible. This so that:
1. It doesn't break other user apps.  E.g. if they used it as an API for a web app, then were playing with it on a notebook and setting the output to DF. We don't want that setting to break the web app.
2. If another user runs code from this user, if the setting was system-wide then it wouldn't be visible, which impacts the UX.

So far my PR is having the intended consequences which are generating a discussion on this topic, which ultimately will lead to a big improvement in the user experience 😄 (once you actually try to use it for a bit, it's easy to realize how painful typing `.to_df()` every time is).

montezdesousa:
> > Yes here "openbb_sdk/openbb/**init**.py". Is there a way to make it persistent on subsequent cells?
> 
> Looks like: `IPython.core.getipython.get_ipython().events.register` can do it.
> 
> Since this requires IPython, would it make sense to make this an extension, like openbb-magics ?

It would be nice to have it right away as an option without another package. Do you see potential problems in including IPython into dependency tree?

DidierRLopes:
> > > Yes here "openbb_sdk/openbb/**init**.py". Is there a way to make it persistent on subsequent cells?
> > 
> > 
> > Looks like: `IPython.core.getipython.get_ipython().events.register` can do it.
> > Since this requires IPython, would it make sense to make this an extension, like openbb-magics ?
> 
> It would be nice to have it right away as an option without another package. Do you see potential problems in including IPython into dependency tree?

image

piiq:
> Do you see potential problems in including IPython into dependency tree?

Ah yes - lines 27-99 in this file https://github.com/ipython/ipython/blob/main/setup.cfg 🤣
Don't get me wrong, I love jupyter and use it every day, but i think it's just an unnecessary dependency for the core and should be an extension.

montezdesousa:
> > Do you see potential problems in including IPython into dependency tree?
> 
> Ah yes - lines 27-99 in this file https://github.com/ipython/ipython/blob/main/setup.cfg 🤣 Don't get me wrong, I love jupyter and use it every day, but i think it's just an unnecessary dependency for the core and should be an extension.

alright, seems valid reason for different package lol

jmaslek:
Another idea here: #5441

deeleeramone:
>     * we lose `OBBject` methods (specially the charting capabilities) + metadata available with it
 
You get to do, `DataFrame.plot()`, instead. You might actually gain more flexibility where it comes to charting because of the ability to set a preferred backend - Plotly, MPL, Seaborne.

99% of the time the metadata is of no interest, or use, to the user.  With the exception of when that metadata is actually the contents of the Series/DataFrame used on the input to a function and the results are not a time series.

hjoaquim:
Can we close this in detriment of https://github.com/OpenBB-finance/OpenBBTerminal/pull/5471#issuecomment-1735338784?

montezdesousa:
See https://github.com/OpenBB-finance/OpenBBTerminal/pull/5471

PR #5428, opened 2023-09-11
fix/allow-date - allow date to be returned when it is just a date

deeleeramone:
Screenshot 2023-09-10 at 10 37 20 PM


the-praxs:
@deeleeramone ready fo review?

PR #5430, opened 2023-09-11
Feature/improved helpers

IgorWounds:
Improve the `to_dataframe` and `to_dict` helpers.


openbb_sdk/sdk/core/openbb_core/app/model/obbject.py
montezdesousa, line 1: Here why do we need to sort/not-sort the columns?

openbb_sdk/sdk/core/openbb_core/app/utils.py
montezdesousa, line 1: Do we need to specify the columns order here as well? The dict will preserve them iiuc

openbb_sdk/openbb/package/crypto.py
montezdesousa, line 1: Is it possible to generate this with python 3.8 so that we don't get changes?

openbb_sdk/openbb/package/__extensions__.py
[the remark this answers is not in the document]
IgorWounds: Might have been a mistake in my env. I'll make sure it is rebuilt.

openbb_sdk/sdk/core/openbb_core/app/model/obbject.py
IgorWounds: We don't want to sort the columns where we follow the Pydantic model ordering. Where we don't follow it, we sort the columns for consistency and also to make our tests more deterministic. I hope this answers the question.

openbb_sdk/openbb/package/crypto.py
IgorWounds: Yes, I'll do that. Thanks!

openbb_sdk/sdk/core/openbb_core/app/utils.py
IgorWounds: Just saw that it was an artifact from my previous testing. Thanks for spotting this!

PR #5433, opened 2023-09-12
Test for provider field duplication check

hjoaquim:
Add a test that checks for provider field duplication - those that either should be standard or optional standard.

How it works:

1. Find all standard models
   - Inheriting from `QueryParams`
   - Inheriting from `Data`

2. Find all provider models 

3. Create a dict that match provider models with its parents:

```
{
     'StandardModel_1': 
       [
          {'FMPSTandardModel_1': ['field_1', 'field_2', ...]},
          {'FredSTandardModel_1': ['field_1', 'field_2', ...]},
          ...
       ],
     'StandardModel_2': 
       [
          {'FMPSTandardModel_2': ['field_1', 'field_2', ...]},
          {'FredSTandardModel_2': ['field_1', 'field_2', ...]},
          ...
       ],
     ...
}
```

4. Merge all fields of all provider models under each standard model

6. Find duplicates on the list created on (4)


hjoaquim:
Do not merge yet, tests should be failing on the GH actions.

openbb_sdk/providers/tests/test_provider_field_dupes.py
```diff
@@ -0,0 +1,189 @@
...
+                continue
+
+            # Use fnmatch to find .py files
+            for filename in fnmatch.filter(files, "*.py"):
+                provider_model_files.append(os.path.join(root, filename))
+
+        child_parent_dict = {}
+
+        for file in provider_model_files:
+            splited_path = file.split("/")
```
the-praxs, line 153: Please modify this to also work with Windows ('\')

openbb_sdk/providers/tests/test_provider_field_dupes.py
```diff
@@ -0,0 +1,189 @@
...
+
+                fields = []
+                provider_models = []
+
+                for provider_cls in providers_w_fields:
+                    provider_models.extend(list(provider_cls.keys()))
+                    fields.extend(list(provider_cls.values())[0])
+
+                seen = set()
+                dupes = [x for x in fields if x in seen or seen.add(x)]
```
the-praxs, line 184: Please create a set out of this to resolve duplicate fields in the output.

openbb_sdk/providers/tests/test_provider_field_dupes.py
```diff
@@ -0,0 +1,189 @@
...
+
+                assert not dupes, (
+                    f"The following fields are duplicated and should be standardized: {dupes}.\n"
+                    f"Standard model: {std_cls}, Provider models: {', '.join(provider_models)}\n"
```
the-praxs, line 188: We could show the providers and the fields collectively that are found rather than just the fields!

the-praxs (changes requested):
@hjoaquim you are a testing god!

hjoaquim:
@jmaslek it looks like only tests inside `openbb_sdk/sdk` are being collected, which causes tests on `openbb_sdk/provider` to not run. Do we need to adjust something on the actions?

jmaslek:
> @jmaslek it looks like only tests inside `openbb_sdk/sdk` are being collected, which causes tests on `openbb_sdk/provider` to not run. Do we need to adjust something on the actions?

We can actually just change the noxfile. I can push an edit shortly

the-praxs:
@henriqueio tests are failing on the CI because of `import providers` -
```python
ModuleNotFoundError: No module named 'providers'
```
Runs fine locally but not sure about CI.

hjoaquim:
You da best, thank you, @jmaslek

the-praxs (approved):
lgtm. Merge it!!

PR #5437, opened 2023-09-13
Feature/sdk v4 pydantic v2 POC

tehcoderer:
POC working with pydantic v2


openbb_sdk/openbb/package/crypto.py
```diff
@@ -22,54 +21,43 @@ def __repr__(self) -> str:
     @validate_arguments
     def load(
         self,
-        symbol: typing_extensions.Annotated[
-            Union[str, List[str]],
-            OpenBBCustomParameter(description="Symbol to get data for."),
```
montezdesousa, line 27: @hjoaquim do we need to keep `OpenBBCustomParameter` ?

IgorWounds: @montezdesousa We use it so that argparse can work with the Platform on the Terminal.

openbb_sdk/sdk/core/openbb_core/app/router.py
```diff
@@ -479,8 +514,7 @@ def from_extensions() -> Router:
                     router=entry_point.load(), prefix=f"/{entry_point.name}"
                 )
             except Exception as e:
-                raise LoadingError(
-                    f"Invalid extension '{entry_point.name}': {e}"
-                ) from e
+                traceback.print_exception(type(e), e, e.__traceback__)
```
montezdesousa, line 517: @tehcoderer nice, can we have it on `RegistryLoader` as well?

openbb_sdk/sdk/provider/openbb_provider/registry_map.py
```diff
@@ -87,8 +90,19 @@ def _get_map(self, registry: Registry) -> Tuple[MapType, MapType]:
                     "QueryParams": extra_query,
                     "Data": extra_data,
                 }
+
+                in_return_map = return_map.get(model_name, return_type)
+                if union_return_map.get(model_name, None) is None and (
+                    get_origin(return_type) != list
+                    and get_origin(in_return_map) == list
```
montezdesousa, line 97: @tehcoderer  here can't we just check if `get_origin(return_type) != get_origin(in_return_map)` so it works for any type besides `list`?

openbb_sdk/sdk/provider/openbb_provider/abstract/data.py
```diff
@@ -1,23 +1,51 @@
...
+    )
 
-    class Config:
-        """Pydantic configuration."""
+    @model_validator(mode="before")
+    @classmethod
+    def _use_alias(cls, values):
+        """Use alias for error locs."""
+        # set the alias dict values keys
+        alises = {orig: alias for alias, orig in cls.__alias_dict__.items()}
```
montezdesousa, line 47:
```suggestion
        aliases = {orig: alias for alias, orig in cls.__alias_dict__.items()}
```

openbb_sdk/sdk/provider/openbb_provider/abstract/data.py
```diff
@@ -1,23 +1,51 @@
...
+        if alises:
```
montezdesousa, line 48:
```suggestion
        if aliases:
```

openbb_sdk/sdk/provider/openbb_provider/abstract/data.py
```diff
@@ -1,23 +1,51 @@
...
+            return {alises.get(k, k): v for k, v in values.items()}
```
montezdesousa, line 49:
```suggestion
            return {aliases.get(k, k): v for k, v in values.items()}
```

openbb_sdk/openbb/package/crypto.py
montezdesousa: I'm aware, is there any good alternative @IgorWounds @hjoaquim? Otherwise we need to keep this @tehcoderer

tehcoderer: yeah I took it out init cause it didn't work with basemodel, just pushed an update that uses `dataclass` instead and works

openbb_sdk/providers/benzinga/openbb_benzinga/models/stock_news.py
```diff
@@ -18,8 +18,11 @@ class BenzingaStockNewsQueryParams(StockNewsQueryParams):
     Source: https://docs.benzinga.io/benzinga/newsfeed-v2.html
     """
 
-    class Config:
-        fields = {"symbols": "tickers", "limit": "pageSize"}
+    pageSize: int = Field(
```
montezdesousa, line 21: Here think we can just use the alias dict and map "limit" to "pageSize", then in `extract_data` we `model_dump(by_alias=True)` ?

openbb_sdk/providers/polygon/openbb_polygon/models/stock_news.py
```diff
@@ -93,7 +91,7 @@ def extract_data(
         api_key = credentials.get("polygon_api_key") if credentials else ""
 
         base_url = "https://api.polygon.io/v2/reference/news"
-        querystring = get_querystring(query.dict(by_alias=True), [])
+        querystring = get_querystring(query.model_dump(), [])
```
montezdesousa, line 94: I was checking these cases where `query.model_dump()`, don't we need to keep the `by_alias=True` to send the query fields with the alias instead of the field name?

openbb_sdk/openbb/package/stocks.py
```diff
@@ -514,28 +511,28 @@ def multiples(
...
             ),
-        ] = 0,
-        limit: typing_extensions.Annotated[
-            Union[pydantic.types.NonNegativeInt, None],
+        ],
+        limit: Annotated[
+            int,
             OpenBBCustomParameter(description="Number of results to return per page."),
-        ] = 15,
+        ],
```
montezdesousa, line 529: Defaults missing on these arguments

openbb_sdk/sdk/provider/openbb_provider/abstract/query_params.py
```diff
@@ -1,18 +1,18 @@
...
+    __alias_dict__: Dict[str, str] = {}
 
     def __repr__(self):
         """Return the string representation of the QueryParams object."""
-        return f"{self.__class__.__name__}({', '.join([f'{k}={v}' for k, v in self.dict().items()])})"
+        return f"{self.__class__.__name__}({', '.join([f'{k}={v}' for k, v in self.model_dump().items()])})"
+
+    model_config = ConfigDict(
+        extra="allow", populate_by_name=True, alias_generator=alias_generators.to_snake
+    )
```
montezdesousa, line 18:
```suggestion
    )
    
        def model_dump(self, *args, **kwargs):
        original = super().model_dump(*args, **kwargs)
        by_alias = kwargs.get("by_alias", False)
        if by_alias:
            return {
                self.__alias_dict__.get(key, key): value
                for key, value in original.items()
            }
        return original
```

If we add this one here we can `model_dump(by_alias=True)` when exporting the query to send to the provider's API, wdyt?

openbb_sdk/providers/polygon/openbb_polygon/models/stock_news.py
```diff
@@ -58,12 +57,11 @@ class PolygonPublisher(BaseModel):
...
-            "text": "description",
-            "date": "published_utc",
-        }
+    __alias_dict__ = {
+        "url": "article_url",
+        "text": "description",
+        "date": "published_utc",
+    }
 
     amp_url: Optional[str] = Field(description="AMP URL.")
```
montezdesousa, line 66:
```suggestion
    amp_url: Optional[str] = Field(default=None, description="AMP URL.")
```

Apparently v2 requires us to explicitly set the default as `None` when type is `Optional[T]`
https://docs.pydantic.dev/2.0/migration/#required-optional-and-nullable-fields

openbb_sdk/sdk/provider/openbb_provider/abstract/query_params.py
tehcoderer: nice this is a good implementation 

```python
    model_config = ConfigDict(
        extra="allow", populate_by_name=True, alias_generator=alias_generators.to_snake
    )

    def model_dump(self, *args, **kwargs):
        by_alias = kwargs.pop("by_alias", False)
        original = super().model_dump(*args, **kwargs)
        if by_alias:
            return {
                self.__alias_dict__.get(key, key): value
                for key, value in original.items()
            }
        return original
```

just tested and allows `news?provider=benzinga&display_output=full&channels=All&page_size=50&symbols=PLTR` to work using snakecase

openbb_sdk/providers/polygon/openbb_polygon/models/stock_news.py
the-praxs: Yes, apparently `Optional` without a default value is a required field. Setting the default as `None` indicates it's not required.

openbb_sdk/sdk/provider/openbb_provider/abstract/query_params.py
the-praxs: I think we can set different alias_generator based on the providers e.g. FMP uses lower camel case and Intrinio uses snake case for their returned fields.

openbb_sdk/providers/benzinga/openbb_benzinga/models/stock_news.py
the-praxs: That would be the preferred method for the standardization to work.

PR #5441, opened 2023-09-15
Make my attempt at returning defined function when called in nb

jmaslek:
This is my take on how to solve the issue addressed in #5424.  My proposal (actually was @montezdesousa idea) is to make magic functions in ipython for these and to group them in an extension.

Example usage:
Screenshot 2023-09-15 at 4 31 58 PM

Cons are that you have to import the magics and that its just in a notebook so scripts will still need something else more clever, and you need it on each cell, I couldnt figure out how to globally do it.


DidierRLopes:
I like it Picasso 🖼️

deeleeramone:
Is there an elegant way to insert a user preference into `command_runner`?  

Screenshot 2023-09-15 at 2 48 49 PM

montezdesousa:
> Is there an elegant way to insert a user preference into `command_runner`?
> 
> Screenshot 2023-09-15 at 2 48 49 PM

Think this comment fits better https://github.com/OpenBB-finance/OpenBBTerminal/pull/5424 discussion.

Anyway, you have the preferences there (see line 369), but I don't think it's a good idea doing it inside `CommandRunner`. We w'd be messing the backend to fix a pure frontend issue. If we go this route I'd say doing it in the `Container` class is better. More context here https://github.com/OpenBB-finance/OpenBBTerminal/pull/5424#discussion_r1320885509

hjoaquim:
Can we close this in detriment of https://github.com/OpenBB-finance/OpenBBTerminal/pull/5471#issuecomment-1735338784?

piiq:
resolved by #5471

PR #5444, opened 2023-09-17
hotfix/fix-candle - Fix `max().max()`

deeleeramone:
This PR fixes an error with `/stocks/candle`which appears to only be present in the PyPI distribution.

```
Error: 'float' object has no attribute 'max'
```


deeleeramone:
Resolved in #5446

PR #5450, opened 2023-09-19
Print message if yahoo finance load fails

### LINKED ISSUE #5392: [Bug] `load --source YahooFinance` fails silently while every other `source` prints a warning or the request's `status_code`
To recreate, simply load any bad ticker with YahooFinance as the source.

Screenshot 2023-08-30 at 2 42 22 PM

Light-JH:
Fixes #5392

Currently, in the /stocks/load command, if yahoo finance does not find data it doesn't print any indication to the user.

This PR adds a message if yahoo finance does not find data：

Screenshot 2023-09-19 at 2 43 11 PM

### DIFF
--- a/openbb_terminal/stocks/stocks_helper.py
+++ b/openbb_terminal/stocks/stocks_helper.py
@@ -411,6 +411,7 @@ def load(
 
             # Check that loading a stock was not successful
             if df_stock_candidate.empty:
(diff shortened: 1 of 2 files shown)


piiq (approved):
Thank you very much for your contribution @Light-JH 🚀

PR #5453, opened 2023-09-20
Add more input types to data processing commands

hjoaquim:
^ 

In this PR we can find a proposal for allowed supported types for the `data` argument on data processing commands.
Supported types:

- Data
- List[Data]
- DataFrame
- List[DataFrame]
- Series
- List[Series]
- np.ndarray
- List
- Dict


hjoaquim:
While we allow this types for the `data` argument, it is a very different thing to then process that data.
Atm, we were only using `DataFrame` and `List[Data]`, which were the only supported types, the following way:

image

Now we'll need something like `data_type_to_df()` instead of `basemodel_to_df()` that should be capable of transform any of the above-mentioned types into a `DataFrame`.

What's your take on this?
@montezdesousa @the-praxs @piiq @IgorWounds @jmaslek

the-praxs:
> ^
> 
> In this PR we can find a proposal for allowed supported types for the `data` argument on data processing commands. Supported types:
> 
> * Data
> * List[Data]
> * DataFrame
> * List[DataFrame]
> * Series
> * List[Series]
> * np.ndarray
> * List
> * Dict

`Data` is a generic term here if there is no context given. By `Data` I am assuming the primitive data types (`int`, `float`, `str`, `bool`) and non-primitive data types (`List`, `Dict`, `Set`, `Tuple`). I am unsure about `Files` as I haven't seen a use case yet but we might after having a database connector implemented.

hjoaquim:
@the-praxs by `data` I mean the data argument (the data to be processed) in a data processing command (as shown in the print screen and also by looking at the changed files).
I don't follow you comment as it doesn't really address the question I've stated on the first comment :grimacing:

hjoaquim:
> > @the-praxs by `data` I mean the data argument (the data to be processed) in a data processing command (as shown in the print screen and also by looking at the changed files). I don't follow you comment as it doesn't really address the question I've stated on the first comment 😬
> 
> Gotcha. It essentially means what I said. I feel these current data types are sufficient.

Thing is, atm we only accept either `DataFrame` or `List[Data]`

the-praxs:
> > > @the-praxs by `data` I mean the data argument (the data to be processed) in a data processing command (as shown in the print screen and also by looking at the changed files). I don't follow you comment as it doesn't really address the question I've stated on the first comment 😬
> 
> > 
> 
> > Gotcha. It essentially means what I said. I feel these current data types are sufficient.
> 
> 
> 
> Thing is, atm we only accept either `DataFrame` or `List[Data]`

I meant the current proposed ones.

hjoaquim:
> > @the-praxs by `data` I mean the data argument (the data to be processed) in a data processing command (as shown in the print screen and also by looking at the changed files). I don't follow you comment as it doesn't really address the question I've stated on the first comment 😬
> 
> Gotcha. It essentially means what I said. I feel these current data types are sufficient.

Oh yes, I also think those should be enough. Any thoughts on `filter_inputs()`?

the-praxs:
> > > @the-praxs by `data` I mean the data argument (the data to be processed) in a data processing command (as shown in the print screen and also by looking at the changed files). I don't follow you comment as it doesn't really address the question I've stated on the first comment 😬
> > 
> > 
> > Gotcha. It essentially means what I said. I feel these current data types are sufficient.
> 
> Oh yes, I also think those should be enough. Any thoughts on `filter_inputs()`?

I think its appropriate but @IgorWounds will be the right person to provide feedback on it.

deeleeramone:
> Oh yes, I also think those should be enough. Any thoughts on `filter_inputs()`?

So I think we need to establish and communicate what `List[Data]` specifically means in the context of use.  Does it represent multiple sets of data, for gang-processing, or in the context of the Fetcher outputs and representing a single data set?

An example is, the `symbol` parameter.  When a user sees an input as accepting, `List[str]`, they will assume that a list of tickers can be passed to the request - which is only true for certain providers and functions.

the-praxs:
@hjoaquim tests are failing

PR #5457, opened 2023-09-20
Stock Historical standardization and UX improvements

the-praxs:
**Changelog** -

* Modified `validator` to `field_validator` with updated arguments in the `StockHistorical` standard model
* Standardized params in AlphaVantage, FMP, Intrinio, Polygon, and Yahoo!Finance
* Modified AlphaVantage API params to `PrivateAttr` to optimize model schema and readability
* Added a helpers file in AlphaVantage to make provider URL requests and get data from dynamic keys
* Modified return type to JSON in AlphaVantag that removed usage of pandas
* Modified the description for CBOE `interval` param to use the in-built helper dict
* Modified the Intrinio API params to `PrivateAttr` to optimize model schema and readability
* Added EOD data URL for Intrinio with additional standard interval params
* Standardized `interval`/`frequency` data field in Intrinio
* Modified `multiplier` and `timespan` Polygon API params to `PrivateAttr` to optimize model schema and readability
* Modified Yahoo!Finance `period`, `progress`, `keepna`, `repair`, `rounding`, and `group_by` args to `PrivateAttr` to optimize model schema and readability

**Checklist** -
- [x] Unit Tests Pass Python Interface
- [x] Unit Tests Pass API Interface
- [x] Integration Tests Python Interface
- [x] Integration Tests API Interface
- [ ] Duplicate Tests


deeleeramone:
It appears that start_date is no longer accepting a string.

```
ValidationError: 1 validation error for load
standard_params.start_date
  Input should be a valid datetime, invalid datetime separator, expected `T`, `t`, `_` or space [type=datetime_parsing, input_value='2023-01-01', input_type=str]
    For further information visit https://errors.pydantic.dev/2.4/v/datetime_parsing
```

deeleeramone:
Polygon is returning empty. 

Screenshot 2023-10-05 at 2 55 04 PM

even with no interval:

Screenshot 2023-10-05 at 2 56 28 PM

deeleeramone:
Intrinio is stuck in a perpetual await state. This has not timed out. and it's been over 1 hour.

Screenshot 2023-10-05 at 4 02 04 PM

the-praxs:
Currently Intrinio requires `start_date`/`end_date` and `start_time`/`end_time` separately for params. I tried to standardize these into one and make the time params private for better UX but Pydantic throws a validation error `date_from_datetime_inexact` because of strict type check for `date` (standard model) and `datetime` (intrinio model).

There should be a way I believe to get this functionality rather than having to modify the standard model params to another type.

deeleeramone:
`extended_hours` and `prepost` do the same thing, but for different providers.  I would vote for `prepost` because it is fewer characters.

deeleeramone:
```
adjusted : Optional[bool]
    Output time series is adjusted by historical split and dividend events.Only available for intraday data. (provider: alpha_vantage, polygon); Adjust all OHL
```

This parameter is not constrained to intraday.

deeleeramone:
With Polygon, starting at `1900-01-01` breaks it for some reason.  `2000-01-01` works despite their data going back only to 2003.

```
obb.stocks.load("AAPL", provider="polygon", start_date="1900-01-01").to_df()

OpenBBError: Results not found.
```

deeleeramone:
Intrinio intraday is still stuck in a perpetual await:

```
In [27]: obb.stocks.load("AAPL", provider="intrinio", interval="1m").to_df()
```

deeleeramone:
AlphaVantage intraday does not work.

deeleeramone:
> Currently Intrinio requires `start_date`/`end_date` and `start_time`/`end_time` separately for params.

So does Polygon. They should be separate params from start_date/end_date.

the-praxs:
> `extended_hours` and `prepost` do the same thing, but for different providers. I would vote for `prepost` because it is fewer characters.

extended_hours is more readable though. Although we have the upper hand to creating a better term so please let me know.

the-praxs:
> With Polygon, starting at `1900-01-01` breaks it for some reason. `2000-01-01` works despite their data going back only to 2003.
> 
> ```
> obb.stocks.load("AAPL", provider="polygon", start_date="1900-01-01").to_df()
> 
> OpenBBError: Results not found.
> ```

This is the response from Polygon
```json
{
  "status": "ERROR",
  "request_id": "ec2a8403a5f58d2510281f3fb65f19bf",
  "error": "The parameter's 'from' and/or 'to' cannot be before Unix epoch time."
}
```

the-praxs:
> Intrinio intraday is still stuck in a perpetual await:
> 
> ```
> In [27]: obb.stocks.load("AAPL", provider="intrinio", interval="1m").to_df()
> ```

I checked the URLs and the data fetched. All of it is working as expected. The reason its taking time is the interval value is too small. 1-minute data by default over a period of 1 year is a lot!

the-praxs:
> AlphaVantage intraday does not work.

What's the exact issue?

deeleeramone:
> > Intrinio intraday is still stuck in a perpetual await:
> > ```
> > In [27]: obb.stocks.load("AAPL", provider="intrinio", interval="1m").to_df()
> > ```
> 
> I checked the URLs and the data fetched. All of it is working as expected. The reason its taking time is the interval value is too small. 1-minute data by default over a period of 1 year is a lot!

How many minutes does it take to execute? I give up after five minutes, but even let it go over an hour. 

Loading about 200,000 bars from Polygon takes only a couple of seconds. Without extended hours there are only 113,400 minutes in a trading year.

the-praxs:
> > > Intrinio intraday is still stuck in a perpetual await:
> > > ```
> > > In [27]: obb.stocks.load("AAPL", provider="intrinio", interval="1m").to_df()
> > > ```
> > 
> > 
> > I checked the URLs and the data fetched. All of it is working as expected. The reason its taking time is the interval value is too small. 1-minute data by default over a period of 1 year is a lot!
> 
> How many minutes does it take to execute? I give up after five minutes, but even let it go over an hour.
> 
> Loading about 200,000 bars from Polygon takes only a couple of seconds. Without extended hours there are only 98,280 minutes in a trading year (calculated as (6.5 * 60) * 252).

I get the results in ~5 min on average from multiple runs.

image

PR #5459, opened 2023-09-20
Hotfix/ext map test

montezdesousa:
* Tests the extension map against pyproject.toml to ensure:
1. non required extensions are not built
2. required extensions are built

* Removes package folder before rebuilding to avoid leaving unused files there


IgorWounds (approved):

the-praxs (approved):
LGTM 🚀

PR #5460, opened 2023-09-20
Feature/etf-v4

deeleeramone:
This PR creates an ETF extension with asset class-specific commands, and four new community providers.

To install, navigate into the `openbb_sdk` folder and enter:

```
python dev_install.py -e
```

- Adding new providers
  - TMX
  - Blackrock (US and Canada supported)
  - Invesco (US only)
  - BMO (Canada only)

- Extending existing providers
  -  FMP

- Adding features supported by most providers in the ETF universe.
  -  Holdings
      - Historical daily holdings available from Blackrock.  Some funds only have EOM/EOQ files, but ~15 years history.
      - TMX provides current top 10 holdings.
      - Invesco does not support historical daily holdings, but provides the complete current holdings for each fund.
   - Sectors
     - Sector allocation breakdown with multiple tickers supported.
     - Not yet supported for Invesco.
   - Countries
     - Country allocation breakdown with multiple tickers supported.
     - Not yet supported for Invesco.
   - Info
     - General info for an ETF or compare multiple tickers.
     - Not yet supported for Invesco.
  - Search
    - Fuzzy text query.
  - Historical NAV
    - Currently supported by BMO and Invesco.

- Extending coverage on existing commands:
  - stocks.fa.divs(provider="tmx")
    - `historical_dividends` standard model updated to allow for other providers than FMP.
    - `date` -> `ex_date` to reflect the proper label of the `ex-div` trading date.
    - tests updated for FMP and added for TMX.
  - stocks.fa.filings(provider="tmx")
  - stocks.fa.ins(provider="tmx")
  - stocks.fa.pt(provider="tmx")
  - stocks.news(provider="tmx")
  - stocks.search(provider="tmx")
    - updated standard model for `symbol` to be Optional
  - stocks.earnings_calendar(provider=["fmp", "tmx"]
    - Market earnings calendars with start/end dates, or single date.
    - FMP was mapping to historical EPS, so this has been updated to be the earnings calendar end point.
  - stocks.fa.historical_eps(provider="fmp")
    - This is essentially a rename of `stocks.fa.earning`, with a new standard model that is a copy of the old one.
  - `economy.available_indices(provider="tmx")`
  - `economy.index_constituents()`
    - Remaps what was `economy.const` to group with index-functions and provides more clarity in the name.
    - Converts FMP to use `index_constituents` as the Standard Model.
 
- Other new commands in the Router:
  - `economy.index.info()`
    - General information on an index - description, document links, general performance and stats.


openbb_sdk/extensions/etf/openbb_etf/etf_definitions.py
```diff
@@ -0,0 +1,300 @@
...
+
+from typing import Dict, List, Optional
+
+import pandas as pd
+from openbb_core.app.model.obbject import OBBject
+from openbb_core.app.utils import basemodel_to_df
+from pydantic import Field
+
+
+class EtfBase(OBBject):
```
montezdesousa, line 11: Why create child classes and not return `OBBject` directly?

deeleeramone: Thanks for the feedback.

The initial reason was because I needed a way to return data outside of the "results" field. The solution I came up with for "etf_holdings" was to return a Tuple from the Provider (holdings data and additional information such as the recorded date, number of outstanding shares, and value), which satisfies the output requirements there, and re-assemble it in the next phase.  

It could just pass that directly to the command output, but this format would not present usable data to the user. Some things are just not appropriate for storing in "results".  Additional info is just a simple dictionary that cannot be included within a table belonging under "results". This additional info is also not suitable for the "metadata" field because that field is specifically for the command-execution context.

The output of the Provider forces it to be a list of Dicts, and as-is, makes `to_df()` not usable at the final stage.  The output from the Provider needs to be validated through OBBject (before it hits the command output in the Router), and those additional fields I wish to include are rejected because they are not defined in the OBBject definitions.  Additional fields are accepted only within the context of "results".

I want to return additional properties that are immediately keyable for future downstream processes:

- Model (I want the ability to query a field that knows it is an EtfHoldings model)
- Fields (I want the ability to accept/reject an object in POST based on the present fields, which vary by provider)
- Info (I want this key information to be captured which otherwise would be lost.  Having it keyable is a functionality bonus for future purposes - i.e, to_chart() for the titles.)

Customizing `to_df` so that it always makes a great DataFrame - using function-specific logic - is also win-win.


I also would like to have docstrings attached to the returned object which are meaningful, and this needs to be done at the point of class definition.  i.e, if I validate an object through OBBject only, I am forced to accept all of the docstrings that come along with it.


Looking forward, view methods - `to_chart(**params)` - can be made specifically for the data that exists in the object - pie and bar charts - and include any parameters that can make for better visualizations.

montezdesousa: Following the 3 point list:

* Model: can you provide an example of ideal usage and result?
* Fields: I think extracting fields from a response object is quite straightforward, for example this would work without additional attributes in the response.

```
out = obb.etf.holdings('SPY')
out.to_df().columns
```

* Info: iiuc this update https://github.com/OpenBB-finance/OpenBBTerminal/pull/5471 might help by introducing the `extra` field which should allow you to send additional information there without modifying the `OBBject` structure.

deeleeramone: if df.model == "StockHistorical"
    n_trading_days = 252
if df.model == "CryptoHistorical"
    n_trading_days = 365

deeleeramone: Regarding info, this also needs to come out of the provider fetcher independent of the serialized data model.  Currently possible to do, but not within the current testing framework.

piiq:
Hi, some of the functionality of this PR has been cherry-picked and is already available in the v4 branch.
It would be better to move tmx and other providers to an extension in a separate repo. Feel free to close this PR once you backed up the code of this PR

PR #5461, opened 2023-09-20
Hotfix/fix-quandl-regression

deeleeramone:
`cot_search()` was edited to return only the first item in the list or results instead of the entire list.  This fixes that regression.

### DIFF
--- a/openbb_sdk/providers/quandl/openbb_quandl/models/cot_search.py
+++ b/openbb_sdk/providers/quandl/openbb_quandl/models/cot_search.py
@@ -22,7 +22,7 @@ class QuandlCotSearchData(CotSearchData):
     """Quandl CFTC Commitment of Traders Reports Search data."""
 
 
-class QuandlCotSearchFetcher(Fetcher[CotSearchQueryParams, QuandlCotSearchData]):
+class QuandlCotSearchFetcher(Fetcher[CotSearchQueryParams, List[QuandlCotSearchData]]):
     """Quandl CFTC Commitment of Traders Reports Search Fetcher."""
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


IgorWounds:
Hi,

Can you please check that the type hinting withing the Fetcher implementation is correct? This test fails now:

`FAILED openbb_sdk/providers/quandl/tests/test_quandl_fetchers.py::test_quandl_cot_search_fetcher - AttributeError: type object 'QuandlCotSearchData' has no attribute '__args__'`

deeleeramone:
> Hi,
> 
> Can you please check that the type hinting within the Fetcher implementation is correct? This test fails now:
> 
> `FAILED openbb_sdk/providers/quandl/tests/test_quandl_fetchers.py::test_quandl_cot_search_fetcher - AttributeError: type object 'QuandlCotSearchData' has no attribute '__args__'`

Why would a data object have an arguments attribute?

Screenshot 2023-09-20 at 3 07 35 PM

Screenshot 2023-09-20 at 3 04 53 PM

deeleeramone:
@IgorWounds Are we sure that these tests are even doing anything? Quandl uses a Python client and does not have "headers".  This function doesn't even call an HTTP endpoint, it references a curated list in the `utils` folder, "series_ids.py". 



Screenshot 2023-09-20 at 9 44 47 PM

IgorWounds:
Hi @deeleeramone,

Yes, the tests are doing a lot of things and checking that the fetcher is properly implemented. Please navigate to the `test` method of the abstract `Fetcher` class to see how it checks each step of the pipeline. In order to not expose our API key, we filter the headers that are made through the request to Quandl. You can easily inspect this inside the record folder. The configuration to filter out our API keys is exposed to the whole test file, we are not cherry picking functions that might or might not do an HTTP request as they all live in the same file.

Screenshot 2023-09-21 at 11 17 44

the-praxs:
@IgorWounds @deeleeramone test was failing because we need to return a `List[QuandlCotSearchData` in the Fetcher as the `transform_data` returns the same data structure. Fixed now.

We will merge this. Thanks for the PR @deeleeramone and @IgorWounds for helping in the tests 🚀

PR #5462, opened 2023-09-21
Include extensions in package __repr__

montezdesousa:
Screenshot 2023-09-21 at 13 19 10

Screenshot 2023-09-21 at 13 24 56


the-praxs:
We could add a little space between the version and the name

PR #5464, opened 2023-09-21
Standardize `date` and `datetime` for all historical models

the-praxs:
This PR allows `date` and `datetime` of all formats for **only** the `StockHistorical`, `MajorIndicesHistorical`, `CryptoHistorical`, and `ForexHistorical` models. 

This is achieved using `dateutil.parser` that removes the need for validators that check whether its a `date` or `datetime` of specific format(s). The only requirement is that the object should be of class `datetime`, `dateutil` or `pd.Timestamp`.


deeleeramone:
This is a tough question to answer, ngl, and might be related to other PRs; what time is it?

Screenshot 2023-09-21 at 8 21 44 PM

the-praxs:
That is a good question, which will be pondered upon in another PR. This PR is only to fix the regressions and issues of date/datetime conversion simultaneously. Merging this PR will allow us to release openbb v4.0.0a2 package.

deeleeramone (approved):
This is working as advertised.

PR #5465, opened 2023-09-21
Add chat with OBBject

jmaslek:
Title.

@montezdesousa @IgorWounds @hjoaquim @piiq  is this an acceptible way to build extensions?  I would be limited that I could not add more than one of these :/

anyways.  need help to pass through obb.user....


montezdesousa:
This might be an option. Only returns user_settings in python and needs package to use `Container.run` method instead of going directly to `CommandRunner`

Screenshot 2023-09-22 at 10 36 38

Screenshot 2023-09-22 at 10 33 16

Then...
Screenshot 2023-09-22 at 10 35 35

PR #5470, opened 2023-09-22
Add Versioned Documentation

piiq:
This PR is used instead of https://github.com/OpenBB-finance/OpenBBTerminal/pull/5376 

~~This PR is branched of develop so that we're able to merge the versioned docs before the release of v4~~

This branch points to the v4 branch to facilitate a merge to develop sooner

additionally this PR touches thousands


deeleeramone:
I know it's a WIP, but just to point it out:

Install instructions for V4 will be different than V3, so they should have their own dedicated page.  Having old content as a placeholder may lead to confusion.

PR #5471, opened 2023-09-22
OBBject updates for python package

montezdesousa:
Allows access to credentials inside OBBject methods https://github.com/OpenBB-finance/OpenBBTerminal/pull/5465

This will also allow to set user_preference `obbject_method` to return `to_df()` by default from `Container.run` as proposed by @hjoaquim


openbb_sdk/sdk/core/openbb_core/app/static/container.py
```diff
@@ -1,9 +1,22 @@
...
         self._command_runner = command_runner
+
+    def run(self, *args, **kwargs) -> Union[OBBject, pd.DataFrame, dict]:
+        """Run a command in the container."""
+        obbject = self._command_runner.run(*args, **kwargs)
+        user_settings = self._command_runner.user_settings
+        python_output = user_settings.preferences.python_output
+        if python_output:
+            return getattr(obbject, python_output)()
+        obbject.extra["user_settings"] = user_settings.dict()
```
hjoaquim, line 21: Isn't it a bit "dangerous" to put the raw credentials there?
Also, need to test to ensure that this doesn't end up on logs!

hjoaquim: Wouldn't this be a more "cautious" approach to use the `to_chat()` method on the OBBject:

```python

# Never mind if syntax is not exactly correct

openai_key = obb.credentials.openai()
my_data = obb.stocks.load(...)

my_data.to_chat(query="explain me this stuff", key=openai_key)
```

As I type this, I also feel it's kinda meh because it would pretty much be the only functionality that makes you deal with api keys.

jmaslek:
So how would this work with an extension that requires a key.  I imagine we wont be able to use the hub, so will it have to be

```
from openbb import obb
import extension

obb.user_stuff.new_key = ...
```

?

openbb_sdk/sdk/core/openbb_core/app/static/container.py
montezdesousa: Yes need to think a bit more about implications of allowing this before merging.

montezdesousa:
> So how would this work with an extension that requires a key. I imagine we wont be able to use the hub, so will it have to be
> 
> ```
> from openbb import obb
> import extension
> 
> obb.user_stuff.new_key = ...
> ```
> 
> ?

At first I did that, but not sure if having that alternative way of registering a provider is good. It's certainly an option and probably the fastest, we'd just need to `Extra.allow` in the credentials model. The usual way is to define a `Provider` iiuc, need to explore that as well.

hjoaquim (approved):

PR #5478, opened 2023-09-26
Feature/integration marker

hjoaquim:
Adding `@pytest.mark.integration` to integration tests, so we can leverage `pytest` to easily run:
 - unit + integration : `pytest`
 - unit : `pytest -m "not integration"`
 - integration : `pytest -m integration`

Note that, for the CI, we don't rely on this since we specify the directories directly.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,7 +2,7 @@ import nox
 
 test_locations = [
     "openbb_sdk/tests",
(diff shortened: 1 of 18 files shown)


the-praxs (approved):
@hjoaquim man you are god

IgorWounds (approved):

PR #5481, opened 2023-09-27
Replace install_all files with Juan's dev_install script

piiq:
# Description

This replaces the platform specific `install_all` scripts to juan's platform agnostic python dev_install script.
This script will install the dependencies in editable mode for development purposes

# How has this been tested?

I've set up my env with this script


piiq:
I'm late

PR #5482, opened 2023-09-27
Remove old scripts and update the readme

piiq:
this 👆


the-praxs (changes requested):
@piiq can you fix the incorrect numbering (commented) and add the line `python dev_install.py -e` wherever relevant?

IgorWounds (approved):

PR #5484, opened 2023-09-27
Idea for extending obbject

jmaslek:
openbb_obbject1 is a dummy extension

Pros - can have multiple extensions with custom obbject method

Cons - specific naming convention.

Just an idea.  Do with it what you will

Screenshot 2023-09-26 at 8 47 34 PM


jmaslek:
Closing in favor of #5612

PR #5491, opened 2023-09-28
Hotfix/crypto hist yf

montezdesousa:
Fixes `/crypto/load` is broken for yf with symbol `BTC`
Adds checks for empty data coming from yf


the-praxs (approved):
lgtm 🚀

PR #5504, opened 2023-10-02
Hotfix/int tests python fixture

hjoaquim:
Python integration tests need to import the `obb` app behind a fixture, otherwise even when running tests with the "not integration" marker, the `CommandRunner` will be instantiated, causing the unit tests to be slower and to generate unwanted logs.

Also, when running tests locally, because the app was being instantiated (and as consequence some other internal components) this was causing unexpected behaviors on classes that are singletons (namely the `LoggingService`) resulting in unit tests to fail.


IgorWounds (approved):
This is very good!

PR #5508, opened 2023-10-02
Hotfix/econometrics export filename fix

ssahaxd:
Added --export option for exporting to a desired file in /econometrics menu


deeleeramone (changes requested):
Hello!

Unfortunately, this doesn't work for me, still getting the same error as before.

```
(🦋) /econometrics/ $ export loaded_data -t cleaned_file.xlsx

usage: export [-n NAME] [-t {xlsx,csv}] [-h] [--export EXPORT] [--sheet-name SHEET_NAME [SHEET_NAME ...]]
export: error: argument -t/--type: invalid choice: 'cleaned_file.xlsx' (choose from 'xlsx', 'csv')
```

openbb_terminal/econometrics/econometrics_controller.py
```diff
@@ -471,15 +471,15 @@ def call_export(self, other_args: List[str]):
         if other_args and "-" not in other_args[0][0]:
             other_args.insert(0, "-n")
         ns_parser = self.parse_known_args_and_warn(
-            parser, other_args, export_allowed=NO_EXPORT
+            parser, other_args, export_allowed=EXPORT_BOTH_RAW_DATA_AND_FIGURES
```
deeleeramone, line 474: The parser is defined, in the lines above, with constrained `choices`.  Without including your changes, I can comment out line 466 and I can set the filename accordingly.

ssahaxd: I have added the `--export` option, which I observed being used in the following format:
`/stocks/ $ load aapl --start 2022-01-01 --end 2022-12-31 --weekly --export aapl.xlsx --sheet-name weekly`

Now, the export command in the `econometrics` section will resemble the following:
`/econometrics/ $ export aapl --export test.xlsx`

ssahaxd: Found a better solution.
Added a -f/--file option to provide the filename for export.
Made -t/--type and -f/--file mutually exclusive.
If -f is not given, -t is required and the filename will be generated based on the time and module.

openbb_terminal/econometrics/econometrics_controller.py
```diff
@@ -458,7 +458,15 @@ def call_export(self, other_args: List[str]):
             type=str,
         )
 
-        parser.add_argument(
+        export_group = parser.add_mutually_exclusive_group(required=True)
+        export_group.add_argument(
+            "-f",
```
deeleeramone, line 463: The `--help`, `-h` flags do not work now.

The way that this works across the rest of the Terminal is: `command --export filename.xlsx`, or `command --export xlsx`.  The main difference here is that the dataset must be selected as a choice, and the command name is called `export`.  There should, however, only be one argument which covers both situations, with a way to write to a specific `--sheet-name`.

```
(🦋) /econometrics/ $ export --help

usage: export [-n NAME] (-f FILE | -t {xlsx,csv}) [-h]
export: error: one of the arguments -f/--file -t/--type is required


(🦋) /econometrics/ $ export -h

usage: export [-n NAME] (-f FILE | -t {xlsx,csv}) [-h]
export: error: one of the arguments -f/--file -t/--type is required
```

ssahaxd: Fixed.

openbb_terminal/econometrics/econometrics_controller.py
```diff
@@ -479,7 +487,7 @@ def call_export(self, other_args: List[str]):
                 console.print("Please enter a valid dataset.")
             else:
                 export_data(
-                    ns_parser.type,
+                    ns_parser.file if ns_parser.file else ns_parser.type,
                     os.path.dirname(os.path.abspath(__file__)),
                     ns_parser.name,
                     self.datasets[ns_parser.name],
```
deeleeramone, line 493: Add this here to export to a specified sheet name.
```
sheet_name=" ".join(ns_parser.sheet_name)
```

openbb_terminal/econometrics/econometrics_controller.py
```diff
@@ -479,7 +487,7 @@ def call_export(self, other_args: List[str]):
                 console.print("Please enter a valid dataset.")
```
deeleeramone, line 487: Line 482 changes to: 

```
export_allowed=EXPORT_ONLY_RAW_DATA_ALLOWED
```

deeleeramone (changes requested):
couple of small changes and we're off to the races!

openbb_terminal/econometrics/econometrics_controller.py
deeleeramone: Sorry, my bad, we need to include a caveat for if `sheet-name` is None:

```
sheet_name=" ".join(ns_parser.sheet_name) if ns_parser.sheet_name else None
```

deeleeramone (approved):
Thanks for the fix!

PR #5510, opened 2023-10-03
Add EXTENSIONS.md

piiq:
This PR moves the lists of available extensions and providers to an EXTENSIONS.md file in the platform root.
Outside contributors would be asked to add their extensions to the list in that file

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -48,7 +48,7 @@ jobs:
           pip install types-pytz types-requests types-termcolor types-tabulate types-PyYAML types-python-dateutil types-setuptools types-six
       - run: bandit -x ./tests -r . || true
       - run: black --diff --check .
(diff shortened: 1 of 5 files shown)


the-praxs (commented):
I am unsure if the providers should be in a different `PROVIDERS.md` file or should be in this file. 

The extensions and providers are fundamentally different, as the extension is responsible for an action that produces some output. The provider is for bringing in the data to the extension. This means that both should have their own files.

On the other hand, if we are assuming that the extensions comprise both command and data providers then in this file, we need to create two tables, one for extensions and the data where each has 2 sub-sections: official and community.

andrewkenreich:
I think we should have a separate providers and extensions .md file as well - I just foresee that file getting extremely large eventually with both tables there - better to do it now then later

jmaslek:
I think we should definitely differentiate:

- Providers
- Extensions

I think that having substructure of these (community/openbb/etc.) does make sense, but for the time being, I think author attribution is enough.

piiq:
Thanks for the feedback everyone, I've incorporated it and updated the PR. Now we have

`./openbb_platform/EXTENSIONS.md` for the extension list and
`./openbb_platform/PROVIDERS.md` for the providers list

PR #5535, opened 2023-10-09
Fixes FMP index endpoint to get 1 day interval data

the-praxs:
The downside is that the start and end dates don't work when the interval is 1 day.


deeleeramone:
> The downside is that the start and end dates don't work when the interval is 1 day.

This is an easy fix, filter it right before the output of `extract_data`.  This is how other providers work where there is no start/end parameter on the data request.

deeleeramone:
Constrained values are problematic here, and we will need to allow 0 or None.

```
obb.economy.index("^105833-USD-STRD")

OpenBBError: 1 validation error for FMPMajorIndicesHistoricalData
open
  Input should be greater than 0 [type=greater_than, input_value=0, input_type=int]
    For further information visit https://errors.pydantic.dev/2.4/v/greater_than

```

openbb_platform/platform/provider/openbb_provider/standard_models/major_indices_historical.py
```diff
@@ -8,7 +8,7 @@
 from typing import List, Optional, Set, Union
 
 from dateutil import parser
-from pydantic import Field, PositiveFloat, validator
+from pydantic import Field, PositiveFloat, NonNegativeFloat, field_validator
```
deeleeramone, line 11: NonNegativeFloat does not actually fix the underlying issue.  Negative values are OK, and in some cases are the expected values, they should be allowed.

openbb_platform/platform/provider/openbb_provider/standard_models/major_indices_historical.py
```diff
@@ -38,15 +38,15 @@ class MajorIndicesHistoricalData(Data):
     """Major Indices end of day price data."""
 
     date: datetime = Field(description=DATA_DESCRIPTIONS.get("date", ""))
-    open: PositiveFloat = Field(description=DATA_DESCRIPTIONS.get("open", ""))
+    open: NonNegativeFloat = Field(description=DATA_DESCRIPTIONS.get("open", ""))
```
deeleeramone, line 41: The same principle applies to "high", "low", "close" values.

deeleeramone (changes requested):
Why just "open"? Why not make them pure floats? That would be a more robust validation solution.

openbb_platform/platform/provider/openbb_provider/standard_models/major_indices_historical.py
the-praxs: Done.

openbb_platform/platform/provider/openbb_provider/standard_models/major_indices_historical.py
the-praxs: Done.

PR #5540, opened 2023-10-10
Make Intrinio options chains great again

the-praxs:
^


deeleeramone:
CBOE is now broken and the standard model is expecting mandatory fields that shouldn't be there.

- `symbol` does not need to exist at all.  It shouldn't be in the table of data, it is the same underlying ticker symbol repeated on every row.
- `date` should not be called date.  It should be `eod_date` because otherwise `to_df()` indexes that column, and it is the more descriptive label. Moreover, this field should definitely not be mandatory.
- `type` was `option_type`, because `type` is a special word used in the standard Python library and it confuses things when a variable is also the name of an operative - similar to `date` and `dateType`.

```
OpenBBError: 3 validation errors for CboeOptionsChainsData
symbol
  Field required [type=missing, input_value={'expiration': '2023-10-1...lose': 128.82, 'dte': 3}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/missing
type
  Field required [type=missing, input_value={'expiration': '2023-10-1...lose': 128.82, 'dte': 3}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/missing
date
  Field required [type=missing, input_value={'expiration': '2023-10-1...lose': 128.82, 'dte': 3}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/missing
```

deeleeramone:
On a bright note, it's nice to see that Intrinio has removed `$` from being required in the query string for options on indices.  No more helper with handling the 12 or so symbols covered!

the-praxs:
> CBOE is now broken and the standard model is expecting mandatory fields that shouldn't be there.
> 
> * `symbol` does not need to exist at all.  It shouldn't be in the table of data, it is the same underlying ticker symbol repeated on every row.
> * `date` should not be called date.  It should be `eod_date` because otherwise `to_df()` indexes that column, and it is the more descriptive label. Moreover, this field should definitely not be mandatory.
> * `type` was `option_type`, because `type` is a special word used in the standard Python library and it confuses things when a variable is also the name of an operative - similar to `date` and `dateType` but on a higher order.
> 
> ```
> OpenBBError: 3 validation errors for CboeOptionsChainsData
> symbol
>   Field required [type=missing, input_value={'expiration': '2023-10-1...lose': 128.82, 'dte': 3}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.4/v/missing
> type
>   Field required [type=missing, input_value={'expiration': '2023-10-1...lose': 128.82, 'dte': 3}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.4/v/missing
> date
>   Field required [type=missing, input_value={'expiration': '2023-10-1...lose': 128.82, 'dte': 3}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.4/v/missing
> ```

* `symbol` is present in the `option` key of Intrinio. According to current standardization principles we have to include it in the data. Due to CBOE, we can modify the `symbol` field to be `Optional[str]`.
* Agreed. Standard model now has `eod_date` instead of `date`.
* Good catch! Now we have `option_type` in the standard model.

the-praxs:
> On a bright note, it's nice to see that Intrinio has removed `$` from being required in the query string for options on indices. No more helper with handling the 12 or so symbols covered!

Removed! 🚀

jmaslek (approved):

Issue #5539, opened 2023-10-10
[Bug] Revenue History and Estimations table has incorrectly formatted dates


adrianmck1:
Hi,

I am looking at revenue history and estimations data for various tickers in /stocks/fa/revfc and when the table is displayed, the FISCALYEAR column only shows the date "1970-01-01 00:00:02". The rest of the data in the table is fine. 

When I export the CSV file, it shows the correct year, so I'm guessing it's just some sort of formatting bug on fiscal year. 

I have attached an image showing the issue and a second image showing the data from the CSV file.

**To Reproduce**
stocks/fa
load <TICKER>
revfc

**Screenshots**
![image](https://github.com/OpenBB-finance/OpenBBTerminal/assets/122870179/6b47914d-0db5-4ab8-ab79-a0321f775e77)

![image](https://github.com/OpenBB-finance/OpenBBTerminal/assets/122870179/694bccb9-c5b7-4cc1-a5c6-c16b2dec49a5)

**Desktop (please complete the following information):**

- OS: Windows 10
- Python version: N/A
- OpenBB executable v3.2.2

**Additional context**
As well as the executable, I also have the SDK installed and it's fine, but it's not using the table viewer.

![image](https://github.com/OpenBB-finance/OpenBBTerminal/assets/122870179/1888cb0d-bc97-4dff-a59e-5f1fac82615f)

Thanks.

deeleeramone:
Nice find!

sohinidas1092:
Hi @deeleeramone , can you please assign this to me?

PR #5545, opened 2023-10-11
commiting changes as defined in #5539

dhruvmalik007:
# Description

## Summary 

Resolving the issue concerning the #5539 where the fiscal year tends to define by the starting of the 'datetime' value (based on suggestion by @deeleeramone ).  along with that in the test files for the seek_alpha , I saw that the the test function was incorrectly defined.

# How has this been tested?

I havent tested given that there was no access of the mock data of Seeking Alpha.

# Checklist:

- [ ] I have adhered to the GitFlow naming convention and my branch name is in the format of `feature/feature-name` or `hotfix/hotfix-name`.
- [ ] Update [our documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).  Update any user guides that are affected by the changes.
- [ ] If a feature was added make sure to add it to the corresponding [integration test script](https://github.com/OpenBB-finance/OpenBBTerminal/tree/develop/openbb_terminal/miscellaneous/integration_tests_scripts).

# Others


openbb_terminal/stocks/fundamental_analysis/seeking_alpha_view.py
```diff
@@ -68,6 +68,7 @@ def display_rev_estimates(
             show_index=False,
             title=f"{symbol.upper()} Revenue History and Estimations",
             export=bool(export),
+            columns_keep_types=["fiscalyear"],
```
deeleeramone, line 71: Thanks for jumping in!

A slight oversight on my part, we also need to convert the column to a string. One more line to add above.

Line 66:
```
rev_estimates["fiscalyear"] = rev_estimates["fiscalyear"].astype(str)
```

While we are here, let's also apply the same solution to the other function, `display_eps_estimates`.

deeleeramone (changes requested):
Slipped my mind that there is a second function right above this which also needs treatment. :)

openbb_terminal/stocks/fundamental_analysis/seeking_alpha_view.py
deeleeramone: I think this is actually because you are not running this from the environment that is not where OpenBB is installed.  With the environment active, you would only need to run `pytest {path} {--params}`.

It seems to test fine over here, and oddly enough, the cassettes already captured reflect EPS and Rev as you changed them to be. :)

The linters might yell at us for not having an empty line at the end of the test file, but let's find out!

Thanks!

deeleeramone (approved):

deeleeramone:
Git Flow is putting up a fight, it wants your branch to be named, `/hotfix/issue-#5539`.

openbb_terminal/stocks/fundamental_analysis/seeking_alpha_view.py
dhruvmalik007: ok, I thought that if I build from source and then setup the env using `poetry env use python3.10` then things should've worked as mentioned. I will recheck on my side.


You're welcome and thanks again for the description, Indeed exicted to start my contribution streak to OpenBB and in general to contribute to the openBB ecosystem.

PR #5549, opened 2023-10-12
Standard convention for multi-ticker symbols 

the-praxs:
Standardizes the symbol input for the Crypto and Forex Historical standard models.

Accepted formats - `CURR1CURR2` ; `CURR1-CURR2`


IgorWounds (changes requested):
"There should be one-- and preferably only one --obvious way to do it." - The Zen of Python

IgorWounds (approved):

the-praxs:
> "There should be one-- and preferably only one --obvious way to do it." - The Zen of Python

"Explicit is better than implicit." - The Zen of Python

PR #5557, opened 2023-10-13
Feature/obbject to dict

DidierRLopes:
@jmaslek this should fix the issue with the `to_dict()` which seemed that it was hitting a regression?

Before
image

After
image

### DIFF
--- a/openbb_platform/platform/core/openbb_core/app/model/obbject.py
+++ b/openbb_platform/platform/core/openbb_core/app/model/obbject.py
@@ -1,6 +1,6 @@
 """The OBBject."""
 from re import sub
-from typing import Any, Dict, Generic, List, Literal, Optional, TypeVar
+from typing import Any, Dict, Generic, List, Literal, Optional, TypeVar, Union
 
 import pandas as pd
 from numpy import ndarray
@@ -183,37 +183,19 @@ class OBBject(Tagged, Generic[T]):
         """Convert results field to numpy array."""
         return self.to_dataframe().reset_index().to_numpy()
 
-    def to_dict(self) -> Dict[str, List]:
-        """Convert results field to list of values.
+    def to_dict(self) -> Union[Dict[str, Any], List[Dict[str, Any]]]:
+        """Convert results field Data models to dict.
 
         Returns
         -------
-        Dict[str, List]
-            Dictionary of lists.
+        Union[Dict[str, Any], List[Dict[str, Any]]]
+            Dict or list of dicts.
         """
-        if isinstance(self.results, dict) and all(
-            isinstance(v, dict) for v in self.results.values()
-        ):
-            results: Dict[str, List] = {}
-            for _, v in self.results.items():
-                for kk, vv in v.items():
-                    if kk not in results:
-                        results[kk] = []
-                    results[kk].append(vv)
-
-            return results
-
-        df = self.to_dataframe().reset_index()  # type: ignore
-        results = {}
-        for field in df.columns:
-            f = df[field].tolist()
-            results[field] = f[0] if len(f) == 1 else f
-
-        # remove index from results
-        if "index" in results:
-            del results["index"]
-
-        return results
+        return (
+            self.results.model_dump()
+            if not isinstance(self.results, list)
+            else [d.model_dump() for d in self.results]
+        )
 
     def to_chart(self, **kwargs):
         """
--- a/openbb_platform/platform/core/openbb_core/app/model/obbject.py
+++ b/openbb_platform/platform/core/openbb_core/app/model/obbject.py
@@ -191,11 +191,18 @@ class OBBject(Tagged, Generic[T]):
         Union[Dict[str, Any], List[Dict[str, Any]]]
             Dict or list of dicts.
         """
-        return (
-            self.results.model_dump()
-            if not isinstance(self.results, list)
-            else [d.model_dump() for d in self.results]
-        )
+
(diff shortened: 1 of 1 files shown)


openbb_platform/platform/core/openbb_core/app/model/obbject.py
```diff
@@ -183,37 +183,26 @@ def to_numpy(self) -> ndarray:
...
+            if hasattr(item, "model_dump"):
+                return item.model_dump()
+            elif isinstance(item, list):
+                return [nested_model_to_dict(i) for i in item]
+            elif isinstance(item, dict):
+                return {k: nested_model_to_dict(v) for k, v in item.items()}
+
+            return item
+
+        return nested_model_to_dict(self.results)
```
montezdesousa, line 205: Does it cover your use case if we just implement in one line?

```
def to_dict(self) -> Dict
    ...
    return self.model_dump()["results"]
```

deeleeramone:
> @jmaslek this should fix the issue with the `to_dict()` which seemed that it was hitting a regression?

I think this was a choice because as a dict like this, it is actually more of a list and the "fields" are not keyable, you need to convert from a List, whereas like a Dict of Lists, you can key and entire field, like `df.to_dict()["close"]`

DidierRLopes:
> > @jmaslek this should fix the issue with the `to_dict()` which seemed that it was hitting a regression?
> 
> I think this was a choice because as a dict like this, it is actually more of a list and the "fields" are not keyable, you need to convert from a List, whereas like a Dict of Lists, you can key and entire field, like `df.to_dict()["close"]`
> 
> A dict("records") gives you essentially the same thing as `df.results`. Can we have `to_dict()` parameterized so that the user define as a preference?

We've had this discussion before in the past, and this was the final decision. List of dicts it's what is used in practice by everyone. Dicts of list is unnatural, so it's fine for us to support it but as `to_dict` we should have the default one.

And we don't want to leave this as user criteria for the same method since we want to have as much deterministic as possible .

deeleeramone:
> We've had this discussion before in the past, and this was the final decision. List of dicts it's what is used in practice by everyone. Dicts of list is unnatural, so it's fine for us to support it but as `to_dict` we should have the default one.
> 
> And we don't want to leave this as user criteria for the same method since we want to have as much deterministic as possible .

The only reason I suggest it was a choice somewhere, instead of a regression, is that it has been added to the documentation under `/core/README.md`:

Screenshot 2023-10-13 at 6 53 25 PM

DidierRLopes:
It was several months ago, but then we changed that decision. Maybe it never got updated.

CC: @jmaslek @piiq

jmaslek:
This is coming back to me.  Darren is correct that this was the reasoning and is indeed not a regression. 

The fiasco with List[Dict] vs Dict[List], was with what was being served via api.   If you look at obbject.results, it is List[Models], which gets converted into a nice json format.

Screenshot 2023-10-14 at 7 18 32 AM


> Dicts of list is unnatural, so it's fine for us to support it but as to_dict we should have the default one.

This is actually misleading.  The default to_dict() method (via pandas) is actually returning Dict[Dict], so neither case is default.  When I see a to_dict, I assume I am getting back a dictionary of values, which is what we have, rather than a nice typical records format, which is what the api does. 

We can add as many different methods to return the data as we want, lets just make sure we document them and clearly explain what they are/how they work.

openbb_platform/platform/core/openbb_core/app/model/obbject.py
tehcoderer: 🤦🏻‍♂️Lmao, this what happens when doing a @DidierRLopes task at 1am.

Yeah this for sure the best way to go!!

DidierRLopes:
Gotcha. Sorry @deeleeramone.

@jmaslek , I'm not getting that output with `obb.stocks.load("AAPL").results`:

Screenshot 2023-10-14 at 9 34 08 AM

Can we add an argument `orient` to `to_dict`, so a user can also do `.to_dict(orient="results")`?

Screenshot 2023-10-14 at 9 36 17 AM

jmaslek:
> Gotcha. Sorry @deeleeramone.
> 
> 
> 
> @jmaslek , I'm not getting that output with `obb.stocks.load("AAPL").results`:
> 
> 
> 
> Screenshot 2023-10-14 at 9 34 08 AM
> 
> 
> 
> Can we add an argument `orient` to `to_dict`, so a user can also do `.to_dict(orient="results")`?
> 
> 
> 
> Screenshot 2023-10-14 at 9 36 17 AM
> 
> 
> 
> 

Yeah the results output you see is python/pydantic stuff, but on the fastapi end it's converted to what I showed.

Yup we can either add an orient or just skip straight to a to_records/to_json method. Dealers choice.

DidierRLopes:
> > Gotcha. Sorry @deeleeramone.
> > @jmaslek , I'm not getting that output with `obb.stocks.load("AAPL").results`:
> > Screenshot 2023-10-14 at 9 34 08 AM
> > Can we add an argument `orient` to `to_dict`, so a user can also do `.to_dict(orient="results")`?
> > Screenshot 2023-10-14 at 9 36 17 AM
> 
> Yeah the results output you see is python/pydantic stuff, but on the fastapi end it's converted to what I showed.
> 
> Yup we can either add an orient or just skip straight to a to_records/to_json method. Dealers choice.

So I suggest doing both:
* to_dict having the orient='results' as an option
* to_json returning a JSON with list of dicts

@andrewkenreich @jose-donato any comment on this?

PR #5559, opened 2023-10-13
Allow single item inputs for list of literal parameters

montezdesousa:
Screenshot 2023-10-13 at 10 29 18


montezdesousa:
No solution seems to be good for this problem as of now, will close and iterate later.

PR #5563, opened 2023-10-13
Fix most tests and bugs

IgorWounds:
Fix most tests and bugs for stocks and futures.


hjoaquim (changes requested):
Futures:

image

Stocks:

image





News:

image


Maybe I require all api keys for this to work?

PR #5568, opened 2023-10-16
Publish versions `a3` to Pypi

hjoaquim:
- Update versions
- Create script that automate publishing


the-praxs:
@tehcoderer updated the versions of the files in the latest commit so we only gotta create the publishing script.

IgorWounds (approved):

PR #5572, opened 2023-10-17
organizing build folder

hjoaquim:
This PR proposes a re-organization of the `build` folder on the root of the repo.

Adds two subfolders: 
- terminal
- platform

Changed the documentation and the workflows that mention the path to such files.


hjoaquim:
@piiq can I close this PR as it doesn't really seem necessary for now? Also, there were some adjustments done on  #5579 that might do it for now.

PR #5576, opened 2023-10-18
Automated examples

IgorWounds:
Automated examples.


openbb_platform/openbb/package/economy.py
```diff
@@ -1570,7 +1677,13 @@ def sp500_multiples(
         date : str
             The date data for the time series.
         value : float
-            The data value for the time series."""  # noqa: E501
+            The data value for the time series.
+        Example
+        --------
+
+        >>> from openbb import obb
+        >>> obb.economy.sp500_multiples(series_name=PE Ratio by Month, collapse=monthly)
```
hjoaquim, line 1685: This should be inside quotes. Maybe we can generalize that if it's a Literal and its options are strings, we need to put quotes in the example.

hjoaquim (changes requested):
Other than the small comment, everything looks good to me.

hjoaquim (approved):

PR #5578, opened 2023-10-18
use SecretStr for sensitive fields

montezdesousa:
* replaces `str` for `SecretStr` to hide keys on notebook __repr__, same for access_token
* filters out provider keys to remove unused providers

Screenshot 2023-10-18 at 17 17 54

To get the key `obb.user.credentials.fred_api_key.get_secret_value()`
To  set the key just passing string works, pydantic coerces  `obb.user.credentials.fred_api_key = 123`


the-praxs (approved):
SHIP IT

PR #5579, opened 2023-10-18
Feature/v4 pypi

hjoaquim:
^


openbb_platform/dev_utils/PUBLISH.md
piiq, line 1: Do you think the whole content of this folder would be better off in the build folder in the root of the repo? We have a pypi related folder there and this content and code directly relates to that

hjoaquim: Yes! I agree

.github/workflows/pypi-nightly.yml
```diff
@@ -12,37 +12,52 @@ jobs:
...
+          --sdist
+          --wheel
+          --outdir dist/
+          .
+
+      - name: OpenBBTerminal - Publish distribution 📦 to Test PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: dist/
+          password: ${{ secrets.TEST_PYPI_API_TOKEN_OBB_TERMINAL_NIGHTLY }}
```
hjoaquim, line 47: @piiq we need to create an `openbb-terminal` project on PyPI and then create a project scoped API token and add it to the GitHub secrets.

hjoaquim: cc @luqmanbello

hjoaquim:
@luqmanbello just tagging you to ask if you see any problems with the changes on the actions

PR #5580, opened 2023-10-18
Hotfix/fix-v4-economic-calendar: Add FMP to Econ Calendar and fix some TE bugs

deeleeramone:
This PR adds FMP to the `economy.econcal()` function.  Adding another provider to the endpoint required renovating the standard model so that it is not specifically for TE. 

This PR also fixes the validation errors that existed in the TE model, for example:

```
obb.economy.econcal(provider="tradingeconomics")

ValidationError: 2 validation errors for econcal
standard_params.importance
  Input should be 'Low', 'Medium' or 'High' [type=literal_error, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.4/v/literal_error
standard_params.group
  Input should be 'interest rate', 'inflation', 'bonds', 'consumer', 'gdp', 'government', 'housing', 'labour', 'markets', 'money', 'prices', 'trade' or 'business' [type=literal_error, input_value=None, input_type=NoneType]
  ```

Now it works without entering any parameters, and multiple countries are accepted from a Fast API input. Country is valid using: "New Zealand", "New_Zealand", "new_Zealand", "new zealand".  Multiple countries are accepted in the Python interface as both a List and string-list, or singular.

Screenshot 2023-10-18 at 4 18 40 PM

What this does not fix is the URL generator in TE, it seems that no combination of parameters suffices:

```
OpenBBError: No url generated. Check combination of input parameters.
```


deeleeramone:
@jmaslek, an example of an error generating a URL where there is known data to return:

Screenshot 2023-10-18 at 4 41 49 PM

deeleeramone:
Any thoughts to calling this endpoint, `calendar`, instead of, `econcal`?  `obb.economy.calendar`

jmaslek (approved):
As long as the tests pass, this lgtm.  Except that fmp returns countries as 2 letters, but c'est la vie

jmaslek:
> Any thoughts to calling this endpoint, `calendar`, instead of, `econcal`? `obb.economy.calendar`

@deeleeramone Im behind the time.  Im fine with that.

PR #5588, opened 2023-10-21
Add didiers to_records method

jmaslek:
Allows you to do 

- obbject.to_dict(orient = " ") just has to be valid arg
- obbject.to_records() -> gives the json type stuff

Remove polars so that can be an extension

### DIFF
--- a/openbb_platform/platform/core/openbb_core/app/model/obbject.py
+++ b/openbb_platform/platform/core/openbb_core/app/model/obbject.py
@@ -1,6 +1,6 @@
 """The OBBject."""
 from re import sub
-from typing import Any, Dict, Generic, List, Literal, Optional, TypeVar
+from typing import Any, Dict, Generic, List, Literal, Optional, TYPE_CHECKING, TypeVar
 
 import pandas as pd
 from numpy import ndarray
@@ -13,6 +13,12 @@ from openbb_core.app.model.abstract.warning import Warning_
 from openbb_core.app.model.charts.chart import Chart
 from openbb_core.app.provider_interface import ProviderInterface
 from openbb_core.app.utils import basemodel_to_df
+
+if TYPE_CHECKING:
+    try:
+        from polars import DataFrame as PolarsDataFrame
+    except ImportError:
+        PolarsDataFrame = None
 
 T = TypeVar("T")
 PROVIDERS = Literal[tuple(ProviderInterface().available_providers)]  # type: ignore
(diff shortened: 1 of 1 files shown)


DidierRLopes:
image

DidierRLopes (approved):

IgorWounds (changes requested):
This looks good!

To not introduce redundancy. Shall we just have the to_dict() method with a literal of these:

['dict', 'list', 'series', 'split', 'tight', 'records', 'index']

This way, we can set the records to be the default orientation and people can easily change to other orientations based on their needs. This will make the to_records() method not needed.

jmaslek:
> This looks good!
> 
> To not introduce redundancy, shall we just have the to_dict() method with a literal of these:
> 
> ['dict', 'list', 'series', 'split', 'tight', 'records', 'index']
> 
> This way, we can set the records to be the default orientation and people can easily change to other orientations based on their needs. This will make the to_records() method not needed.

Was just following this: https://github.com/OpenBB-finance/OpenBBTerminal/pull/5557#issuecomment-1763055940

I like the to_dict returning an actual dictionary, that we can index.  I can add the literals, but would rather just have the to_records option.

IgorWounds:
> Was just following this: [#5557 (comment)](https://github.com/OpenBB-finance/OpenBBTerminal/pull/5557#issuecomment-1763055940)
> 
> I like the to_dict returning an actual dictionary, that we can index. I can add the literals, but would rather just have the to_records option.

Got it. I think that you can just change the default to be the "list" option which is essentially what the pure to_dict() is doing right now. If you really want to split off the to_records, to_X, ...; we could do it with shortcuts the same way we did with to_df(). 

Personally, I wouldn't say I like having too many helper functions that can easily be united and many Python users use Pandas which will make it easy to understand that they can change the to_dict() output in the same way they would do with Pandas.

jjfantini:
I have been loving the `to_polars()` method. I had a custom method `to_polars()` with v3. I think if anything def make it an extension...users are slowly converting to Polars, I think its best to prepare for that future 😄

jmaslek:
> > Was just following this: [#5557 (comment)](https://github.com/OpenBB-finance/OpenBBTerminal/pull/5557#issuecomment-1763055940)
> > I like the to_dict returning an actual dictionary, that we can index. I can add the literals, but would rather just have the to_records option.
> 
> Got it. I think that you can just change the default to be the "list" option which is essentially what the pure to_dict() is doing right now. If you really want to split off the to_records, to_X, ...; we could do it with shortcuts the same way we did with to_df().
> 
> Personally, I wouldn't say I like having too many helper functions that can easily be united and many Python users use Pandas which will make it easy to understand that they can change the to_dict() output in the same way they would do with Pandas.

Adjuisted this to reflect this.  @DidierRLopes you good?

I also brought back polars, since why not

IgorWounds (approved):

DidierRLopes (approved):
🚢it

jjfantini:
v happy you decided to keep `polars` :) @jmaslek

Issue #5582, opened 2023-10-19
[Bug] `/stocks/fa/growth`: A value is trying to be set on a copy of a slice from a DataFrame


deeleeramone:
Recreate with Python 3.10 and the command:

```
/stocks/load aapl/growth
```

```
/OpenBBTerminal/openbb_terminal/stocks/fundamental_analysis/fmp_model.py:663: SettingWithCopyWarning:

A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
```

PR #5591, opened 2023-10-22
fix: A value is trying to be set on a copy of a slice from a DataFrame

abhishek-amar:
- Fixes warning caused by setting a value on copy of a DataFrame by explicitly creating a copy.
- Link #5582


deeleeramone:
Thanks for the patch! This works, however, to merge your branch we need to rebase it to the `/develop` branch.  You should be able to change it with the "Edit" button.

abhishek-amar:
> Thanks for the patch! This works, however, to merge your branch we need to rebase it to the `/develop` branch. You should be able to change it with the "Edit" button.

Sorry for the confusion, I have changed the PR to merge with the `develop` branch and rebased it.

deeleeramone (approved):
Thanks for the PR!

PR #5593, opened 2023-10-23
user preferences descriptions

hjoaquim:
^


deeleeramone:
I like the table!  There are a couple of open PRs relating to the output preferences. We should probably update this to reflect them, assuming we know the outcome already.

hjoaquim:
@deeleeramone do you think this provides enough context on how, when, why use each preference? Or should we try to expand more?
Regarding the output preference, I think it's fine, I see a couple of PRs that will eventually expand it but not really change it. I don't think we should list the options on the docs (with the risk it might get easily obsolete).
Wdyt?

deeleeramone:
> @deeleeramone do you think this provides enough context on how, when, why use each preference? Or should we try to expand more? Regarding the output preference, I think it's fine, I see a couple of PRs that will eventually expand it but not really change it. I don't think we should list the options on the docs (with the risk it might get easily obsolete). Wdyt?

If we don't list the options in the Docs, where is the complete list going to be?  The nature of docs is for them to become obsolete as soon as they are finished. 😢


So, thinking out loud here, the CSS files for table and plots provide a way for users to customize the way things look.  Do we need to reference these file paths and formats?

deeleeramone (approved):
LGTM and we can iterate further over on the `/feature/improve-sdk-v4-docs-struct` branch.

PR #5610, opened 2023-10-26
Add ETF historical price from yfinance

piiq:
Integration tests are coming in a separate subsequent PR for all etf data points


the-praxs (approved):
lgtm

PR #5617, opened 2023-10-27
Feature/polygon-stock-quote: Add NBBO Quotes, with historical, to `obb.stocks.quote()`

deeleeramone:
This PR adds Polygon as a provider to the `obb.stocks.nbbo()` function.  This has a default limit of 25 results, but can return an unstable amount of data if desired.  I have put an internal cap at 10M entries.  The limit argument becomes important when requesting historical data, especially for the most trafficked tickers.

A typical day of SPY NBBO quotes will be around 5 million rows and nearly 1GB as a raw, uncompressed, CSV file.  It is not recommended to go about it this way, but it is possible.  Best practice would be to break up the request into smaller chunks - like an hour at a time - using the greater/less than parameters to bracket the period.

Just the default setting with only a ticker entered returns the most recent 25 NBBO Quotes:

**Note** examples here have a not-current name. The function and model have changed to `nbbo`, from `quote`.

```
obb.stocks.quote("TSLA", provider="polygon").to_df()
```

Screenshot 2023-10-27 at 3 44 09 PM

I looked into converting the condition and indicator codes but the reference tables in the Polygon documentation appear to be missing some entries and/or are not accurate.

For any query that includes a time, it must be formatted like this:
```
2023-10-27T16:01:00.000000000-04:00
```

Example of a bracketed request:

```python
obb.stocks.quote("SPY", provider="polygon", timestamp_gt="2023-10-26T15:00:00.000000000-04:00",timestamp_lt="2023-10-26T15:30:00.000000000-04:00").to_df()
```

Yields 317,491 records between 3:00 and 3:30 yesterday:

Screenshot 2023-10-27 at 3 38 48 PM

Lower volume stocks are way less heavy, the entire day for CLOV yesterday produces only 11,046 rows.

```
obb.stocks.quote("CLOV", provider="polygon", timestamp="2023-10-26").to_df()
```

Screenshot 2023-10-27 at 3 41 53 PM


jmaslek (dismissed):
I would like to see this as a different endpoint than just quote.  My biggest reasoning is just the pure amount of data you can get.  Trying to just get a clean look, and oops i am now dealing with 6 GB

openbb_platform/providers/polygon/openbb_polygon/models/stock_nbbo.py
```diff
@@ -146,13 +122,15 @@ class PolygonStockNBBOData(StockNBBOData):
...
     def date_validate(cls, v):  # pylint: disable=E0213
         """Return formatted datetime."""
-        return (
-            to_datetime(v, unit="ns", origin="unix", utc=True).tz_convert("US/Eastern")
-            if v
-            else None
-        )
+
+        if v:
+            dlt = timedelta(hours=-5)
```
deeleeramone, line 130: @hjoaquim, can we do this with the TZ name, "America/New_York", instead of numbers?  If we rely on N hours, then this will be incorrect for half of the year, when daylight savings knocks off an hour.

openbb_platform/providers/polygon/openbb_polygon/models/stock_nbbo.py
```diff
@@ -29,36 +31,36 @@ class PolygonStockNBBOQueryParams(StockNBBOQueryParams):
             + " Splitting large requests into chunks is recommended for full-day requests of high-volume symbols."
         ),
     )
-    timestamp: Optional[Union[datetime, str]] = Field(
+    timestamp: Optional[datetime] = Field(
```
deeleeramone, line 34: Unfortunately, these changes break the function.  The example syntax of a bracketed function, in the PR, does not work with these changes.

```
obb.stocks.quote("SPY", provider="polygon", timestamp_gt="2023-10-26T15:00:00.000000000-04:00",timestamp_lt="2023-10-26T15:30:00.000000000-04:00")
```

```
OpenBBError: Error in Polygon request -> failed to parse timestamp query params: invalid time string: parsing time "2023-10-26 15:00:00-04:00": extra text: " 15:00:00-04:00"
```

openbb_platform/providers/polygon/openbb_polygon/models/stock_nbbo.py
hjoaquim: I agree. Was trying to get rid of the pandas import, but maybe that's the way!

the-praxs:
@deeleeramone @hjoaquim please don't merge until @piiq is done with the great reanme. Thanks!

the-praxs:
@deeleeramone I have made the following changes in the code to make it leaner and improve readability -

* Removed using multiple tickers since we only fetched data for a single ticker. I suppose this command should allow the use of a single ticker as multithreading could cause overload on a user's system given the request size.
* Removed the use of the `max_` variable in favor of `limit`. The `limit` parameter is validated in the `QueryParams` class, and in the `extract_data` is set in the query string if it's less than `50000` otherwise checked against the number of results being obtained.
* Added the alias for `date` as `timestamp`.
* Removed the `EXCHANGE_ID_MAP` in `helpers.py` and added the same functionality in the `extract_data` function. I did this to improve the readability of the code. I think that using a cache would be beneficial instead of making the request for the `exchanges` data for this.

the-praxs (approved):

PR #5618, opened 2023-10-28
Feature/v4-price-performance: % returns for various horizons, comparable across multiple tickers.

deeleeramone:
This PR adds a `price_performance` endpoint to the Stocks and ETF extensions.  The same model is mapped to both places for the sake of convenience, the data returned is not specific to either and the response does not change.

In the future, and as other providers find their way into it, a CAGR parameter would be a nice variation.

Screenshot 2023-10-27 at 10 38 48 PM


jmaslek (approved):
I would expect this to be an extension along the qa fronts, but since its a direct endpoint, I have no issue with it.

PR #5628, opened 2023-10-31
economy.calendar nasdaq

montezdesousa:
Screenshot 2023-10-31 at 18 54 00


openbb_platform/providers/quandl/openbb_quandl/models/economic_calendar.py
```diff
@@ -0,0 +1,124 @@
...
+        while start <= end:
+            dates.append(start.strftime("%Y-%m-%d"))
+            start = start + timedelta(days=1)
+
+        data = []
+        for date in dates:
+            try:
+                response = (
+                    make_request(
+                        url=f"https://api.nasdaq.com/api/calendar/economicevents?date={date}",
```
jmaslek, line 91: this is actually nasdaq and not quandl/nasdaq datalink 😓

the-praxs: that means another provider

the-praxs (approved):

the-praxs:
@montezdesousa I modified the model to be more pythonic, by using the pydantic way of handling things and enforcing TET pipeline.

PR #5630, opened 2023-10-31
Add stocks disc endpoints

IgorWounds:
Adds active, gainers, and losers endpoints to stocks/disc.


openbb_platform/extensions/stocks/integration/test_stocks_api.py
```diff
@@ -1000,3 +1000,61 @@ def test_stocks_disc_active(params, headers):
     result = requests.get(url, headers=headers, timeout=10)
     assert isinstance(result, requests.Response)
     assert result.status_code == 200
+
+
+@pytest.mark.parametrize(
+    "params",
+    [
+        ({"symbol": "AAPL", "provider": "sec"}),
```
piiq, line 1008: will this not fail?

piiq (approved):

openbb_platform/extensions/stocks/integration/test_stocks_api.py
IgorWounds: If we bump the timeout to 30 it will not

Screenshot 2023-10-31 at 14 50 41

PR #5634, opened 2023-10-31
Check for empty data globally

montezdesousa:
Since we are doing this check in a lot of places, wdyt having it here? Or even at results level?

### DIFF
--- a/openbb_platform/platform/provider/openbb_provider/abstract/fetcher.py
+++ b/openbb_platform/platform/provider/openbb_provider/abstract/fetcher.py
@@ -6,6 +6,7 @@ from typing import Any, Dict, Generic, Optional, TypeVar, get_args, get_origin
 
 from openbb_provider.abstract.data import Data
 from openbb_provider.abstract.query_params import QueryParams
+from openbb_provider.utils.errors import EmptyDataError
 
 Q = TypeVar("Q", bound=QueryParams)
 D = TypeVar("D", bound=Data)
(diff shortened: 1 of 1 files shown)


the-praxs:
In the `get_data` function that is provided in the each of the provider utils and is almost the same for all 
```python
def get_data(url: str, **kwargs: Any) -> Union[list, dict]:
    """Get data from Intrinio endpoint."""

    try:
        r: Union[requests.Response, BasicResponse] = helpers.make_request(url, **kwargs)
    except SSLError:
        r = request(url)
    if r.status_code == 404:
        raise RuntimeError("Intrinio endpoint doesn't exist")

    data = r.json()
    if r.status_code != 200:
        error = data.get("error")
        message = data.get("message")
        value = error or message
        raise RuntimeError(f"Error in Intrinio request -> {value}")

    if "Error Message" in data:
        raise RuntimeError("Intrinio Error Message -> " + data["Error Message"])

    if len(data) == 0:
        raise EmptyDataError()

    return data
```

It makes sense to have them here rather than scoop it out and put it globally. But if we overhaul the way we make requests and check all the others along with this one, I think it would be more robust.

the-praxs (commented):
In the `get_data` function that is provided in the each of the provider utils and is almost the same for all 
```python
def get_data(url: str, **kwargs: Any) -> Union[list, dict]:
    """Get data from Intrinio endpoint."""

    try:
        r: Union[requests.Response, BasicResponse] = helpers.make_request(url, **kwargs)
    except SSLError:
        r = request(url)
    if r.status_code == 404:
        raise RuntimeError("Intrinio endpoint doesn't exist")

    data = r.json()
    if r.status_code != 200:
        error = data.get("error")
        message = data.get("message")
        value = error or message
        raise RuntimeError(f"Error in Intrinio request -> {value}")

    if "Error Message" in data:
        raise RuntimeError("Intrinio Error Message -> " + data["Error Message"])

    if len(data) == 0:
        raise EmptyDataError()

    return data
```

It makes sense to have them here rather than scoop it out and put it globally. But if we overhaul the way we make requests and check all the others along with this one, I think it would be more robust.

PR #5640, opened 2023-10-31
Disc router for equities.

IgorWounds:
Adds the disc router to stocks.


the-praxs (approved):
@IgorWounds u da man 🚀

PR #5644, opened 2023-11-01
Remove code that wrongfully infers the filing date as end of quarter

piiq:
This is a tricky one. The holdings are made public through a regulatory filing that usually comes at the end of each quarter.
There is no consistency between companies filing their filings when end of quarter falls onto a weekend. Some file on Saturdays, some don’t and I’ve not found data for Q1 2019 on FMP which was the last time end of quarter was a Sunday.

The initial code was inferring that the date of filing is the last day of the quarter (which is false for some companies' filings). Then I made code to infer that the date of the filing is the last working day of the quarter (which is false for other companies' filings)
Both of the solutions above lead to a situation when you can't consistently get the holdings even if you know the exact date of the filing.

In this PR i remove the code that adjusts the date and point the user to the holdings_dates endpoint to get the exact dates of the filings

### DIFF
--- a/openbb_platform/extensions/etf/integration/test_etf_api.py
+++ b/openbb_platform/extensions/etf/integration/test_etf_api.py
@@ -106,18 +106,36 @@ def test_etf_sectors(params, headers):
 @pytest.mark.parametrize(
     "params",
     [
+        ({"symbol": "IOO"}),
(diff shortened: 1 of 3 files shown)


piiq:
I'm going to close this PR in favour of https://github.com/OpenBB-finance/OpenBBTerminal/pull/5645 . That one has the updated test data in addition to the changes that are mentioned here.

PR #5651, opened 2023-11-01
Add ETF Holder command with FMP data

piiq:
There might be something wrong with FMP's data and it might not mean what we think it means. Putting this up here to test. Don't merge


piiq:
closing this PR and not implementing the endpoint

Issue #5633, opened 2023-10-31
[Bug] `/crypto/disc/fees` - The values for one-day fees are presented as a unix Timestamp.


deeleeramone:
Recreate with:

```console
/crypto/disc/fees
```

![Screenshot 2023-10-31 at 10 21 05 AM](https://github.com/OpenBB-finance/OpenBBTerminal/assets/85772166/4d53d9c3-bf6b-4854-9c36-97baab816c76)

PR #5652, opened 2023-11-01
#5633 adding comma every 1000

ramazanguvenc:
- issue #5633 

If fees are too long, it starts to convert to timestamps. If we add comma for every 1000, we can solve the problem and make it readable. This is how it looks after changes: 

image

How it looked before:

image


deeleeramone (approved):
Thanks for the PR!

PR #5655, opened 2023-11-02
Expanding the `fixedincome` menu

hjoaquim:
^


IgorWounds (approved):
Aside from what we discussed, looks good!

PR #5659, opened 2023-11-02
Add ETF holdings performance with FMP data

piiq:


IgorWounds (approved):

deeleeramone (changes requested):
I get an error with the symbol, IWM:

Screenshot 2023-11-02 at 2 32 24 PM

piiq:
Thanks for a great find @deeleeramone . IWM has over 1800 symbols in the holdings list and the request for price performance would break if you pass over 1730ish symbols. We're probably hitting the max url length with such a long array. I've added code to split the list into chunks of 500 symbols and this should solve the error.
This command combines 3 other commands available on the platform that can be used separately

deeleeramone:
> Thanks for a great find @deeleeramone . IWM has over 1800 symbols in the holdings list and the request for price performance would break if you pass over 1730ish symbols. We're probably hitting the max url length with such a long array. I've added code to split the list into chunks of 500 symbols and this should solve the error. This command combines 3 other commands available on the platform that can be used separately

@piiq, the number of symbols is not the only cause of errors - see the second screenshot as an example.

piiq:
Yes, this happens when it's not a valid symbol supported by fmp. The data here is as good as it is. IMO this command can give a nice overview but for detailed analysis calling the commands separately and filtering or cleaning the holdings list is safer and potentially more accurate

deeleeramone:
> Yes, this happens when it's not a valid symbol supported by fmp. The data here is as good as it is. IMO this command can give a nice overview but for detailed analysis calling the commands separately and filtering or cleaning the holdings list is safer and potentially more accurate

It's because they are not mapping by ISIN, and are converting only US-listed stocks to symbols.  IMO, it paints a false picture for an overview because there are too many holes and caveats.

PR #5665, opened 2023-11-03
Update settings.md

DidierRLopes:
Small update to trigger docs update


the-praxs (approved):
@DidierRLopes is da man

PR #5668, opened 2023-11-04
hotfix/fix-econ-cal: Nasdaq Econ Calendar Fetcher

deeleeramone:
This PR addresses an issue where there are an inconsistent number of results returned when the same request is made, and some other edge cases.

## Repeating the same request returns different results

Before:

image

After:

image

This is resolved by using the Requests library directly, and keeping the same headers and user agent for all URL requests the function makes.  Headers are defined in the helpers file.

## Validation error parsing the datetime of the event

Before:

```console
OpenBBError: 1 validation error for NasdaqEconomicCalendarData
date
  Input should be a valid datetime, invalid character in hour [type=datetime_parsing, input_value='2021-10-27 Tentative', input_type=str]
    For further information visit https://errors.pydantic.dev/2.4/v/datetime_parsing
```

After:

Screenshot 2023-11-03 at 10 06 57 PM

Edge case resolved by replacing, "Tentative", for, "00:00".

## Allow underscores in country names

After:

Screenshot 2023-11-03 at 10 03 35 PM

## HTML tags and artifacts in "description"

Before:

```console
"Gross Domestic Product (GDP) measures the annualized change in the inflation-adjusted value of all goods and services produced by the economy. It is the broadest measure of economic activity and the primary indicator of the economy's health.<BR/><BR/>A higher than expected reading should be taken as positive/bullish for the EUR, while a lower than expected reading should be taken as negative/bearish for the EUR."
```

After:

```console
"Gross Domestic Product (GDP) measures the annualized change in the inflation-adjusted value of all goods and services produced by the economy. It is the broadest measure of economic activity and the primary indicator of the economy's health.  A higher than expected reading should be taken as positive/bullish for the EUR, while a lower than expected reading should be taken as negative/bearish for the EUR."
```

Resolved with additional regex and replace in the validator.


IgorWounds (approved):

the-praxs:
I will have a look at it today and get this merged today itself

PR #5670, opened 2023-11-05
feature/compare-etf-countries: Add multi-ticker support to etf_countries

deeleeramone:
This PR proposes a change to the `etf.countries()` endpoint so that multiple tickers can be requested and compared with the same request.

- "weight" becomes {symbol}
- "weight" is removed from standard model, leaving only "country".
-  FMP fetcher modified to iterate over a list of symbols.

For other providers to follow along, the required column is "country" and the column for weight is the ticker symbol.  The symbol parameter will accept a string, `"IWM,VOO"`, or a list, `["IWM", "VOO"]`.

Screenshot 2023-11-04 at 10 08 43 PM


openbb_platform/platform/provider/openbb_provider/standard_models/etf_countries.py
```diff
@@ -26,7 +26,6 @@ def upper_symbol(cls, v: Union[str, List[str], Set[str]]):
 class EtfCountriesData(Data):
     """ETF Countries Data."""
 
-    country: str = Field(description="The country of the exposure.")
-    weight: float = Field(
-        description="Exposure of the ETF to the country in normalized percentage points."
+    country: str = Field(
+        description="The country of the exposure.  Corresponding values are normalized percent points."
```
deeleeramone, line 30: @piiq, if this is a reasonable place for describing the scale, I will apply the same treatment to Sectors and Info as a separate PR.

piiq (approved):

PR #5687, opened 2023-11-08
Rename Forex to Currency

IgorWounds:
Renames Forex to Currency


openbb_platform/extensions/currency/openbb_currency/price/price_router.py
```diff
@@ -9,28 +11,16 @@
 from openbb_core.app.router import Router
 from pydantic import BaseModel
 
-router = Router(prefix="")
-
-
-# pylint: disable=unused-argument
-@router.command(model="ForexPairs")
```
hjoaquim, line 16: It's supposed to drop this command?

IgorWounds: Yes, we renamed it to search.

the-praxs (approved):

openbb_platform/extensions/currency/openbb_currency/price/price_router.py
the-praxs: LGTM 🚀

PR #5689, opened 2023-11-08
Rename Crypto

the-praxs:
^


openbb_platform/extensions/crypto/openbb_crypto/price/price_router.py
```diff
@@ -0,0 +1,27 @@
+# pylint: disable=W0613:unused-argument
+"""Comparison Analysis Router."""
```
deeleeramone, line 2: Highlighting that there is an incorrect label at the top of the router here.

the-praxs: Thanks! Fixed.

jmaslek (approved):

PR #5693, opened 2023-11-09
Docs/v4-working-docs: PR to Work on V4 Docs

deeleeramone:
This PR opens a current place to iterate on V4 docs. FYI @DidierRLopes, @hjoaquim, @andrewkenreich, @piiq


piiq:
I've updated the v4 reference and data model specs according to the command org structure. 
There were very many files changed introduced so I'd like to merge this PR, if you don't mind

PR #5700, opened 2023-11-10
Hotfix/etf-multi-ticker: Update ETF sectors/info for multi-ticker support.

deeleeramone:
This PR updates `etf.sectors()` and `etf.info()` with the same multi-ticker support as `etf.countries()`.

Screenshot 2023-11-09 at 7 51 22 PM

Screenshot 2023-11-09 at 7 53 55 PM


piiq:
Hi, thanks
I'm not sure about this one yet. We're making a tradeoff on the standardization/validation and speed sides in favour of the ability to pass multiple tickers. Is this tradeoff worth that? It would be a no-brainer if there would be such an endpoint on the provider level or if this would be a separate command for comparing etfs, but as the only command that gets us etf's sector/country exposure, this seems a risky move. @jmaslek @IgorWounds your thoughts?

What if an user gives a list of 100 symbols? What if an api API would return NaNs or strings for one of them?

deeleeramone:
> What if an user gives a list of 100 symbols?

- Note that "info" is a little different, and just appends the results concurrently.

deeleeramone:
@piiq, in the event of a bad symbol being passed, the expected behaviour is for it to pass.

Screenshot 2023-11-10 at 9 23 12 AM

piiq:
My main points is that in general I oppose introducing patterns that remove standardization. Before we had a column that we were sure would come out a float if there's data there. Now we don't
This will be referenced as an example on other commands and it's going to be soon enough someone from the team will do that on a date column or something.

If this was a separate command (an opinionated shortcut if I can call it that way) or if the provider would have supported passing multiple tickers as a request parameter this would definitely be a feature worth looking at, but for now I don't think it brings enough value for the tradeoff.

I'll leave this as an opinion, but I will ask @jmaslek and @IgorWounds to share their thoughts.

deeleeramone:
> Before we had a column that we were sure would come out a float if there's data there. Now we don't

So what about `info` then?

deeleeramone:
Closing this for now.

PR #5703, opened 2023-11-10
Add devtools metapackage

piiq:
What this does is adds a package that has the developer tools we use. It's a package that has no logic in it and that's being installed only as a part of the dev_install script

### DIFF
--- a/openbb_platform/dev_install.py
+++ b/openbb_platform/dev_install.py
@@ -14,6 +14,7 @@ PYPROJECT = PLATFORM_PATH / "pyproject.toml"
 LOCAL_DEPS = """
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
+openbb-devtools = { path = "./extensions/devtools", develop = true }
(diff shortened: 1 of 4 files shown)


hjoaquim (commented):
Thank you, @piiq, looks amazing :smile: 

Only thing missing is to actually add it to the `dev_install.py` right?

Adding:
`openbb-devtools = { path = "./extensions/devtools", develop = true }`
on line 17 should be enough!

Also, should we consider adding the following to that script?:

```python

try:
    import toml
    import poetry  # noqa: F401
except ImportError:
    subprocess.run(
        [sys.executable, "-m", "pip", "install", "toml", "poetry"], check=True
    )
    import toml

```

So one doesn't have to install `toml` and `poetry` manually before running the script?

piiq:
i've added it to dev install 

> So one doesn't have to install `toml` and `poetry` manually before running the script?

I would leave this to the docs for now, let's talk about this after the beta release

PR #5705, opened 2023-11-10
Rename `economy` and create `index` extension

the-praxs:
^


the-praxs:
@piiq @IgorWounds @hjoaquim the commands `cot`, `cot_search`, and `sp500_multiples` are left as is in the `index` extension.

deeleeramone:
> @piiq @IgorWounds @hjoaquim the commands `cot`, `cot_search`, and `sp500_multiples` are left as is in the `index` extension.

I think it will be more correct to place, `cot` and `cot_search`, under, `regulators/cftc` because the data is not specific to indices.

the-praxs:
> > @piiq @IgorWounds @hjoaquim the commands `cot`, `cot_search`, and `sp500_multiples` are left as is in the `index` extension.
> 
> I think it will be more correct to place, `cot` and `cot_search`, under, `regulators/cftc` because the data is not specific to indices, or any specific asset class.

@minhhoang1023 wdyt? it will take a couple of minutes to do it.

piiq (approved):

PR #5709, opened 2023-11-10
Feature/v4-earnings-calendar: V4 Earnings Calendar

deeleeramone:
This PR adds the Equity Earnings Calendar for both FMP and Nasdaq.

The default state is to get the events from today until three-days from now, with historical data available from both providers.

Screenshot 2023-11-10 at 11 20 18 AM


piiq:
hi, thanks, is this the same as `obb.equity.estimates.historical`?

deeleeramone:
> hi, thanks, is this the same as `obb.equity.estimates.historical`?

Not at all. This is the market calendar.

deeleeramone:
This PR has moved to #5718

PR #5225, opened 2023-07-17
OpenBB Platform v4

piiq:
This PR adds the code for the refactored and rebranded OpenBB Platform v4. This is a major re-write of the platform that makes it extendable.

The PR includes:

- Core components: `openbb_core` and `openbb_provider`
- Extensions in `openbb_platform/extensions` folder
- Data providers integrations in `openbb_platform/providers` folder

See the readme in `openbb_platform` folder for details on how to install and use the v4

Fixes #5226

PR #5710, opened 2023-11-10
Add `derivatives` menu for `options` commands

the-praxs:
Moved `equity.options` to `derivatives.options` as a separate package.


the-praxs:
This should be merged into #5705 before that is merged into #5225.

piiq:
i think we'll keep options in the equity section for now. we don't have many datapoints for derivatives as a separate class and as we get those we'll introduce the menu. I'll be wrapping up the economy/index menu for now

the-praxs:
> i think we'll keep options in the equity section for now. we don't have many datapoints for derivatives as a separate class and as we get those we'll introduce the menu. I'll be wrapping up the economy/index menu for now

We have `futures` and `options` so we can include both in the `derivatives` extension.

the-praxs:
Closing in favor of #5742!

PR #5715, opened 2023-11-12
Rename etf discovery

piiq:


piiq:
Merging because unit tests pass. The jsons that fail the linting tests are related to the documentation and this is addressed in another PR

PR #5721, opened 2023-11-13
cool cards

andrewkenreich:
Didier: Don't generate index anymore @andrewkenreich, went through each manually. Except the reference ones and platform/data models. If you are happy w what we have I suggest using the new style on reference for all docs!

Only thing missing is removing this "Introduction" and making it as the index of each top menu. But I think this needs some tweaks on the sidebar - @jose-donato can u check this please?

image


hjoaquim:
I believe this proposal will not move forward.

Issue #5649, opened 2023-11-01
[Bug] `/stocks/gov/qtrcontracts`: Total column is summing string values instead of numbers.


deeleeramone:
Recreate with:

```
/stocks/gov/qtrcontracts
```

![Screenshot 2023-11-01 at 11 32 23 AM](https://github.com/OpenBB-finance/OpenBBTerminal/assets/85772166/6b61538d-355f-4d66-a50d-49f77991177b)

PR #5722, opened 2023-11-13
bug #5649

ramazanguvenc:
issue #5649 

convert Amount column to numeric

before:
before

after:
after


deeleeramone (changes requested):
Hi,

Thanks for the PR!

When I use the `--analysis` flags I get this error:


```
qtrcontracts --analysis downmom

Error: unsupported operand type(s) for /: 'str' and 'int'
```

ramazanguvenc:
This error existed before&after the changes I made, but cause is basically same so I took the same approach.

deeleeramone (approved):
Works as advertised, thanks!

PR #5725, opened 2023-11-14
Run tests on develop

jmaslek:
^^

### DIFF
--- a/.github/workflows/platform-core.yml
+++ b/.github/workflows/platform-core.yml
@@ -3,7 +3,7 @@ name: Test Platform V4
 on:
   pull_request:
     branches:
-      - feature/openbb-sdk-v4
+      - develop
 
 concurrency:
     group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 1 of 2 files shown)


deeleeramone (approved):
LGTM

.github/workflows/platform-core.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         fail-fast: false
         matrix:
           python_version:
-            ["3.8", "3.9", "3.10", "3.11"]
```
hjoaquim, line 19: Are we dropping 3.8 support??

piiq:
you think we can make a condition to run the tests only if the diff only has files in openbb_platform? or we have that already

.github/workflows/platform-core.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request:
     branches:
       - develop
+    paths:
+      - 'openbb_platform/**'
```
piiq, line 8: That was fast

piiq (approved):

PR #5737, opened 2023-11-15
Skip checking creds

montezdesousa:
# Pull Request OpenBB

Allows query executor to skip checking provider credentials if fetcher explicitly says so. By default it raises if credential is not set.

This is useful when 1 or more fetchers under the same provider don't actually use the credentials.

It happens with nasdaq, since economic calendar does not use api_key, but other fetchers do.

Screenshot 2023-11-15 at 13 15 47

Screenshot 2023-11-15 at 13 03 29


IgorWounds (approved):
Nice one! Can we also write this down somewhere in the docs so that people know how to explicitly state it?

PR #5738, opened 2023-11-15
Docs/contributing guidelines

hjoaquim:
Improving the contributing guidelines section on the website docs.


hjoaquim:
@piiq I generated the keywords and descriptions with the help of Cursor. Is there any other custom process I should follow instead for an SEO optimized set of keywords and descriptions?

deeleeramone:
@hjoaquim, this is good stuff. Let's port this over to the branch I'm working on.  I do have a question though, under Developer Guidelines we have Expectations for Developers, which is the same as Expectations for Contributors.  As we have separated the personas here, are we under the assumption that developers are not going to be submitting PRs but are building on top for their own purposes?

hjoaquim:
> @hjoaquim, this is good stuff. Let's port this over to the branch I'm working on. I've merged this PR into #5720.
> 
> I do have a question though, under Developer Guidelines we have Expectations for Developers, which is the same as Expectations for Contributors. As we have separated the personas here, are we under the assumption that developers are not going to be submitting PRs but are building on top for their own purposes?

As I view it, yes, that's it!

hjoaquim:
Closing this PR in favor of #5720.

PR #5739, opened 2023-11-15
Lint python package folder

montezdesousa:
To see linters being applied when building

```
import openbb
openbb.build(verbose=True)
```


hjoaquim (approved):
Works like a charm.

PR #5744, opened 2023-11-16
Rename the fmp search -> screener

jmaslek:
Title.  This was different than search as this allows to filter on different params.


deeleeramone:
> Title. This was different than search as this allows to filter on different params.

This was titled as Search because it resembles more of a search than a screener.

the-praxs (approved):
LGTM 🚀

PR #5746, opened 2023-11-16
feature/us-treasury-auctions: Adds Treasury Auctions endpoint and provider for US Government Data.

deeleeramone:
This PR adds `treasury_auctions()` to the router: `obb.fixedincome.government.treasury_auctions()` with data coming from the TreasuryDirect API as a provider named, `government_us`.   The purpose of naming the provider this is to provide clarity when other countries are added to government-specific endpoints.

This also renames the `treasury()` endpoint to be `treasury_rates()`, for more clarity.

There are a lot of fields returned (over 100) from the source, so to effectively use this function one would need to discard unwanted fields.  The default start date is three months ago, default end date is today.

Screenshot 2023-11-16 at 3 12 24 PM

For example, the bid-to-cover ratio is a really great field.

```
data = obb.fixedincome.government.treasury_auctions().to_df()
data[["cusip","original_issue_date","issue_date","maturity_date","security_term", "first_payment_date","bid_to_cover_ratio", "price_per_100"]]
```

Screenshot 2023-11-16 at 3 19 13 PM

```
bond_data = obb.fixedincome.government.treasury_auctions(security_type="Bond", start_date="2023-01-01").to_df()

bond_data[["cusip","original_issue_date","issue_date","maturity_date","security_term", "first_payment_date","bid_to_cover_ratio", "price_per_100"]]
```

Screenshot 2023-11-16 at 3 22 38 PM


openbb_platform/providers/government_us/openbb_government_us/models/treasury_auctions.py
```diff
@@ -0,0 +1,94 @@
...
+            query.end_date = datetime.now()
+
+        _query = query.model_dump()
+        _query["startDate"] = (
+            _query["startDate"].strftime("%m/%d/%Y") if query.start_date else None
+        )
+        _query["endDate"] = (
+            _query["endDate"].strftime("%m/%d/%Y") if _query["endDate"] else None
+        )
+        query_string = get_querystring(_query, [])
```
IgorWounds, line 76: Can you handle this Query part in the transform_query step please?

deeleeramone: Actually, if I do this step in the `transform_query` it simply returns the date format to the original formatting, YYYY-MM-DD.

The Output, for the URL, needs to be formatted specifically like that.

IgorWounds: What about the start_date and end_date logic?

deeleeramone: I can move the part that sets the default start/end.

IgorWounds (approved):

PR #5770, opened 2023-11-21
Add support for multiple ticker

IgorWounds:
As stated above.

### DIFF
--- a/openbb_platform/providers/intrinio/openbb_intrinio/models/company_news.py
+++ b/openbb_platform/providers/intrinio/openbb_intrinio/models/company_news.py
@@ -1,6 +1,7 @@
 """Intrinio Company News Model."""
 
 
+from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


openbb_platform/providers/intrinio/openbb_intrinio/models/company_news.py
```diff
@@ -61,12 +62,21 @@ def extract_data(
...
+            url = f"{base_url}/{symbol}/news?{query_str}&api_key={api_key}"
+            data = get_data_many(url, "news", **kwargs)
+            data = [{**d, "symbol": symbol} for d in data]
+            results.append(data)
 
-        return get_data_many(url, "news", **kwargs)
+        with ThreadPoolExecutor() as executor:
+            executor.map(get_data, [s.strip() for s in query.symbols.split(",")])
+
+        results = [item for sublist in results for item in sublist]
```
the-praxs, line 78: What is the use of this line of code?

the-praxs:
Except the comment I left, LGTM 🚀

the-praxs (approved):

openbb_platform/providers/intrinio/openbb_intrinio/models/company_news.py
IgorWounds: It makes sure that we don't end up with a List[List[Dict]]

PR #5778, opened 2023-11-22
Add github action with the same functionality as the reviewpad

piiq:
Replacing reviewpad's sanity checks with github actions following reviewpad being acquired by snyk

### DIFF
--- a/.github/workflows/sanity-review.yml
+++ b/.github/workflows/sanity-review.yml
@@ -1,0 +1,72 @@
+name: PR Sanity Review
+
+on: [pull_request]
+
+jobs:
+  label-pull-request-with-size:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Calculate PR size and label
+        id: pr_size
+        run: |
+          #!/bin/bash
+          # Patterns to ignore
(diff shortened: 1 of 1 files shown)


piiq:
I've actually tried to implement all the functionality we had offered by reviewpad, but working with diffs seemed to take more time than it's worth.
The challenge is that when I want to work on the contents of the diff I need to fetch the base branch which takes a lot of time given the size of our repo. Given the checks that reviewpad did were meant to be quick ones that preceded the tests and other actions waiting a few minutes to be bounced is not great.

luqmanbello:
> I've actually tried to implement all the functionality we had offered by reviewpad, but working with diffs seemed to take more time than it's worth. The challenge is that when I want to work on the contents of the diff I need to fetch the base branch which takes a lot of time given the size of our repo. Given the checks that reviewpad did were meant to be quick ones that preceded the tests and other actions waiting a few minutes to be bounced is not great.

When I was working with `git diff`, it worked best with v1 and you have to define the depth of what to fetch to improve speed. I would work on it and see what I can achieve

jmaslek:
@piiq @luqmanbello Is this mergable ?

piiq:
For now this only adds the label. But in general yes. Reviewpad will be switched off pretty soon

piiq:
This is not relevant anymore

PR #5787, opened 2023-11-23
New Intrinio endpoints w/ tests

the-praxs:
# New endpoints added
* equity.info
* index.market
* equity.discovery.filings
* equity.ownership.institutional
* equity.fundamental.latest_attributes
* equity.fundamental.historical_attributes
* equity.fundamental.metrics
* equity.ownership.insider_trading
* equity.calendar.dividend
* equity.ownership.share_statistics

# Changelog
* EquityInfo standard model has been modified to favor Intrinio's data. CBOE may need to define a new standard model that matches the data they provide.
* MarketIndices standard model is based on the FMP data. Currently Intrinio is the de-facto provider but its data is defined in the provider model so that FMP doesn't break.
* DiscoveryFilings model has been modified to balance between FMP and Intrinio data. However, FMP returns data for all symbols while Intrinio returns data for a specified symbol. The model needs to be modified to handle this difference.
* InstitutionalOwnership model has been redefined with the common values. It needs standardization b/w FMP and Intrinio.
* FinancialAttributes model has been renamed to HistoricalAttributes as per the latest requirements for Excel.
* LatestAttributes model has been added to get the latest value of the intrinio tags.
* KeyMetrics model has been redefined to include only the common fields between FMP and Intrinio.
* InsiderTrading model has been redefined to favor Intrinio's data.


deeleeramone:
> * DiscoveryFilings model has been modified to balance between FMP and Intrinio data. However, FMP returns data for all symbols while Intrinio returns data for a specified symbol. The model needs to be modified to handle this difference.

DiscoveryFilings is for non-specific data, i.e., all.  CompanyFilings is for symbol-specific data of the same nature.

the-praxs:
> > * DiscoveryFilings model has been modified to balance between FMP and Intrinio data. However, FMP returns data for all symbols while Intrinio returns data for a specified symbol. The model needs to be modified to handle this difference.
> 
> DiscoveryFilings is for non-specific data, i.e., all. CompanyFilings is for symbol-specific data of the same nature.

`FMPFilings` in `discovery_filings.py` was incorrectly defined, it should have been `FMPDiscoveryFilings`.

the-praxs:
This PR will be closed because of the breaking changes that lead to change in so many files.

Another PR will be created with the required files soon.

the-praxs:
Closed due to a lot of file changes possessing a stability risk. Correct feature PR -> #5789

PR #5789, opened 2023-11-24
New Intrinio Endpoints

the-praxs:
# New endpoints added

* equity.info
* index.market
* equity.discovery.filings
* equity.ownership.institutional
* equity.fundamental.latest_attributes
* equity.fundamental.historical_attributes
* equity.fundamental.metrics
* equity.ownership.insider_trading
* equity.calendar.dividend
*  equity.ownership.share_statistics

# Changelog

## Models
* EquityInfo standard model has been modified to favor Intrinio's data. CBOE may need to define their provider model that can be compatible with the standard model.
* MarketIndices standard model is based on the FMP data. Currently Intrinio is the de-facto provider but its data is defined in the provider model so that FMP doesn't break.
* Filings standard model has been removed because of its redundant requirement.
* CompanyFilings standard model has been added to provider SEC filing for the requested symbol(s). All the provider models are compatible with the data in the standard model.
* DiscoveryFilings standard model has been added to provide the latest SEC filings. Currently only FMP uses it.
* InstitutionalOwnership model has been redefined with the common fields. It needs standardization b/w FMP and Intrinio.
* FinancialAttributes model has been renamed to HistoricalAttributes as per the latest requirements for Excel.
* LatestAttributes model has been added to get the latest value of the intrinio tags.
* KeyMetrics model has been redefined to include only the common fields between FMP and Intrinio.
* InsiderTrading model has been redefined to favor Intrinio's data.

## Other
* `regulators.sec.filings` has been removed as the `openbb-sec` provider already caters to `CompanyFilings` standard model.
* `check_missing_integration_tests` has been modified to correctly replace `test` from the names, which includes functions having `test` in their names.
* `get_test_params` has been modified to include provider in `test_params` even if its empty.

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/calendar_dividend.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/calendar_dividend.py
@@ -14,7 +14,7 @@ from openbb_core.provider.utils.descriptions import (
 )
 
 
-class DividendCalendarQueryParams(QueryParams):
+class CalendarDividendQueryParams(QueryParams):
     """Dividend Calendar Query."""
 
     start_date: Optional[dateType] = Field(
@@ -25,14 +25,17 @@ class DividendCalendarQueryParams(QueryParams):
     )
 
 
-class DividendCalendarData(Data):
+class CalendarDividendData(Data):
     """Dividend Calendar Data."""
 
     date: dateType = Field(
         description=DATA_DESCRIPTIONS.get("date", "") + " (Ex-Dividend)"
     )
     symbol: str = Field(description=DATA_DESCRIPTIONS.get("symbol", ""))
-    name: Optional[str] = Field(description="Name of the entity.", default=None)
+    amount: Optional[float] = Field(
+        default=None, description="Dividend amount, per-share."
+    )
+    name: Optional[str] = Field(default=None, description="Name of the entity.")
     record_date: Optional[dateType] = Field(
         default=None,
         description="The record date of ownership for eligibility.",
@@ -44,7 +47,4 @@ class DividendCalendarData(Data):
     declaration_date: Optional[dateType] = Field(
         default=None,
         description="Declaration date of the dividend.",
-    )
-    amount: Optional[float] = Field(
-        default=None, description="Dividend amount, per-share."
     )
--- a/openbb_platform/providers/nasdaq/openbb_nasdaq/__init__.py
+++ b/openbb_platform/providers/nasdaq/openbb_nasdaq/__init__.py
@@ -1,6 +1,6 @@
 """Nasdaq provider module."""
 from openbb_core.provider.abstract.provider import Provider
-from openbb_nasdaq.models.calendar_dividend import NasdaqDividendCalendarFetcher
+from openbb_nasdaq.models.calendar_dividend import NasdaqCalendarDividendFetcher
 from openbb_nasdaq.models.calendar_earnings import NasdaqCalendarEarningsFetcher
 from openbb_nasdaq.models.calendar_ipo import NasdaqCalendarIpoFetcher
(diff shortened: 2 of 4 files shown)


deeleeramone:
I found a bug in, `obb.equity.ownership.institutional("AAPL", provider="intrinio", limit=10000).to_df()`, the name of the institution is all the same:

Screenshot 2023-11-26 at 12 00 25 PM

deeleeramone:
When I look at, `obb.equity.fundamental.metrics("AAPL", provider="intrinio").to_df()`,  I do not see the data from Intrinio and FMP being in the same category at all.  The FMP stuff belongs under the `fundamental` module, but the Intrinio data does not. That belongs more under the main `equity` section, or `price`.

the-praxs:
> I found a bug in, `obb.equity.ownership.institutional("AAPL", provider="intrinio", limit=10000).to_df()`, the name of the institution is all the same:
> 
> Screenshot 2023-11-26 at 12 00 25 PM

Fixed!
image

the-praxs:
> When I look at, `obb.equity.fundamental.metrics("AAPL", provider="intrinio").to_df()`, I do not see the data from Intrinio and FMP being in the same category at all. The FMP stuff belongs under the `fundamental` module, but the Intrinio data does not. That belongs more under the main `equity` section, or `price`.

I am also unclear on why that Intrinio endpoint was included in the `KeyMetrics` standard model.

@minhhoang1023 can you please check if this data should be in the metrics module?

the-praxs:
> I find it unclear how to use the Intrinio Market Indices endpoint. All of the other providers return data with just the symbol:
> 
> ```
> obb.index.market("NDX")
> ```
> 
> Screenshot 2023-11-26 at 12 21 17 PM
> 
> But the Intrinio returns an error:
> 
> ```
> OpenBBError: 1 validation error for IntrinioMarketIndicesQueryParams
> tag
>   Field required [type=missing, input_value={'symbol': 'NDX', 'start_...time.date(2023, 11, 26)}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.5/v/missing
> ```
> 
> What is the `tag` parameter and where do I find out what the valid tags are?
> 
> ```
> tag : Optional[str]
>     Index tag. (provider: intrinio)
> ```
> 
> Everything I try comes back with:
> 
> ```
> OpenBBError: Intrinio endpoint doesn't exist.
> ```
> 
> How do I get the data?
> 
> Looking in the fetcher test file, I can see that "tag" needs to be "level". What are the other tags and is the data still relevant to this function when it is not "level"?
> 
> Another notable item is that the `symbol` starts with a `$`. This should be added internally so that the user just enters, `SPX`, normally because it's unlikely they will know that this prefix needs to be there.

The tag parameter is essentially the name of the data field as set by Intrinio. For e.g., if you want the EBITDA value _only_ of a particular entity, you can use the `ebitda` tag in this [endpoint](https://docs.intrinio.com/documentation/web_api/get_data_point_number_v2). If suppose you want the name of the CEO _only_ of a company, you can use the `ceo` tag in this [endpoint](https://docs.intrinio.com/documentation/web_api/get_data_point_text_v2) instead.

After discussing with @montezdesousa, we found out that the `level` tag in question provides the `close` data for the market index. However, information on other supported tags is vague in Intrinio's docs and we suspect that it's an internal issue with them. Hence, I need to contact them to provider clarity on this endpoint and the supported tags for it. 

For now, you will have to provide the `tag=level`param in the function to make it work.

As per our previous discussions, the ability to enter a market symbol with/without the provider-required character was reverted and we decided that the symbol has to be prefixed with the provider-required character (FMP - `^`, Intrinio - `$`). 

I see that `NDX` provides data for other providers but this is not the case for `DJI`. We need to observe this behavior in detail for all providers and then add this functionality later. However, this is not the focus of this PR.

IgorWounds (approved):

PR #5788, opened 2023-11-23
Feature/ more async, replace `ThreadPoolExecutor` with `asyncio.gather`

tehcoderer:
WIP
image

PR #5801, opened 2023-11-27
Set `level` as default `tag` value for `index.market` in Intrinio

the-praxs:
^


openbb_platform/providers/intrinio/openbb_intrinio/models/market_indices.py
```diff
@@ -1,91 +1,91 @@
...
+        default=None,
+        description="Index type.",
+    )
+    sort: Literal["asc", "desc"] = Field(
+        default="desc",
+        description="Sort order.",
+        alias="sort_order",
+    )
+    limit: NonNegativeInt = Field(
+        default=100,
```
deeleeramone, line 35: I think while we are here, making this default to 10,000 will allow start/end dates to work better.

the-praxs: It took ~4 min when the limit was 4000, 10000 would be ~10 min or so. 

Would need to add the async code in place of the `ThreadPoolExecutor` to improve time complexity. #5788 is taking care of this issue so I'll ask @tehcoderer to make the limit 10000 on his PR after this is merged to develop.

the-praxs: The `start_date` and `end_date` should be taken into account if they are provided, regardless of the `limit`. This means a change in the code logic which is out of the scope of this PR.

deeleeramone: > It took ~4 min when the limit was 4000, 10000 would be ~10 min or so.

No, it allows the entire history to be returned by a single query.

Screenshot 2023-11-27 at 1 17 50 PM

the-praxs: I mistook this for the institutional ownership endpoint where it takes much more time to fetch the data.

But I will reiterate that the date logic should be implemented in the code which is more robust. The limit value by default is fine for now.

the-praxs: As per the recent team's discussion, the API params are to be included in the provider model, whether its used or not.

`limit` param is here to stay and the better user experience would be to implement the logic for the dates, as mentioned.

deeleeramone: I'm simply asking to increase the limit to an arbitrarily high number because that allows results to return without the need for pagination.  The "logic" for dates already exists, increasing the `page_size` is all that needs to happen.

the-praxs: On further testing, I find that 1000 is the maximum number of entries returned. I will set it as 1000 for the time being.

deeleeramone: How are you arriving at 1000 being the max?  I quite easily get 8000, no pagination required.

Here's the URL: 

"https://api-v2.intrinio.com/indices/stock_market/$NDX/historical_data/level?start_date=1990-11-27&end_date=2023-11-27&sort_order=desc&page_size=100000&api_key="

Screenshot 2023-11-27 at 4 20 54 PM

the-praxs:
@deeleeramone we need to merge this so that the command works out-of-the-box.

deeleeramone (approved):

PR #5808, opened 2023-11-28
Fix typos

omahs:
Fix typos


hjoaquim:
@luqmanbello any idea why is the "Branch Name Check" failing? I think the branch name adheres to the convention.

luqmanbello:
> @luqmanbello any idea why is the "Branch Name Check" failing? I think the branch name adheres to the convention.

image
The branch name showing in the error is `patch-1` so it would fail. maybe the branch was not renamed properly .

hjoaquim (approved):

PR #5811, opened 2023-11-28
Fix nasdaq top retail working without a key resulting in an error

the-praxs:
Without the api key, `equity.discovery.top_retail` will try to fetch the data but will run into

``` python
openbb_core.app.model.abstract.error.OpenBBError: 'datatable'
```

This PR fixes this by raising an error if Nasdaq API key is not found.


jmaslek:
Should we also add the catch around the datatable not found in the response?

montezdesousa:
@the-praxs I get this. If the key is not being required by default that's would be a bug in the `QueryExecutor`, why should we require the key manually in the fetcher?

Screenshot 2023-11-28 at 18 41 29

the-praxs:
> @the-praxs I get this. If the key is not being required by default that's would be a bug in the `QueryExecutor`, why should we require the key manually in the fetcher?
> 
> Screenshot 2023-11-28 at 18 41 29

I discovered with James that if there's no API key (i.e. its `""`) in this particular model, you will receive the `KeyError` (wrapped as `OpenBBError`. However, there exists a vulnerability that it will accept anything except an empty string in the API key for the link to work.

We can implement this hack for this case or if you have a long-term solution, please let me know!

montezdesousa:
> > @the-praxs I get this. If the key is not being required by default that's would be a bug in the `QueryExecutor`, why should we require the key manually in the fetcher?
> > Screenshot 2023-11-28 at 18 41 29
> 
> I discovered with James that if there's no API key (i.e. its `""`) in this particular model, you will receive the `KeyError` (wrapped as `OpenBBError`). However, there exists a vulnerability that it will accept any string in the API key for the link to work.
> 
> We can implement this hack for this case or if you have a long-term solution, please let me know!

If you check for `None` and `""` line 52 it should work for every provider

Screenshot 2023-11-28 at 18 55 19

the-praxs:
> Should we also add the catch around the datatable not found in the response?

I think we should put a custom exception in try-catch for the `dateable`. Any ideas in mind?

jmaslek (approved):

PR #5820, opened 2023-11-29
Make pro fetchers async

piiq:
This PR picks up from https://github.com/OpenBB-finance/OpenBBTerminal/pull/5788 because the original one is on a fork that prevents me from pushing


tehcoderer:
Closing in favour of #5822, it's upstream now

PR #5822, opened 2023-11-29
Make Pro Fetchers async by replacing `ThreadPoolExecutor` with `asyncio.gather`

tehcoderer:
image
_______

Implemented Fetchers:

- [x] Polygon
- [x] FMP
- [x] Benzinga
_______

While many changes are just adding async to files - here are some highlights of bigger ones
_______
### **`openbb_platform/core/openbb_core/provider/utils/client.py`**

- This Python code defines a custom client for making HTTP requests using the `aiohttp` library. It includes two main classes: `ClientResponse` and `ClientSession`.

- The `ClientResponse` class extends `aiohttp.ClientResponse` and adds functionality to obfuscate sensitive information from the request info. It also provides methods to get the first item from the response (`get_one`) and to get the JSON response (`json`).

- The `ClientSession` class extends `aiohttp.ClientSession` and sets the response class to `ClientResponse`. It provides methods for sending GET and POST requests (`get`, `post`), getting a JSON response (`get_json`), getting the first item from a JSON response (`get_one`), and a general request method (`request`).

- The request method in `ClientSession` includes functionality to handle gzip and deflate content encodings. If the `auto_decompress` flag is set to False, it manually decompresses the response body.

- The `get_user_agent` function returns a randomly chosen user agent string from a predefined list.

- The `obfuscate` function replaces sensitive information (like keys, tokens, and auth) in a dictionary with asterisks. This is used to obfuscate sensitive information in request parameters and headers.

_______

### **`openbb_platform/core/openbb_core/provider/utils/helpers.py`**

- This Python code provides a set of utility functions for making HTTP requests and manipulating data in the openbb_core project.
- The `async_request` function is an asynchronous helper function that makes HTTP requests to a given URL with potential headers and parameters. It supports both GET and POST methods.
- The `async_requests` function is similar to async_request but allows for making multiple requests asynchronously.
- The `make_request` function is a synchronous version of `async_request`, using the requests library to make HTTP requests.
- The `maybe_coroutine` function checks if a given function is a coroutine and runs it accordingly.
- The `run_async` function runs a coroutine function in a blocking context using the `anyio`library's `start_blocking_portal` function.
- The code also includes some type hinting and error handling to ensure the correct types of arguments are passed to the functions and to handle invalid HTTP methods.

_______

### **`openbb_platform/providers/fmp/openbb_fmp/models/equity_historical.py`**

- The `extract_data` method is asynchronous and uses the `async_requests` helper function to make concurrent requests to the FMP API for each symbol in the query. It also includes logic to increase the request timeout if there are more than 20 symbols in the query.

The callback method in the `extract_data` function of the `FMPEquityHistoricalFetcher` class is an asynchronous function that processes the response from each HTTP request made to the Financial Modeling Prep (FMP) API.

1. It takes two arguments: response and _. The response is the HTTP response from the FMP API, and _ is a placeholder for an argument that is not used in this function.

2. It extracts the JSON data from the response using the `await response.json()` command.

3. It extracts the symbol from the URL of the response. This is the stock symbol for which the historical data was fetched.

4. If the data is a dictionary, it retrieves the "historical" key's value, which is a list of historical data. If the "historical" key does not exist, it defaults to an empty list.

5. If the query symbol contains multiple symbols (i.e., there is a comma in query.symbol), it adds the symbol to each data dictionary in the list.

6. Finally, it returns the list of data dictionaries.

This callback function is used in the `async_requests` function, which makes asynchronous HTTP requests to the FMP API for each symbol in the query. The callback function processes the response from each request.

_______

### **`openbb_platform/providers/fmp/openbb_fmp/models/equity_quote.py`**

- The `extract_data` method splits the symbols into groups of 10 and makes asynchronous requests to the FMP API for each group of symbols.

_______

### Before
image

### After
image


deeleeramone:
Question: How do we get the output from the `response` when `.json()` is not actually expected to exist? 

i.e, "response.text", "response.content.decode()".

deeleeramone:
Additional questions:

Is there any specific reason for a small, defined, list of user-agents that may not cover all bases; or, can we use the `random_user_agent` library to broaden the choices?

If the user-agent is automatically placed in the request, does the user-agent stay the same for each subsequent request, or is it changing every time?

What HTTP caching library do we need to use, and how do we go about making the request using the pre-defined cached session object?

tehcoderer:
@deeleeramone 

1. depending on the context
```python
async def callback(response: ClientResponse, _: Any) -> str:
    """Return the response."""
    text = await response.text()
    ...  # do something with text

    return # return something

some_text = await async_request(urls, response_callback=callback, headers={"Accept": "text/html"})
````

2. something like [aiohttp-client-cache](https://github.com/requests-cache/aiohttp-client-cache) and passing it to `async_request(url,  response_callback=callback,, session=session)`

deeleeramone:
>something like [aiohttp-client-cache](https://github.com/requests-cache/aiohttp-client-cache) and passing it to
async_request(url,  response_callback=callback,, session=session)

Can we have `aiohttp-client-cache` added to the Core dependencies please?

IIUC, the only difference between `async def get_one` and `async def request` is that `get_one` does the `.json()` step?

> 3\. and the user_agents is there to match what sync `make_request` has

The list of ten user agents defined are very obsolete versions of Firefox, will this present issues for some requests?

piiq (changes requested):
Thanks, I’ve reviewed the PRr. It’s pretty clean, I have 2 questions/comments:

1. Can we have unit tests for the added openbb_core/provider/utils/client.py ? In general this file has a bunch of mypy errors that will be a pain down the road. Is it possible to get those fixed?
2. The changes to the field names like `actual_eps` -> `eps_actual`  in calendar earnings and historical eps. These can be breaking changes for existing users. Any context why this change is needed?

tehcoderer:
> Thanks, I’ve reviewed the PRr. It’s pretty clean, I have 2 questions/comments:
> 
> 1. Can we have unit tests for the added openbb_core/provider/utils/client.py ? In general this file has a bunch of mypy errors that will be a pain down the road. Is it possible to get those fixed?
> 2. The changes to the field names like `actual_eps` -> `eps_actual`  in calendar earnings and historical eps. These can be breaking changes for existing users. Any context why this change is needed?

1. donee!

2.  the `standard_model` has 
image
so I made sure we were following the standardized schema
image

PR #5829, opened 2023-11-30
Move pytest dependencies from openbb level to to core dev deps level

piiq:
This PR cleans up the leftovers of the core+provider merge. The pytest packages end up in openbb_core dev dependencies


hjoaquim:
Not really related, but, do we need `openbb-core` in the `openbb-devtools` package dependencies?

piiq:
> Not really related, but, do we need `openbb-core` in the `openbb-devtools` package dependencies?

top of my head - no

hjoaquim (approved):
Dependency wizard :mage_man: 
Thanks @piiq

PR #5839, opened 2023-11-30
hotfix/fix-fundamentals-symbol-validator: Fixes error created by validator.

deeleeramone:
This PR fixes a symbol validator that was raising errors in the Balance, Income, Cash standard models when valid symbols are being passed. 

Before:
image

After:
Screenshot 2023-11-30 at 10 40 35 AM


the-praxs:
@jmaslek Do we allow users to use both symbols and CIK or limit them to symbols?

jmaslek:
> @jmaslek Do we allow users to input either symbol(s) or CIK or limit them to symbol(s)?

That would be the ideal case to allow symbol or CIK (or others such as intrinio_id with intrinio commands)

the-praxs (approved):
LGTM 🚀

PR #5844, opened 2023-12-01
Small change in extensions doc 

jmaslek:
Try to make it more clear what is installed when we run pip install openbb.

Also edited the readme for the pypi version.


deeleeramone (approved):
LGTM

PR #5845, opened 2023-12-01
feature/yfinance-multi-ticker: Adds support to openbb-yfinance for multi-ticker download.

deeleeramone:
This PR makes the multi-ticker download functionality within the yFinance library compatible with the OpenBB format.  This is particularly useful for users updating from V3 SDK who are used to having this capability via `openbb.stocks.ca.hist` or `openbb.economy.index`.

Before:

```
obb.equity.price.historical(symbol="AAPL,MSFT", provider="yfinance").to_df()

OpenBBError: 'Open'
```

After:

Screenshot 2023-12-01 at 9 26 54 AM

Screenshot 2023-12-01 at 9 39 17 AM

Screenshot 2023-12-01 at 10 29 36 AM

Screenshot 2023-12-01 at 9 11 39 PM


the-praxs (changes requested):

openbb_platform/providers/yfinance/openbb_yfinance/models/crypto_historical.py
```diff
@@ -72,12 +72,21 @@ def extract_data(
         **kwargs: Any,
     ) -> List[Dict]:
         """Return the raw data from the Yahoo Finance endpoint."""
-        if "-" not in query.symbol:
-            position = len(query.symbol) - 3
-            query.symbol = query.symbol[:position] + "-" + query.symbol[position:]
+
+        tickers = query.symbol.split(",") if "," in query.symbol else [query.symbol]
```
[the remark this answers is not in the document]
deeleeramone: This cannot happen there because it needs to be used as a List internally, `transform_query` will just convert it back to a string-list.

jmaslek, line 76: split method will actually return a list if its not in there so you can drop the conditional

openbb_platform/providers/yfinance/openbb_yfinance/models/crypto_historical.py
```diff
@@ -72,12 +72,21 @@ def extract_data(
...
+        new_tickers = []
+        for ticker in tickers:
+            if "-" not in ticker:
+                position = len(ticker) - 3
+                new_ticker = ticker[:position] + "-" + ticker[position:]
```
jmaslek, line 81: You can also just combine this as:

new_ticker = ticker[:-3] + "-" + ticker[-3:]

The negative slicing gets rid of the need for the intermediate calc.

openbb_platform/providers/yfinance/openbb_yfinance/models/currency_historical.py
```diff
@@ -48,12 +48,22 @@ def transform_query(
...
+        symbols = (
+            params["symbol"].split(",")
+            if "," in params["symbol"]
+            else [params["symbol"]]
         )
+        for symbol in symbols:
+            new_symbol = (
+                f"{symbol.upper()}=X" if "=X" not in symbol.upper() else symbol.upper()
+            )
+            new_symbols.append(new_symbol)
```
jmaslek, line 64: you could replace this whole loop with a single list comprehension:

new_symbols = [f"{s.upper()}=X" if "=X" not in s.upper() else s.upper() for s in symbols]

openbb_platform/providers/yfinance/openbb_yfinance/models/market_indices.py
```diff
@@ -58,6 +58,31 @@ def transform_query(params: Dict[str, Any]) -> YFinanceMarketIndicesQueryParams:
...
+            else [params.get("symbol")]
+        )
+
+        new_tickers = []
+        for ticker in tickers:
+            _ticker = ""
+            indices = pd.DataFrame(INDICES).transpose().reset_index()
+            indices.columns = ["code", "name", "symbol"]
+
+            if ticker in indices["code"].to_list():
```
jmaslek, line 73: all of these to_list() can just be `.values` to optimize the usage of pandas

jmaslek (approved):
lgtm.  Just added a couple stylistic pythonic comments.

openbb_platform/providers/yfinance/openbb_yfinance/models/currency_historical.py
deeleeramone: replaced

openbb_platform/providers/yfinance/openbb_yfinance/models/market_indices.py
deeleeramone: replaced

Issue #5631, opened 2023-10-31
[Bug] `/crypto/defi/vaults` - Error: Coindix api exception: <html> <head><title>404 Not Found</title></head>


deeleeramone:
Recreate with:

`/crypto/defi/vaults`

and seemingly any combination of parameter.

```console
Error: Coindix api exception: <html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.18.0 (Ubuntu)</center>
</body>
</html>

Error: 'list' object has no attribute 'empty'
```

davidtran001:
I believe I have figured out a fix, please assign this issue to me.

PR #5849, opened 2023-12-02
Update API endpoint from deprecated api.coindix to new api.nanoly

davidtran001:
This PR addresses issue #5631, it fixes the `/crypto/defi/vaults` command.

The api endpoint `https://api.coindix.com/` and `https://apiv2.coindix.com/` has been replaced with `https://api.nanoly.com/` since coindix was renamed to nanoly last year.

Without the fix running `/crypto/defi/vaults` results in:
Screenshot 2023-12-01 at 11 18 44 PM

Running `/crypto/defi/vaults` after the fix resolves this issue:
Screenshot 2023-12-01 at 11 19 36 PM


deeleeramone:
Hi there! 

Thanks for the PR.  This looks pretty good, but the code linters are picking up something. 

`ruff --fix` might do the trick for the item it has highlighted, but there may be more after that.

 You can install the pre-commit hooks by running:

```
pre-commit install
```

This will run the series of linters, as well as the testing suite, locally before the commit goes to GitHub. 


As the name of the data source is changing, we need to reflect that change on-screen.  This is accomplished by updating this file:
https://github.com/OpenBB-finance/OpenBBTerminal/blob/develop/openbb_terminal/miscellaneous/sources/openbb_default.json

Doing this will also trigger the need to recapture the controller test output because the contents of the menu screen will have changed.

davidtran001:
> Hi there!
> 
> Thanks for the PR. This looks pretty good, but the code linters are picking up something.
> 
> `ruff --fix` might do the trick for the item it has highlighted, but there may be more after that.
> 
> You can install the pre-commit hooks by running:
> 
> ```
> pre-commit install
> ```
> 
> This will run the series of linters, as well as the testing suite, locally before the commit goes to GitHub.
> 
> As the name of the data source is changing, we need to reflect that change on-screen. This is accomplished by updating this file: https://github.com/OpenBB-finance/OpenBBTerminal/blob/develop/openbb_terminal/miscellaneous/sources/openbb_default.json
> 
> Doing this will also trigger the need to recapture the controller test output because the contents of the menu screen will have changed.

Hi, thanks for the feedback!

I ran `ruff -fix` and received the following output:
```sh
openbb_platform/providers/alpha_vantage/tests/test_alpha_vantage_fetchers.py:25:5: PLW0127 Self-assignment of variable `params`
openbb_platform/providers/cboe/tests/test_cboe_fetchers.py:86:5: PLW0127 Self-assignment of variable `params`
openbb_terminal/keys_model.py:947:9: PLW0127 Self-assignment of variable `r`
openbb_terminal/stocks/fundamental_analysis/yahoo_finance_model.py:295:11: S310 Audit URL open for permitted schemes. Allowing use of `file:` or custom schemes is often unexpected.
Found 4 errors.
```
It found a few errors but these are files that I have not made changes to so some I would appreciate some guidance on what to do here.

I also installed `pre-commit` and ran the linters and testing suite before commits on 12/5/2023.

openbb_terminal/cryptocurrency/defi/defi_controller.py
```diff
@@ -9,8 +9,8 @@
 
 from openbb_terminal.core.session.current_user import get_current_user
 from openbb_terminal.cryptocurrency.defi import (
-    coindix_model,
-    coindix_view,
+    nanoly_model,
```
deeleeramone, line 12: It is right here where Ruff is tripping.  It wants to see these sorted alphabetically.

deeleeramone: There is an error with the related tests.  I think you may have to re-capture the tests for the model, view and controller.

Screenshot 2023-12-06 at 7 43 13 PM

`pytest tests/openbb_terminal/path_to/test_file.py --record-mode once --rewrite-expected` should get us there, allegedly.

davidtran001: Hi, I have tried running the command you suggested but I can't get it to run on my machine locally.
The command:
```sh
≈```

and receive this error:
```sh
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --record-mode once
  inifile: /Users/davidtran/Code/OpenBBTerminal/pytest.ini
  rootdir: /Users/davidtran/Code/OpenBBTerminal
```

I have also ran
```sh
pytest tests/openbb_terminal/cryptocurrency/test_defi_controller.py --record-mode=once --rewrite-expected
```
But received the same error.

Do you have any suggestions where to go from here?

deeleeramone: oooo..  looks like I missed a sub-folder (defi), sorry about that. 

tests/openbb_terminal/cryptocurrency/defi/test_defi_controller.py

hjoaquim:
We're not maintaining the OpenBB Terminal anymore. Please refer to https://github.com/OpenBB-finance/LegacyTerminal if you still want to use that code.

PR #5872, opened 2023-12-07
Allow yf to be run in joblib parallel

### LINKED ISSUE #5869: [Bug] multi-ticker option - load many stock prices
In practice it is rarely the case that people will use just on stock. Instead people usually want to download the prices for all stocks in their portfolio. 

I have tried to parallelise myself the loading of historical prices, but it seems the openBB function is not well integrated with parallelize tools. At least the way I tried. 

Here an example: 

```
from joblib import Parallel, delayed
tickers = ['AAPL', 'TSLA']
func = obb.equity.price.historical
results = Parallel(n_jobs=2, prefer="threads")(delayed(func)(symbol = [tic], provider = 'yfinance') for tic in tickers)
```

This gives an error: 

`openbb_core.app.model.abstract.error.OpenBBError: 'AAPL'`

I have tried with other libraries to download yahoo prices and there was no problem in using `joblib`

jmaslek:
^^^ Fixes #5869


deeleeramone (approved):
LGTM

Screenshot 2023-12-07 at 9 01 40 PM

PR #5888, opened 2023-12-14
Fix bug. Add router test.

IgorWounds:
Test routers. Fix bug.

### DIFF
--- a/openbb_platform/extensions/tests/utils/router_testers.py
+++ b/openbb_platform/extensions/tests/utils/router_testers.py
@@ -124,23 +124,16 @@ def check_router_model_functions_signature() -> List[str]:
                 os.path.join(*router_name.split(".")) + ".py",
                 function.__name__,
             )
-            args = list(function.__code__.co_varnames)
-            if args != expected_args and decorator not in decorator_filer:
-                missing_args.append(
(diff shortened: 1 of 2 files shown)


openbb_platform/extensions/tests/utils/router_testers.py
```diff
@@ -8,13 +8,22 @@
 from openbb_core.app.provider_interface import ProviderInterface
 
 
-def collect_routers(path) -> List:
-    """Collect all routers in the path."""
+def collect_routers(target_dir: str) -> List[str]:
+    """Collect all routers in the target directory."""
+    current_dir = os.path.dirname(__file__)
+    base_path = os.path.abspath(os.path.join(current_dir, "../../../"))
```
hjoaquim, line 14: Is there another way to grab the `base_path`? In case of anything changing locations, this might break.

hjoaquim (changes requested):

openbb_platform/extensions/tests/utils/router_testers.py
IgorWounds: Not the best way, I know. Not sure how to do it right now. But I don't envision the test file moving around the codebase either :D

hjoaquim (approved):

PR #5890, opened 2023-12-14
Add REST API docs page with advanced settings configuration

piiq:
This PR moves the FastAPI docs to a dedicated page in the Usage section.
Also removes/cleans up minor things here and there


deeleeramone (approved):
LGTM!  I bumped this new rest page to the top of the usage section tree and added an example notebook link in the index file.

PR #5895, opened 2023-12-15
Fix pylint scope

piiq:
This PR should scope pylint work to only the fiels that were changed by the PR


luqmanbello (approved):
Looks good!!

PR #5900, opened 2023-12-18
feature/cboe-async: Refactor Cboe for Async

deeleeramone:
This PR refactors the Cboe provider extension to be fully async for all HTTP requests, and does some general maintenance along the way. This provides improved performance for the multi-ticker outputs, and generally cleans up the code.  Some highlights are listed below.

- Symbol directory caching library changed to `aiohttp-client-cache` - all related tests can be set to `"use_cache":False` for clean cassette captures.
- All Cboe fields related to a `%` have been converted to normalized percentage points.
- Consolidate the European Indices with the others.
- `profile()` --> `quote()`.
- Makes `IndexConstituents` usable for other providers and not only FMP with three choices.
- Fixes Intrinio bug in Options Chains for the `date`.
  - `OpenBBError: unsupported operand type(s) for -: 'str' and 'datetime.timedelta'`
- Arranges the fields in the OptionsChains standard model so the order is not hijacked by a single provider.
- Renovates EquityHistorical model so that `volume` is `Optional`, and `date` allows a pure date in addition to a datetime.


openbb_platform/openbb/package/__extensions__.py
```diff
@@ -11,13 +11,16 @@ class Extensions(Container):
     /crypto
     /currency
     /derivatives
+    /econometrics
```
hjoaquim, line 14: We need to keep the community extensions out of the static assets - extensions and providers.

hjoaquim (changes requested):
Looks great overall - only comment I have is asking for smaller PR next time 🥹

hjoaquim (approved):

hjoaquim:
Removed the changes on the static assets - that will be done on the release branch.

hjoaquim:
Added do not merge tag - let's merge only after the release is done.

deeleeramone:
> Looks great overall - only comment I have is asking for smaller PR next time 🥹

This can be a challenge because some things need to be done "all or nothing" and sprinkling PRs over the course of months and version increments can end up being more work.

PR #5902, opened 2023-12-19
Update USD Liquidity Example Notebook for V4

deeleeramone:
This PR updates the example notebook of the USD Liquidity Index for use with OpenBB Platform V4.  It also adds a `.tsx` file with and `webp` preview images.  The images cannot be added directly to the `MD` files like we do in the documentation.


DidierRLopes:
I updated @jmaslek as a reviewer 🤝

deeleeramone:
@jose-donato, will this TSX file and webp images be compatible with Hub if they are kept and maintained here?

jose-donato:
> @jose-donato, will this TSX file and webp images be compatible with Hub if they are kept and maintained here?

i would make it a json so it's easier to fetch from the hub

deeleeramone:
> i would make it a json so it's easier to fetch from the hub and also we would need to have the full path for the image to fetch it easily from the hub

More better now?

PR #5917, opened 2024-01-03
Add OECD unemployment (+more) + data cache idea

jmaslek:
2 fold:

1) Add all the stuff supported in OECD for unemployment rates (@deeleeramone- I added all the goodies from the previous PR).  Can do either:
```
obb.economy.unemployment().to_df()
```

Or have some fun:

```
obb.economy.unemployment(country="all", frequency="annual").to_df().reset_index(drop=False).pivot(index='date', columns='country', values='value')
```

2) Add idea for local caching.  I used this data example because pulling in the entire database is slow (will take >10 seconds the first run).  The API call can be edited each time to account for different parameters - but the request still takes longer than loading from memory.

@IgorWounds I left it in the provider specific one for now, but the idea would be the same.  Create functions in core, similar to obbject, that would handle whatever files/db one would want, and we can get fancy and have fun with it.  

have tests.

added composite leading indicator from the data explorer + short term + long term interest rates


openbb_platform/providers/oecd/openbb_oecd/utils/helpers.py
```diff
@@ -46,3 +54,163 @@ def fetch_data(url: str, csv_kwargs: Optional[Dict] = None, **kwargs: Any) -> Da
...
+        The method used for caching the data.
+
+    Raises
+    ------
+    NotImplementedError
+        If the cache_method is not 'parquet'.
+    """
+    if cache_method == "parquet":
+        cache_path = f"{cache}/{function}.parquet"
+        data.to_parquet(cache_path, engine="pyarrow")
```
deeleeramone, line 177: "pyarrow" requires Visual C++ Redistributable - X-Code and GCC for Mac - is there an engine that can avoid this process?

jmaslek: I think that's actually a parquet requirement even more than pyarrow. So if we wanted to avoid that I would go either with a different file or some error handling around it. And also the file handling should be moved into core helpers, I just did it here as a quick win.

Ultimately I'll let @IgorWounds weigh in

jmaslek: I made csv the default + removed pyarrow from deps.  Idea would be that a user setting could select the file to use.

IgorWounds:
Is this PR ready, or do you want to do more changes based on the added comments in the code?

jmaslek:
> Is this PR ready, or do you want to do more changes based on the added comments in the code? @jmaslek

Ready for review.  The idea is that if the solution (csv/parquet) was okay, I could move it to the core helpers.

IgorWounds (approved):

PR #5925, opened 2024-01-08
Introduces a limit argument for the `equity.fundamental.trailing_dividend_yield` endpoint

hjoaquim:
Issue description by @mnicstruwig:

> Currently the obb.equity.fundamental.trailing_divident_yield function returns all the values for the default argument. 
> 
> In the case of symbol="ASML" for example, this will return data all the way to 1995. Without the ability to limit these values, we often exceed context limits when executing function calls with an LLM.
> 
> As a proposed solution, can we utilize a limit kwarg with a default value of 252 (which is the number of trading days in a year). This means that, by default, a year’s worth of data will be returned.

Fix:

I introduced the limit argument in the following way:

1) If the limit kwarg is not specified, defaults to 252.
2) Any other int can be used to specify the limit.
3) If limit=None (explicitly) → gives you all available entries.

> the limit will give you the N most recent entries of the data (not the oldest)


openbb_platform/providers/tiingo/openbb_tiingo/models/trailing_dividend_yield.py
```diff
@@ -60,4 +60,5 @@ def transform_data(
         **kwargs: Any,
     ) -> List[TiingoTrailingDivYieldData]:
         """Return the transformed data."""
+        data = data[-query.limit :] if query.limit else data
```
mnicstruwig, line 63: Just double checking that this is deliberate (the `-`) to make sure we return the 252 _most recent_ days? So Tiingo returns data in chronological order (i.e. most-recent rows are last?)

IgorWounds (approved):

openbb_platform/providers/tiingo/openbb_tiingo/models/trailing_dividend_yield.py
hjoaquim: Yes, it's deliberate! That's exactly that.

hjoaquim:
Tagging @andrewkenreich @tehcoderer to make sure this is not a problem on Pro's side.

andrewkenreich:
we have it noted to change this - ty

PR #5928, opened 2024-01-08
Add filtering to AV historical price

hjoaquim:
Issue:

> `obb.equity.price.historical` doesn't follow `start_date` or `end_date` when using "alpha_vantage" as provider 

Fix:

Adding a helpers function for filtering the data with the start and end dates.

Note:
I didn't find a way of doing it directly on the request, thus adding the function.
Tagging @the-praxs to double-check this is correct.


openbb_platform/providers/alpha_vantage/openbb_alpha_vantage/models/equity_historical.py
```diff
@@ -180,5 +186,6 @@ def transform_data(
             {"date": date, **{extract_key_name(k): v for k, v in values.items()}}
             for date, values in data.items()
         ]
+        data = filter_by_dates(data, query.start_date, query.end_date)
```
the-praxs, line 189: I think we should do the filtering in the `extract_data` function as technically it's more suitable over there. The `filter_by_dates` function would then require a little tweak to work for a `dict` type object.

the-praxs:
Thanks for the PR @hjoaquim. Makes the most used command more versatile ❤️

the-praxs (approved):
LGTM 🚀

PR #5932, opened 2024-01-09
`equity.fundamental.dividends`: allow filter by date

montezdesousa:
^

```python
from openbb import obb
o = obb.equity.fundamental.dividends(
    symbol="AAPL", 
    start_date="2023-01-01", 
    end_date="2023-05-31",
    provider="intrinio"
)
o.to_df()
```

```python
from openbb import obb
o = obb.equity.fundamental.dividends(
    symbol="AAPL", 
    start_date="2023-01-01", 
    end_date="2023-05-31",
    provider="fmp"
)
o.to_df()
```


deeleeramone:
This PR works as-advertised, and also seems to reveal a difference between providers and data quality.  The number of results should be the same between both providers, the first/last dates in the DataFrames are the same.

Screenshot 2024-01-11 at 5 26 41 PM

montezdesousa:
> This PR works as-advertised, and also seems to reveal a difference between providers and data quality. The number of results should be the same between both providers, the first/last dates in the DataFrames are the same. The payment amount is also not the same.
> 
> Screenshot 2024-01-11 at 5 26 41 PM

Definitely some weird differences

the-praxs (approved):
LGTM 🚀

PR #5937, opened 2024-01-10
Disable auto_build on test run

IgorWounds:
^


hjoaquim (changes requested):
Can we have it on the conftest instead, so the behavior is the same locally?

hjoaquim (approved):

PR #5938, opened 2024-01-10
Feature/charting obbject

hjoaquim:
# Charting Extension Refactor

This PR is a POC/proposal for us to consider moving the `openbb-charting` extension into the `OBBject` Extension realm (https://docs.openbb.co/platform/development/developer-guidelines/obbject_extensions).

### Pros

- current setup for charting extensions might be difficult to understand: it uses the `ChartingService` which is not that straightforward - this proposal completely drops the need for having it.
- it's very easy and straightforward to understand how the extension interacts with the `OBBject` being one of its accessors.
- it's easy to discover/understand that the new views should be added to `charting_router`
- plug-and-play that might set a precedent for other charting extensions to come in
- `OBBject.to_chart()` method is removed in detriment of `OBBject.charting.to_chart()`, which exposes `ChartIndicators` - this was identified as a main pain point bc a user couldn't really know how to use it.
- introduces `OBBject.charting.indicators` to easily discover TA indicators

### Cons

- no type hints: might still be difficult to discover the extension is there and to leverage its methods - this is overcome by looking into `OBBject.accessors`

> [!WARNING] 
> This has breaking changes on the charthing extension - we need to bump to version to `2.0` and publish before when the PR is ready.

## TODO: 

- [x] Discover commands with views
- [x] Remove ChartingService infra
- [x] Revamp tests: remove old ones (charting int tests + charting service unit tests) /// create tests for the new extension 
- [x] Cleanup `extensions/charting` - reuse the function that generate boilerplate for charting functions
- [x] Ensure dev install works properly
- [x] Ensure nightly-pypi works properly
- [x] Check if logs are properly generated - also on PyWry window
- [x] Improve discoverability: 
   - [x] make accessors explicit
   - [x] documented
- [x] Review README
- [x] Rename to `openbb-charting` so we can reuse the pypi project

## Questions:

1. ~~Drop `OBBject.show()`?~~ No, it adds visibility to the fact that the chart is there.
2. ~~Where to document `OBBject` accessors?~~ Website docs and README.


hjoaquim:
Installation note:
1. Install OBBject extension: go to `OpenBBTerminal/openbb_platform/obbject_extensions/charting` and do `poetry install` - this will cause the core to not be installed from local.
2. Re-install core from local: navigate to `OpenBBTerminal/openbb_platform/core` and do `poetry install`

> this is obviously a temporary workaround for testing purposes

hjoaquim:
Example: 
image

openbb_platform/obbject_extensions/charting/charting/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+    """Charting extension."""
+
+    def __init__(self, obbject):
+        """Initialize Charting extension."""
+        # pylint: disable=import-outside-toplevel
+        from openbb_core.app.model.charts.charting_settings import ChartingSettings
+        from openbb_core.app.model.obbject import OBBject
+
+        self._obbject: OBBject = obbject
+        self._obbject = obbject
```
montezdesousa, line 24:
```suggestion
```

montezdesousa:
> Example: image

lgtm, can we reuse the https://pypi.org/project/openbb-charting/ ?

openbb_platform/core/openbb_core/app/model/obbject.py
```diff
@@ -59,6 +61,8 @@ class OBBject(Tagged, Generic[T]):
         description="Extra info.",
     )
     _credentials: ClassVar[Optional[BaseModel]] = None
```
montezdesousa, line 63: we prolly don't need to have these `credentials`, since we've access to `user_settings` which contain them

hjoaquim: Yes! You're right - didn't change it yet for the sake of simplicity but if moving forward that's something to change

hjoaquim:
> > Example: image
> 
> lgtm, can we reuse the https://pypi.org/project/openbb-charting/ ?

that could be an approach indeed. although I would advocate for dropping it in favor of the new one - no need to have to maintain 2 extensions imo

montezdesousa:
> > > Example: image
> > 
> > lgtm, can we reuse the https://pypi.org/project/openbb-charting/ ?
> 
> that could be an approach indeed. although I would advocate for dropping it in favor of the new one - no need to have to maintain 2 extensions imo

I meant replacing but using the same project name in papi, not maintaining the 2

hjoaquim:
> > > > Example: image
> > > 
> > > 
> > > lgtm, can we reuse the https://pypi.org/project/openbb-charting/ ?
> > 
> > 
> > that could be an approach indeed. although I would advocate for dropping it in favor of the new one - no need to have to maintain 2 extensions imo
> 
> I meant replacing but using the same project name in papi, not maintaining the 2

ohh gotcha, sorry for the misinterpretation! yes, absolutely.

deeleeramone:
@hjoaquim, with `to_chart()`, do you think the output should be suppressed - like `show(render=False)`?  Are these equivalent actions, and if yes, do we need both methods?

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+
+        fig, content = charting_function(**kwargs)
+        self._obbject.chart = Chart(
+            fig=fig, content=content, format=charting_router.CHART_FORMAT
+        )
+        if render:
+            fig.show()
+
+    def to_chart(
+        self,
```
deeleeramone, line 87: In `to_chart`, and `show`, I'm looking for an entry point that can take `List[Data]` and some variants.  A data param, something like: `data: Optional[Union[Data, List[Data], Dict, List[Dict], DataFrame, Series]] = None`, where `data` would override `self` before it created the figure, when supplied.

```
data = obb.equity.price.historical("AAPL", start_date="2000-01-01")
data.charting.to_chart(data = data.to_df().loc["2022-01-01":])
```

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+        indicators: Optional[Union[ChartIndicators, Dict[str, Dict[str, Any]]]] = None,
+        symbol: str = "",
+        candles: bool = True,
```
deeleeramone, line 90: I also find myself looking for a way to make this useful for other types of data, or where the type of chart desired is a different format.  Like the entry point for data, but then a way to define X/Y based on a list of fields within Data.

hjoaquim:
> @hjoaquim, with `to_chart()`, do you think the output should be suppressed - like `show(render=False)`? Are these equivalent actions, and if yes, do we need both methods? Currently, if you do not catch the output of `to_chart()` in a variable, a browser window opens up to display the JSON.

I actually like the idea of instead of returning the figure, save it to the OBBject instead, just like `show()` does - will keep rendering it, unless said otherwise (`render=False`).

Regarding the need for both methods, in theory, we could have a single one, in practice I believe it's better to have them separate:
 - `show()` out-of-the-box views 
 - `to_chart()` user customized new view

> tldr: different functionalities, different methods

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
hjoaquim: I like this idea, this is something we already have in the data processing commands so it's really low effort to implement - will do it on this PR.

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
hjoaquim: I like the idea but for now, I think we need to keep adding changes, PR is too big already! 
IMO, the primary focus regarding the charting extension moving forward should be to actually implement more views.
@deeleeramone would you mind documenting this in our usual Jira ticket so that we don't lose this insight?

openbb_platform/core/openbb_core/app/model/obbject.py
```diff
@@ -39,6 +41,10 @@
 class OBBject(Tagged, Generic[T]):
     """OpenBB object."""
 
+    accessors: ClassVar[Set[str]] = set()
```
IgorWounds, line 44: Is there a better or more intuitive naming than accessors?

openbb_platform/obbject_extensions/charting/README.md
```diff
@@ -8,6 +8,9 @@ The library includes:
 - a set of charting components
 - prebuilt charts for a set of commands that are built-in OpenBB extensions
 
+>[!NOTE]
+> The charting library is in fact an [`OBBject` extension](https://docs.openbb.co/platform/development/developer-guidelines/obbject_extensions) which means you'll have the functionality it exposes on every command result.
```
IgorWounds, line 12: "in fact" not needed

IgorWounds:
@CodiumAI-Agent /review

deeleeramone:
@hjoaquim, I'm finding the "Add Text" button to be not working.

openbb_platform/core/openbb_core/app/model/obbject.py
hjoaquim: I think the reason @montezdesousa picked that nomenclature was due to the fact `pandas` uses it.

hjoaquim:
> * If

Thanks for the careful review @deeleeramone !
All the mentioned issues are already on the current charting extension, but the one related to the `techinical` extension views.
That's fixed now, and I'll add the others to the backlog.

hjoaquim:
@IgorWounds agree we need to release `2.0` on merging?

hjoaquim:
@deeleeramone does this introduce potential issues in docs/content? I'm guessing notebook examples and PyQuantNews materials.

deeleeramone:
> @deeleeramone does this introduce potential issues in docs/content? I'm guessing notebook examples and PyQuantNews materials.

Thanks, @hjoaquim, I'll take a look through and see if there's anything to update. There hasn't been a lot of focus on charting in the examples yet.

PR #5939, opened 2024-01-10
Excel docs update

montezdesousa:
* Updates parameter order
* Renames 'Optional' column to 'Required' in reference


montezdesousa:
wrong branch

PR #5945, opened 2024-01-11
Hotfix/remove unused workflow

IgorWounds:
^


noxfile.py
```diff
@@ -20,6 +20,10 @@ def tests(session):
     )
     session.install("pytest")
     session.install("pytest-cov")
+    session.run(
```
hjoaquim, line 23: You've probably checked out from the wrong branch

hjoaquim (changes requested):

hjoaquim (approved):

PR #5950, opened 2024-01-12
Feature/chart improvements

hjoaquim:
- Tries to make the `ChartIndicators` more visible and usable.
- Experiment with the signatures of charting extension `to_chart()` and `OBBject.to_chart()`
   - idea was to wrap `OBBject.to_chart()` with `to_chart()` signature so that user can leverage the `kwargs` - although it works and make the updated signature available, I'm not sure if it enhances UX.

This is a different approach than the one proposed [here](https://github.com/OpenBB-finance/OpenBBTerminal/pull/5938)  with the goal of enhancing UX.

### DIFF
--- a/openbb_platform/extensions/charting/README.md
+++ b/openbb_platform/extensions/charting/README.md
@@ -90,4 +90,35 @@ The user can use any number of `**kwargs` that will be passed to the `PlotlyTA` 
 Refer to the [`to_chart` implementation](openbb_charting/core/to_chart.py) for further details.
 
 > Note that, this method will only work to some limited extent with data that is not standardized.
-> Also, it is currently designed only to handle time series data.
+> Also, it is currently designed only to handle time series (OHLCV) data.
+
+Example usage:
+
+- Plotting a time series with TA indicators
+
+  ```python
+
+    from openbb import obb
+    res = obb.equity.price.historical("AAPL")
+
+    indicators = dict(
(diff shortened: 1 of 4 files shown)


openbb_platform/extensions/charting/README.md
```diff
@@ -90,4 +90,35 @@ The user can use any number of `**kwargs` that will be passed to the `PlotlyTA`
...
+    )
+    res.to_chart(**{"indicators": indicators})
+
+  ```
+
+- Get all the available indicators
+
+    ```python
+
+    from openbb_charting.core.plotly_ta.data_classes import ChartIndicators
```
deeleeramone, line 121: What do you think about a Charti object method within the extension framework of your other PR, `data.chart.available_indicators()`?

hjoaquim: Can easily be implemented as well! I mean, everything in this PR can easily be translated into the other!

deeleeramone: It might better to reflect on all chart improvements together.

hjoaquim: Haven't group them bc not sure if the OBBject extension should be the way moving forward. 😬

hjoaquim:
Closing in detriment of https://github.com/OpenBB-finance/OpenBBTerminal/pull/5938.

PR #5986, opened 2024-01-23
Improve UX for `obb.news`

hjoaquim:
Improving the UX of `obb.news.company` and `obb.news.world` by adding date filtering.


openbb_platform/core/openbb_core/provider/standard_models/company_news.py
```diff
@@ -25,13 +28,39 @@ class CompanyNewsQueryParams(QueryParams):
     limit: Optional[NonNegativeInt] = Field(
         default=20, description=QUERY_DESCRIPTIONS.get("limit", "")
     )
+    date: Optional[dateType] = Field(
```
montezdesousa, line 31: Did you consider removing this parameter `date` since we now have `start_date` and `end_date` it feels duplicated

montezdesousa, line 31: Did you consider removing this parameter `date` since we now have `start_date` and `end_date` it feels duplicated?

hjoaquim: It was intended! This is because `benzinga` has those 3 parameters, so I modeled after that.
Do you think it would be a better approach to have start and end dates as standard and then `date` only in bezinga, as a provider specific parameter?

IgorWounds (approved):

montezdesousa (approved):

PR #5987, opened 2024-01-23
hotfix/intrinio-index: Drop None Rows From intrinio Historical Index

deeleeramone:
This PR removes empty rows of data in the response from Intrinio's historical index levels. A couple of enhancements have been made for UX:

- Multi-ticker support
- `$` is no longer required to be in the symbol, both forms of the symbol are accepted.
- Remove params that don't serve any real purpose:
  - `tag`
  - `type`
- Allow for a date only to be returned, instead of forcing a datetime for non-intraday data.

Screenshot 2024-01-23 at 12 37 15 PM


deeleeramone:
Apparently I added the package folder to the commit, will undo that.

IgorWounds (approved):
LGTM

@andrewkenreich

andrewkenreich:
we use fmp here too - so this fine

PR #5993, opened 2024-01-24
[Feature] - Support for custom examples in router commands

IgorWounds:
Why?
- To allow developers to add their custom handmade examples besides the autogenerated ones.

What?
- Removed parts of the `DocstringGenerator` to its own class for generating examples `ExampleGenerator`. Added a new field in the `openapi_extra` to store the examples. The examples are now generated on every launch of the API interface.

To add custom example you can do:

```
@router.command(model="SOFR", examples=["example_1", "example_2"])
async def sofr( ...
```

Impact:
- This is expected to impact all docstring on which devs will write custom examples. Currently, it might impact the import time. Impact score: 5

Testing Done:
- Tested it by adding an example to a router command and observed the expected change in the static asset. Wrote unit tests for the new `ExampleGenerator` class.


montezdesousa:
This approach works, but adding python examples directly in the summary will pollute the RestAPI descriptions.

These descriptions should be language agnostic, otherwise client applications will not be able to reuse them directly.

Screenshot 2024-01-25 at 14 32 48


I'd explore using the `openapi_extra` that we already use to store the command model inside the `router.command` decorator. We could use that `dict` to store "python_examples" in a way that it can be ignored by non python clients.

IgorWounds:
Thanks for the review @montezdesousa, would this alternative cleanup do the trick as we already do it for the parameters and return fields?

montezdesousa (approved):

PR #5997, opened 2024-01-26
hotfix/intrinio-options-date: Fix comparing values between str and datetime

deeleeramone:
This PR fixes:

```python
obb.derivatives.options.chains("dxc", date="2022-04-01")
```
image

Didn't quite fix it all the way in the last effort, this should finish it off.

Screenshot 2024-01-25 at 5 32 53 PM


tehcoderer (approved):
perfect phone call !

PR #5999, opened 2024-01-26
[Feature] - Field order added to the OBBject extra as preference

IgorWounds:
Why?

To fix the issue of the openapi.json being an unordered collection which resulted in frontends that are parsing it not having the right order of fields that are declared in the Pydantic models.

What?

Updated the `OBBject` extra that it has a `field_order` dictionary which preserves the information. It is also a preference that can be toggled on/off.

Impact:

This is expected to impact the look of the `OBBject` and the preferences. Impact score: 3
Testing Done:

Tested it by executing the income statement command and observing that the field order was added.


hjoaquim (approved):

minhhoang1023 (approved):

montezdesousa:
Hey don't forget to include the new preference in the docs page

PR #6015, opened 2024-01-30
Deprecation warning on the reference docs

hjoaquim:
Adds a deprecation warning in the reference docs:

image

> [!NOTE] 
> I tried to make it strike-through the title (fastapi style), but it's not trivial - if we want do also do it, we'll need to allocate more time to it!

> [!IMPORTANT] 
> I needed to bring back some retired methods on the `PackageBuilder` - this might need some revision - added to backlog.


deeleeramone:
@hjoaquim, we will also need to have the procedure for deprecating a function or parameter documented in a page under the Contributor Guidelines.

PR #6022, opened 2024-02-01
[BugFix] Fix package builder

montezdesousa:
1. **Why**?:

    - Bug when running `openbb.build()`

2. **What**?:

    - Specify the import following previous pattern

3. **Impact**:

    - Specific imports in other files are not loaded, need to be explicit now

4. **Testing Done**:

    - `import openbb; openbb.build()`
    - `obb.index.market("^IBEX")`


deeleeramone:
This fixes the issue I encountered with macOS Python 3.11, at least for me.

hjoaquim (approved):

PR #6023, opened 2024-02-01
[BugFix] - hotfix/polygon-tz-aware - Add tz-aware to timestamp conversion.

deeleeramone:
-    Fix polygon timestamps:
       - Format: [BugFix] - Adds a timezone to the Unix timestamp conversion to correct localization errors.

-    Why? (1-3 sentences or a bullet point list):
       - Bug with the conversion of Unix timestamp that made the start of day wrong depending on what timezone you are in. 

-    What? (1-3 sentences or a bullet point list):
      -  Stocks and Indices are stamped as "America/New_York" - Polygon only services American Markets.
      -  Crpyto and currencies are stamped as "UTC".

-    Impact (1-2 sentences or a bullet point list):
      -  None

-    Testing Done:
       - obb.index.price.historical("NDX")
       - obb.equity.price.historical("AAPL")
       - obb.currency.price.historical("USDJPY")
       - obb.crypto.price.historical("BTCUSD")
       - Intraday and daily

Before:

Screenshot 2024-02-01 at 1 22 05 PM

Dates are all shifted back one day.

After:

Screenshot 2024-02-01 at 1 24 31 PM

### DIFF
--- a/openbb_platform/core/openbb_core/app/static/package_builder.py
+++ b/openbb_platform/core/openbb_core/app/static/package_builder.py
@@ -345,6 +345,7 @@ class ImportDefinition:
         code += "\nfrom openbb_core.app.static.utils.decorators import validate\n"
         code += "\nfrom openbb_core.app.static.utils.filters import filter_inputs\n"
         code += "\nfrom openbb_core.provider.abstract.data import Data"
+        code += "\nfrom openbb_core.app.deprecation import OpenBBDeprecationWarning\n"
         if path.startswith("/quantitative"):
             code += "\nfrom openbb_quantitative.models import "
             code += "(CAPMModel,NormalityModel,OmegaModel,SummaryModel,UnitRootModel)"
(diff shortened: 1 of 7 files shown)


IgorWounds:
@deeleeramone 
Shouldn't we allow users that want the power of Polygon to have any type of candle interval to use it? I personally wouldn't hide the timespan and similar from the provider-specific QueryParams as that is the power unique to Polygon and doesn't really cost us anything.

deeleeramone:
> @deeleeramone Shouldn't we allow users that want the power of Polygon to have any type of candle interval to use it? I personally wouldn't hide the timespan and similar from the provider-specific QueryParams as that is the power unique to Polygon and doesn't really cost us anything.

@IgorWounds - You can enter any interval you like, but we want standardized params.  "equity.price.historical" already is this, so why is it an issue when adopting this for essentially the exact same query elsewhere?  "Timespan" + "multiplier" were being used **instead** of "interval", so to get a 1 minute interval you have to do something completely different, unique to Polygon.  "1m" should always be "1m", not "1min" or "timespan='minute', multiplier=1"

IgorWounds (approved):

PR #6062, opened 2024-02-09
[Feature]  - Comma separated string and list inputs

montezdesousa:
1. **Why**?

    - Consistent behaviour in validation, raises exception if user tries to pass multiple symbols and that is not allowed
    - Easier for downstream applications to get an error than look into warnings
    - `upper` validator was parsing the symbols, better to keep single responsibility

2. **What**?

    - Use `__json_schema_extra__ = {"<field_name>": ["multiple_items_allowed"]}` to be explicit about which providers support multiple items or not for a given field
    - Then we merge the content into the field's json schema extra, to be reflected in openapi.json
    - This is used by `PackageBuilder` to join list items into comma-separated string, depending on the provider
    -  Updates description and docstring type stating when field allows multiple values
    - Mark all fetchers where multiple values allowed -> https://github.com/OpenBB-finance/OpenBBTerminal/pull/6074
    - This can be reused for other purposes and will equally populate the field schema:
```python
__json_schema_extra__ = {
    "<field_name_A>": ["some_prop", "another_prop"], 
    "<field_name_B>": ["yet_another_prop"]
}
```

openapi.json
image

3. **Impact**:

    - This PR mainly affects python users
    - We don't use the first symbol in the list if the input is strictly single symbols
    - Fixed inconsistent descriptions and docstrings against signature

4. **Testing Done**:

    - Passing multiple symbols to different endpoints

```python
# multiple values not allowed, raises exception
obb.equity.price.historical(["AAPL", "MSFT"], provider="intrinio")
# multiple values allowed
obb.equity.price.historical(["AAPL", "MSFT"], provider="yfinance")
```


IgorWounds (approved):

deeleeramone:
We need to allow `List[str]` in the Python interface where multiple symbols are allowed.

Sorry, but not allowing this subject has already introduced a massive breaking change that impacts all users, along with our own syntax and documentation.

montezdesousa:
> We need to allow `List[str]` in the Python interface where multiple symbols are allowed.
> 
> Sorry, but not allowing has already introduced a massive breaking change that impacts all users, along with our own syntax and documentation.

@deeleeramone can you pls test the current state of this PR and tell if it eliminates the breaking changes?

IgorWounds (approved):

PR #6064, opened 2024-02-10
[BugFix] Fix World News

deeleeramone:
1. **Why**? :
    - Only Benzinga and Tiingo will accept start/end dates as a query parameter. Filters applied to every provider on output resulted in nothing being returned. 

    - Remove `filter_by_date`
    - Make default `start_date` more practical - 2 weeks. There are thousands of results per-day, 1-year is beyond the capabilities of any provider to return as a single query.
    - Make default limit size 2500, which is about 1-2 days of results from Intrinio.

    - It works now.

4. **Testing Done**:

Before:

Screenshot 2024-02-09 at 9 48 22 PM

After:

Screenshot 2024-02-09 at 10 44 05 PM

Screenshot 2024-02-09 at 10 46 47 PM

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/world_news.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/world_news.py
@@ -21,9 +21,9 @@ class WorldNewsQueryParams(QueryParams):
     """World News Query."""
 
     limit: NonNegativeInt = Field(
-        default=20,
+        default=2500,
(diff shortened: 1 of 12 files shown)


IgorWounds (approved):

IgorWounds:
Good spot @deeleeramone ! Let's just fix the test and we're good to go. @andrewkenreich will the change in the standard model affect you in any way?

PR #6072, opened 2024-02-14
[BugFix] - Fix QA Examples

IgorWounds:

    - The examples had a bug that misled users.

    - Changed all data that is passed to QA commands to dataframes.

    - Only changes the docstrings.

4. **Testing Done**:

    - Tested the QA commands.

### DIFF
--- a/openbb_platform/extensions/quantitative/openbb_quantitative/quantitative_router.py
+++ b/openbb_platform/extensions/quantitative/openbb_quantitative/quantitative_router.py
@@ -55,7 +55,7 @@ def normality(data: List[Data], target: str) -> OBBject[NormalityModel]:
     Examples
     --------
     >>> from openbb import obb
-    >>> stock_data = obb.equity.price.historical(symbol="TSLA", start_date="2023-01-01", provider="fmp")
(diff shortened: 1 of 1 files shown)


hjoaquim (approved):
This works perfectly. Another approach could be to use the results directly without converting; example: 

```
>>> stock_data = obb.equity.price.historical(symbol="TSLA", start_date="2023-01-01", provider="fmp")
 >>> obb.quantitative.normality(data=stock_data.results, target="close")
```

^ this is the approach used in the `technical` extension examples, should we follow to add consistency or, on the other hand, should we do it with the dataframe conversion for diversity?

IgorWounds:
> This works perfectly. Another approach could be to use the results directly without converting; example:
> 
> ```
> >>> stock_data = obb.equity.price.historical(symbol="TSLA", start_date="2023-01-01", provider="fmp")
>  >>> obb.quantitative.normality(data=stock_data.results, target="close")
> ```
> 
> ^ this is the approach used in the `technical` extension examples, should we follow to add consistency or, on the other hand, should we do it with the dataframe conversion for diversity?

Diversity and inclusion, lol :D

PR #6078, opened 2024-02-14
Add dates to /quantitative as needed.

jmaslek:
Also I ran the new black formatter
What?

Add dates to rolling functions

Why?

Because there werent dates in the functions

How?

Added dates to the functions

Who?

me

Where?

at home


jmaslek:
hmmmmm
Screenshot 2024-02-14 at 4 05 54 PM

IgorWounds:
@jmaslek Can you please push only the router and tests? The statics will pollute as it has extra things and we don't need the statics of this router.

jmaslek:
> @jmaslek Can you please push only the router and tests? The statics will pollute as it has extra things and we don't need the statics of this router.

my b

openbb_platform/extensions/quantitative/openbb_quantitative/quantitative_router.py
```diff
@@ -415,8 +432,10 @@ def sortino_ratio(
 
 
 @router.command(methods=["POST"])
-def skewness(data: List[Data], target: str, window: PositiveInt) -> OBBject[List[Data]]:
-    """Get Skewness.
+def skewness(
+    data: List[Data], target: str, window: PositiveInt, index: str = "date"
```
deeleeramone, line 436: Sharpe and Kurtosis have a default state for `window`, so should there be one here and Quantile?

openbb_platform/extensions/quantitative/openbb_quantitative/quantitative_router.py
```diff
@@ -494,12 +517,14 @@ def quantile(
...
 
-    df = basemodel_to_df(data)
+    df = basemodel_to_df(data, index=index)
     series_target = get_target_column(df, target)
 
     df_median = ta.median(close=series_target, length=window).to_frame()
     df_quantile = ta.quantile(series_target, length=window, q=quantile_pct).to_frame()
-    results = pd.concat([df_median, df_quantile], axis=1).dropna()
+    results = (
+        pd.concat([df_median, df_quantile], axis=1).dropna().reser_index(drop=False)
```
deeleeramone, line 526: `reser_index()`.  I guess codespell doesn't catch these kinds of spelling typos.

deeleeramone (commented):
It wouldn't be a terrible idea to have a check that verifies the length of the data index is longer than the window, otherwise Empty is returned as a silent fail.

openbb_platform/extensions/quantitative/openbb_quantitative/quantitative_router.py
jmaslek: surprised other linters doesnt pick it up

openbb_platform/extensions/quantitative/openbb_quantitative/quantitative_router.py
jmaslek: added

jmaslek:
> It wouldn't be a terrible idea to have a check that verifies the length of the data index is longer than the window, otherwise Empty is returned as a silent fail.

Threw a value error instead.  Nothing worse than not knowing why its returning empty.

IgorWounds (approved):

PR #6086, opened 2024-02-16
[BugFix] Fix Equity Integration Tests

deeleeramone:

    - Failing integration tests for API/Python Equity.

    - Fixes serialization errors because of bad value inputs into number fields.

    - Missing "allow_multiple_symbols"

    - No real impact.

4. **Testing Done**:

    - `133 passed, 3 skipped in 146.27s (0:02:26)`


IgorWounds (changes requested):
Hi, I see you use a very similar, if not identical, `def df_apply(data):` function. This should live in the helpers so that you can call it without having duplicate code.

IgorWounds (approved):

PR #6087, opened 2024-02-16
Update the quantitative extension to make more sense

jmaslek:

0. **QA Extension**:

    - This PR breaks up the quantitative extension into different sub-routers that make more sense.

    - Previously, I was not sure what to expect when I was using some functions.  We returned everything rolling, but I wanted to get a basic statistic of returns, not the rolling.
    - I also broke performance into a different router, because it doesnt make sense to have all these ratios if I am looking at economic data for example.  These would be better suited for a portfolio extension (maybe I'll do that next)
    - The other functions I am still not sure they belong, but I left them

    - I have broken the previous qa extension into sub routers that make more sense to me as a user.

    - There is no usage in Pro or PQN notebooks.  I just need to update some documentation

4. **Testing Done**:

    - I tested everything with a variety of inputs.  I also wrote some tests for my statistics functions.  

    - Does this make sense?


IgorWounds (approved):
To me this makes sense and makes it nicer for navigation.

jmaslek:
> To me this makes sense and makes it nicer for navigation. Just make sure to lint and add integration tests for this by using our framework.

How do we do 'toolkit' integration tests with the framework.  The integration test generator doesn't create the tests when I run that file since there is no fetcher

PR #6095, opened 2024-02-19
[BugFix] - Fix POST docstring examples

IgorWounds:

    - The POST commands should follow the same convention as other router commands when adding examples for consistency.

    - Moved POST command router examples from docstring to the command decorator.
    - Fixed a bug where the POST examples wouldn't be added to the static assets docstrings.

    - Will change the look of the router files and add `examples` to the `openapi.json`. Shouldn't change anything substantial as these extension aren't yet widely used.

4. **Testing Done**:

    - Ensured that the examples are indeed on the static assets as expected.


the-praxs (approved):
LGTM 🚀

PR #6099, opened 2024-02-20
[BugFix] Fixes to_df() where the date series contains multiple TZ-offsets.

deeleeramone:
Fixes `to_df()` where the date series contains multiple TZ-offsets.

    - `to_df()` was raising an error when the time series was TZ-aware and contained multiple UTC offsets.

    - Couldn't get around it by setting "index=None" because it would replace None with "date".

    - Changing `pd.to_datetime(df["date"])` to `df["date"].apply(to_datetime)`

    - Restore the TZ-aware datetime returned from Polygon as this was made naive to workaround this issue.

    - Allow `None` to be a valid choice in `to_df(index=None)`.

    - Ensure that all providers return the historical price time series data as ascending dates because that is how most downstream processes expect it.  The outputs from providers were mixed. 
 

    - No real impact other than bug squashing.

4. **Testing Done**:

    - Loading intraday data that is TZ-aware, starting before Nov 3.  

    - Use `to_df()` and then `to_df(index=None)` to check each provider is returning the same direction.


montezdesousa (changes requested):
LGTM

Can you pls add 1 or 2 unit tests to make sure this does not break again?

IgorWounds (approved):
Nice job! I'm in line with Diogo's request in adding a few unit tests.

montezdesousa (approved):

PR #6094, opened 2024-02-19
[Feature] - Platform V4 Markdown Generator V2

the-praxs:
# [Feature] - Platform V4 Markdown Generator V2
1. **Why**?

    - To improve the markdown generation by removing the use of docstrings of the static files
    - To remove dependency on various classes
    - To improve the code structure and readability

2. **What**?

    - Re-built the markdown generator from the ground up mainly using the `ProviderInterface` and `APIRouter` classes
    - Reduced imports of several packages and classes thereby reducing dependencies
    - Improved the code structure by creating functions for specific use cases
    - Improved code readability by using clear variable and function names, adding comments and docstrings wherever desired.

3. **Impact**

    - High Impact
    
    - Impact Analysis breakdown:
        - _UX_ = 0 (Developer use only)
        - _Performance_ = 3 (Reduced package imports)
        - _Compatibility_ = 5 (Re-built from scratch)
        - _Security_ = 3 (File System Operations, Extra JSON generation)
        - _Scalability_ = 4 (Scalable to software leveraging the Platform Python interface)
        - _Maintenance_ = 2 (Maintainance shifted towards the Platform core)
        - _Compliance_ = 0 (Compliant with the current license)
        - _Dependency_ = 5 (Reduced dependency on external packages and several internal classes)

    - _Total Impact Score_ = 22

4. **Testing Done**

    - Static built and validated for all the extensions and providers
    - Functions tested individually manually
    - Script tested for performance and edge cases manually
    - Generated files tested extensively by locally building the documentation files via `npm`

5. **Reviewer Notes**

    Testing procedure - 
    - Ensure you are on the latest version of the `develop` branch
    - Use the following command to install all the extensions and providers locally -
      ````bash
      python openbb_platform/dev_install.py -e
      ````

    - Go to the `website` directory and run the following command to generate the platform markdown files -
      ```bash
      python generate_platform_v4_markdown.py
      ```

    - Build the documentation for the generated packages using the following command -
      ```bash
      npm run dev
      ```

PR #6101, opened 2024-02-20
[Feature] - Create openbb assets folder

montezdesousa:

1. **Why**?

    - json files were mixed with code
    - we will add more metadata json files later like `reference.json` https://github.com/OpenBB-finance/OpenBBTerminal/pull/6094, so it will be contained here

2. **What**?

    - Create  `openbb/assets` folder to separate code from json or other files

```shell
openbb/
     assets/
             *.json
             *.[no_code_file]
     package/
             *.py
```

3. **Impact**:

    - PackageBuilder might break if paths are not changed properly
    - Auto build might break if paths are not changed properly
    - Unit tests rely of these files

4. **Testing Done**:

    - Install/uninstall new extension and `import openbb` to check build gets triggered
    - Manual build `openbb.build()`
    - Manual build single module `openbb.build("/economy")`


the-praxs (approved):
LGTM 🚀

PR #6107, opened 2024-02-21
[Feature] New endpoint `economy.markets.historical`

hjoaquim:

1. **Why**?:

    - Access to yields across the world - data needed for the OpenBB Terminal PRO.

2. **What**? :

    - Added new endpoint `economy.markets.historical`
    - New standard model (`MarketHistorical`)
    - New fetcher for Trading Economics (`TEMarketHistoricalFetcher`)

    - TBD

    > [!TIP]
    > Refer to the Impact Analysis confluence (internal) document for more information.

4. **Testing Done**:

    - TBD

    - TBD


openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/market_historical.py
```diff
@@ -0,0 +1,94 @@
...
+        query_str = get_querystring(query.model_dump(), ["symbol"])
+        urls = []
+
+        for symbol in symbols:
+            urls.append(f"{base_url}/{symbol}?{query_str}&c={api_key}")
+
+        async def callback(response: ClientResponse, _: Any) -> Union[dict, List[dict]]:
+            """Return the response."""
+            if response.status != 200:
+                raise RuntimeError(f"Error in TE request -> {await response.text()}")
```
deeleeramone, line 83: Because we are allowing multiple symbols, the failure of one symbol should communicate via warnings, raising an `EmptyDataError` if all fail.

We don't want to throw out the entire request because of one bad apple.

openbb_platform/extensions/economy/openbb_economy/markets/markets_router.py
```diff
@@ -0,0 +1,53 @@
...
+async def historical(
+    cc: CommandContext,
+    provider_choices: ProviderChoices,
+    standard_params: StandardParams,
+    extra_params: ExtraParams,
+) -> OBBject:
+    """
+    Get historical price data for a symbol within a market category.
+
+    The market categories available are exchange rates, stock market indexes,
```
deeleeramone, line 50: Should this functionality be split up into the existing router functions, like `index.price.historical`, `crypto.price.historical`, `currency.price.historical`?

jmaslek: I agree with darren.  This api is wacky, but we can get snapshots without a ticker (f'https://api.tradingeconomics.com/markets/index?c={api_key}' or for the specific symbol:

f'https://api.tradingeconomics.com/markets/historical/aapl:us,gac:com?c={api_key}'

So this should go into snapshot and historical for crypto/index/stock. This also probably warrants a new bond/government/snapshot  for  f'https://api.tradingeconomics.com/markets/bond?c={api_key}'

openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/market_historical.py
hjoaquim: Agreed.

openbb_platform/extensions/economy/openbb_economy/markets/markets_router.py
hjoaquim: That's a very fair point. I don't know tbh, although it covers those assets, it's not limited to them - and TE doesn't seem to offer any options for filtering.
Thoughts @jmaslek @minhhoang1023 ?

hjoaquim: The page was outdated when I 1st commented! I agree with that approach @jmaslek. Will restructure accordingly.
So, iiuc: 
   - TE Historical will serve crypto/index/equity historical.
   - TE Snapshot will serve economy/index snapshot. --> should we create a snapshot endpoint under economy for the bonds?

The expectation is that the user can then use (eg) `index.historical` to get historical data for any given bond?

PR #6111, opened 2024-02-22
[BugFix] Fix Company News

deeleeramone:
1. **Why**?:

    - Providers allowing multiple symbols did not have `__json_schema_extra__`

    - Default limit size was arbitrarily small that resulted in no data.

    - Every provider has a different field name for "images"

    - Can't use alternative lookup params (CUSIP, ISIN) from Benzinga because "symbol" is mandatory.

    - Allow multiple symbols for all providers.
    - All images are now returned to standardized field, "images"
    - Add pagination to Polygon fetcher.
    - Standardize start/end date with Polygon and simplify the model by using `__alias_dict__` instead of a private attribute.
    - Remove start/end date filter from yFinance as it is not doing anything.
    - Symbol made Optional to allow lookup by other params.

    - Doesn't change anything for Pro other than you can lookup news by ISIN/CUSIP now.
    - Providers with images now all return to the same field, "images"

4. **Testing Done**:

    - Recaptured test cassettes
    - Updated integration tests and verified they all pass.
    - Manually checking with multiple symbols and dates/limit combinations.


IgorWounds (approved):

IgorWounds:
@andrewkenreich are the changes to the standard model okay with you?

PR #6119, opened 2024-02-23
[BugFix] Fix Intraday Check in Charting `ta_class`

deeleeramone:

Fix for:

```
>       self.intraday = df_stock.index[-2].time() != df_stock.index[-1].time()
E       AttributeError: 'datetime.date' object has no attribute 'time'

../../obbject_extensions/charting/openbb_charting/core/plotly_ta/ta_class.py:186: AttributeError
```

    - Apply `to_datetime()` at the entry point in the `ta_class` in a way that allows the index to be tz-aware and contain daylight savings.

    - No real impact.

4. **Testing Done**:

    - The error does not occur now.

    - OBBject integration test passes.


openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
```diff
@@ -178,7 +178,7 @@ def to_chart(
         index = (
             data.index.name
             if has_data and isinstance(data, (pd.DataFrame, pd.Series))
-            else ""
```
hjoaquim, line 181: Do we need this? Empty string will also evaluate to `False`

openbb_platform/obbject_extensions/charting/openbb_charting/core/plotly_ta/ta_class.py
```diff
@@ -182,6 +182,12 @@ def __plot__(
         if not isinstance(indicators, ChartIndicators):
             indicators = ChartIndicators.from_dict(indicators or dict(dict()))
 
+        # Apply to_datetime to the index in a way that handles daylight savings.
```
hjoaquim, line 185: Would `df_stock.index = pd.to_datetime(df_stock.index)` suffice?

hjoaquim (changes requested):
How can I reproduce the issue?

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
deeleeramone: None is the valid setting for Index.

openbb_platform/obbject_extensions/charting/openbb_charting/core/plotly_ta/ta_class.py
deeleeramone: No, this is not compatible with series containing tz-aware timestamps with multiple offsets (daylight savings).

deeleeramone:
> How can I reproduce the issue?

Sorry, I guess I forgot that key piece of information! 

```
obb.equity.price.historical("MSFT", provider="yfinance", chart=True)
```

openbb_platform/obbject_extensions/charting/openbb_charting/core/plotly_ta/ta_class.py
jmaslek: I don't follow that.  The apply operation is the same as doing what henrqiue suggested:

```
one_way = pd.to_datetime(df_stock.index)

df_stock.loc[:, "date"]  = df_stock.index
df_stock["date"] = df_stock["date"].apply(pd.to_datetime)
df_stock.index = df_stock["date"]

all(one_way == df_stock.index)
Out[14]: True
```

deeleeramone: There's a subtle difference, `.apply()` will do it by element. You can see the PR here which cleared the raised error which was blocking `to_df()` from working when a TZ-aware index contained multiple UTC offsets (daylight savings): https://github.com/OpenBB-finance/OpenBBTerminal/pull/6099

hjoaquim: wouldn't that be the case if we were using the `apply()` method with more arguments (i.e., adding further manipulation)? since that's not the case, doing it row by row or vector-like should accomplish the same?
also, tried it out with `obb.equity.price.historical("MSFT", provider="yfinance", chart=True)` and both of them work well!

deeleeramone: Your example is not actually going to encounter the problem. yFinance does not return a tz-aware timestamp.

This handles the exception described above, specifically. In the charting extension, if the DataFrame being passed to `df_stock` has been run through `to_df()` or `basemodel_to_df()`, then it has already been handled before the entry point. This ensures that the edge case will not create an error at this point, even when data is not passed directly from the OBBject methods.

hjoaquim (approved):

PR #6122, opened 2024-02-25
[Feature] Add `form_13f` Endpoint to `equity.ownership`

deeleeramone:

Form 13F-HR submissions are newsworthy events, happening every calendar quarter. This is the time of year that Michael Burry remains relevant, or Berkshire Hathaway. 

    - `obb.equity.ownership.form_13f()`

    - Data is available quarterly, by calendar year, to 2013-06-30.

    - The `symbol` parameter accepts either, a stock ticker or CIK.

    - The `date` parameter returns the filing for the quarter ending the date entered, and overrides the `limit` param.

    - The `limit` parameter returns the most recent N number of reports.
 

    - Not a breaking change.

    - Improved data coverage.

4. **Testing Done**:

    - Integration and unit tests added.

    - Variety of symbols and CIKs with limit/date combinations.

Screenshot 2024-02-24 at 10 51 15 PM

Screenshot 2024-02-24 at 10 53 40 PM


openbb_platform/providers/sec/openbb_sec/models/form_13FHR.py
```diff
@@ -0,0 +1,97 @@
...
+            date = parse_13f.date_to_quarter_end(query.date.strftime("%Y-%m-%d"))
+            urls = [filings.loc[date]]
+
+        results = []
+
+        async def get_filing(url):
+            """Get a single 13F-HR filing and parse it."""
+
+            # Broadcast the source URL for each filing.
+            _warn(f"Source: {url}")
```
the-praxs, line 77: Why is an explicit warning raised here?

openbb_platform/core/openbb_core/provider/standard_models/form_13FHR.py
```diff
@@ -0,0 +1,88 @@
...
+        description="The date of the reporting period ended."
+    )
+    issuer: str = Field(description="The name of the issuer.")
+    cusip: str = Field(description="The CUSIP of the security.")
+    asset_class: str = Field(
+        description="The title of the asset class for the security."
+    )
+    security_type: str = Field(
+        description="The type of security"
+        + " 'SH' for shares. 'PRN' for principal amount. 'Call' or 'Put' for options.",
```
the-praxs, line 60: If these are the only options then it should be a `Literal`.

openbb_platform/core/openbb_core/provider/standard_models/form_13FHR.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    investment_discretion: str = Field(
+        description="The nature of the investment discretion held by the Manager."
+        + " One of: 'SOLE', 'DFND' (defined), 'OTR' (other).",
```
the-praxs, line 64: If these are the only options then it should be a `Literal`.

the-praxs (commented):
Thank you for adding another insightful endpoint to our Platform! 

I have left some comments that cater to code quality. Also, I would kindly request to add comments in the code to indicate why that particular LOC/snippet is needed. This will make it for other developers to understand them.

openbb_platform/providers/sec/openbb_sec/models/form_13FHR.py
deeleeramone: This provides the user with transparency for the exact source of the data. We can yeet it if it's not appropriate, but warnings are currently the only injection point to the OBBject from within the provider fetcher. If there is another way, please let me know!

openbb_platform/core/openbb_core/provider/standard_models/form_13FHR.py
deeleeramone: Thinking about it a bit more, this field doesn't really need to exist because it is broken out into the number of shares by discretion.

openbb_platform/providers/sec/openbb_sec/models/form_13FHR.py
the-praxs: I see. Maybe add it as a field under `extra`?

openbb_platform/core/openbb_core/provider/standard_models/form_13FHR.py
the-praxs: If it's able to give more granular control over data, then I think it should be added.

openbb_platform/providers/sec/openbb_sec/models/form_13FHR.py
deeleeramone: I don't want to serialize the URL for each row in the data, that's way too much bloat.

deeleeramone:
> Also, I would kindly request to add comments in the code to indicate why that particular LOC/snippet is needed. This will make it for other developers to understand them.

Thanks for the review, I have added some comments to the `parse_13f.py` file. Is that what you were looking for?

the-praxs:
> > Also, I would kindly request to add comments in the code to indicate why that particular LOC/snippet is needed. This will make it for other developers to understand them.
> 
> Thanks for the review, I have added some comments to the `parse_13f.py` file. Is that what you were looking for?

Yes. Thank you.

deeleeramone:
> Warning is not a good method to inform the user about the exact source of the data. The extra I am talking about is the OBBject extra attribute.

Do we have anything documented that outlines what this is and how to use it?

the-praxs:
> > Warning is not a good method to inform the user about the exact source of the data. The extra I am talking about is the OBBject extra attribute.
> 
> Do we have anything documented that outlines what this is and how to use it?

`Extra` -> https://docs.openbb.co/platform/development/developer-guidelines/obbject

deeleeramone:
> > > Warning is not a good method to inform the user about the exact source of the data. The extra I am talking about is the OBBject extra attribute.
> > 
> > 
> > Do we have anything documented that outlines what this is and how to use it?
> 
> `Extra` -> https://docs.openbb.co/platform/development/developer-guidelines/obbject

This is does not tell you how to inject something there, or anything at all other than a very basic definition of the field in the OBBject.

the-praxs:
> > > > Warning is not a good method to inform the user about the exact source of the data. The extra I am talking about is the OBBject extra attribute.
> > > 
> > > 
> > > Do we have anything documented that outlines what this is and how to use it?
> > 
> > 
> > `Extra` -> https://docs.openbb.co/platform/development/developer-guidelines/obbject
> 
> This is does not tell you how to inject something there, or anything at all other than a very basic definition of the field in the OBBject.



> > > > Warning is not a good method to inform the user about the exact source of the data. The extra I am talking about is the OBBject extra attribute.
> > > 
> > > 
> > > Do we have anything documented that outlines what this is and how to use it?
> > 
> > 
> > `Extra` -> https://docs.openbb.co/platform/development/developer-guidelines/obbject
> 
> This is does not tell you how to inject something there, or anything at all other than a very basic definition of the field in the OBBject.

Currently we don't have a way to do it. In the future we might if it warrants a strong need.

deeleeramone:
> Currently we don't have a way to do it. In the future we might if it warrants a strong need.

So until there is a way to do it, I will continue to use warnings, because that is better than nothing.  You can turn off warnings, or with `to_df()` you won't even see them. There is absolutely a strong need for being able to communicate metadata from the provider fetcher to the OBBject that doesn't involve applying it to every single instance of the results object.

the-praxs:
> > Currently we don't have a way to do it. In the future we might if it warrants a strong need.
> 
> So until there is a way to do it, I will continue to use warnings, because that is better than nothing. You can turn off warnings, or with `to_df()` you won't even see them. There is absolutely a strong need for being able to communicate metadata from the provider fetcher to the OBBject that doesn't involve applying it to every single instance of the results object.

Why must this particular endpoint among others should display the source url? What is the critical need for the user to know this? How is the user going to benefit from this information?

I believe that using warnings as a workaround method to display the source url for the data is not a good practice. Warnings are supposed to convey to the user about possible issues but here this is not an issue.

@IgorWounds @montezdesousa thoughts please?

montezdesousa:
I agree that warnings are a bad place to display that.

What's the practical use case for this feature at the moment? If it's not a common issue I'd wait to have a proper place to put it.

deeleeramone:
> I agree that warnings are a bad place to display that.
> 
> What's the practical use case for this feature at the moment? If it's not a common issue I'd wait to have a proper place to put it.

Metadata, I cannot stress this enough and have been saying it from the beginning. This is super practical, and extremely useful - i.e., FRED series. You need to know what the units are, along with other things related to the series and you absolutely do not want to serialize this across every row of data.

IgorWounds (approved):

the-praxs (approved):
LGTM 🚀

PR #6127, opened 2024-02-26
[Bug fix] - explicit multiple items docstring

montezdesousa:

1. **Why**?
    - Explicit docstring when multiple items allowed.

2. **What**?

    - Change docstrings

    - NA

4. **Testing Done**:

    - Confirm docstrings changed


openbb_platform/openbb/package/regulators_sec.py
```diff
@@ -30,7 +30,7 @@ def cik_map(
         symbol: Annotated[
             Union[str, List[str]],
             OpenBBCustomParameter(
-                description="Symbol to get data for. Multiple items allowed: fmp, intrinio, yfinance."
+                description="Symbol to get data for. Multiple items allowed for provider(s): fmp, intrinio, yfinance."
```
deeleeramone, line 33: This one went rogue!

deeleeramone: ah, I see. There is a need for some "generic" standard models that can be easily shared by providers when the input/output is very simple, without needing to add new standard models or mess with the repo.

montezdesousa: Makes sense, but we cannot cross reference models `SecCikMap` -> `EquityInfo`

PR #6129, opened 2024-02-26
[Bug fix] - Fix regulator cik map standard model

montezdesousa:

1. **Why**?

    - Wrong standard model used for query params
    - No standard model for data

2. **What**?

    - Create standard and use it

    - Fix docsring

4. **Testing Done**:

```python
from openbb import obb
obb.regulators.sec.cik_map("AAPL").to_df()
```

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/cik_map.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/cik_map.py
@@ -1,4 +1,4 @@
-"""Equity Info Standard Model."""
+"""Cik Map Standard Model."""
 
 from typing import Optional, Union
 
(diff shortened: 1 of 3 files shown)


openbb_platform/core/openbb_core/provider/standard_models/cik_map.py
```diff
@@ -0,0 +1,32 @@
+"""Equity Info Standard Model."""
```
jmaslek, line 1: title

jmaslek (approved):

jmaslek:
Feel like we could do something fun with the data where we actually keep it in the repo and update it every week or so.  Start our own master security list.  Food for thought.

PR #6130, opened 2024-02-26
[Feature] Pretty Pydantic `ValidationError` exceptions

the-praxs:
# [Feature] Pretty Pydantic `ValidationError` exceptions

1. **Why**?

    Make `ValidationError` more explicit for users

2. **What**?

    - Add `ValidationError` except block
    - Raise `OpenBBError` with custom message

3. **Impact**

    - Moderate Impact
    
    - Impact Analysis breakdown:
        - _UX_ = 3 (Reduced traceback call)
        - _Performance_ = 2 (Additional exception handling)
        - _Compatibility_ = 2 (New exception block)
        - _Security_ = 1 (Exception handling)
        - _Scalability_ = 2 (Adapts to the Platform changes)
        - _Maintenance_ = 1 (May need to check for errors)
        - _Compliance_ = 0 (Compliant with the current license)
        - _Dependency_ = 0 (No additional imports)

    - **Total Impact Score** = 11

4. **Testing Done**:

    Tested with `DEBUG_MODE` enabled and disabled


openbb_platform/core/openbb_core/app/static/utils/decorators.py
```diff
@@ -44,14 +44,36 @@ def exception_handler(func: Callable[P, R]) -> Callable[P, R]:
...
+                        f"input_value={error['input']}]\n"
+                    )
+                    error_info = f"    For further information visit {error['url']}\n"
+                    error_list.append(arg_error + error_details + error_info)
+
+                error_list.insert(0, validation_error)
+                error_str = "\n".join(error_list)
+                raise OpenBBError(
+                    f"\nType -> ValidationError \n\nDetails -> {error_str}"
+                ) from None
```
hjoaquim, line 72: imo, this should be `from e`, otherwise it will point to the decorator instead of the actual code that's breaking

openbb_platform/core/openbb_core/app/static/utils/decorators.py
```diff
@@ -44,14 +44,36 @@ def exception_handler(func: Callable[P, R]) -> Callable[P, R]:
...
+
+            # If the error is not a ValidationError, then it is a generic exception
             raise OpenBBError(
-                f"\nType -> {e.__class__.__name__}\n\nDetail -> {str(e)}"
+                f"\nType -> {e.original.original.__class__.__name__}\n\nDetail -> {str(e)}"
             ) from None
```
hjoaquim, line 77: same applies as above

hjoaquim (changes requested):
Looks good to me, but the comments.
I assume that suppressing the traceback will come on a different PR? If you end up adding the traceback suppression here, then the comments are irrelevant bc traceback will not be shown anyway

openbb_platform/core/openbb_core/app/static/utils/decorators.py
the-praxs: If we do `from e` then we are adding the context of the original exception which is wrapped by `OpenBBError`. To reduce the traceback call we use `from None` so that the context of the original exception is not used and so that is suppressed.

openbb_platform/core/openbb_core/app/static/utils/decorators.py
the-praxs: same answer as above.

the-praxs:
> Looks good to me, but the comments. I assume that suppressing the traceback will come on a different PR? If you end up adding the traceback suppression here, then the comments are irrelevant bc traceback will not be shown anyway

Thank you for the feedback. I need to introspect the exception object deeply to print the first occurrence of the traceback while the other frames are being suppressed.

openbb_platform/core/openbb_core/app/static/utils/decorators.py
hjoaquim: let's keep it that way until we find a good solution for the traceback suppression, then!

hjoaquim (approved):

PR #6134, opened 2024-02-27
[Docs] Platform Markdown Generator V2

the-praxs:
# [Docs] Platform Markdown Generator V2

1. **Why**?

    To check the Platform Markdown Generator V2 live

2. **What**?

    Create docs using the `generate_platform_v4_markdown.py` script

3. **Impact**

    High UX impact

4. **Testing Done**

    Tested live on the https://www.docs.openbb.co website

5. **Additional Notes**
    
    The test file was removed due to pytest issues.


deeleeramone (commented):
The only change is removing the test, shouldn't it be replaced by something for the new script?

the-praxs:
> The only change is removing the test, shouldn't it be replaced by something for the new script?

Tests were created for the functions used to generate the `reference.json` file. Since we are moving it to the `package_builder.py` as a new class, tests will be added for that file.

deeleeramone:
> > The only change is removing the test, shouldn't it be replaced by something for the new script?
> 
> Tests were created for the functions used to generate the `reference.json` file. Since we are moving it to the `package_builder.py` as a new class, tests will be added for that file.

I don't know what I'm supposed to be reviewing here.

the-praxs:
> > > The only change is removing the test, shouldn't it be replaced by something for the new script?
> 
> > 
> 
> > Tests were created for the functions used to generate the `reference.json` file. Since we are moving it to the `package_builder.py` as a new class, tests will be added for that file.
> 
> 
> 
> I don't know what I'm supposed to be reviewing here.

Updated docs I believe. Although the files are not committed and I am not sure if we are to do that.

the-praxs:
Closing this since it has served its purpose.

PR #6147, opened 2024-02-29
Fix get reference

montezdesousa:
^ Fix OBB.GET examples https://github.com/OpenBB-finance/OpenBBTerminal/pull/6145


the-praxs (approved):
LGTM 🚀

PR #6152, opened 2024-02-29
Improve the OECD requests

jmaslek:

0. **Title**:

    - When I implemented OECD I was pulling the entire database and caching it.  The files are huge so its like a 20-30 second call.  This makes the url more granular and implements a different cache idea that makes the key depend on the query params (ignoring dates because they arent jsonable and the filtering is a fast step anyways)

    - Change the URL to get specific fields instead of all data

    - Change the cache to have a key that depends on the input query params.

    - Should not have any impact especially since its not on pro yet.

4. **Testing Done**:

    - Trying the endpoint with many different types of 

Funcs done:

- [x] Unemployment
- [ ] LTIR
- [ ] STIR
- [ ] CLI


openbb_platform/providers/oecd/openbb_oecd/models/composite_leading_indicator.py
```diff
@@ -84,13 +84,22 @@ def extract_data(
         **kwargs: Any,
     ) -> Dict:
```
deeleeramone, line 85: Here the output is a `List[Dict]`, the mismatch will trigger a test failure.  The same applies to the Input of the `transform_data` method.

deeleeramone (changes requested):
The start/end date thing needs to be applied to the GDP functions as well. You could also add an "all" to these.

Generally, it's downloading much better now. I made sure to delete all the existing cache.

The test cassettes need to be recaptured and the types in the Fetcher need to be a `List[Dict]` instead of `Dict`.

jmaslek:
> The start/end date thing needs to be applied to the GDP functions as well. You could also add an "all" to these.

The GDP functions actually need to be updated from the old stats.oecd endpoints to the newer sdmx.  That is out of the scope here (those also don't touch cache) -- but is 100% on the roadmap.  The start/end are simple enough

openbb_platform/providers/oecd/openbb_oecd/models/composite_leading_indicator.py
jmaslek: should be all set now

openbb_platform/providers/oecd/openbb_oecd/models/gdp_forecast.py
```diff
@@ -94,11 +96,20 @@ def extract_data(
...
         data_df = data_df[data_df["country"] == query.country]
         data_df = data_df[["country", "date", "value"]]
-        return data_df.to_dict(orient="records")
+        data = data_df.to_dict(orient="records")
+        start_date = query.start_date.strftime("%Y-%m-%d")  # type: ignore
+        end_date = query.end_date.strftime("%Y-%m-%d")  # type: ignore
+        data = [
+            x
+            for x in data
+            if f"{x['date']}-01-01" >= start_date and f"{x['date']}-12-31" <= end_date
```
deeleeramone, line 105: This is funky.

Screenshot 2024-02-29 at 12 43 07 PM

jmaslek: You sent screenshot of STIR, but commented on gdp_forecast.  

This gdp was a quick hack because it was returning just a year and itll need to be updated anyways as stats. is being deprecated soon. Will try to figure why your screenshot isnt working

jmaslek: Aha its because it hasnt been validated yet.

```
{'country': 'ireland', 'date': '1984-02', 'value': 12.69}
```

This returns YYYY-mm so it cant exactly filter.  This means I need to get creative.  Hang tight.

deeleeramone: We can simplify the date validator, no need to regex it.

```
if "Q" in str(in_date):
    return to_datetime(in_date).to_period("Q").to_timestamp("Q").date()
if len(str(in_date) == 7:
    return pd.to_datetime(in_date).to_period("M").to_timestamp("M").date()
```

jmaslek: yeah thats smarter.

openbb_platform/providers/oecd/openbb_oecd/models/gdp_forecast.py
```diff
@@ -94,11 +96,18 @@ def extract_data(
         )
         data_df = data_df[data_df["country"] == query.country]
         data_df = data_df[["country", "date", "value"]]
+        data_df["date"] = data_df["date"].apply(helpers.oecd_date_to_python_date)
+        data_df = data_df[
+            (data_df["date"] <= query.end_date) & (data_df["date"] >= query.start_date)
+        ]
+        data_df["date"] = data_df["date"].apply(
+            lambda x: x.year
+        )  # Validator won't accept datetime.date?
```
deeleeramone, line 105: I think this is because the validator is expecting a string representation. Although, because you have already conformed the date field in the extract process, the validator is now redundant.

deeleeramone (approved):

openbb_platform/providers/oecd/openbb_oecd/models/gdp_forecast.py
jmaslek: I thought that too but it accepts str|date so it was unclear what was actually going wrong

PR #6154, opened 2024-03-01
[Feature] - refactor examples

montezdesousa:
1. **Why**?

    - Examples were written in python but API clients are not bound to python language
    - Excel wants to use the examples, parsing python code into Excel examples would be prone to failure and hard to maintain since you have a lot ways to call the same function in python
    - Examples are inconsistent, sometimes with `to_df`, other times accessing internal obbject attributes, etc. This will only get worse with time if we don't impose any restrictions on how the example should look
    - There's no policy on which types of examples we should have for each endpoint added

2. **What**?

    - Creates 2 example models `APIEx` and `PythonEx`
    - These must implement a `to_python` method that renders the example into python code (pacakge_builder.py then calls this to create docstrings)
    - `APIEx` are more structured aiming to be language agnostic - provide less freedom
    - `PythonEx` give more freedom to create complex examples
    - When a given provider is not installed, its example will be excluded from `openapi.json` and python docstrings

3. **Impact**:

    - For the general Python/API users this will have no impact
    - Platform documentation: `generate_platform_markdown.py` needs a few tweaks to get the examples @the-praxs 
    - Excel documentation: `generate_excel_markdown.py` will start using these examples when this reaches production

4. **Testing Done**:

    - The examples should look almost the same as before
    - @IgorWounds will add unit tests to follow a policy with 2 rules:
    1. at least one example for all endpoints
    2. at least one API example with the required parameters for GET endpoints


openbb_platform/extensions/economy/openbb_economy/economy_router.py
```diff
@@ -145,13 +166,20 @@ async def money_measures(
...
-        "#### Demographics for the statistics are selected with the `age` and `sex` parameters. ####",
-        "obb.economy.unemployment(",
-        'country="all", frequency="quarterly", age="25-54"',
-        ').to_df().pivot(columns="country", values="value")',
+        APIEx(parameters={"provider": "oecd"}),
+        APIEx(
+            parameters={"country": "all", "frequency": "quarterly", "provider": "oecd"}
+        ),
+        APIEx(
+            description="Demographics for the statistics are selected with the `age` and `sex` parameters.",
```
IgorWounds, line 175: We don't show the sex param here 😉

IgorWounds (approved):

PR #6156, opened 2024-03-01
[Docs] Cleanup README Files

deeleeramone:

    - Manually maintaining individual README files is prone to error and being out-of-sync.

    - Refer to the documentation website so we can maintain a single source of truth.

    - We may need to find a better way to generate the content that is visible on PyPI.

4. **Testing Done**:

    - N/A


hjoaquim (approved):
100% agree. but, do we need the changes on `sdk_helpers.py` and `controller_doc_classes.py` - seems completely unrelated to the PR?

hjoaquim:
I believe this is now outdated, after the recent changes on the docs.

PR #6157, opened 2024-03-01
Adding OECD Endpoints (CPI + Share Price Index)

jmaslek:
1. **Why**?:

    - Expand macroeconomic data and provider coverage.
    
- [x] CPI
- [x] Share Price Index 

2. **What**?:

    - Added CPI from oecd which doesn't require an api key - but the input parameters are a bit more confusing.  Nothing like 29 possible expenditures.
    - Edited the standard model
    - Expanded cpi -> consumer_price_index in files
    - Added `share_price_index` standard model and Economy route.

3. **Impact** :

    - I did change the standard model to be compatible between fred + oecd, but the default values should work.
    - New standard model and provider endpoint to maintain. 

4. **Testing Done**:

    - Unit/Integration tests.
    - CLI 
    - Various combinations of countries.


openbb_platform/core/openbb_core/provider/standard_models/share_price.py
```diff
@@ -0,0 +1,40 @@
...
+
+class SharePriceData(Data):
+    """Share Price Data."""
+
+    date: Optional[dateType] = Field(
+        default=None, description=DATA_DESCRIPTIONS.get("date")
+    )
+    value: Optional[float] = Field(
+        default=None,
+        description="Interest rate (given as a whole number, i.e 10=10%)",
```
deeleeramone, line 35: Description here is from something else.  Is the field going to represent a price? 

Maybe there's a better name for this standard model? "GlobalSharePriceIndex" or something more intuitive than "SharePriceData" which could easily be construed for data related to TSLA or whatever else.

jmaslek: looks llike its not a requirement anymore, so may yank it

deeleeramone:
I have formatted the two providers to be uniform, and added unit/integration tests.

Screenshot 2024-05-28 at 11 19 43 AM

IgorWounds (approved):
Nice!

Issue #6163, opened 2024-03-04
[Bug] Installing openbb-core also installs ruff, prevents upgrade to ">=0.2.0


samueldg:
**Describe the bug**

The `openbb-core` package, required to install `openbb`, defines `ruff` as a dependency in the main group, and not the dev group.

Compare:

https://github.com/samueldg/OpenBBTerminal/blob/47541d4c957d7ab366e3b5f07615b4a7a9916c7e/openbb_platform/core/pyproject.toml#L25

with the (correct) top-level `pyproject.toml`:

https://github.com/samueldg/OpenBBTerminal/blob/47541d4c957d7ab366e3b5f07615b4a7a9916c7e/pyproject.toml#L178

Concretely, this means that you end up with the linter/formatter installed even if you don't want to use it, but more importantly it means that you can't upgrade it to the latest available version (`0.3.0`) because it clashes with the version rule in the `openbb-core` dependencies.

**To Reproduce**

1. Create the following `pyproject.toml` in a new directory:
    
    ```toml
    [tool.poetry]
    name = "openbb-core-test"
    version = "0.1.0"
    description = ""
    authors = ["Test <test@example.com>"]
    
    [tool.poetry.dependencies]
    python = "^3.11"
    openbb-core = "^1.1.2"
    
    [tool.poetry.group.dev.dependencies]
    ruff = "^0.3.0"
    
    [build-system]
    requires = ["poetry-core"]
    build-backend = "poetry.core.masonry.api"
    ```
2. Try to `poetry install`
3. Get the following error

    ```txt
    Because openbb-core (1.1.2) depends on ruff (>=0.1.6,<0.2.0)
     and no versions of openbb-core match >1.1.2,<2.0.0, openbb-core (>=1.1.2,<2.0.0) requires ruff (>=0.1.6,<0.2.0).
    So, because openbb-core-test depends on both openbb-core (^1.1.2) and ruff (^0.3.0), version solving failed.
    ```

**Desktop (please complete the following information):**

- OS: Mac Sonoma 14.3
- Python version: Tried on 3.11 and 3.10

deeleeramone:
Hi there, 

Ruff and Black are both used in the PackageBuilder class, which is responsible for creating the static assets required to run the Python interface.

[...]

PR #6164, opened 2024-03-04
[BugFix] Move ruff to dev dependencies

samueldg:
Like pytest, ruff isn't required to run the code, only in development.

(Similar to the main `pyproject.toml`, which has `ruff` in the same dependency group.)

Closes https://github.com/OpenBB-finance/OpenBBTerminal/issues/6163


samueldg:
Closing per https://github.com/OpenBB-finance/OpenBBTerminal/issues/6163#issuecomment-1977694708

Fixing this would require a different approach.

PR #6165, opened 2024-03-04
[Feature] - Excel help page

montezdesousa:
1. **Why**?

    - Need for help page to suggest fixes for excel issues

2. **What**?

    - Creates https://docs.openbb.co/excel/help

3. **Impact**:

    - NA

4. **Testing Done**:

    - `cd website` > `npm start` and check how it looks

### DIFF
--- a/website/content/excel/data-connectors.md
+++ b/website/content/excel/data-connectors.md
@@ -1,6 +1,6 @@
 ---
 title: Data connectors
-sidebar_position: 4
+sidebar_position: 3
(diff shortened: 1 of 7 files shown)


DidierRLopes (approved):
I would also add here something like

"Cannot retrieve the data that you added to the backend of Pro on Excel" -> "this is because you are not running https on Safari or Mac.. blabla".

This helps users debug this known issue while at the same time also highlighting this feature.

PR #6169, opened 2024-03-06
[Feature] Custom choices

hjoaquim:
1. **Why**?:

    - Platform clients need more info on the available choices for certain parameters, which can't use `Literal` directly.

2. **What**?:

    - Adds a way of custom adding choices to the arguments so that Platform clients can leverage this info.
       1. Adds `OpenBBCustomChoices` which is then used on the statics assets to store the choice information.
       2. By leveraging `json_schema_extra` this info automatically is set up on the `openapi.json`.

    - Minor change in the UX, only perceivable when navigating to the command implementation (python) or the `openapi.json`. 
    - Also might introduce some maintenance impact as adds a "new" way of adding choices.
    - Impact score: 5

<details>
    User Experience (UX) Impact: 1
    Performance Impact: 0
    Compatibility Impact: 1
    Security Impact: 0
    Scalability Impact: 1
    Maintenance Impact: 2
    Compliance Impact: 0
    Dependency Impact: 0
</details>

4. **Testing Done**:

    - Running the `PackageBuilder`
    - Running the Rest API and navigating to `/openapi.json`


openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -500,7 +501,15 @@ def get_extra(field: FieldInfo) -> dict:
...
-            return getattr(field_default, "json_schema_extra", {})
+            keys_to_remove = ["choices"]
+
+            json_schema_extra = {
+                k: v
+                for k, v in getattr(field_default, "json_schema_extra", {}).items()
+                if k not in keys_to_remove
+            }
+
+            return json_schema_extra
```
montezdesousa, line 512: Here we can avoid the iterations

```python
json_schema_extra = getattr(field_default, "json_schema_extra", {})
json_schema_extra.pop("choices", None)
return json_schema_extra
```

montezdesousa (approved):

openbb_platform/core/openbb_core/app/static/package_builder.py
hjoaquim: Was thinking that more keys could be added to `keys_to_remove` in the future, but yes we can do that and eventually to smt like this if needed!

IgorWounds (approved):

PR #6172, opened 2024-03-06
[Bugfix] - Lower case validator in standard models

montezdesousa:
1. **Why**?

    - Allow users to send anycase inputs to case insensitive fields, like `sort: Literal["asc", "desc"]` should support `"Asc"`

2. **What**?

    - Create `to_lower` validator to include standard model fields that can handle this feature
    - Renamed some `upper_symbol` validators to `to_upper` for consistency across files and future use cases besides `symbol`

3. **Impact**:

    - If there are provider models that don't support lower case fields they will break, so I made sure only included literals or obvious non breaking fields like `sort` or `period`

4. **Testing Done**:

    - Unittests

    - Gradually this should be extended to provider fields, but will keep this contained for this PR

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/equity_historical.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/equity_historical.py
@@ -37,7 +37,7 @@ class EquityHistoricalQueryParams(QueryParams):
     @field_validator("symbol", mode="before", check_fields=False)
     @classmethod
     def to_upper(cls, v: str) -> str:
-        """Convert symbol to uppercase."""
+        """Convert field to uppercase."""
         return v.upper()
 
     @field_validator("interval", mode="before", check_fields=False)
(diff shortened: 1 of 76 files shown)


openbb_platform/core/openbb_core/provider/standard_models/equity_historical.py
```diff
@@ -36,10 +36,16 @@ class EquityHistoricalQueryParams(QueryParams):
 
     @field_validator("symbol", mode="before", check_fields=False)
     @classmethod
-    def upper_symbol(cls, v: str) -> str:
-        """Convert symbol to uppercase."""
+    def to_upper(cls, v: str) -> str:
+        """Convert field to uppercase."""
         return v.upper()
 
+    @field_validator("interval", mode="before", check_fields=False)
```
deeleeramone, line 43: We have "1m" and "1M" for intervals, where upper case M = Month and lower case m = minute.

montezdesousa: Thanks for raising that, I'll remove interval

IgorWounds (approved):

deeleeramone (approved):

PR #6177, opened 2024-03-10
[Enhancement] Improve Intrinio Unusual Options

deeleeramone:
1. **Why**?:

    - The current target endpoint had only a parameter for "symbol", it was returning no results today directly from the Intrinio API.

    - Targets a different Intrinio endpoint "intraday", which is parameterized but very difficult to get a complete data set for a single date.

    - To overcome some of the challenges of the data retrieval:
        - `start_date`/`end_date` are valid only when a symbol is supplied, else `start_date` is the effective "target date".
          - enforced via `model_validator`.
        - pagination occurs at 1000 records, but a typical day has between 50-80K.
          - default "limit" is set at 100K.
          - absolute max limit is 1.25M, based on outlier dates. 

    - Improved UX and data mining capability.
    - Using max ceiling for "limit" will take > 1 min.
    - Using without supplying "symbol" will result in a large number of API calls.

4. **Testing Done**:

    - `obb.derivatives.options.unusual(start_date="2023-11-17")`
    - `obb.derivatives.options.unusual(start_date="2023-11-20")`
    - `obb.derivatives.options.unusual(start_date="2023-11-20", limit=1200000)` 
    - `obb.derivatives.options.unusual(start_date="2023-11-20", symbol="AAPL")`


IgorWounds (approved):

deeleeramone:
Unclear why the test is failing, it's definitely a valid ISO string:

```
from datetime import datetime

v = "2023-11-20T15:45:05.837Z"
datetime.fromisoformat(v)
```

```
datetime.datetime(2023, 11, 20, 15, 45, 5, 837000, tzinfo=datetime.timezone.utc)
```

PR #6191, opened 2024-03-11
[BugFix] Update Nasdaq Econ Calendar Headers

deeleeramone:
1. **Why**?:

    - While doing some testing in Colab, I noticed the economic calendar from Nasdaq wasn't returning, but all the other calendars are. Locally, was fine, but not on Colab. 

2. **What**? :

    - It was using a different schema for the request headers, changed it to be like the other ones.

3. **Impact** :

    - N/A

4. **Testing Done**:

    - Could only test locally, will not know for certain until a nightly build, but still works locally!


the-praxs (approved):
LGTM 🚀

PR #6211, opened 2024-03-14
[Enhancement] Sync `reference.json` and Platform function docstring

the-praxs:
# [Enhancement] Sync `reference.json` and Platform function docstring

1. **Why**? 

    Consistent information between `reference.json` and package docstring

2. **What**?

    - `"None"` is now saved as `None` type in `QueryParams` and `Data` fields to be leverage on the Platform CLI
    - Field types with Pydantic models are processed to show only the model name resulting in cleaner type
    - `OBBject` return data fields are now stored as a list of dictionaries in the `reference.json` for consistency with the `data` and `parameters` fields in the `reference.json`. Updated logic is used to create the markdown content in the `create_reference_markdown_returns_section` function.
    - Fixed provider models to reflect correct field type and description
    - Updated static to reflect the new changes

3. **Impact** 

    - Minor impact on docs, major impact in CX/DX
    - `reference.json` is not backwards compatible with the previous versions

4. **Testing Done**

    - Static generated and tested locally 
    - Website docs generated and tested locally

### DIFF
--- a/openbb_platform/core/openbb_core/app/static/package_builder.py
+++ b/openbb_platform/core/openbb_core/app/static/package_builder.py
@@ -939,11 +939,10 @@ class DocstringGenerator:
     def get_OBBject_description(
         results_type: str,
         providers: Optional[str],
-        target: Literal["docstring", "website"] = "docstring",
     ) -> str:
         """Get the command output description."""
         available_providers = providers or "Optional[str]"
-        indent = 2 if target == "docstring" else 0
+        indent = 2
 
         obbject_description = (
             f"{create_indent(indent)}OBBject\n"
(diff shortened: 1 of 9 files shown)


openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -915,19 +927,14 @@ def get_field_type(
                 .replace("pydantic.types.", "")
                 .replace("datetime.datetime", "datetime")
                 .replace("datetime.date", "date")
+                .replace("datetime.time", "time")
```
hjoaquim, line 930: For future reference: we need to be careful with these things, since it's not a Python type, the CLI might break trying to parse unexpected types.

hjoaquim (changes requested):
All seems good - do we need to test something specifically on the docs? Also, we need to re-generate the `reference.json` so changes are reflected.
Can you update the PR description as well pls?

hjoaquim (approved):

openbb_platform/core/openbb_core/app/static/package_builder.py
the-praxs: I think that it should have the complete type i.e. as `datetime.<type>`. Thoughts @montezdesousa ?

PR #6223, opened 2024-03-14
[BugFix] Fix incorrect rendering of examples using ReferenceGenerator

the-praxs:
# [BugFix] Fix incorrect rendering of examples using ReferenceGenerator

1. **Why**?

    - Fixes the `ReferenceGenerator.get_endpoint_examples` function to fix incorrect rendering of examples that would have reflected on the docs website
    - Fixes an issue with the `interval` field for `/equity/price/historical` shown incorrectly as `str` instead of a `Literal` for `fmp` provider

2. **What**?

    - Removes the `interval` field from `EquityHistorical` standard model to reflect correct field information 
    - Modifies the `DocstringGenerator.generate` function to work with both `docstring` and `website`
    - Updates the static

3. **Impact**

    - Impact Score = 11 (Moderate)
    - Code refactored to get correct examples for docstring and website
    - Syncs correctness in the platform package static and website docs

4. **Testing Done**

    Tested locally


hjoaquim (changes requested):
This doesn't seem correct on the docs - left is deployed docs and right is this branch:
image

Also, why are we removing `interval` from the standard model? It's clearly a standard query parameter - it happens to be in almost all the provider specific models.

the-praxs:
> This doesn't seem correct on the docs - left is deployed docs and right is this branch:
> 
> image
> 
> 
> 
> Also, why are we removing `interval` from the standard model? It's clearly a standard query parameter - it happens to be in almost all the provider specific models.

The docs on the right are from the old generated files. You need to run the Platform Markdown Generator V2 to generate the new files then start the server to check.

`interval` is a standard parameter undoubtedly but since it's being overloaded in provider models, it's not reflected in the package asset correctly and thus in the docs.

hjoaquim (commented):
Thank you for the clarification - all looks good then - we need to address the standardization thingie tho!

hjoaquim (approved):

deeleeramone:
> `interval` is a standard parameter undoubtedly but since it's being overloaded in provider models, it's not reflected in the package asset correctly and thus in the docs.

We could also apply this to "period" for Fundamentals/Financial statements.

the-praxs:
Closed since PR scope has changed. New PR #6227

PR #6243, opened 2024-03-21
[Enhancement] Improve Charting Extension Plots and Indicators

deeleeramone:
1. **Why**?:

    - It is currently way too easy to break it, or draw a terrible chart.

    - API usage is very limited because CommandRunner is not passing through kwargs.

- Revamps Historical Prices & Indicators
  - Does not attempt to draw a candlestick chart when it is not able to do so, or not desired.
  - Determines if multiple symbols are in the `results` object and draws the chart accordingly.
    - If two symbols exist, it will place them on separate Y-axis.
    - If more than two symbols exist, it will default to cumulative returns and draw all on the same axis.
- Adds kwargs for controlling the chart creation, relevant to the specific endpoint returned, accessible from the OBBject:

```python
data = obb.equity.price.historical("AAPL", provider="yfinance")
data.charting.get_params()
```

```
EquityPriceHistoricalChartQueryParams

    Parameters
    ----------

    data (Union[Data, list[Data], NoneType]):
        Filtered versions of the data contained in the original `self.results`.
        Columns should be the same as the original data.
        Example use is to reduce the number of columns, or the length of data, to plot.

    title (Union[str, NoneType]):
        Title of the chart.

    target_column (Union[str, NoneType]):
        The specific column to target.
        If supplied, this will override the candles and volume parameters.

    multi_symbol (bool):
        Flag to indicate whether the data contains multiple symbols.
        This is mostly handled automatically, but if the chart fails to generate try setting this to True.

    same_axis (bool):
        If True, forces all data to be plotted on the same axis.

    normalize (bool):
        If True, the data will be normalized and placed on the same axis.

    returns (bool):
        If True, the cumulative returns for the length of the time series will be calculated and plotted.

    candles (bool):
        If True, and OHLC exists, and there is only one symbol in the data, candles will be plotted.

    heikin_ashi (bool):
        If True, and `candles=True`, Heikin Ashi candles will be plotted.

    volume (bool):
        If True, and volume exists, and `candles=True`, volume will be plotted.

    indicators (Union[ChartIndicators, dict[str, dict[str, Any]], NoneType]):
        Indicators to be plotted, formatted as a dictionary.
        Data containing multiple symbols will ignore indicators.
        Example:
            indicators = dict(
                sma=dict(length=[20,30,50]),
                adx=dict(length=14),
                rsi=dict(length=14),
            )
```

- Similar approach for the Indicators parameters:

Screenshot 2024-03-27 at 11 56 25 PM

```python
data.charting.indicators().rsi.model_dump_json()

>>> '{"length":14,"scalar":100.0,"drift":1}'
```

- Edits `_chart` OBBject class method to allow a `chart_params` kwarg to be passed on command execution.
  - Do not raise waning when "chart_params" is an extra_param.
  - This makes it possible for the API to access all available kwargs for each endpoint supported by the charting extension.
  - In the Python interface, you don't have to draw the chart first before "redrawing" it with parameters.

- Mirrors `equity.price.historical` to all the other `price.historical` endpoints - ETF, Index, Currency, Crypto.

- Fixes most weirdness that would happen when flipping light/dark mode.

- Bug fixes for Indicators in general.

- Fixes a Type issue in the Technical router where `scalar` should be a float.

- Add utility/fallback functions for generic chart creation from any data.
  - Line/Candles
  - Bar

- Add to charting router:
  - equity.price.performance
  - etf.price.performance
  - etf.holdings

    - Does not change anything for users without the charting extension installed.
    - Improved out-of-the-box charts that are much harder to break and offer more flexibility to the end-user.
    - Improved ability to refine the chart without directly manipulating the Plotly Figure object.
    - REST API compliant parameterized chart creation directly from request.

4. **Testing Done**:

    - Load single tickers, and lists of symbols. Draw the chart and then redraw with the `obbject.charting.to_chart()` methods.
    - Switch between light/dark mode, on the chart and in user_settings.
    - Add integration tests for charting router additions.

6. **Any other information**:

    - Documentation is all missing, but it was already.

```python
obb.equity.price.historical("SPY,QQQ,JNK", provider='yfinance', chart=True).show()
```

Screenshot 2024-03-20 at 5 26 36 PM

```python
obb.equity.price.historical("QQQ", provider='yfinance', chart=True, chart_params=dict(indicators=dict(rsi=dict(length=14)))).show()
```

Screenshot 2024-03-20 at 5 33 37 PM

```
obb.equity.price.historical("QQQ", provider='yfinance', chart=True, chart_params=dict(heikin_ashi=True)).show()
```

Screenshot 2024-03-20 at 9 42 58 PM

```
obb.equity.price.performance("SPY,XLB,XLC,XLY,XLU,XLK", provider="finviz", chart=True, chart_params=dict(orientation='h')).show()
```

Screenshot 2024-03-28 at 10 51 05 AM


hjoaquim:
@deeleeramone this is looking great! 
Should this PR have the draf tag removed now? It's already a decent amount of files being changed!

deeleeramone:
> @deeleeramone this is looking great! Should this PR have the draf tag removed now? It's already a decent amount of files being changed!

Probably, yes.

openbb_platform/obbject_extensions/charting/openbb_charting/query_params.py
```diff
@@ -103,3 +218,544 @@ class TechnicalConesChartQueryParams(QueryParams):
...
+    show: bool = Field(
+        default=True,
+        description="Show the support and resistance lines.",
+    )
+
+
+class IndicatorsParams(QueryParams):
+    """Indicators Query Params."""
+
+    sma: SMAIndicatorsQueryParams = Field(
```
hjoaquim, line 664: As commented above, I think it's a lot of overhead to instantiate 2 objects of the same class only to inform the user about the available indicators.

openbb_platform/obbject_extensions/charting/openbb_charting/query_params.py
```diff
@@ -1,47 +1,167 @@
 """Charting Extension Query Params."""
 
-from typing import List, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from openbb_core.provider.abstract.data import Data
 from openbb_core.provider.abstract.query_params import QueryParams
 from pydantic import Field
 
+from openbb_charting.core.to_chart import ChartIndicators
```
hjoaquim, line 9: I love the idea of having everything really explicit and as pydantic models!
But, I believe the chain of inheritance could be a lot simplified, example:

`BaseModel -> QueryParams -> BaseQueryParams -> IndicatorsQueryParams -> MAIndicatorsQueryParams -> SMAIndicatorsQueryParams`

`BaseModel -> QueryParams -> BaseQueryParams -> ChartQueryParams -> EquityPricePerformanceChartQueryParams -> EtfPricePerformanceChartQueryParams`

> imo, bad idea to mix equity and etf things, as discussed in previous situations

Do we need 4/5 levels of inheritance?

As a suggestion, I think this could be enough (1/2 levels of inheritance):

`BaseModel -> MAIndicatorsQueryParams -> SMAIndicatorsQueryParams`

openbb_platform/obbject_extensions/charting/openbb_charting/query_params.py
```diff
@@ -103,3 +218,544 @@ class TechnicalConesChartQueryParams(QueryParams):
...
+class SRLinesIndicatorsQueryParams(IndicatorsQueryParams):
+    """Support and Resistance Lines Indicators Query Params."""
+
+    show: bool = Field(
+        default=True,
+        description="Show the support and resistance lines.",
+    )
+
+
+class IndicatorsParams(QueryParams):
```
hjoaquim, line 661: Indicators should be validated against `openbb_charting.core.plotly_ta.data_classes.TAIndicator.name`: which contains the list of available indicators.

hjoaquim (changes requested):
This is an amazing and super valuable progress to our charting extension - the progess this PR adds is invaluable.

I've added some comments asking for clarification and suggestion on things I believe we could improve - always open for iteration and discussion on those!

Last comment that didn't really had a place to be written in the changed files:

   - I think `chart_params` should to be added to the `PackageBuilder` the same way `chart` is, otherwise will be a bit hidden.

openbb_platform/core/openbb_core/app/command_runner.py
```diff
@@ -293,15 +302,42 @@ async def _command(
...
+        if "extra_params" in kwargs:
+            try:
+                chart_params = kwargs["extra_params"].__dict__.get("chart_params", {})
+            except AttributeError:
+                chart_params = kwargs["extra_params"].get("chart_params")
+        if "chart_params" in kwargs:
+            chart_params.update(kwargs.pop("chart_params", {}))
+        if "kwargs" in kwargs:
+            _kwargs = kwargs.pop("kwargs", {})
+            chart_params.update(_kwargs.get("chart_params", {}))
```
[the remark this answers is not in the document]
deeleeramone: There is a difference between GET and POST params that are passed internally - you don't get CommandContext -  and then there is the undetermined future of the ability to pass **kwargs.

openbb_platform/core/openbb_core/app/command_runner.py
```diff
@@ -350,7 +386,7 @@ async def _execute_func(
                 obbject._standard_params = kwargs.get("standard_params", None)
 
                 if chart and obbject.results:
-                    cls._chart(obbject, **kwargs)
+                    cls._chart(obbject, warning_list, **kwargs)
```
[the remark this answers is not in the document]
deeleeramone: This is responsible for passing the metadata to the chart.

openbb_platform/core/openbb_core/app/command_runner.py
```diff
@@ -293,15 +302,42 @@ async def _command(
...
+                getattr(metadata[0], "message", "{}")
+                if metadata and isinstance(metadata, list) and len(metadata) > 0
+                else "{}"
+            )
+            metadata = json.loads(str(message)) if message else {}
+        except json.JSONDecodeError:
+            metadata = {}
+
+        obbject.charting.show(  # type: ignore
+            render=False, metadata=metadata, **chart_params, **kwargs
```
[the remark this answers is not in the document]
deeleeramone: unpacking **chart_params and **kwargs together essentially flattens them all for the downstream process.  **kwargs encompasses everything passed, so as far as the function signature is concerned **chart_params are **kwargs.

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
```diff
@@ -44,19 +57,18 @@ def __init__(self, obbject):
...
-        from openbb_core.app.model.obbject import OBBject
 
         self._obbject: OBBject = obbject
         self._charting_settings = ChartingSettings(
             user_settings=self._obbject._user_settings,  # type: ignore
             system_settings=self._obbject._system_settings,  # type: ignore
         )
-        self._handle_backend()
 
     @classmethod
```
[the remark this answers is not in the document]
deeleeramone: I removed it here so that the backend is only started when a chart is about to be made. Otherwise it was creating it whenever an OBBject is initialized.

openbb_platform/obbject_extensions/charting/openbb_charting/query_params.py
deeleeramone: Inheriting from BaseModel is not going to cut it here.

- `EquityPricePerformanceChartQueryParams -> EtfPricePerformanceChartQueryParam`
  - This is exactly what needs to happen. For all purposes, there is no difference between the two other than getting a new name and docstring.

`ChartParams` and `IndicatorsQueryParams` do not have the same fields but are both children of QueryParams. 

`BaseQueryParams` exists so that the `__repr__` method is inherited everywhere I need it.

`IndicatorsParams(QueryParams)` exists like this because it should not inherit from `BaseQueryParams`.

openbb_platform/core/openbb_core/app/command_runner.py
```diff
@@ -281,8 +288,10 @@ async def _command(
         cls,
         func: Callable,
         kwargs: Dict[str, Any],
+        show_warnings: bool = True,  # pylint: disable=unused-argument   # type: ignore
     ) -> OBBject:
         """Run a command and return the output."""
+
```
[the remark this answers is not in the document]
deeleeramone: This is inconsistent across everything.

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
```diff
@@ -44,19 +57,18 @@ def __init__(self, obbject):
...
     def indicators(cls):
         """Returns a list of the available indicators."""
-        return ChartIndicators.get_available_indicators()
+        # return ChartIndicators.get_available_indicators()
+        return IndicatorsParams()
```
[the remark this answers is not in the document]
deeleeramone: I don't see how this is helpful, practically or visually. As detailed above, inheriting from BaseModel does not do the job that needs to be done.

```
{'sma': FieldInfo(annotation=SMAIndicatorsQueryParams, required=False, default=SMAIndicatorsQueryParams(length=50, offset=0), description='Simple Moving Average Indicators Query Params.'),
 'ema': FieldInfo(annotation=EMAIndicatorsQueryParams, required=False, default=EMAIndicatorsQueryParams(length=50, offset=0), description='Exponential Moving Average Indicators Query Params.')}
```

The output of `IndicatorsParams.model_fields` looks like this:

Screenshot 2024-04-03 at 9 32 31 AM

It's not readable. I want this because it clearly shows exactly what you need to know and nothing else:

Screenshot 2024-04-03 at 9 34 19 AM

The class needs to be initialized for the `__repr__` to generate. Even if `__repr__` == `__doc__` it still needs to be initialized to extract the information required to replace `__doc__`.

openbb_platform/obbject_extensions/charting/openbb_charting/query_params.py
deeleeramone: The class must be initialized.

deeleeramone:
> Last comment that didn't really have a place to be written in the changed files:
> 
>     * I think `chart_params` should to be added to the `PackageBuilder` the same way `chart` is, otherwise will be a bit hidden.

This won't fully accomplish what you intend because, just like `Data`, it will not get displayed in the Swagger docs.

openbb_platform/obbject_extensions/charting/openbb_charting/charting_router.py
```diff
@@ -1,145 +1,769 @@
 """Charting router."""
 
-import json
+# pylint: disable=too-many-arguments, too-many-locals, too-many-branches, too-many-statements, unused-argument, too-many-lines
```
hjoaquim, line 3: We should try to make our code comply with the linting as much as possible, instead of ignoring it.

openbb_platform/obbject_extensions/charting/openbb_charting/__init__.py
```diff
@@ -44,19 +57,19 @@ def __init__(self, obbject):
...
         # pylint: disable=import-outside-toplevel
         from openbb_core.app.model.charts.charting_settings import ChartingSettings
-        from openbb_core.app.model.obbject import OBBject
 
         self._obbject: OBBject = obbject
         self._charting_settings = ChartingSettings(
             user_settings=self._obbject._user_settings,  # type: ignore
             system_settings=self._obbject._system_settings,  # type: ignore
         )
-        self._handle_backend()
```
hjoaquim, line 54: We should keep this in the constructor instead of having other functions (which responsibility is different) doing it.

deeleeramone: This is to prevent it from running without being asked to.

openbb_platform/obbject_extensions/charting/openbb_charting/charting_router.py
deeleeramone: We could also set the rules of the linter to allow more arguments parameters and locals. "unused-argument" is basically a required import for every file.

PR #5958, opened 2024-01-15
Uppercase excel reference

montezdesousa:
# Pull Request OpenBB

## The OpenBBTerminal

<details>
<summary> Pull Request for the OpenBBTerminal</summary>

### Description

- [ ] Summary of the change/ bug fix.
- [ ] Link # issue, if applicable.
- [ ] Screenshot of the feature or the bug before/after fix, if applicable.
- [ ] Relevant motivation and context.
- [ ] List any dependencies that are required for this change.

### How has this been tested?

- Please describe the tests that you ran to verify your changes.
- Please provide instructions so we can reproduce.
- Please also list any relevant details for your test configuration.

- [ ] Ensure the affected commands still execute in the OpenBB Terminal.
- [ ] Ensure the Platform (previously named SDK) is working as intended.
- [ ] Check any related reports.

### Checklist

- [ ] I ensure I have self-reviewed my code.
- [ ] I have commented/documented my code, particularly in hard-to-understand sections.
- [ ] I have adhered to the GitFlow naming convention and my branch name is in the format of `feature/feature-name` or `hotfix/hotfix-name`.
- [ ] Update [our documentation](https://openbb-finance.github.io/OpenBBTerminal/) following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/website).  Update any user guides that are affected by the changes.
- [ ] Update our tests following [these guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/tree/main/tests).
- [ ] Make sure you are following our [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBBTerminal/blob/main/CONTRIBUTING.md).
- [ ] If a feature was added make sure to add it to the corresponding [integration test script](https://github.com/OpenBB-finance/OpenBBTerminal/tree/develop/openbb_terminal/miscellaneous/integration_tests_scripts).

</details>

## The OpenBB Platform

<details>
<summary> Pull Request for the OpenBB Platform</summary>

### Description

- [ ] Summary of the change/ bug fix.
- [ ] Link # issue, if applicable.
- [ ] Screenshot of the feature or the bug before/after fix, if applicable.
- [ ] Relevant motivation and context.
- [ ] List any dependencies that are required for this change.

### How has this been tested?

- Please describe the tests that you ran to verify your changes.
- Please provide instructions so we can reproduce.
- Please also list any relevant details for your test configuration.

- [ ] Ensure all unit and integration tests pass.
- If you modified/added command(s):
[...]

PR #6246, opened 2024-03-21
[Docs] Regenerate Markdown Files.

deeleeramone:
1. **Why**?:

    - The markdown files under reference and data models never got updated.

    - Ran the generator script.

    - The website should reflect reality.

4. **Testing Done**:

    - Build locally and check in Chrome.


IgorWounds (approved):

IgorWounds (changes requested):
Hello,

Let's delete all these files as their generation is automated. 

The step would be to add the folder to the `.gitignore` file. 

@DidierRLopes made good points about this here: https://github.com/OpenBB-finance/OpenBBTerminal/pull/5958

piiq:
@IgorWounds we still need to sync the docs. Right now the docs that are live don't reflect the changes to the content introduced by @deeleeramone

IgorWounds (approved):

PR #6248, opened 2024-03-21
[BugFix] Make **kwargs Get Inserted Last On Build

deeleeramone:
1. **Why**?:

    - Fixes an order of insertion error where **kwargs was not placed last argument in the function signature.
    - This raised a SyntaxError.

    - Add kwargs to the end of the params ordered dict.
    - In `def validate_kwargs`, don't validate the "kwargs" field.
    - In `def _execute_func`, unpack the "kwargs" dictionary when it is a key in "kwargs".

    - **kwargs + chart can now exist together.

4. **Testing Done**:

Before:

image

After:

Screenshot 2024-03-21 at 1 30 23 PM

### DIFF
--- a/openbb_platform/core/openbb_core/app/static/package_builder.py
+++ b/openbb_platform/core/openbb_core/app/static/package_builder.py
@@ -561,14 +561,14 @@ class MethodDefinition:
                 ],
                 default=False,
             )
-
+        kwargs = None
         formatted: Dict[str, Parameter] = {}
 
         for name, param in parameter_map.items():
             if name == "extra_params":
                 formatted[name] = Parameter(name="kwargs", kind=Parameter.VAR_KEYWORD)
             elif name == "kwargs":
-                formatted["**" + name] = Parameter(
+                kwargs = Parameter(
                     name="kwargs", kind=Parameter.VAR_KEYWORD, annotation=Any
                 )
             elif name == "provider_choices":
@@ -628,7 +628,8 @@ class MethodDefinition:
                     annotation=updated_type,
                     default=param.default,
                 )
-
+        if kwargs:
+            formatted["**kwargs"] = kwargs
         return MethodDefinition.reorder_params(params=formatted)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -628,7 +628,8 @@ def format_params(
                     annotation=updated_type,
                     default=param.default,
                 )
-
+        if kwargs:
+            formatted["**kwargs"] = kwargs
         return MethodDefinition.reorder_params(params=formatted)
```
montezdesousa, line 633: To be consistent `reorder_params` should take care of the order

keeping the insertion
```
formatted["**" + name] = Parameter(
                    name="kwargs", kind=Parameter.VAR_KEYWORD, annotation=Any
                )
```

ensuring that `**kwargs` come last
```
    @staticmethod
    def reorder_params(params: Dict[str, Parameter]) -> "OrderedDict[str, Parameter]":
        """Reorder the params."""
        formatted_keys = list(params.keys())
        for k in ["provider", "extra_params", "**kwargs"]:
            if k in formatted_keys:
                formatted_keys.remove(k)
                formatted_keys.append(k)
        od: OrderedDict[str, Parameter] = OrderedDict()
        for k in formatted_keys:
            od[k] = params[k]
        return od
```

IgorWounds (approved):

montezdesousa:
While reviewing this issue I remembered that Python `kwargs` should use `extra_params: Optional[dict] = None` in the API, just like we do for other router endpoints, where we have `extra_params: ExtraParams`.

In Python we map `**kwargs` to `extra_params`. Currently there is a warning being raised which I'll fix in this PR https://github.com/OpenBB-finance/OpenBBTerminal/pull/6250. 

So I think this PR should be closed and use the different signature instead.

deeleeramone:
> While reviewing this issue I remembered that Python `kwargs` should use `extra_params: Optional[dict] = None` in the API, just like we do for other router endpoints, where we have `extra_params: ExtraParams`.

POST commands are different though, right? They do not use CommandContext and params are not defined in  models which then separate them into "standard_params" and "extra_params" like they do with GET fetchers.

deeleeramone:
Close this in favour of #6250

PR #6252, opened 2024-03-23
[BugFix] Set Default State For **kwargs in API to Make Them Optional.

deeleeramone:
This Tags on to #6250, #6248, #6240.

When `parameter.kind == Parameter.VAR_KEYWORD`, it correctly cannot be assigned a default state because **kwargs cannot have one. 

However, Fast API does not see it the same way and will make it a required field. In both, the body, and URL parameters.

To get around this, we can set:

`parameter.kind == Parameter.POSITIONAL_OR_KEYWORD`, then assign it as `Optional[Dict[str, Any]]`, and give it a default state of `None`.

The end result is that `**kwargs` can be a valid parameter for the API that gets assigned to the body of the request.

Note that `**kwargs` could be anything - `**extra_params`.


hjoaquim (changes requested):
Why do we need to change FastAPI behavior? I believe it will cause some unexpected results in the future.

There are some references on this matter, namely: https://github.com/tiangolo/fastapi/issues/251 and https://stackoverflow.com/questions/59929028/python-fastapi-error-422-with-post-request-when-sending-json-data/70636163#70636163.

Why not go for the simplest approach and have something like this:
(example signature):
`def correlation_matrix(data: List[Data], payload: Dict[Any, Any] = None) -> OBBject[List[Data]]:`

Which could then be used as follows on a POST request:

```
{
  "data": [
    {
      "additionalProp1": {}
    }
  ],
  "payload": {"chart_arg_1":"value1","chart_arg_2":"value2"}
}
```

Which results in having the following values inside the function during execution:

image

> Note taht this approach would avoid touching the `commands.py` file.

PR #6261, opened 2024-03-27
[BugFix] Add `sort` to tiingo equity historical

hjoaquim:
1. **Why**?:

    - Tiingo requires the `sort` argument. The endpoint breaks without it.

    - Added the `sort` argument.

    - 5 (positive UX impact fixing an unusable endpoint)

4. **Testing Done**:

    - Run the endpoint for tiingo, before and after the fix.

### DIFF
--- a/openbb_platform/providers/tiingo/openbb_tiingo/models/equity_historical.py
+++ b/openbb_platform/providers/tiingo/openbb_tiingo/models/equity_historical.py
@@ -21,7 +21,7 @@ from openbb_core.provider.utils.helpers import (
     amake_requests,
     get_querystring,
 )
-from pydantic import Field, PrivateAttr, model_validator
+from pydantic import Field, PrivateAttr, field_validator, model_validator
 
 _warn = warnings.warn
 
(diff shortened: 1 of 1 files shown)


openbb_platform/providers/tiingo/openbb_tiingo/models/equity_historical.py
```diff
@@ -44,6 +44,24 @@ class TiingoEquityHistoricalQueryParams(EquityHistoricalQueryParams):
     _frequency: Literal["daily", "weekly", "monthly", "annually"] = PrivateAttr(
         default=None
     )
+    sort: Literal[
```
deeleeramone, line 47: Isn't this a fruitless effort to expose this param? Sorting occurs by date, ALWAYS for this endpoint, and results are always sorted in the same direction, ascending for every provider.

deeleeramone:
I'm not able to recreate the issue you raise on the develop branch:

```
obb.equity.price.historical("AAPL", provider="tiingo").to_df()
```

Screenshot 2024-03-27 at 10 43 46 AM

openbb_platform/providers/tiingo/openbb_tiingo/models/equity_historical.py
hjoaquim: I got this error without using the sort:
`"Error: Sort column name is not a valid column"`

So, apparently the `sort` needs to be explicit in the URL.

deeleeramone: How come it works for me right now, without this?

hjoaquim:
Just figured that #6259 fixes this issue - the URL is built differently and then the `sort` is not mandatory (apparently). Not sure if we should keep it anyway. 
@deeleeramone @montezdesousa

openbb_platform/providers/tiingo/openbb_tiingo/models/equity_historical.py
deeleeramone: The URL appears to work fine without a sort column.

```
https://api.tiingo.com/tiingo/daily/AAPL/prices?startDate=2023-03-27&endDate=2024-03-27&resampleFreq=monthly&token=
```

Screenshot 2024-03-27 at 10 54 01 AM

PR #6277, opened 2024-04-02
[Feature] Add Relative Rotation To `openbb-technical` and `openbb-charting`

deeleeramone:
1. **Why**?:

    - To increase the breadth of analysis and charts.

    - `obb.technical.relative_rotation()`

    - With `chart=True`

    - Not a breaking change.

    - Requires #6243 to allow chart_params on initial request.

4. **Testing Done**:

    - Load multiple symbols and then pass through the function.
    - Integration tests.
      - API integration tests do not pass because of a discovered bug in the Fast API response for historical prices.

```
constituents = obb.index.constituents("^TX60", provider="tmx").to_df()
symbols = constituents.symbol.to_list()+["^TX60"]
data = obb.equity.price.historical(symbols, provider="tmx", start_date="2022-01-01")
rrg = obb.technical.relative_rotation(data.results, benchmark="^TX60")
rrg.charting.to_chart(show_tails=False)
```

Screenshot 2024-04-01 at 5 34 22 PM


openbb_platform/obbject_extensions/charting/openbb_charting/charting_router.py
```diff
@@ -1201,3 +1207,75 @@ def economy_fred_series(  # noqa: PLR0912
...
+    **kwargs: Any,
+) -> Tuple["OpenBBFigure", Dict[str, Any]]:
+    """Relative Rotation Chart."""
+
+    ratios_df = basemodel_to_df(kwargs["obbject_item"].rs_ratios, index="date")  # type: ignore
+    momentum_df = basemodel_to_df(kwargs["obbject_item"].rs_momentum, index="date")  # type: ignore
+    benchmark_symbol = kwargs["obbject_item"].benchmark  # type: ignore
+    study = (
+        str(kwargs.get("study"))
+        if "study" in kwargs and kwargs.get("study") is not None
```
IgorWounds, line 1222: You can make this cleaner by only getting `study` and `show_tails` once.

IgorWounds (changes requested):
This looks good to me, only a couple of small edits needed which are in the comment.

openbb_platform/extensions/technical/openbb_technical/relative_rotation.py
```diff
@@ -0,0 +1,546 @@
...
+    """ "Min/Max ScalingNormalization Method"""
+    return (data - data.min()) / (data.max() - data.min())
+
+
+def z_score_standardization(data: Series) -> Series:
+    """Z-Score Standardization Method."""
+    return (data - data.mean()) / data.std()
+
+
+def normalize(data: DataFrame, method: Literal["z", "m", "a"] = "z") -> DataFrame:
```
hjoaquim, line 33: Should this functions (normalize, standard_deviation, calculate_relative_strength_ratio, calculate_momentum, etc.) be inside the class or if not and the intention is to use them as utilities in the future, let's move them to another file.

hjoaquim:
I was unable to test:

image

openbb_platform/extensions/technical/openbb_technical/relative_rotation.py
```diff
@@ -0,0 +1,546 @@
...
+                    f"    {k} : {_get_type_name(v.annotation)}\n        {v.description}"
+                    for k, v in fields.items()
+                ]
+            )
+            + "\n"
+        )
+        self.__doc__ = doc_str
+
+
+class RelativeRotationFetcher(
```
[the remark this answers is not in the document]
deeleeramone: Sure we are, the data is being extracted from the supplied `Data`, and then calculations are made, and then it is transformed into the required shape. Then you can also use the class methods of convenience within the Fetcher class.

openbb_platform/extensions/technical/openbb_technical/relative_rotation.py
deeleeramone: They are excluded specifically from the class because they should not get bound to it. Those functions, however, are specifically tailored to this use. They are meant to serve this function only.

deeleeramone:
> I was unable to test:

Yup, pretty sure something else is doing this, probably related to the broken API.

Screenshot 2024-04-12 at 10 58 12 AM

deeleeramone:
> I was unable to test:

Try again, should be good now.

hjoaquim:
> > I was unable to test:
> 
> Try again, should be good now.

@deeleeramone I'm running into the same error:
```bash
OpenBBError: 
Type -> AttributeError

Detail -> '_LiteralSpecialForm' object has no attribute '__name__'
```

IgorWounds (approved):

PR #6279, opened 2024-04-02
[BugFix] - Serialise response before FastAPI

montezdesousa:
1. **Why**?

    - API responses are excluding the symbol response field when multiple symbols are passed into `symbol` query parameter

2. **What**?

    - Serialising the response before FastAPI prevents Pydantic from dropping extra fields, thus dropping `symbol` from API response
    - This was present before https://github.com/OpenBB-finance/OpenBBTerminal/pull/6024, but discriminated unions require us to pass the provider field in the model

3. **Impact**

    - Behaviour is consistent in both interfaces
    - It also returns the provider where the data comes from, removing this in both interfaces introduces unnecessary complexity since the information can be ignored by frontend

4. **Testing Done**:

image

Screenshot 2024-04-02 at 16 13 45

### DIFF
--- a/openbb_platform/core/openbb_core/api/router/commands.py
+++ b/openbb_platform/core/openbb_core/api/router/commands.py
@@ -117,7 +117,7 @@ def build_new_signature(path: str, func: Callable) -> Signature:
     )
 
 
-def validate_output(c_out: OBBject) -> OBBject:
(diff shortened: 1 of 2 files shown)


IgorWounds (approved):

IgorWounds:
@montezdesousa we keeping the provider field too?

PR #6283, opened 2024-04-03
[BugFix] Drop 'provider' in `to_df()`

deeleeramone:
1. **Why**?:

    - Takes care of an impact detailed in #6279, ignoring the 'provider' column introduced, for the Python Interface.

    - Drops that column in `to_df()` and `basemodel_to_df()`

    - Maintains a cleaner output and presentation.

4. **Testing Done**:

Screenshot 2024-04-02 at 10 57 10 PM

### DIFF
--- a/openbb_platform/core/openbb_core/app/model/obbject.py
+++ b/openbb_platform/core/openbb_core/app/model/obbject.py
@@ -226,6 +226,9 @@ class OBBject(Tagged, Generic[T]):
         except Exception as ex:
             raise OpenBBError(f"An unexpected error occurred: {ex}") from ex
 
+        if "provider" in df.columns:
+            df.drop(columns=["provider"], inplace=True)
+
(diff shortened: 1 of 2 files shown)


hjoaquim (changes requested):
Doesn't seem a good idea to patch it like this - should be fixed on the route instead.
CC @montezdesousa

montezdesousa:
> Doesn't seem a good idea to patch it like this - should be fixed on the route instead. CC @montezdesousa

Yes, I prefer the suggestion you made in the channel @deeleeramone 

We explicitly include the `symbol` field in the models that need to show it, `exclude=True` the field "provider" in `registry_map.py` and remove the `model_dump` in `commands.py::validate_output` function.

deeleeramone:
Closing in favour of a better, future, solution.

PR #6295, opened 2024-04-09
[BugFix] Handling negative timestamps on Windows

hjoaquim:

    - Having a negative timestamp would break the model validation thus corrupting the command response.

    - Used `datetime.timedelta` for a safe timestamp converstion when on Windows platform.

    - NA - done at the provider level allowing users to use the affected endpoint.

4. **Testing Done**:

    - Running the identified command that was running into issues for all the available providers: `obb.equity.profile("IBM", provider="yfinance").to_df()`.

    - If the reviewer can't test on Windows:
    - Before:
image

    - After:
image

### DIFF
--- a/openbb_platform/core/openbb_core/provider/utils/helpers.py
+++ b/openbb_platform/core/openbb_core/provider/utils/helpers.py
@@ -1,8 +1,9 @@
 """Provider helpers."""
 
 import asyncio
+import os
 import re
-from datetime import date, datetime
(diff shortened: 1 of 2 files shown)


deeleeramone (approved):
Nice, maybe we need to apply this to everywhere that uses `datetime.fromtimestamp()`, just to be safe!

hjoaquim:
Closing in favor of #6298 - this one started from the wrong head.

PR #6301, opened 2024-04-10
[Feature] Limit number of countries in TE request

hjoaquim:

    - Allowing more than N countries in a TE request would break with a http code of 400 - this PR enhances the UX.

    - Added a validation for the number of countries in the fetcher.

    - Low - small improvement on UX.

4. **Testing Done**:

    - Experiment with any amount of countries from `openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/utils/countries.py`:
```python
from openbb import obb
countries = [...]
obb.economy.calendar(provider="tradingeconomics", country=countries).to_df()
```

    - It looks like it depends on the countries we're sending. Sometimes it fails for 28, other times 31/32. I wasn't able to get more than 32 with any combination of countries. Also, couldn't find any information related with this limit on TE API docs. cc @andrewkenreich

### DIFF
--- a/openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/economic_calendar.py
+++ b/openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/economic_calendar.py
@@ -57,6 +57,13 @@ class TEEconomicCalendarQueryParams(EconomicCalendarQueryParams):
         for v in values:
             check_item(v.lower(), COUNTRIES)
             result.append(v.lower())
+
+            if len(result) > 32:
+                raise ValueError(
+                    "Too many countries. "
+                    "Trading Economics API allows a maximum of 32 countries."
+                )
+
         return ",".join(result)
 
     @field_validator("importance")
--- a/openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/economic_calendar.py
+++ b/openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/economic_calendar.py
@@ -31,6 +31,7 @@ GROUPS = Literal[
     "trade",
     "business",
 ]
+TE_COUNTRY_LIMIT = 32
 
 
 class TEEconomicCalendarQueryParams(EconomicCalendarQueryParams):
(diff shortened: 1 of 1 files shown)


deeleeramone:
> 1. **Why**? (1-3 sentences or a bullet point list):
>        
>        * Allowing more than N countries in a TE request would break with a http code of 400 - this PR enhances the UX.
> 
>     2. **What**? (1-3 sentences or a bullet point list):
>        
>        * Added a validation for the number of countries in the fetcher.
> 
>     3. **Impact** (1-2 sentences or a bullet point list):
>        
>        * Low - small improvement on UX.
> 
>     4. **Testing Done**:
>        
>        * Experiment with any amount of countries from `openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/utils/countries.py`:
> 
> 
> ```python
> from openbb import obb
> countries = [...]
> obb.economy.calendar(provider="tradingeconomics", country=countries).to_df()
> ```
> 
>     5. **Reviewer Notes** (optional):
>        
>        * It looks like it depends on the countries we're sending. Sometimes it fails for 28, other times 31/32. I wasn't able to get more than 32 with any combination of countries. Also, couldn't find any information related with this limit on TE API docs. cc @andrewkenreich

Is there a reason why "All" is not used instead of trying to get a huge list? It would be pretty simple to filter post-request, and storing that data in cache would dramatically reduce the number of API calls.

IgorWounds (approved):

IgorWounds:
@deeleeramone

deeleeramone:
> Is `all` even an option? Couldn't find that on TE API docs. I like the idea, if possible.
> 
> Although, we can still set a limit for number of countries in one request, if TE has one (as it look it does), using the validator.

All is what you get when you only enter a start/end date.  A different way of going about restricting the length of the URL would be by the character count instead of the number of countries. 

2000 characters is a general "safe" limit for URL length, although this can vary by platform.

hjoaquim:
> > Is `all` even an option? Couldn't find that on TE API docs. I like the idea, if possible.
> > Although, we can still set a limit for number of countries in one request, if TE has one (as it look it does), using the validator.
> 
> All is what you get when you only enter a start/end date. A different way of going about restricting the length of the URL would be by the character count instead of the number of countries.
> 
> 2000 characters is a general "safe" limit for URL length, although this can vary by platform.

That seems to be the case - let's do that instead.

PR #6308, opened 2024-04-11
[Feature] - LLM mode

IgorWounds:

    - The goal of the OpenBB Platform is to be LLM-friendly and easily used and accessed by LLM frameworks such as Magentic.

    - Adds the `llm` `output_type` user preference
    - Adds the `docstring_sections` to the Python system settings that control which sections of the docs are added
    - Adds the `docstring_max_length` to the Python system settings that control the context size of the docstrings

    - Minimal impact to users that don't utilize LLMs

4. **Testing Done**:

    - Tested that it does what is expected.

5. **Reviewer notes**:

   - To try it out please use the following approach
   
```
from openbb import obb

obb.user.preferences.output_type="llm"
obb.system.python_settings.docstring_sections=['description', 'examples']
obb.system.python_settings.docstring_max_length=1024

import openbb
openbb.build()
```


openbb_platform/core/openbb_core/app/model/preferences.py
```diff
@@ -3,35 +3,48 @@
...
         return f"{self.__class__.__name__}\n\n" + "\n".join(
             f"{k}: {v}" for k, v in self.model_dump().items()
         )
+
+    @field_validator("output_type", mode="before")
+    @classmethod
+    def llm_mode(cls, value: str) -> str:  # pylint: disable=no-self-argument
+        """Set LLM mode."""
+        if Env().LLM_MODE:
+            return "llm"
```
mnicstruwig, line 49: This didn't seem to work on my end for this branch:

```
# .zshrc
export OPENBB_LLM_MODE=True
```
Screenshot 2024-04-15 at 11 45 02

The output type is supposed to be `llm`, right?

openbb_platform/core/openbb_core/app/model/obbject.py
mnicstruwig, line 1: I'm not sure if it was intended to be part of this PR, but `from openbb.llm import obb` is missing.

openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -1079,6 +1079,10 @@ def get_param_info(parameter: Parameter) -> Tuple[str, str]:
 
         docstring = summary.strip("\n").replace("\n    ", f"\n{create_indent(2)}")
         docstring += "\n\n"
+        if Env().LLM_MODE:
+            if len(docstring) > 1024:
+                docstring = docstring[:1021] + "..."
+            return docstring
```
mnicstruwig, line 1085: This also doesn't seem to work correctly, eg. calling `obb.equity.price.quote.__doc__` returns the full docstring.

openbb_platform/core/openbb_core/app/model/preferences.py
mnicstruwig: Given that other parts of the code aren't working, I suspect that the `OPENBB_LLM_MODE` detection is _probably_ broken, which is causing the rest of the functionality to fail.

mnicstruwig: Manually trying to set the `output_type` raises the following error:

Screenshot 2024-04-15 at 11 57 42

Is it possible somehow that things aren't being "registered" correctly somehow?

I can confirm I'm on the correct branch, and have installed using `pip install -e .`:

Screenshot 2024-04-15 at 11 59 01

openbb_platform/core/openbb_core/app/model/obbject.py
IgorWounds: This won't work as it triggers the __init__.py from the usual openbb and breaks the pattern and automated builds of static assets.

openbb_platform/core/openbb_core/app/static/package_builder.py
IgorWounds: Did you rebuild?

mnicstruwig:
Thanks for the updates, a few comments:
- Use `Field`, instead of `FieldInfo` -- both achieve the same thing, but `pydantic` explicitly [recommends](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.FieldInfo) using `Field` to annotate. 
  - Also keep in mind that setting default values **inside of an `Annotated` type hint is [not supported](https://docs.pydantic.dev/latest/concepts/fields/#using-annotated)**! (Instead, assign the default value like we do in regular Python, eg. `param: Annotated[type, <metadata>] = "default_value"`. All of the LLM frameworks will retrieve the default value from the actual signature, not from the `Field` (this is how Pydantic also works).
  - also, the setting a type as `Optional` [does not imply](https://docs.pydantic.dev/latest/concepts/fields/#default-values) a default value of `None`.
  - So we'll have to make sure we follow this rules and guidelines before merging.
- The `required` value is parsed from the actual default value specified for the argument (_not_ from the Annotation!), which means that using `Field` _will_ solve the issue with `magentic`.

BUT... it does make the signatures looks a bit ugly:

```python
def get_weather_forecast(
        location: Annotated[str, Field(description="The location")],
        num_days: Annotated[int, Field(description="The number of days")] = 3
        ) -> str:
    """Get the weather for a certain number of days in advance"""
    return "Sunny"

get_weather_forecast?
# Output
Signature:
get_weather_forecast(
    location: typing.Annotated[str, FieldInfo(annotation=NoneType, required=True, description='The location')],
    num_days: typing.Annotated[int, FieldInfo(annotation=NoneType, required=True, description='The number of days')] = 3,
) -> str
Docstring: Get the weather for a certain number of days in advance
File:      [/var/folders/gq/jl16rc4171g3102f40n98w0m0000gn/T/ipykernel_38014/188702632.py](https://file+.vscode-resource.vscode-cdn.net/var/folders/gq/jl16rc4171g3102f40n98w0m0000gn/T/ipykernel_38014/188702632.py)
Type:      function
```

Those `annotations` in the `FieldInfo` look quite ugly if you ask me, and could be misleading to users (since the `annotation` is misleading _and_ the `required=True` is confusing, especially if a default value already exists).

What I suggest is the following:
- Despite what the documentation says, we don't use `Field` (which is a function, despite the name).
- Instead, we create an `OpenBBCustomParameter` that _inherits_ from `FieldInfo`.
- This `OpenBBCustomParameter` defines a custom `__repr__` method, to make the documentation look the way it currently does.
- We annotate _everything_ using the "new" `OpenBBCustomParameter`

This should allow us to keep both the LLM framework functionality, as well as our existing annotations and documentation. A literal win-win.

Demo:

```python
class OpenBBCustomParameter(FieldInfo):
    def __repr__(self):
        return f"OpenBBCustomParameter(description={self.description})"

def get_weather_forecast(
        location: Annotated[str, OpenBBCustomParameter(description="The location")],
        num_days: Annotated[int, OpenBBCustomParameter(description="The number of days")] = 3
        ) -> str:
    """Get the weather for a certain number of days in advance"""
    return "Sunny"
```

This gives us the following tool schema for LLMs:

```python
{'type': 'function',
 'function': {'name': 'get_weather_forecast',
              'parameters': {'properties': {'location': {'description': 'The '
                                                                        'location',
                                                         'title': 'Location',
                                                         'type': 'string'},
                                            'num_days': {'default': 3,
                                                         'description': 'The '
                                                                        'number '
                                                                        'of '
                                                                        'days',
                                                         'title': 'Num Days',
                                                         'type': 'integer'}},
                             'required': ['location'],
                             'type': 'object'},
              'description': 'Get the weather for a certain number of days in '
                             'advance'}}
```

As well as a beautiful documentation signature:

```python
Signature:
get_weather_forecast(
    location: typing.Annotated[str, OpenBBCustomParameter(description=The location)],
    num_days: typing.Annotated[int, OpenBBCustomParameter(description=The number of days)] = 3,
) -> str
Docstring: Get the weather for a certain number of days in advance
File:      [/var/folders/gq/jl16rc4171g3102f40n98w0m0000gn/T/ipykernel_38014/308173038.py](https://file+.vscode-resource.vscode-cdn.net/var/folders/gq/jl16rc4171g3102f40n98w0m0000gn/T/ipykernel_38014/308173038.py)
Type:      function
```

Let me know what you think!

montezdesousa:
Hi @mnicstruwig, maybe you wrote your comment based on a different commit, but I don't see the problems in the latest, except being ugly. It doesn't use `Field`, `annotation` is consistent, `required` is not there and the default is only there to make `is_required()` consistent with actual signature (it is inferred from the default).

Anyways, your solution looks much better and if we won't need any `FieldInfo` property except the description I'd go with your approach.

mnicstruwig:
@montezdesousa 

Yes, I was specifically referring to an older commit before the "ugly" fix -- I saw the proposed changes to the generation code and I agreed that it was ugly, which is why I proposed the alternative solution :) . If we can get away with "not ugly", I'd much prefer it!

mnicstruwig:
Much better 💪 

Mentioned on Slack, but bears repeating here, the `required` fields are only correctly reported for `pydantic>=2.7.0`:

```
# Pydantic 2.7.0

{'properties': {'symbol': {'anyOf': [{'type': 'string'},
    {'items': {'type': 'string'}, 'type': 'array'}],
   'choices': None,
   'description': 'Symbol to get data for. Multiple comma separated items allowed for provider(s): fmp, intrinio, yfinance.',
   'title': 'Symbol'},
  'provider': {'anyOf': [{'enum': ['fmp', 'intrinio', 'yfinance'],
     'type': 'string'},
    {'type': 'null'}],
   'choices': None,
   'default': None,
   'description': "The provider to use for the query, by default None.\n    If None, the provider specified in defaults is selected or 'fmp' if there is\n    no default.",
   'title': 'Provider'},
  'kwargs': {'default': {}, 'title': 'Kwargs', 'type': 'object'}},
 'required': ['symbol'],  # <-- required field
 'title': 'FuncModel',
 'type': 'object'}

# Pydantic 2.6.4 -- notice the missing `required` field.
{'properties': {'symbol': {'anyOf': [{'type': 'string'},
    {'items': {'type': 'string'}, 'type': 'array'}],
   'choices': None,
   'description': 'Symbol to get data for. Multiple comma separated items allowed for provider(s): fmp, intrinio, yfinance.',
   'title': 'Symbol'},
  'provider': {'anyOf': [{'enum': ['fmp', 'intrinio', 'yfinance'],
     'type': 'string'},
    {'type': 'null'}],
   'choices': None,
   'default': None,
   'description': "The provider to use for the query, by default None.\n    If None, the provider specified in defaults is selected or 'fmp' if there is\n    no default.",
   'title': 'Provider'},
  'kwargs': {'default': {}, 'title': 'Kwargs', 'type': 'object'}},
 'title': 'FuncModel',
 'type': 'object'}
```

Unfortunately, since we don't pin pydantic I _think_ we end up with `pydantic==2.6.4` because of dependencies somewhere.

openbb_platform/core/openbb_core/app/model/field.py
```diff
@@ -0,0 +1,27 @@
...
+        # We use repr() to avoid decoding special characters like \n
+        if self.choices:
+            return f"OpenBBField(description={repr(self.description)}, choices={repr(self.choices)})"
+        return f"OpenBBField(description={repr(self.description)})"
+
+    def __init__(self, description: str, choices: Optional[List[Any]] = None):
+        """Initialize OpenBBField."""
+        super().__init__(
+            description=description, json_schema_extra={"choices": choices}
+        )
```
mnicstruwig, line 22: This will force `choices=None` when no choices are provided, which pollutes the resulting pydantic model schema of the function signature. Instead, we should do the following: 
```suggestion
        """Initialize OpenBBField."""
        json_schema_extra = {"choices": choices} if choices else None
        super().__init__(
            description=description, json_schema_extra=json_schema_extra
        )
```

cc @montezdesousa

mnicstruwig (approved):
LGTM. Fantastic work!

PR #6323, opened 2024-04-19
Bugfix/reference breaking when no model

piiq:

## Description

- Prevent package builder from failing when building a router-only extension that does not use standard models

## How has this been tested?

Ran commands from a router-only extension that have and haven't got query parameters


piiq:
closed in favor of https://github.com/OpenBB-finance/OpenBBTerminal/pull/6324

PR #6327, opened 2024-04-19
hotfix/Frontend sync

tehcoderer:
Frontend sync


openbb_platform/providers/fmp/openbb_fmp/utils/helpers.py
```diff
@@ -144,3 +149,27 @@ def get_interval(value: str) -> str:
     }
 
     return f"{value[:-1]}{intervals[value[-1]]}"
+
+
+# some fmp endpoint return date in EST without a timezone, this function will parse it
```
deeleeramone, line 154: They actually return the date in the timezone of the exchange, without a timezone. For example: 

https://financialmodelingprep.com/api/v3/historical-chart/5min/MNDI.L?
```
  {
    "date": "2024-04-08 08:05:00",
    "open": 1400.5,
    "low": 1396.5,
    "high": 1400.5,
    "close": 1398,
    "volume": 5298
  },
  {
    "date": "2024-04-08 08:00:00",
    "open": 1395,
    "low": 1395,
    "high": 1407.5,
    "close": 1399,
    "volume": 4883
  }
```

The London Stock Exchange is open Monday through Friday from 8:00 am to 4:30 pm British Summer Time (GMT+01:00).

You are localizing that as 8:00 AM EST, which is not a correct TZ conversion.

tehcoderer: smh they really are the best 💀

deeleeramone: This is exactly the reason why it's not localized, it's better to leave it alone than to make it incorrectly tz-aware.

tehcoderer: then I need to figure out the current timezone per symbol, or else all dates will be treated as UTC on the frontend. RIP

openbb_platform/providers/fmp/openbb_fmp/models/financial_ratios.py
```diff
@@ -226,37 +237,54 @@ async def aextract_data(
...
-            results[0].update(
-                {"period": "TTM", "date": datetime.now().date().strftime("%Y-%m-%d")}
-            )
-        for item in results:
-            item.pop("symbol", None)
-            item.pop("dividend_yiel_percentage", None)
-        return [FMPFinancialRatiosData.model_validate(d) for d in results]
+        results: List[FMPFinancialRatiosData] = []
+        for item in data:
+            new_item = {to_snake_case(k).replace("ttm", ""): v for k, v in item.items()}
```
deeleeramone, line 280: Screenshot 2024-04-19 at 4 18 17 PM

Looks like we need to add "_" to the str.replace(), and "dividend_yiel_percentage" is a specialty for the "TTM" response. It's popped because it is redundant.

tehcoderer: nice yup done!

tehcoderer: also have no idea why this one keeps failing on the test runs 
image
lol

openbb_platform/providers/fmp/openbb_fmp/models/financial_ratios.py
```diff
@@ -24,22 +27,30 @@ class FMPFinancialRatiosQueryParams(FinancialRatiosQueryParams):
     Source: https://financialmodelingprep.com/developer/docs/#Company-Financial-Ratios
     """
 
+    __json_schema_extra__ = {"symbol": ["multiple_items_allowed"]}
+
     period: Literal["annual", "quarter", "ttm"] = Field(
         default="annual", description=QUERY_DESCRIPTIONS.get("period", "")
     )
+    with_ttm: Optional[bool] = Field(
```
deeleeramone, line 35: This was purposefully left as an either/or situation.  Combining "TTM" with "quarter" or "annual" makes a huge mess of the response and they don't necessarily play nicely together.

Screenshot 2024-04-19 at 4 28 58 PM

openbb_platform/providers/fmp/openbb_fmp/utils/helpers.py
deeleeramone: Seems like something frontend needs to deal with independently of the fetcher. These should stay tz-unaware from the fetcher.

openbb_platform/providers/fmp/openbb_fmp/models/financial_ratios.py
deeleeramone: > nice yup done!

"dividend_yield_percentage" can be dropped. This is redundant.

PE ratio and peg ratio, maybe some others are also being duplicated.

Screenshot 2024-04-19 at 10 00 58 PM

openbb_platform/providers/fmp/openbb_fmp/models/index_historical.py
```diff
@@ -57,6 +57,12 @@ class FMPIndexHistoricalData(IndexHistoricalData):
         json_schema_extra={"x-unit_measurement": "percent", "x-frontend_multiply": 100},
     )
 
+    @field_validator("date", mode="before", check_fields=False)
+    @classmethod
+    def date_validate(cls, v):  # pylint: disable=E0213
+        """Return formatted datetime."""
+        return parse_date(v, "America/New_York")
```
deeleeramone, line 64: This will not work here, for the same reasons it won't in `equity.historical()`. They need to be TZ-unaware.

openbb_platform/providers/fmp/openbb_fmp/utils/helpers.py
andrewkenreich: @deeleeramone if we have to do this on FE can you get us a list of all exchanges and the time zone they in

 Ty

deeleeramone: Not from FMP directly, at least AFAICT.

openbb_platform/providers/fmp/openbb_fmp/models/financial_ratios.py
deeleeramone: > also have no idea why this one keeps failing on the test runs 

I think this is because of way the requests are structured - the nested async requests within the callback.

openbb_platform/providers/fmp/openbb_fmp/models/financial_ratios.py
```diff
@@ -226,37 +237,57 @@ async def aextract_data(
...
-        )
-        results = await amake_request(
-            url, response_callback=response_callback, **kwargs
-        )
+        include_ttm = query.period != "ttm" and query.with_ttm
+
+        async def response_callback(
+            response: ClientResponse, session: ClientSession
+        ) -> List[Dict]:
+            results: List[dict] = await response.json()  # type: ignore
```
deeleeramone, line 247: This effectively undoes the efforts to expose the actual error message from the FMP response - https://github.com/OpenBB-finance/OpenBBTerminal/pull/6237. 

Instead of clearly communicating the issue to the user:
```
OpenBBError: FMP Error Message -> Invalid API KEY. Feel free to create a Free API Key or visit https://site.financialmodelingprep.com/faqs?search=why-is-my-api-key-invalid for more information.
```

It is now a validation error that cannot be traced back to the root cause:

```
OpenBBError: 2 validation errors for FMPFinancialRatiosData
periodEnding
  Field required [type=missing, input_value={'error_message': 'Invali... for more information.'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.7/v/missing
fiscalPeriod
  Field required [type=missing, input_value={'error_message': 'Invali... for more information.'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.7/v/missing
```

deeleeramone (changes requested):

deeleeramone:
Closing because these have been integrated elsewhere and this is now stale.

PR #6342, opened 2024-04-25
[BugFix] - Fix lowercase symbols

IgorWounds:

    - The `field_validator` from the standard model wasn't being picked up by the `CBOEDerivatesFetcher` which resulted in errors when the symbol was lowercase.

    - Ensures all symbols passed to the `CBOEDerivatesFetcher` are uppercase.

    - N/A

4. **Testing Done**:

    - Ensured it works.


deeleeramone:
This impacts all the other providers as well. Intrinio symbols are all case-sensitive.

PR #6344, opened 2024-04-26
Feature/move terminal files

montezdesousa:
* Deletes unused files and folders
* Moves cli related files into `openbb_terminal`
* Creates singleton `Session` class to handle all interactive tasks that a user can affect by changing it's preferences, settings, account, etc.


openbb_platform/providers/yfinance/openbb_yfinance/__init__.py
```diff
@@ -27,9 +27,6 @@
 from openbb_yfinance.models.key_executives import YFinanceKeyExecutivesFetcher
 from openbb_yfinance.models.key_metrics import YFinanceKeyMetricsFetcher
 from openbb_yfinance.models.losers import YFLosersFetcher
-from openbb_yfinance.models.market_indices import (
```
hjoaquim, line 30: Was this change to the platform intended?

hjoaquim (approved):

PR #6345, opened 2024-04-26
[BugFix] Docker yamls are missing in main

hameliknaoh:
Docker yamls are missing in main. Updated the setup doc to point to the yamls in latest release.


deeleeramone:
Hello,

These have been intentionally removed from the development pipeline.

PR #6358, opened 2024-05-01
[BugFix] Restore Platform api.dockerfile

deeleeramone:
1. **Why**?:

    - Fallen casualty to big changes.


deeleeramone:
Need to take a different approach..

PR #6359, opened 2024-05-02
[BugFix] SEC ETF Holdings - Try Catch for RemoteDisconnect Error

deeleeramone:
1. **Why**?:

This is an attempt to mitigate a RemoteDisconnect error which can occasionally get raised from an SEC search query. In the handful of times I have encountered, the next request is successful. 

image

    - Implement up-to 3 retries with a one-second sleep. If a retry is successful, a warning will be set as a notification.

    - Example: "Updated the risk calculation algorithm to factor in real-time market fluctuations."

    - This may extend the timeout period, but if a response cannot be read within a couple of seconds, there is likely a network error. 

4. **Testing Done**:

    - Extremely hard to recreate,  @montezdesousa, are you able to reliably reproduce and/or does this resolve?

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/models/etf_holdings.py
+++ b/openbb_platform/providers/sec/openbb_sec/models/etf_holdings.py
@@ -2,6 +2,7 @@
 
 # pylint: disable =[unused-argument,too-many-locals,too-many-branches]
 
+import asyncio
 from datetime import date as dateType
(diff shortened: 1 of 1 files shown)


montezdesousa (approved):
Cannot reproduce, but changes lgtm

IgorWounds (approved):

PR #6380, opened 2024-05-09
[BugFix] Fix Currency Search

deeleeramone:
1. **Why**?:

    - This command has 3 different providers but then it doesn’t have any standard argument and all the available provider arguments belong to polygon only.

2. **What**?:

    - There is now one parameter, "query", and it is valid for all 3 providers.

3. **Impact**:

    - This endpoint is pretty unlikely to be used, is for generating lists of valid symbols for a source, and a basic text search through every field returned is all its capable of.  

4. **Testing Done**:

    - Updated integration tests and Polygon unit test.
 
```python
obb.currency.search("euro", provider="polygon")
```

```console
OBBject

id: 0663c5e6-9a47-7b78-8000-84fc8542c7a8
results: [{'symbol': 'AEDEUR', 'name': 'Euro', 'currency_symbol': 'EUR', 'base_curr...
provider: polygon
warnings: None
chart: None
extra: {'metadata': {'arguments': {'provider_choices': {'provider': 'polygon'}, 'st...
```

```python
obb.currency.search("europe", provider="polygon")
```

```console
Detail -> No results were found with the query supplied. -> europe
```

```python
obb.currency.search("europe", provider="fmp")
```

```console
OpenBBError: 
Type -> EmptyDataError

Detail -> No results were found with the query supplied. -> europe Hint: Names and descriptions are not searchable from FMP, try 3-letter symbols.
```


openbb_platform/providers/polygon/openbb_polygon/models/currency_pairs.py
```diff
@@ -22,75 +25,39 @@ class PolygonCurrencyPairsQueryParams(CurrencyPairsQueryParams):
     Source: https://polygon.io/docs/forex/get_v3_reference_tickers
     """
 
-    symbol: Optional[str] = Field(
```
hjoaquim, line 25: Why removing this Polygon query parameters?

hjoaquim (commented):
Works well for all providers. I believe having the "query" facilitates understading how the command works, nice!

openbb_platform/providers/polygon/openbb_polygon/models/currency_pairs.py
deeleeramone: Because we can do everything post-request with no real impact on performance while making it operate like everything else and simplifying it dramatically for the end-user.

PR #6382, opened 2024-05-09
[Feature] Update chart creation so it doesn't break the command execution

hjoaquim:

    - Creating a chart shouldn't break the whole creation of the `OBBject` (unless in debug mode).

    - Encapsulate in a try/except block.

    - NA / minor UX.

4. **Testing Done**:

    - Run any chart - manually insert an exception for testing purposes.

### DIFF
--- a/openbb_platform/core/openbb_core/app/command_runner.py
+++ b/openbb_platform/core/openbb_core/app/command_runner.py
@@ -304,31 +304,38 @@ class StaticCommandRunner:
         **kwargs,
     ) -> None:
         """Create a chart from the command output."""
-        if "charting" not in obbject.accessors:
-            raise OpenBBError(
-                "Charting is not installed. Please install `openbb-charting`."
-            )
-        chart_params = {}
-        extra_params = kwargs.get("extra_params", {})
+        try:
+            if "charting" not in obbject.accessors:
+                raise OpenBBError(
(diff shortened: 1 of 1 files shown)


deeleeramone:
LGTM, on a side-note, which function was failing when `chart=True` and under what conditions? It would be good to know this so it can be looked into.

hjoaquim:
> LGTM, on a side-note, which function was failing when `chart=True` and under what conditions? It would be good to know this so it can be looked into.

Actually, any! This come up on a random experiment and just thought it would be a good idea to not break everything just bc the chart couldn't be computed!

deeleeramone:
> Actually, any! This come up on a random experiment and just thought it would be a good idea to not break everything just bc the chart couldn't be computed!


WDYM, any? I understand what this is doing, but am looking for a repeatable scenario where `chart=True` fails.

deeleeramone (approved):

hjoaquim:
> > Actually, any! This come up on a random experiment and just thought it would be a good idea to not break everything just bc the chart couldn't be computed!
> 
> WDYM, any? I understand what this is doing, but am looking for a repeatable scenario where `chart=True` fails.

It does not fail in any function afaik; I forced an exception to check something, not sure about the context anymore tho

PR #6384, opened 2024-05-09
[Feature] - Create repo assets directory

montezdesousa:
1. **Why**?

    - `providers.json` needed to be manually updated
    - Information about extensions and providers is required by OpenBB Hub

2. **What**?

    - Instructions and `v3_credentials` are now defined in the `Provider` class
    - Creates `assets` folder for the repo. Applications that need to know any repo information should rely on this folder only.
    - This avoids downstream applications to scrape the repo and then be broken by a file that is randomly deleted.
    - We know which files need to be maintained
    - Deletes `PROVIDERS.md` and `EXTENSIONS.md`. These were not maintained and info is duplicated in multiple places

3. **Impact**:

    - `providers.json` was already read by OpenBB Hub to provide credential hints, need to update that website before merging this code to `main`

4. **Testing Done**:

```shell
python assets/scripts/generate_extensions_data.py
```


deeleeramone:
@montezdesousa, should we include the `/examples` folder too? Would we need to move the JSON file there?

https://github.com/OpenBB-finance/OpenBBTerminal/blob/develop/examples/content.json

IgorWounds (approved):
This is good!

Knowing that V3 is deprecated and discontinued, for how long should we keep the v3 credential mapping in our v4 architecture and mapping?

Should we migrate to the mapping of v4 at a certain point? cc: @andrewkenreich

montezdesousa:
> @montezdesousa, should we include the `/examples` folder too? Would we need to move the JSON file there?
> 
> https://github.com/OpenBB-finance/OpenBBTerminal/blob/develop/examples/content.json

Sounds good, is it auto-generated or we can just move it and redirect on the Hub?

PR #6387, opened 2024-05-10
Security updates

piiq:
This PR is for addressing most of the security alerts raised by Dependabot

bumps aiohttp, python-multipart and urllib3

only lock files (and core/pyproject.toml) changed

this should remove a few hundred alerts


IgorWounds:
@piiq I ran all unit and integration tests. Things are looking good. I've noticed that some integration tests that are using the TMX provider fail:

- test_fixedincome_corporate_bond_prices
- test_fixedincome_government_treasury_prices

This provider uses two dependencies that might be linked to the change:

- aiohttp-client-cache = "^0.10.0"
- aiosqlite = "^0.19.0"

piiq:
@IgorWounds I've bumped those and the integration tests for tmx are passing on my side.
Can you check please?
image

piiq:
I am done with commits in this PR. I will create another one to remove python-jose and bump cryptography and starlette

IgorWounds (approved):

PR #6388, opened 2024-05-10
[Feature] Fallback to to_df() method when results are a string

hjoaquim:

    - So that the `to_df()` method doesn't break if the results are a string.

    - Wraps the string in a list so that it's data framable.

    - Minor UX improvement.

4. **Testing Done**:

    - `res = obb.equity.price.historical(...)`
    - `res.results = "Hello!"`
    - `res.to_df()`

### DIFF
--- a/openbb_platform/core/openbb_core/app/model/obbject.py
+++ b/openbb_platform/core/openbb_core/app/model/obbject.py
@@ -105,6 +105,9 @@ class OBBject(Tagged, Generic[T]):
         - Dict[str, List]
         - Dict[str, BaseModel]
 
+        Other supported formats:
+        - str
+
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


deeleeramone:
If the results are a simple string, let's just print it directly to the screen.

Converting to a DataFrame in this instance doesn't really provide any benefit or utility.

montezdesousa (approved):

hjoaquim:
> If the results are a simple string, let's just print it directly to the screen.
> 
> Converting to a DataFrame in this instance doesn't really provide any benefit or utility.

Not related with the CLI, but this is actually a nice fallback to have, since it makes the `to_df()` method a bit more robust imo:

Before:

image


After:

image

PR #6392, opened 2024-05-10
[BugFix] Econ Calendar

deeleeramone:
1. **Why**?:

Started as needing to add the list of countries to `json_schema_extra` in the TradingEconomics model, but ran into other items needing to be taken care of.

  - Standard model contained parameters and data fields that were clearly only valid for TE, the description even named 'TradingEconomics'.
  - Missing field definitions from TE response.
  - Can't lookup by country, date, and importance.
  - Literal choices not showing up in API docs.

2. **What**?:

    - Remove TE-only items from the standard model.
    - Add `calendar_id` as a parameter for TE.
    - Can now lookup by country, date, and importance.
    - Defined the missing fields in the TE model
    - Removed `Optional` from Literals and they now are proper choices in the API docs.
    - "Importance" parameter made lower-case.
    - Chunk dates for FMP so that start/end dates are not bound by a 3-month window.
    - TE `group` parameter had one choice with a space in it.
 
3. **Impact**:

    - @andrewkenreich, this renovates the models. Most notably, the choices for "importance" are now lower-case.
      - You will notice more fields, but I don't think you use them anyways.
    - New TE request parameter: `calendar_id`
    - TE `country` parameter now has: `json_schema_extra={"choices": COUNTRIES}`
    - TE requests can now do something like:

```
obb.economy.calendar(country="united_states", provider="tradingeconomics", start_date="2023-01-01", end_date="2023-12-31", importance="high")
```

4. **Testing Done**:

    - Unit and integration tests.
    - Check all providers.

Note that FMP does not have any query parameters other than start/end dates.

Screenshot 2024-05-10 at 1 37 19 PM


openbb_platform/providers/fmp/openbb_fmp/models/economic_calendar.py
```diff
@@ -89,11 +94,46 @@ async def aextract_data(
 
         base_url = "https://financialmodelingprep.com/api/v3/economic_calendar?"
 
-        url = f"{base_url}from={query.start_date}&to={query.end_date}&apikey={api_key}"
-
-        return await amake_request(url, **kwargs)
+        # FMP allows only 3-month windows to be queried, we need to chunk to request.
+        if not query.start_date or query.start_date is None:
```
IgorWounds, line 98: I think that the first part of the conditional statement is all you need. The second one is redundant.

openbb_platform/providers/tradingeconomics/openbb_tradingeconomics/models/economic_calendar.py
```diff
@@ -130,7 +198,8 @@ async def aextract_data(
     ) -> Union[dict, List[dict]]:
         """Return the raw data from the TE endpoint."""
         api_key = credentials.get("tradingeconomics_api_key") if credentials else ""
-
+        if query.group is not None:
```
IgorWounds, line 201: It should be enough to do `if query.group`

deeleeramone: PEP8 recommends always using `if foo is not None` for testing when a variable that defaults to None was set to another value.

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.
Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

Source: https://peps.python.org/pep-0008/#programming-recommendations

openbb_platform/providers/fmp/openbb_fmp/models/economic_calendar.py
deeleeramone: Moved this section to `transform_query`.

IgorWounds (approved):

deeleeramone:
> For some reason I can't get tradingeconomics to work inside the CLI
> 
> cc: @hjoaquim

So in the CLI I was getting:

```
2 validation errors for calendar
extra_params.importance
  Input should be 'low', 'medium' or 'high' 
    For further information visit https://errors.pydantic.dev/2.7/v/literal_error
extra_params.group
  Input should be 'interest_rate', 'inflation', 'bonds', 'consumer', 'gdp', 'government', 'housing', 'labour', 'markets', 'money', 'prices', 'trade' or 'business' 
    For further information visit https://errors.pydantic.dev/2.7/v/literal_error
```

Adding `None` to the Literal for each resolved that.

IgorWounds:
Don't we want the Literal to be Optional? Sending a None through the API does not always work and can break.

@hjoaquim you already have some screenshots on this.

hjoaquim (approved):

PR #6401, opened 2024-05-13
[BugFix] Fix broken `--sheet-name` argument

hjoaquim:

    - Broken `--sheet-name` argument

    -  Better handling of results (which may or may not be `OBBjects`)
    - `sheet-name` needed to be converted from a list to a string.

    - Buf fix -> ux improvement.

4. **Testing Done**:

    - Writing to an Excel with a sheet name: `/equity/fundamental/ $ balance --symbol MSFT --period quarter --export msft_financials.x
lsx --sheet-name balance`
    - Writing to an existing Excel (with another sheet name): `/equity/fundamental/ $ cash --symbol MSFT --period quarter --export msft_financials.xlsx
 --sheet-name cash`

    - Never mind the confusion with the number of commits, this was branched off in the wrong place.


deeleeramone:
When I don't specify `sheet-name`, I get an error.

```
cannot access local variable 'sheet_name' where it is not associated with a value
```

IgorWounds (approved):

hjoaquim:
> When I don't specify `sheet-name`, I get an error.
> 
> ```
> cannot access local variable 'sheet_name' where it is not associated with a value
> ```

I believe this fixes: https://github.com/OpenBB-finance/OpenBBTerminal/pull/6401/commits/45bfac29bb22540e524d24c6c6f0a73f381cab1d

Thank you @deeleeramone 

If it still fails somehow, please share the exact command!

deeleeramone (approved):
`--sheet-name` works, on to the next!

PR #6424, opened 2024-05-16
[BugFix] - `dev_install.py` updates

montezdesousa:
1. **Why**?

    - `dev_install.py` uses `toml` which needs to be manually installed in each environment, it's not in the dependency tree

2. **What**?

    - Replace it with `tomlkit` which is used by `poetry`, thus it's already in the dependency tree
    - Added a `Loading...` print statement to cli `main` since the application cold start takes a while and looks hanged

3. **Impact**

    - No need to `pip install toml` anymore, `pip install poetry` and it will work

4. **Testing Done**:

    - `python dev_install.py`
    - `python dev_install.py -e`

### DIFF
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -19,7 +19,7 @@ python = "^3.8.1,<3.12"
 # OpenBB dependencies
 openbb = { version = "^4.2.0", extras = ["all"] }
 
-[tool.poetry.group.external.dependencies]
+# CLI dependencies
 prompt-toolkit = "^3.0.16"
 rich = "^13"
 python-dotenv = "^1.0.0"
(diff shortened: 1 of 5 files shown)


IgorWounds (approved):

cli/pyproject.toml
```diff
@@ -19,7 +19,7 @@ python = "^3.8.1,<3.12"
 # OpenBB dependencies
 openbb = { version = "^4.2.0", extras = ["all"] }
 
-# Terminal dependencies
+[tool.poetry.group.external.dependencies]
```
piiq, line 22: group will not be picked up by pip

piiq (changes requested):

cli/pyproject.toml
montezdesousa: Yeah, I removed that then.

piiq (approved):

dijonkitchen:
Not sure this works with the official isolated installer: https://python-poetry.org/docs/#installing-with-the-official-installer

PR #6432, opened 2024-05-18
[BugFix] FMP Screener - Convert sector param choices to lower_snake_case

deeleeramone:
1. **Why**?:

    - Valid sector choices to the FMP API include spaces, but in practice we don't want those as parameter choices in the Literals.
    - SEC Form Types also get same treatment.

2. **What**?:

    - Defined the SECTORS Literal in `lower_snake_case` and then converted back before the URL generation.
    - Benzinga aliases as-per -> #6327 

3. **Impact**:

    - This should reduce the chance of unexpected behaviour across all interfaces.
    - The sector parameter will now need to be entered as lower_snake_case, everything else remains the same.

4. **Testing Done**:

    - Used the function and params. No change to the response.


IgorWounds (approved):

hjoaquim:
is this a potential breaking change for Platform clients (like PRO)?
cc @andrewkenreich

openbb_platform/providers/sec/openbb_sec/utils/definitions.py
```diff
@@ -163,8 +163,8 @@
     "CORRESP",
     "CT ORDER",
```
hjoaquim, line 164: this one got unnoticed

PR #6437, opened 2024-05-20
[BugFix] Only update credentials on login

montezdesousa:
1. **Why**?

    - Login resets the user preferences to their default

    - If the user changes some preference in `user_settings.json` file and then logs in, it ignores the file changes and uses defaults instead

2. **What**?

    - Make sure only credentials are updated (merged with local credentials) when login

3. **Impact**

    - NA

4. **Testing Done**:

    - Change preferences in `user_settings.json`. Login into the Hub with different credentials than you have in `user_settings.json`. Confirm no preferences were updated and local credentials were updated with values set in the hub.


IgorWounds (approved):

IgorWounds:
This works well

PR #6439, opened 2024-05-20
Remove xlsx export from CLI's interactive tables

piiq:
This PR removes the option to export tables to excel from the UI of the interactive table.

The xlsx export in CLI's interactive tables relied on a vulnerable version of the xlsx npm package.

The SheetsJS project that is behind xlsx package does not distribute the latest packages via npm making it a bit harder to follow updated.

Since we have xlsx export functionality on the CLI and platform level and there is still an option to export the data from a the interactive table in machine readable format, I decided to remove the vulnerable dependency instead of trying to patch or find a replacement.

cc @jose-donato @tehcoderer 

---

Additional changed following review:

- remove posthog from template html of interactive tables
- fix deploy script
- update css to include better word wrap
- update the table.html in charting extension


hjoaquim (commented):
If it's a table, I still have the options to download it as an Excel file (and it works):
image

piiq:
@hjoaquim I've added a few commits that update the built html in the charting library and the corresponding build script.

I've also removed the posthog code and incorporated the css that @deeleeramone used in the template html into the index.css file.

hjoaquim (approved):
Tables look good, and the export functionalities are working perfectly now. Building the `openbb-charting` wheel and use it also works as expected.

IgorWounds:
Tested this and it works and performs well! Thanks @piiq !

PR #6443, opened 2024-05-20
[BugFix] Bump anyio & sync pre-commit, openbb-devtools and GitHub workflows

montezdesousa:
1. **Why**?

    - Last release `anyio` version had to be pinned to v3->v4 to avoid import error
    - `pre-commit` hooks were out of sync vs `openbb-devtools` which defines GitHub workflow dependency versions
    - `ruff` version use in `openbb-core` was too restrictive and outdated
    - `mypy` errors were different in GitHub workflow and locally

2. **What**?

    - Bump `anyio` ^3.7.0 -> ^4.3.0
    - Bump `ruff` ^0.1.6 -> >=0.1.6,<=0.4.4 (devtools uses ^0.4.4)
    - Bump `black` 24.1.0 -> 24.4.2 (devtools)
    - sync `pre-commit` hooks with `openbb-devtools`
    - Small `ruff`, `mypy`, `pydocstyle` fixes
    - Move dev dependency group from core to devtools `pyptoject.toml`
 
3. **Impact**:
    
    - Fixes https://github.com/OpenBB-finance/OpenBBTerminal/issues/6163 

5. **Testing Done**:

    - NA


openbb_platform/extensions/devtools/pyproject.toml
```diff
@@ -9,16 +9,19 @@ packages = [{ include = "openbb_devtools" }]
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"               # scipy forces <4.0 explicitly
-ruff = "^0.1.6"
+ruff = "^0.4.4"
 pylint = "^3.0.2"
 mypy = "^1.6.1"
 pydocstyle = "^6.3.0"
-black = "^24.1.0"
+black = "24.1.0"
```
piiq, line 16: This pin won't work because black < 24.3.0 is vulnerable
image

piiq (changes requested):
Please bump black to latest

piiq (approved):
Let's give it a try. I would also remove the devtools section from cli, but that's for a different PR

IgorWounds (approved):

PR #6446, opened 2024-05-21
[Feature] Set provider fallback based on credentials

montezdesousa:
1. **Why**?

    - First time users are immediately blocked if they don't setup credentials before using `openbb`
    - v3 used to have a sources priority configuration
    
2. **What**?

Create a provider fallback mechanism that checks if required credentials are set before using a given provider.

The priority is the following:

#### I. Provider specified in the command (if you have no credentials it will display the same message as before)
```python
obb.equity.price.quote("AAPL", provider="fmp") -> "fmp"
```

#### II. Configured priority list (or just 1 provider as before)

```json
{
    "credentials": {},
    "preferences": {},
    "defaults": {
        "commands": {
            # 1 provider
            "equity.price.historical": { "provider": "fmp" },
            # priority list
            "equity.price.quote": { "provider": ["foo", "bar", "fmp"] }
        }
    }
}
```

```python
obb.equity.price.historical("AAPL") -> "fmp"
obb.equity.price.quote("AAPL") -> "fmp" (tries "foo" and "bar" first)
```

#### III. Default priority is the list of providers for the command, sorted alphabetically
         
```python
obb.equity.price.quote("AAPL") -> "yfinance" (tries "fmp" and "intrinio" first or other providers if they are installed)
```

#### IV. Raises exception

```json
{
    "credentials": {},
    "preferences": {},
    "defaults": {
        "commands": {
            "equity.price.quote": { "provider": ["foo", "bar", "baz"] }
        }
    }
}
```

```python
obb.equity.price.quote("AAPL")

OpenBBError: 
[Error] -> Provider fallback failed, please specify the provider or update credentials.
[Providers]
  * 'foo' -> not found
  * 'bar' -> not found
  * 'baz' -> not found
```

- Replaced `routes` by `commands` in the `Defaults` configuration class, since this feature is relevant to Python users only. Routes is an API concept. Will keep it as alias, accept slash split commands and set deprecation warning.

3. **Impact**:

    - Reduces friction for first time users. If the command supports a provider without credential it will be used, no setup needed.
    - LLMs can rely on configuration or just fallback to whatever provider works.

4. **Testing Done**:

    - Examples above and some combinations


IgorWounds (approved):
We could wait for the docs to be merged here so we write about it.

PR #6451, opened 2024-05-21
[BugFix] Add `json_schema_extra` With Choices To Provider 'period' Field

deeleeramone:
1. **Why**?:

    - Parameters defined as a Literal do not have a list of `choices` for the schema.

2. **What**?:

    - Defines the Literal choices in `json_schema_extra` for each provider's `period` field in balance/cash/income endpoints.
    - Cleans up the "growth" models that are only used by FMP to be more like the balance/cash/income models.
    - Adds `json_schema_extra={"x-unit_measurement"}` wherever was missing.

3. **Impact**:

    - Does not change any params.
    - Pro does not use the growth endpoints, the cleanup here will not impact existing widgets.
    - Improved schema definitions that will help anyone relying on `openapi.json` or `reference.json`
    - Rename a couple of Polygon fields to align with others.

4. **Testing Done**:

    - Integration and unit tests.
    - 'regular' usage.


IgorWounds (approved):

hjoaquim:
A lot of changes. Please merge only after release 4.2.1

openbb_platform/providers/polygon/openbb_polygon/models/income_statement.py
```diff
@@ -22,41 +23,44 @@ class PolygonIncomeStatementQueryParams(IncomeStatementQueryParams):
 
     __alias_dict__ = {"symbol": "ticker", "period": "timeframe"}
 
-    period: Literal["annual", "quarter", "ttm"] = Field(default="annual")
-    filing_date: Optional[date] = Field(
+    period: Literal["annual", "quarter", "ttm"] = Field(
+        default="annual",
+        json_schema_extra={"choices": ["annual", "quarter", "ttm"]},
```
hjoaquim, line 28: For future reference:
For the CLI to leverage custom choices, one only need to populate the `json_schema_extra` with them if it's not possible to use such options inside the `Literal` - cases where we want to leverage comma separated lists and as such the type need to be `str` instead of `Literal`.
Here, we're just duplicating the same info.

hjoaquim (approved):

openbb_platform/providers/polygon/openbb_polygon/models/income_statement.py
deeleeramone: From what I've experienced, the act of defining almost any Literal is an exercise in duplication, and this appears to be a limitation of Python 3.9, because you cannot dynamically define a Literal.

PR #6459, opened 2024-05-23
StreamFetcher POC

IgorWounds:

    - The OpenBB Platform had a downside of not being able to support out-of-the-box WebSocket connections.
    - It was mentioned by a couple of users of the Platform and also OpenBB Terminal Pro.

    - Adapt the OpenBB Platform code to support WebSocket connections.
    - Introduces the `atransform_data` method.
    - Introduces a Binance live crypto price data example.
    - Standardizes data on the fly.

    - Allows for handling WebSockets and increases the complexity of the routers.

4. **Testing Done**:

To test on the API interface do:

```python

import requests
import json
url = "http://127.0.0.1:8000/api/v1/crypto/price/live?symbol=ethbtc&lifetime=10"

def get_json_events_sync():
    with requests.get(url, stream=True) as r:
        for line in r.iter_lines():
            yield json.loads(line.decode("utf-8"))

for event in get_json_events_sync():
    print(event)
```

And on the Python interface do:

```python
from openbb import obb

ws = obb.crypto.price.live(symbol="ethbtc", lifetime=15)
async for data in ws.body_iterator:
    print(data)
```

    - This can be further polished by introducing a custom decorator for handling routing when multiple providers share the same WebSocket. For now, we don't need this until we see there is usage and requests for such a feature.
    - Need to check how this is handled on the CLI. @hjoaquim


openbb_platform/extensions/crypto/openbb_crypto/crypto_router.py
```diff
@@ -1,5 +1,7 @@
 """Crypto Router."""
 
+# import asyncio
+
 from openbb_core.app.model.command_context import CommandContext
```
montezdesousa, line 5:
```suggestion

from openbb_core.app.model.command_context import CommandContext
```

hjoaquim:
Thoughts:

- add new provider to pyproject.toml and dev install
- use OBBject as returning type in router (maybe add a prop to OBBject to keep the streaming fetcher)
- the new command live  should've provider as argument; should also appear on the reference

openbb_platform/core/openbb_core/app/static/package_builder.py
deeleeramone, line 1: This will be a blocker:

Screenshot 2024-05-29 at 3 31 05 PM

openbb_platform/extensions/crypto/openbb_crypto/price/price_router.py
```diff
@@ -56,3 +61,15 @@ async def historical(
 ) -> OBBject:
     """Get historical price data for cryptocurrency pair(s) within a provider."""
     return await OBBject.from_query(Query(**locals()))
+
+
+@router.command(methods=["GET"])
+async def live(
+    symbol: str = "ethbtc", lifetime: int = 10
+) -> BinanceCryptoHistoricalData:
```
deeleeramone, line 69: How can we make this a Provider Interface method? There could be any number of providers that have a WS connection, like here - https://site.financialmodelingprep.com/developer/docs#crypto-websocket

openbb_platform/providers/binance/openbb_binance/models/crypto_historical.py
```diff
@@ -0,0 +1,98 @@
...
+        alias="e",
+    )
+    quote_asset_volume: Optional[str] = Field(
+        default=None,
+        description="Total traded quote asset volume",
+        alias="q",
+    )
+
+
+class BinanceCryptoHistoricalFetcher(Fetcher):
```
deeleeramone, line 55: I think this model name and Fetcher class is deceiving. It is not at all like the other provider/standard models with this name.  This is not historical data, and the pattern itself is significantly different. It needs to be differentiated. 
Like how you have added an additional `atransform` method, there should be something like `wsconnect` that accepts a URL and **kwargs where the core logic can be reused.


There should be a WebSocket standard model which ships with the connection logic that others can

openbb_platform/providers/binance/openbb_binance/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    name="binance",
+    website="https://api.binance.com",
+    description="""BizToc uses Rapid API for its REST API.
+You may sign up for your free account at https://rapidapi.com/thma/api/binance.
+
+The Base URL for all requests is:
+
+    https://binance.p.rapidapi.com/
+
+If you're not a developer but would still like to use Biztoc outside of the main website,
```
deeleeramone, line 18: This description is half BizToc and half maybe Binance.

openbb_platform/providers/binance/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
+name = "openbb-binance"
+version = "1.2.1"
```
deeleeramone, line 3: Version should be 1.0.0

openbb_platform/providers/binance/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Biztoc Provider tests."""
```
deeleeramone, line 1: Biztoc -> Binance

openbb_platform/extensions/crypto/openbb_crypto/price/price_router.py
```diff
@@ -56,3 +60,16 @@ async def historical(
 ) -> OBBject:
     """Get historical price data for cryptocurrency pair(s) within a provider."""
     return await OBBject.from_query(Query(**locals()))
+
+
+@router.stream(methods=["GET"])
+async def live(symbol: str = "ethbtc", lifetime: int = 10, tld: str = "us") -> OBBject:
+    """Connect to Binance WebSocket Crypto Price data feed."""
```
deeleeramone, line 67: This endpoint will not generate any documentation or descriptions because it is not using the ProviderInterface.

PR #6463, opened 2024-05-24
[Feature] Optional choices

hjoaquim:

    - There are some providers that allow custom choices other than the finite amount of choices other providers offer - in order not to limit usage, we need to be able to also use those custom choices. i.e., for those providers, the choices are merely a suggestion.

    - Introduction of `optional_choices` attribute as a way of setting custom choices 
        - this will still trigger auto-completion but set the `choices` of a given argument to `None` before parsing, allowing any value to be user
        - this is still fine because all providers are pydantic models and will perform their own validation

    - Took the opportunity to do some improvements in structure / refactoring, although nothing major, it improves readability.

    - Usage and validation remains, as well as UX. Small enhancement since the users will not be able to query with custom values for those fields and providers.

4. **Testing Done**:

    - `equity/price/historical --symbol aapl --provider polygon --interval 2h`
    
    > `equity/price/historical` for `tmx` and `polygon` with the `interval` argument are the only place where this applies afaik.

    - Please try different scenarios for other commands and providers and check if autocompleting and validation is still accurate.


hjoaquim:
@deeleeramone maybe wrong env? seems good on my end:
image

deeleeramone:
> @deeleeramone maybe wrong env? seems good on my end

@hjoaquim, is it possible there is a discrepancy in behaviours between operating systems and Python versions?

Fresh environment install, Python 3.11, macOS:

Screenshot 2024-05-27 at 8 39 55 AM

hjoaquim:
> > @deeleeramone maybe wrong env? seems good on my end
> 
> @hjoaquim, is it possible there is a discrepancy in behaviours between operating systems and Python versions?
> 
> Fresh environment install, Python 3.11, macOS:
> 
> Screenshot 2024-05-27 at 8 37 26 AM
> 
> Screenshot 2024-05-27 at 8 39 55 AM

so strange! i'm getting good results with the exact same arguments:

image

are you sure right branch with latest changes? just double checking
the os could matter, but this is such a small change and at a logical level that i can't believe that's it

montezdesousa:
Works fine for me.

Unrelated topic: the data sorting is not good for previews (Excel also suffers from this).

If you limit output to the first X rows by default you see the oldest data first. I'd change the sorting to display the most recent data first.

montezdesousa (approved):

deeleeramone:
> Unrelated topic: the data sorting is not good for previews (Excel also suffers from this).
> 
> If you limit output to the first X rows by default you see the oldest data first. I'd change the sorting to display the most recent data first or have some sorting preference, because the argument is only there for polygon.

Time series data requires being sorted by ascending for most downstream processes, this is why they are all conformed to this. If sorting to show the newest first is implemented, it should not impact exporting. Unrelated still, sorting does not work on numbers because they have been converted to strings.

deeleeramone:
> are you sure right branch with latest changes? just double checking the os could matter, but this is such a small change and at a logical level that i can't believe that's it

@hjoaquim, for it works with Python 3.9, but not 3.11.

deeleeramone (changes requested):
When I use Python 3.9 it works, but it does not when I use Python 3.11

hjoaquim:
> When I use Python 3.9 it works, but it does not when I use Python 3.11

it was indeed not working as expected for python 3.11; i believe just out of coincidence was working for 3.9
it's now fixed (hopefully) with the latest commit!

deeleeramone:
> it was indeed not working as expected for python 3.11; i believe just out of coincidence was working for 3.9 it's now fixed (hopefully) with the latest commit!

That did the job!

deeleeramone (approved):

PR #6470, opened 2024-05-27
[BugFix] Add openbb-charting to CLI's pyproject.toml

deeleeramone:
1. **Why**?:

    - This package is not declared as a dependency, although it is required.

2. **What**?:

    - Add to `pyproject.toml`
    - Update lock file.


hjoaquim (changes requested):
Why is this needed? The `extras = ["all"]` already installs the charting extension.
On a clean env after `pip install openbb-cli`:
image

deeleeramone:
> Why is this needed? The `extras = ["all"]` already installs the charting extension. On a clean env after `pip install openbb-cli`

I did a CLI install without explicitly installing `all` packages, and I got a ModuleNotFound error.

`openbb-charting` is a direct requirement of CLI, regardless of what provider/extras packages you choose to install. So this guarantees that it is installed.

hjoaquim:
> > Why is this needed? The `extras = ["all"]` already installs the charting extension. On a clean env after `pip install openbb-cli`
> 
> I did a CLI install without explicitly installing `all` packages, and I got a ModuleNotFound error.
> 
> `openbb-charting` is a direct requirement of CLI, regardless of what provider/extras packages you choose to install. So this guarantees that it is installed.
> 
> @hjoaquim, it is "optional" in the `lock` file, instead of required.
> 

Looks like that section `[package.dependencies]` is referring to how those dependencies are specified to the `[package]` (which is `openbb`).

Also on the lock file, you then have each of those dependencies specified:

image

On a clean colab notebook, everything goes accordingly:

image

How can I replicate the issue?

deeleeramone:
> How can I replicate the issue?

Try: `python dev_install.py --cli`

hjoaquim:
> > How can I replicate the issue?
> 
> Try: `python dev_install.py --cli`

Ohhh, that's the issue, you're not supposed to install the CLI without the extra dependencies! If you want to install it you should use `python dev_install.py --extras --cli`

If you look into `dev_install.py`, what it does in order to install the CLI is to "pop" the `openbb` dependency from the `pyproject.toml` and install the remaining; which assumes all the other Platform dependencies were already installed.

deeleeramone:
> Ohhh, that's the issue, you're not supposed to install the CLI without the extra dependencies!

Hence why the required components need to be explicitly declared in the TOML and lock files.

hjoaquim:
> > Ohhh, that's the issue, you're not supposed to install the CLI without the extra dependencies!
> 
> Hence why the required components need to be explicitly declared in the TOML and lock files. The charting extension is like `openpyxl`, a required component that is not a required item for Platform. But, I should still be able to install CLI with no provider or router extensions installed.

The `openbb-charting` dependency will always be installed for the end user, as all the other `openbb` extensions.
It doesn't make sense to me that we add it to the TOML file just because we're manipulating the file in the dev install script so that it's easier to get everything installed for **development** purposes. 
For the users it will be the exact same result in the end, for development you'll get the dependency in non-editable mode (unless you also change the `dev_install.py` script) which is not desired.

deeleeramone:
> The `openbb-charting` dependency will always be installed for the end user, as all the other `openbb` extensions.

This goes against the extension framework and the ability to pick-and-choose what components you actually want installed. Life is way easier if you don't have to install Econometrics, which requires Visual C++ or GCC (maybe 20 GB of 'stuff' that just has to live on your machine) before you can even install it. Macs and PCs do not come with these out-of-the-box.

hjoaquim:
> This goes against the extension framework and the ability to pick-and-choose what components you actually want installed. Life is way easier if you don't have to install Econometrics, which requires Visual C++ or GCC (maybe 20 GB of 'stuff' that just has to live on your machine) before you can even install it. Macs and PCs do not come with these out-of-the-box.

Although I tend to agree, it was an internal decision to make the CLI ship with all the available `openbb` extensions.

PR #6490, opened 2024-06-06
[Enhancement] Fixed Income Enhancements

deeleeramone:
1. **Why**?:

    - SOFR belongs under the `rate` section with the other similar items.
    - Ambiguous standard model for EFFR and CP.
    - Moody's bond indices can be moved to be with the other bond indices.

2. **What**?:

    - Renames the "FED" standard model as: `federal_funds_rate`.
    - Renames the "CP" standard model as: `commercial_paper`.
      - multiple items allowed. 
    - Add model definitions for all fields.
    - `EFFR` & `SOFR`: Standardizes the data from both FRED and the NY Fed API.
    - `openbb-federal-reserve`: Switch the source of the data to the NY Fed API and return a JSON.
    - `openbb-fred`: Add transform and frequency aggregation parameters. Return series metadata.
    - Moves `SOFR` to the `fixedincome.rate` router and add deprecation to the original item.
    - Add `federal_reserve` as provider to `SOFR`.
    - Return percentiles and notional volume for `EFFR` and `SOFR`.
    - Remove discontinued series from Ameribor and define the model with more information.
    - Add frequency and transform to Ameribor and allow multiple items. Renames "parameter" -> "maturity"
    - Adds `overnight_bank_funding_rate` as new route and data from FRED and Federal Reserve.
    - Adds "seasoned_corporate" as an "index" within the "us" category of `bond_indices`.
      - This returns the Moody's bond indices and the other endpoint can be retired - deprecation warning added.
    - Renames "hqm" files as "high_quality_market".
      - Changes the way data is fetched for "hqm" so that only 1 request per-date is made, and gets the 0.5-100 year range of yield curve for the spot rate.
      - Multiple dates allowed for 'hqm'.
    - Add Treasury Minus Fed Funds to Yield Curve as choice for FRED.
    - Renames "estr" -> "euro_short_term_rate"
      - Removes "parameter" paremeter and returns all items to a defined model.
    - Update associated tests.

3. **Impact**:

    - The only breaking change is the FRED parameter named, 'parameter'.
      - The same results are obtained by using the `frequency` or `maturity` parameters.
    - Additional data fields are returned from both providers to this endpoint.
      - Target ranges, percentiles, notional volume.
    - All fields returning a % value are normalized and have `json_schema_extra` definitions for front end.
    - Improved type hinting.
    - Much faster response time from `obb.fixedincome.corporate.hqm`

4. **Testing Done**:

    - Unit, Integration tests.
    - CLI

Screenshot 2024-06-06 at 2 47 33 PM

Screenshot 2024-06-06 at 2 50 14 PM

Screenshot 2024-06-06 at 10 50 36 PM

Screenshot 2024-06-10 at 10 07 19 AM

Screenshot 2024-06-10 at 7 52 09 PM

Screenshot 2024-06-11 at 2 02 44 PM


deeleeramone:
@IgorWounds, gotta cut myself here. No more additions to this PR. 🥲

minhhoang1023 (approved):

openbb_platform/providers/fred/openbb_fred/models/federal_funds_rate.py
```diff
@@ -0,0 +1,205 @@
...
+            transformed_params["start_date"] = (
+                datetime(2016, 1, 1).date()
+                if params.get("effr_only") is False
+                else None
+            )
+        if params.get("end_date") is None:
+            transformed_params["end_date"] = now
+        return FredFederalFundsRateQueryParams(**transformed_params)
+
+        return FredFederalFundsRateQueryParams(**params)
```
IgorWounds, line 145: why do we have a second return statement here?

deeleeramone: Good question, could be merge conflict thing, but I find it amazing that the linters are not complaining about it.

PR #6491, opened 2024-06-07
[Feature] Store defaults in hub account

montezdesousa:
^


assets/extensions/provider.json
```diff
@@ -7,9 +7,9 @@
         "credentials": [
             "alpha_vantage_api_key"
         ],
-        "v3Credentials": [
-            "API_KEY_ALPHAVANTAGE"
-        ],
+        "deprecatedCredentials": {
```
hjoaquim, line 10: why are deprecated keys equal to the "in-use" keys now?

montezdesousa: It maps the old credential name to the new (the idea is to cover providers with multiple keys potentially). Also we can map to `None` if the credential was just removed

hjoaquim (approved):

PR #6505, opened 2024-06-15
[Feature] Intrinio Delayed Options

deeleeramone:
1. **Why**?:

    - This enhances the IntrinioOptionsChainsFetcher to return either current or EOD chains data.

2. **What**?:

    - Small renovation to standard model, but no breaking changes.
    - With no date, Intrinio provider will fetch the current data.
    - Adds additional params to the Intrinio model.

    - Adds new path: `options.snapshots`.
      - Market snapshot for options.

3. **Impact**:

    - `obb.derivatives.options.snapshots(provider="intrinio")` will return a large amount of data from bulk download endpoint.
    - Not advised to have CLI Interactive Tables max rows set to "All" for this, it will definitely crash from overload.

4. **Testing Done**:

    - Snapshots: Skip unit test because the cassette file would be unreasonably large. Integration.

Screenshot 2024-06-14 at 5 22 18 PM

Screenshot 2024-06-14 at 5 43 27 PM

Screenshot 2024-06-15 at 9 51 16 PM


openbb_platform/core/openbb_core/provider/standard_models/options_chains.py
```diff
@@ -31,30 +31,43 @@ def to_upper(cls, v: str) -> str:
...
+    underlying_price: Optional[float] = Field(
+        default=None,
+        description="Price of the underlying stock.",
     )
     contract_symbol: str = Field(description="Contract symbol for the option.")
     eod_date: Optional[dateType] = Field(
         default=None, description="Date for which the options chains are returned."
     )
     expiration: dateType = Field(description="Expiration date of the contract.")
+    dte: Optional[int] = Field(
```
IgorWounds, line 47: I'd vote for a more verbose name on the standard model. I'm fine with `dte` but we did want more verbosity.

IgorWounds (approved):

PR #6506, opened 2024-06-17
[Feature] Improved cached results

hjoaquim:

    - Adds more flexibility to interact with the obtained results.

    - `results` → shows table with all the available results  
    - results always on screen (number of results showing can be managed on settings)
    - using a key instead of the index:
       - `historical --symbol aapl --register_key aapl-hist`
       - `ema --data aapl-hist`
    - `results --idx 0/--key my-key` → displays the actual `OBBject.results` in a table
    - `results --idx 0/--key my-key --chart` → display the chart if available
    - choose which results to keep: `historical --symbol aapl --register_obbject` (the `register_obbject` flag means this result will be opted out, by default we keep the result)

    - 5/40

<details>
  <summary>Impact Analysis</summary>
  
UX: 3
Performance: 0
Compatibility: 0
Security: 0
Scalability: 1
Maintenance: 1
Compliance: 0
Dependency: 0

</details>

4. **Testing Done**:

    - Any of the described above on the `what` section


deeleeramone:
> * choose which results to keep: `historical --symbol aapl --register_obbject` (the `register_obbject` flag means this result will be opted out, by default we keep the result)

This is a lot of keys to type every time (I understand auto complete exists, but still), can we shorten this, something like "--store"?, or "--keep"?

I find the name `--register_obbject` to be counter-intuitive for the action - "False". The name suggests the action would be, "True". Is there an inverse mode where I can keep only the flagged items?

hjoaquim:
> > * choose which results to keep: `historical --symbol aapl --register_obbject` (the `register_obbject` flag means this result will be opted out, by default we keep the result)
> 
> This is a lot of keys to type every time (I understand auto complete exists, but still), can we shorten this, something like "--store"?, or "--keep"?

Given the fact that we have auto-completion, I believe verbosity adds clarity.

> 
> I find the name `--register_obbject` to be counter-intuitive for the action - "False". The name suggests the action would be, "True". Is there an inverse mode where I can keep only the flagged items?

I agree, but taking into consideration the default should be to keep it, what can we name it? Also, despite not being the best name, the message displayed on `--help` clarifies!

IgorWounds (approved):

PR #6510, opened 2024-06-18
[Feature] `to_df` where `results` is a model of arrays.

deeleeramone:
1. **Why**?:

    - `to_df()` does not account for models where the fields are all arrays.
    - Large data sets can be returned as typed arrays, which improves overall execution time.

2. **What**?:

    - Add a condition where if the `results` is a list of 1 item, or just a model, AND all fields in the schema are defined as arrays, use a model dump to create the DataFrame.
    - Implement changes to `options_snapshots` models as a trial case.

3. **Impact**:

    - Improved response time for very large data sets.

    - Changes should only be noticed when the conditions above are realized, and there are not currently any models defined with this pattern.

4. **Testing Done**:

    - Checked `to_df` to ensure that other items have not been affected.
    - The dict[List] pattern is already tested for in `test_obbject`.

Using `relative_rotation` as a test-case that has a mix of arrays and other types, the current behaviour persists while allowing the item as all arrays to be correctly converted.

Screenshot 2024-06-18 at 3 11 39 PM

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/options_snapshots.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/options_snapshots.py
@@ -4,7 +4,7 @@ from datetime import (
     date as dateType,
     datetime,
 )
-from typing import Optional
+from typing import List, Union
 
 from pydantic import Field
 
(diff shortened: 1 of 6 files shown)


montezdesousa (approved):
lgtm

IgorWounds (approved):

openbb_platform/core/openbb_core/provider/standard_models/options_snapshots.py
```diff
@@ -20,56 +20,60 @@ class OptionsSnapshotsQueryParams(QueryParams):
...
-    volume: Optional[int] = Field(
-        default=None,
+    option_type: List[str] = Field(description="The type of option.")
+    volume: List[Union[int, None]] = Field(
+        default_factory=list,
         description=DATA_DESCRIPTIONS.get("volume", ""),
     )
-    open_interest: Optional[int] = Field(
-        default=None,
+    open_interest: List[Union[int, None]] = Field(
```
hjoaquim, line 43: Altough the `reference.json` seems to be well generated (default should be `null`, I believe) this syntax feels weird to me.
Should this be `Optional[List[int]` (and then defaults to `None`)? If we look at the 2nd part of the union it doesn't make sense: `List[None]` ([None,None, None]) - it's possible but odd

PR #6516, opened 2024-06-19
[BugFix] Exclude default values when model dumping

montezdesousa:
1. **Why**?

    - User settings default values are written into `user_settings.json`
    - This is not desirable. If the value is the default, then the user did not change it so there is no point in freezing it into `user_settings.json`
    - It will make it harder to change the defaults in the future if we are writing them to files without user changing it

2. **What**?

    - Exclude defaults when model dumping, did the same for system and when exporting settings to the hub

3. **Impact**

    - NA

4. **Testing Done**:

```python
from openbb import obb
# Check defaults field values are not written into `user_settings.json` when running
obb.account.save()
```


deeleeramone (commented):
I'm not sure that I'm accurately testing this PR because I've found a different bug that is present on develop and this branch. Without understanding what is happening there, I'm not sure what I should be looking for here.

The V3toV4 conversion is no longer working and this is the user-preference that you can currently set on Hub.

Screenshot 2024-06-19 at 12 59 23 PM

deeleeramone:
What I think could be helpful to new users, is if the file is being created ('user_settings.json'), instead of being empty, can we populate the JSON structure with the expected structure instead of an `{}` empty JSON?

Something that helps them when they need to modify the file manually, like below, but with the defined structures for the valid preferences and settings? 

```
  "defaults": {
    "commands": {
      "/route": {"provider":""}
    }
}
```

Where `/route` is a placeholder string that we can evaluate as `""`

montezdesousa:
> I'm not sure that I'm accurately testing this PR because I've found a different bug that is present on develop and this branch. Without understanding what is happening there, I'm not sure what I should be looking for here.
> 
> The V3toV4 conversion is no longer working and this is the user-preference that you can currently set on Hub.
> 
> Screenshot 2024-06-19 at 12 59 23 PM

fixed it now

deeleeramone (approved):

PR #6531, opened 2024-06-21
Security updates

piiq:
- Drop support for python 3.8
- Bump scikit-learn, and urllib3 to patched versions
- Bump ws and braces to patched versions on website and frontend-components
- Bump devtools


piiq:
There is a problem with openbb-devtools. It brings in vcrpy that has a pin of urllib3 that doesn't seem to work nicely with poetry

piiq:
> There is a problem with openbb-devtools. It brings in vcrpy that has a pin of urllib3 that doesn't seem to work nicely with poetry

resolved. vcrpy works fine with later urllib3 on python 3.10. i've bumped the deps in a way our dev install script allows solving the env for 3.9

deeleeramone:
I think the charting extension, and a couple of others, are missing the `.toml` file.

piiq:
need to bump Finviz to 1.0.0 in this security update. the package has been released last week and Python 3.8 support has been dropped. package is `finvizfinance`

openbb_platform/providers/yfinance/pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 packages = [{ include = "openbb_yfinance" }]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.9"
 yfinance = "^0.2.27"
```
deeleeramone, line 12: Current version of yfinance is 0.2.40. This should be bumped.

openbb_platform/providers/tradingeconomics/pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 packages = [{ include = "openbb_tradingeconomics" }]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.9"
 openbb-core = "^1.2.5"
```
deeleeramone, line 12: version of openbb-core should be bumped everywhere, especially if these are not backwards-compatible changes to the library.

deeleeramone:
I'm getting lots of failures with the integration tests. The entire `econometrics` and `technical` modules fail.

deeleeramone:
@piiq, are we enabling 3.12 support in this PR? I'm still being blocked by the project's requirements.

piiq:
> are we enabling 3.12 support in this PR?

yes

IgorWounds (approved):

openbb_platform/providers/tradingeconomics/pyproject.toml
piiq: yes, this is done during the release

PR #6532, opened 2024-06-22
[Feature] Delay/Optimize Imports

deeleeramone:
1. **Why**?:

    - This PR cleans up imports to not run on load, improving load and build time.

2. **What**?:

    - Remove imports of whole modules, only import the items required.
    - Moves non-definition imports to within functions that prevent the import from happening until the code is run.
    - Modifies Package Builder to accept a ForwardRef in "DataProcessingSupportedTypes", and replace with the actual import. The same way it was already done, just replacing "ForwardRef('DataFrame')" with 'DataFrame'
 
3. **Impact**:

    - Faster import times.

    - There *should* be no change, from a user's perspective, in operation.

    - Fundamentally changes the code style to incorporate this pattern.

4. **Testing Done**:

    - Lots of files manually changed, so there is potential for error.
    - WIP

Before:

image

After:

Screenshot 2024-06-26 at 1 58 58 PM

5. **Additional Info:**

    - The CI tests are being held up by, #6531, which is because `finvizfinance` dropped Python 3.8 support and the OpenBB lock files have not been updated in this PR to reflect our dropping of 3.8
    - Test cassettes have been recaptured. I have found one failure, `historical_dividends` from yFinance; this is broken both in the yFinance package and also directly on the Yahoo Finance website.


openbb_platform/core/openbb_core/provider/standard_models/lbma_fixing.py
```diff
@@ -36,12 +33,6 @@ class LbmaFixingQueryParams(QueryParams):
         description=QUERY_DESCRIPTIONS.get("end_date", ""),
     )
 
-    @field_validator("asset", mode="before", check_fields=False)
-    @classmethod
-    def to_lower(cls, v: Optional[str]) -> Optional[str]:
-        """Convert field to lowercase."""
-        return v.lower() if v else v
-
```
montezdesousa, line 44: Why removing this?

montezdesousa (approved):
lgtm

IgorWounds (approved):

openbb_platform/core/openbb_core/provider/standard_models/lbma_fixing.py
deeleeramone: There are defined choices for this, and it's better to avoid computed fields in the standard models unless they are really necessary.

IgorWounds:
@deeleeramone how much does the decrease in import time affect our execution time? If there is a trade-off, is it one we're willing to make?

Cc: @piiq @montezdesousa @hjoaquim

deeleeramone:
> how much does the decrease in import time affect our execution time? If there is a trade-off, is it one we're willing to make?

It only increases time for execution when the imports have not been done yet. Differences should be negligible.

PR #6547, opened 2024-06-27
[Feature] - Add Futures Curve Chart

IgorWounds:

    - To Expand the charting coverage and provide a better UX.

    - Adds a chart to the futures curve endpoint.

    - A new parameter `chart` is available on the command.

4. **Testing Done**:

    - Ensured the chart worked and added tests for it.


deeleeramone:
I have added the possibility for multiple dates, similarly to the `yield_curve` end point, but with a limitation on the historical data from yFinance is for active contracts only. The code isn't the prettiest, but it'll do until we can find a better way to get the entire list of contract symbols associated with the continuation symbol.

Still to do, Cboe VIX Futures.

Screenshot 2024-06-27 at 9 57 40 PM

openbb_platform/extensions/derivatives/integration/test_derivatives_api.py
```diff
@@ -139,7 +151,7 @@ def test_derivatives_futures_curve(params, headers):
 
     query_str = get_querystring(params, [])
     url = f"http://0.0.0.0:8000/api/v1/derivatives/futures/curve?{query_str}"
-    result = requests.get(url, headers=headers, timeout=60)
+    result = requests.get(url, headers=headers, timeout=10)
```
hjoaquim, line 154: might be a bad idea to have such a small timeout on an integration test, will probably cause a false positive.

openbb_platform/extensions/derivatives/openbb_derivatives/derivatives_views.py
```diff
@@ -12,6 +12,215 @@ class DerivativesViews:
...
+        data.charting.to_chart(colors=["green", "red"], title="VIX Futures Curve - 2020 vs. 2024")
+        ```
+        """
+        # pylint: disable=import-outside-toplevel
+        from openbb_charting.core.chart_style import ChartStyle
+        from openbb_charting.core.openbb_figure import OpenBBFigure
+        from openbb_charting.styles.colors import LARGE_CYCLER
+        from openbb_core.app.model.abstract.error import OpenBBError
+        from openbb_core.provider.abstract.data import Data
+        from pandas import DataFrame, to_datetime
```
hjoaquim, line 77: I assume this is due to import time reduction?

openbb_platform/obbject_extensions/charting/openbb_charting/charting.py
```diff
@@ -133,8 +134,13 @@ def get_params(self) -> ChartParams:
         ).replace("/", "_")[1:]
         if hasattr(ChartParams, charting_function):
             return getattr(ChartParams, charting_function)()
-        raise ValueError(
-            f"Error: No chart parameters are defined for the route: {charting_function}"
+
```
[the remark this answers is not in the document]
deeleeramone: It does raise, in the lines above, if there is no route.  After verifying the route exists, it checks ChartParams for that function; if it doesn't exist, then it will return the charting function's help. This allows parameters to be detailed in the code, outside of the charting extension.

hjoaquim (approved):

PR #6553, opened 2024-06-28
[BugFix] Fix weird syntax defining optionals

hjoaquim:
^


openbb_platform/core/openbb_core/provider/standard_models/options_snapshots.py
```diff
@@ -27,53 +27,52 @@ class OptionsSnapshotsData(Data):
...
     )
-    last_size: List[Union[int, None]] = Field(
-        default_factory=list,
+    last_size: Optional[List[int]] = Field(
+        default=None,
         description="Lot size of the last trade.",
     )
-    last_timestamp: List[Union[datetime, None]] = Field(
-        default_factory=list,
+    last_timestamp: Optional[List[datetime]] = Field(
```
deeleeramone, line 55: These need to be defined as a `List[Union[Type, None]` because the "Field" is NOT Optional. Values within the array can be a Union of [Int, None]. `None` is a valid Type.

hjoaquim: Thanks for clarifying, now I understand. Closing the PR then.

openbb_platform/openbb/assets/reference.json
```diff
@@ -2060,119 +2060,119 @@
                 "standard": [
                     {
                         "name": "underlying_symbol",
-                        "type": "str",
+                        "type": "List[str]",
```
hjoaquim, line 2063: @deeleeramone disregarding the changes done in the `.py` files, shouldn't the reference look like this?

deeleeramone: It does. The reference file on `bugfix/security-updates` looks like this. However, it should still be showing as a ListUnion[str, None], and the default value should not be `""`. 

This would be an issue with Reference Generator, and not the model definitions though.

Screenshot 2024-06-28 at 10 46 24 AM

PR #6565, opened 2024-07-08
[Feature] Extract provider to logs

jmaslek:
This is a small PR that touches the logging.

The reason for this is to allow `select provider from data` instead of needing to do weird nested json searching in the logs or having to remember the flattened property name. 

Edge cases: creates an empty DummyProvider if provider_choices is not in kwargs and if provider is not in that model - it will just pass through 'not_passed_to_kwargs'


deeleeramone (commented):
Seems to work, I think you just need to modify the tests to accommodate the new property.

deeleeramone (approved):

IgorWounds (approved):

PR #6577, opened 2024-07-10
moves pro CTA to top

jose-donato:


jose-donato:
> put the tags above the separator.
> 
> Also, you had one job 🤣

i did it directly from github 💀

PR #6585, opened 2024-07-13
[BugFix] Fix `gdp.forecast`

deeleeramone:
1. **Why**?:

    - Running this command downloads a 173MB file, every time.

2. **What**?:

    - Fix the function to hit the endpoint with parameters and exclude non-data from the download.
    - Parameters:
        - `country`: multiple items allowed
        - `frequency`: ["annual", "quarter"]
        - `units`: ["current_prices", "volume", "capita", "growth", "deflator"]

3. **Impact**:

    - File size can now be measured in KBs.
    - Improved response time.
    - New parameters.

4. **Testing Done**:

    - Removed the `skip` in the unit tests.
    - Updated integration tests for parameters.

Screenshot 2024-07-13 at 12 20 35 PM

There is some data that doesn't make any sense, `units='current_prices'`, and no amount of fudge-factor could justify these numbers. Inflation doesn't make Argentina's nominal GDP, expressed in USD, 2X the world's largest economy.

Screenshot 2024-07-13 at 12 33 46 PM


IgorWounds:
@deeleeramone should we send in a warning with this data for users stating that some of it doesn't seem accurate?

deeleeramone:
> @deeleeramone should we send in a warning with this data for users stating that some of it doesn't seem accurate?

Seems like a slippery slope to head down, but I did add a comment in the code highlighting the sus param. Ultimately, it's up to the user to determine the accuracy of the data, correct? I hid it by not returning that as the default.

IgorWounds (approved):

PR #6587, opened 2024-07-15
Create OpenBB Platform Installer with conda constructor

piiq:
This PR adds code for conda constructor that allows creating packaged installer for OpenBB Platform.

The installer will bundle python and package management tools (pip) and install OpenBB Platform from pypi as a part of the installation process.

See `build/conda/README.md` for details


build/conda/README.md
```diff
@@ -0,0 +1,28 @@
+# OpenBB Platform Installer
+
+The code in this folder allows creating a packaged installer for the OpenBB Platform using conda constructor.
+
+The installers can be created for Windows (.exe), MacOS (.pkg) and Linux (.sh).
```
deeleeramone, line 5: For Linux users, they need to have other things installed before attempting to install `openbb[all]`, like the Docker file - [https://github.com/OpenBB-finance/OpenBB/blob/develop/build/docker/platform.dockerfile](https://github.com/OpenBB-finance/OpenBB/blob/develop/build/docker/platform.dockerfile)

deeleeramone: Potentially overcome this with: https://github.com/OpenBB-finance/OpenBB/pull/6590

IgorWounds:
@deeleeramone and @piiq , what is the state of this PR?

piiq:
Good question @IgorWounds!
I would say that this branch is very enjoyable 😂

There are is something missing.

1. We need an update from @andrewkenreich about when the feature to hide (not display) empty columns will be available in Pro. Once that happens @deeleeramone can add the `columnDefs` creation to the script that creates the `widgets.json`.
2. ~~@deeleeramone I've a question. The deepdiff package that the API script uses is not picked up by my language server. I've taken a look and we don't have `deepdiff` mentioned as a dependency and it's not a part of the standard library. Did I miss it, or do we need to add it somewhere?~~ Disregard, I'm smooth brained and I've missed it in one of the pyproject tomls.

piiq:
@deeleeramone I think found a bug. Initially I've deleted the api keys from my user_settings. When launching the api I've used PAT and asked to persist the settings.

Then I've discovered that the credentials were added to the user_settings as escaped quoted strings (see image) (thus they were not picked up by the platform api correctly when i was doing a demo

image

Additionally I see 2 sections "credentials" and "preferences" that have the same content. Not sure if this needs to be this way

deeleeramone:
> @deeleeramone I think found a bug. Initially I've deleted the api keys from my user_settings. When launching the api I've used PAT and asked to persist the settings.
> 
> Then I've discovered that the credentials were added to the user_settings as escaped quoted strings (see image) (thus they were not picked up by the platform api correctly when i was doing a demo

Oh wow, that's not expected at all. Lookiing. Thanks!

deeleeramone:
> @deeleeramone I think found a bug. Initially I've deleted the api keys from my user_settings. When launching the api I've used PAT and asked to persist the settings.
> 
> Then I've discovered that the credentials were added to the user_settings as escaped quoted strings (see image) (thus they were not picked up by the platform api correctly when i was doing a demo

> Additionally I see 2 sections "credentials" and "preferences" that have the same content. Not sure if this needs to be this way

These items should be resolved now, allegedly.

jmaslek (approved):

IgorWounds (approved):

PR #6596, opened 2024-07-20
Highlight agents repo on README

DidierRLopes:


deeleeramone:
I think we're over here now: https://github.com/OpenBB-finance/openbb-docs

deeleeramone (approved):

Issue #6573, opened 2024-07-10
[Bug] [CLI] Argparse can't handle providers where provider #1 field = 'str' and provider #2 field = choices=[CHOICES].


deeleeramone:
**Describe the bug**
^

Constrained choices in one provider impacts the use of all other providers.

**To Reproduce**
```
/derivatives/futures/curve --symbol CL --provider yfinance
```

**Screenshots**

![Screenshot 2024-07-09 at 7 39 19 PM](https://github.com/OpenBB-finance/OpenBB/assets/85772166/84ec6447-1580-4f88-b56d-2e26df259ddd)

hjoaquim:
This might be an edge case and need to look into how a standard required argument deals with provider specific argument with choices - we've solved a similar issue in the past but for non required arguments.

Another thing is, does it make sense for `cboe` to only have those two symbols?
![image](https://github.com/user-attachments/assets/4bb3aef3-24ea-4916-adc0-99c666944792)

deeleeramone:
> Another thing is, does it make sense for `cboe` to only have those two symbols?

Yes, this is explicitly defined because those are the only symbols that can be returned by the provider.

PR #6599, opened 2024-07-22
[BugFix] Fix ReferenceGenerator Unions and Choices

deeleeramone:
1. **Why**?:

    - `reference.json` is creating too many Unions
    - "choices" for a provider contaminate other providers.
    - Expose "period" and "interval" parameters to `reference.json` and `openapi.json`

2. **What**?:

Before:
before1

After:

after1

Before:

before2

After:

Screenshot 2024-07-22 at 7 07 05 PM

3. **Impact**:

    - Where 'multiple_items_allowed', and there are defined choices, they are moved to `__json_schema_extra__` instead of defining under `Field`.
    - Less redundant type references.
    - "choices" for one provider do not force those "choices" on other providers.
    - `reference.json` more closely resembles `openapi.json`.
    - Financial statement "period" choices are now in `openapi.json`.
    - `x.price.historical` "interval" choices are now in `openapi.json`.

4. **Testing Done**:

    - Ran PackageBuilder tests.
    - Manual inspection of docstrings and `reference.json`
    - Check CLI `--help` dialogues.

The schema follows: Param -> schema -> (title has names of all providers) provider (if there are defined choices) -> choices

Screenshot 2024-07-24 at 2 27 23 PM


IgorWounds:
@andrewkenreich can you take a look if this breaks anything on Terminal Pro?

openbb_platform/core/openbb_core/app/provider_interface.py
```diff
@@ -253,14 +253,22 @@ def _create_field(
         annotation = field.annotation
 
         additional_description = ""
+        choices: Dict = {}
```
hjoaquim, line 256: this variable name is misleading as it doesn't take only the choices but the extra info (being it choices and multiple items allowed info)

openbb_platform/core/openbb_core/app/provider_interface.py
```diff
@@ -253,14 +253,22 @@ def _create_field(
...
         if extra := field.json_schema_extra:
-            providers = []
-            for p, v in extra.items():  # type: ignore[union-attr]
+            providers: List = []
+            for p, v in extra.items():  # type: ignore
                 if isinstance(v, dict) and v.get("multiple_items_allowed"):
                     providers.append(p)
+                    choices[p] = {"multiple_items_allowed": True, "choices": v.get("choices")}  # type: ignore
```
hjoaquim, line 262: why are we assuming the `multiple_items_allowed` will be true here? shouldn't it be getting the value as it was before? `v.get("multiple_items_allowed")`

openbb_platform/core/openbb_core/app/provider_interface.py
```diff
@@ -253,14 +253,22 @@ def _create_field(
...
                 elif isinstance(v, list) and "multiple_items_allowed" in v:
                     # For backwards compatibility, before this was a list
                     providers.append(p)
+                    choices[p] = {"multiple_items_allowed": True, "choices": None}  # type: ignore
+                elif isinstance(v, dict) and v.get("choices"):
+                    choices[p] = {
+                        "multiple_items_allowed": False,
+                        "choices": v.get("choices"),
```
hjoaquim, line 270: does it make sense to have the choices here? are they needed for the command's execution?

openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -1742,25 +1754,48 @@ def get_paths(cls, route_map: Dict[str, BaseRoute]) -> Dict[str, Dict[str, Any]]
...
                             standard_model, "QueryParams", provider
                         )
                     )
+
                     # Adds provider specific data fields to the reference
                     reference[path]["data"][provider] = cls._get_provider_field_params(
                         standard_model, "Data", provider
                     )
+
+                    # Remove choices from 'standard' if choices for a parameter exist
```
hjoaquim, line 1776: if there's standard choices why do we remove it in favor of provider choices? i'd say standard should prevail

openbb_platform/core/openbb_core/provider/standard_models/spot.py
```diff
@@ -32,7 +32,6 @@ class SpotRateQueryParams(QueryParams):
     category: str = Field(
         default="spot_rate",
         description="Rate category. Options: spot_rate, par_yield.",
-        json_schema_extra={"choices": ["par_yield", "spot_rate"]},
```
hjoaquim, line 35: why removing the standard choices?

openbb_platform/providers/alpha_vantage/openbb_alpha_vantage/models/equity_historical.py
```diff
@@ -28,7 +28,10 @@ class AVEquityHistoricalQueryParams(EquityHistoricalQueryParams):
     Source: https://www.alphavantage.co/documentation/#time-series-data
     """
 
-    __json_schema_extra__ = {"symbol": {"multiple_items_allowed": True}}
+    __json_schema_extra__ = {
```
hjoaquim, line 31: could this be done at the `Field` level instead of model? the `OpenBBField` is expecting choices to be there - this would then influence the CLI

image

hjoaquim (commented):
Left couple of comments for clarification.

Altough the `choices` are now only available on the reference since they're defined at the model level instead of the field, it looks like the CLI is still able to pick them up since its defined on `reference.json`.

CLI behavior seems good and choices are not conflicting between providers:

image

The help dialog also seems good. Take into consideration that since `argparse` doesn't allow repeated arguments, the help dialog for country will have all the choices regardless the provider - the choices should not be validated at the CLI (by argparse) but instead at the Platform (by pydantic); which seems to be happening as the image above shows.

openbb_platform/core/openbb_core/provider/standard_models/equity_historical.py
```diff
@@ -20,10 +20,6 @@ class EquityHistoricalQueryParams(QueryParams):
     """Equity Historical Price Query."""
 
     symbol: str = Field(description=QUERY_DESCRIPTIONS.get("symbol", ""))
-    interval: Optional[str] = Field(
```
hjoaquim, line 23: This changes on the standard models, namely adding/removing arguments should be done in a different PR unless there's a reason to do it on this one - it seems unrelated.

deeleeramone: This is all related because now these choices are actually in `openapi.json` and `reference.json`, where they were not previously.

openbb_platform/providers/alpha_vantage/openbb_alpha_vantage/models/equity_historical.py
deeleeramone: No, for things to show up in `openapi.json` it must be defined in `__json_schema_extra__`. Not sure why, but that is the way it appears to work.

openbb_platform/core/openbb_core/provider/standard_models/spot.py
deeleeramone: Because these shouldn't be in the standard model.

openbb_platform/core/openbb_core/app/static/package_builder.py
deeleeramone: Because this is a main crux of the issue, combining all choices into one list where there is no distinction between providers. It should always be grouped by provider where multiple  providers exist. "Standard choices" != "valid choices".

openbb_platform/core/openbb_core/app/provider_interface.py
deeleeramone: It's not an assumption, the `if` statement evaluates `v["multiple_items_allowed"]` for truthiness.

deeleeramone:
> Left couple of comments for clarification.
> 
> Altough the `choices` are now only available on the reference since they're defined at the model level instead of the field, it looks like the CLI is still able to pick them up since its defined on `reference.json`.
> 
> CLI behavior seems good and choices are not conflicting between providers:

The behaviour in CLI that is not correct is how when you enter a country for EconDB that is not in the choices for OECD. Choices for OECD are not all valid choices for EconDB, and you can also enter countries far more flexibly in the EconDB provider - i.e, "us", "uk", "USA", "united_states".

deeleeramone:
> Thelp dialog for country will have all the choices regardless the provider - the choices should not be validated at the CLI (by argparse) but instead at the Platform (by pydantic); which seems to be happening as the image above shows.

This is actually not what happens with CLI, otherwise you would be able to enter: `economy --country us --provider econdb`.

openbb_platform/core/openbb_core/app/provider_interface.py
deeleeramone: This is a local variable that for iterating over choices for each the provider.  The only thing ever assigned to it are choices.

andrewkenreich:
> @andrewkenreich can you take a look if this breaks anything on Terminal Pro?

I think we are good - but we can take care of it when we merge this too - wont affect us right away

openbb_platform/core/openbb_core/app/provider_interface.py
hjoaquim: is it? line 262 is `choices[p] = {"multiple_items_allowed": True, "choices": v.get("choices")}` 
sry if i'm misinterpreting

openbb_platform/providers/alpha_vantage/openbb_alpha_vantage/models/equity_historical.py
hjoaquim: apparently is fine since it's allso available on the reference.json

openbb_platform/core/openbb_core/app/static/package_builder.py
hjoaquim: yap, gotcha

openbb_platform/core/openbb_core/provider/standard_models/equity_historical.py
hjoaquim: fair enough

hjoaquim:
> > Thelp dialog for country will have all the choices regardless the provider - the choices should not be validated at the CLI (by argparse) but instead at the Platform (by pydantic); which seems to be happening as the image above shows.
> 
> This is actually not what happens with CLI, otherwise you would be able to enter: `economy --country us --provider econdb`.

Kind of; what happens is that the CLI merges the available choices on all the providers into a unique value list. Since `econdb` has no choices, the `oecd` ones prevail.
Not a showstopper imo, but definitively buggy - thanks for clarifying.

Very much like #6573 (tagging to reference this also on the issue)

hjoaquim (approved):
I'm good with this, thanks @deeleeramone for the clarifications!
Maybe would be good if @montezdesousa run through it also, since he was the one adding the `multiple_items_allowed` feature

openbb_platform/core/openbb_core/provider/standard_models/balance_sheet.py
```diff
@@ -16,10 +16,6 @@ class BalanceSheetQueryParams(QueryParams):
     """Balance Sheet Query."""
 
     symbol: str = Field(description=QUERY_DESCRIPTIONS.get("symbol", ""))
-    period: str = Field(
-        default="annual",
-        description=QUERY_DESCRIPTIONS.get("period", ""),
-    )
```
montezdesousa, line 22: why are we removing `period` from this model?

openbb_platform/openbb/package/crypto_price.py
```diff
@@ -135,11 +135,39 @@ def historical(
...
+                                "90m",
+                                "1h",
+                                "1d",
+                                "5d",
+                                "1W",
+                                "1M",
+                                "1Q",
+                            ],
+                        },
+                    },
```
montezdesousa, line 170: Does this change in the `choices` affect the CLI autocomplete?

hjoaquim: iiuc, no; although it won't pick from the `Field` anymore, it can still pick the choices from the reference.

openbb_platform/core/openbb_core/app/provider_interface.py
deeleeramone: Yes, this is a dictionary of choices for the field, by provider, where "multiple_items_allowed" is part of "choices".

deeleeramone:
> Kind of; what happens is that the CLI merges the available choices on all the providers into a unique value list. Since `econdb` has no choices, the `oecd` ones prevail. Not a showstopper imo, but definitively buggy - thanks for clarifying.
> 
> Very much like #6573 (tagging to reference this also on the issue)

In *this specific instance* it is not entirely a show-stopper but it certainly has the potential to be. `derivatives.futures.curve` comes to mind.

No provider should "prevail" over another, all queries should make it to the provider for it to handle/validate specifically.

openbb_platform/core/openbb_core/provider/standard_models/balance_sheet.py
deeleeramone: Because this is part of the issue, docstrings and type hints do not generate correctly. Long-standing issue being resolved.

IgorWounds (approved):

openbb_platform/core/openbb_core/app/provider_interface.py
montezdesousa: I also find this variable name ambiguous here because `choices` is also used with a different meaning as a key:value pair inside the dictionary

montezdesousa (approved):

PR #6609, opened 2024-08-03
[Feature] Replace Nasdaq SP500 Multiples With Direct Source

deeleeramone:
1. **Why**?:

    - The only provider for this data has stopped updating their endpoint and can be replaced with the direct source.

2. **What**?:

    - Add actionable warning when the command is run with `nasdaq` as the provider.

3. **Impact**:

    - Adds a new provider module, that is not yet published to PyPI.
    - Multiple items allowed for `series_name`.

4. **Testing Done**:

    - Integration/Unit tests.
    - Checked the warning is being broadcast as expected.

image


openbb_platform/core/openbb_core/provider/standard_models/sp500_multiples.py
```diff
@@ -12,7 +12,7 @@
     QUERY_DESCRIPTIONS,
 )
 
-SERIES_NAMES = Literal[
+SeriesNames = Literal[
```
IgorWounds, line 15: Either keep this `SERIES_NAME` or do it all lowercaps to stick with a pythonic way.

PR #6621, opened 2024-08-09
[BugFix] Update `aiohttp` for Dependabot

deeleeramone:
1. **Why**?:

Affects: aiohttp < 3.10.2

Summary
Static routes which contain files with compressed variants (.gz or .br extension) were vulnerable to path traversal outside the root directory if those variants are symbolic links.

Details
The server protects static routes from path traversal outside the root directory when follow_symlinks=False (default). It does this by resolving the requested URL to an absolute path and then checking that path relative to the root. However, these checks are not performed when looking for compressed variants in the FileResponse class, and symbolic links are then automatically followed when performing Path.stat() and Path.open() to send the file.

Impact
Servers with static routes that contain compressed variants as symbolic links, pointing outside the root directory, or that permit users to upload or create such links, are impacted.


piiq (approved):
Thank you

IgorWounds (approved):

PR #6624, opened 2024-08-13
[Feature] Flatten Revenue By Geography and Business Line

deeleeramone:
1. **Why**?:

    - The nested response is incompatible with the Excel add-in.

2. **What**?:

    - We flatten the object and return in the style of a pivot table.
    - Remove unnecessary query parameter - "structure".
    - Move "period" from the standard model to the FMP model.

3. **Impact**:

    - Breaking changes:
      - The response object is now a flat list of records.
      - No more "structure" parameter.
    - Improved response object and better cross-product compatibility.

4. **Testing Done**:

    - Unit/integration tests.

Screenshot 2024-08-13 at 3 01 29 PM


montezdesousa (approved):
lgtm

PR #6635, opened 2024-08-21
Added setup_OpenBB.bat, a batch file to automate full setup in Windows.

Thepathakarpit:

## Description

-Summary:  
 - Automated the setup, installation, and dependencies for the OpenBB Platform.
  - Checks for the presence of Python and Pip, and installs them if they are not found.
  - Installs the OpenBB Platform, OpenBB Platform CLI, and required dependencies.
  - Runs the OpenBB Platform after the setup is complete.
 
- It will make it super easy for users to complete the setup and run the project. After all it only takes one command to run it.

Steps to run: 
1. Save the script to a file, openbb_setup.bat
2. Open the command prompt or PowerShell
3. Run the script by typing openbb_setup.bat and pressing Enter

Thank you

### DIFF
--- a/setup_OpenBB.bat
+++ b/setup_OpenBB.bat
@@ -1,0 +1,57 @@
+@echo off
+
+rem Check if Python is installed
+where python >nul 2>nul
+if errorlevel 1 (
+    echo Python is not installed. Installing Python...
+    goto installPython
+)
+
+rem Check if pip is installed
+python -m pip --version >nul 2>nul
+if errorlevel 1 (
+    echo Pip is not installed. Installing Pip...
+    goto installPip
+)
+
+rem Install OpenBB Platform
+echo Installing OpenBB Platform...
+pip install openbb
+
+rem Install OpenBB Platform CLI
+echo Installing OpenBB Platform CLI...
+pip install openbb-cli
+
+rem Install dependencies
+echo Installing dependencies...
+pip install -r requirements.txt
+
+rem Run the OpenBB Platform
+echo Starting the OpenBB Platform...
+openbb
+
+echo OpenBB Platform setup complete!
+pause
+exit /b
+
+:installPython
+echo Downloading and installing Python...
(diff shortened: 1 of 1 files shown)


deeleeramone:
Hi there, thanks very much for the PR. We are currently working on an installer that will accomplish this task, and more. We don't like the idea of touching the system Python, so the installer will load and setup a full setup, along with additional tooling. You can checkout PR #6587 to see the setup scripts.

Thepathakarpit:
> Hi there, thanks very much for the PR. We are currently working on an installer that will accomplish this task, and more. We don't like the idea of touching the system Python, so the installer will load and setup a full setup, along with additional tooling. You can checkout PR #6587 to see the setup scripts.

Yes it is great but don't you think this PR script is way simpler and intuitive for users to use. The simplicity will help you with people who don't want to read full procedure but just a friend sends them a batch file or they copy paste it and on running it does everything. Consider adding this as a feature too, will help beginners explore too.

deeleeramone:
> Yes it is great but don't you think this PR script is way simpler and intuitive for users to use. The simplicity will help you with people who don't want to read full procedure but just a friend sends them a batch file or they copy paste it and on running it does everything. Consider adding this as a feature too, will help beginners explore too.

Sorry, this is not appropriate script to add to the repository. Installing Python, and Python packages, directly to the system does not follow best practices, and in many cases will create problems for the user and/or break other things they already have configured.

I do appreciate your desire to streamline the process for less technical users, however a signed installer is a much more secure and reputable way to distribute or share this package.

Thepathakarpit:
> > Yes it is great but don't you think this PR script is way simpler and intuitive for users to use. The simplicity will help you with people who don't want to read full procedure but just a friend sends them a batch file or they copy paste it and on running it does everything. Consider adding this as a feature too, will help beginners explore too.
> 
> Sorry, this is not an appropriate script to add to the repository. Installing Python, and Python packages, directly to the system does not follow best practices, and in many cases will create problems for the user and/or break other things they already have configured.
> 
> I do appreciate your desire to streamline the process for less technical users, however a signed installer is a much more secure and reputable way to distribute or share this package.

Okay no issues do let me know if in future you need something like this. Had fun doing this task for the project. Thank you.

PR #6640, opened 2024-08-22
[BugFix] Add `trust_env` to AIOHTTP ClientSession Initialization to Read `.env` for Proxy Mapping

### LINKED ISSUE #6638: [Bug] Missing HTTP Proxy Setting
**Describe the bug**
I notice the document metioned [Proxy Network in User Settings & Enviroment Variables](https://docs.openbb.co/platform/user_guides/settings_and_environment_variables#proxy-networks), but I can't find this setting in source codes anywhere. So is there some thing wrong in the document or the codes?

**To Reproduce**
1. Refer to other enviroment variables, I think `HTTP_PROXY` or `HTTPS_PROXY` should found in `openbb_core/env.py`. However it isn't.
2. Then I check the session definition file `openbb_core/utils/client.py`, there is a class named `ClientSession`, but it don't use `proxies` parameter durning initialization.
3. Then I check the request tool function `amake_request`, `make_request` in file `openbb_core/provider/utils/helpers.py`, it also doesn't use `proxies` parameter.

So, is `HTTP_PROXY` and `HTTPS_PROXY` deprecated?

**Screenshots**

Document
image

`openbb_core/env.py` File
image

`openbb_core/utils/client.py`
image

`openbb_core/provider/utils/helpers.py`
image

- OS: Windows 11
- Python version 3.11.9

**Additional context**
Thanks for answering my problem.

deeleeramone:
1. **Why**?:

`aiohttp` does not read the environment variables by default, whereas `requests` does. This means that only synchronous requests are being directed to the proxy URL when defined as the HTTP_PROXY or HTTPS_PROXY environment variable.

Resolves #6638

2. **What**?:

    - Where aiohttp.ClientSession is initialized, we add `trust_env=True`.

3. **Impact**:

    - Ensures the same proxy behavior for both synchronous and asynchronous requests.

4. **Testing Done**:

    - You can set two environment variables with bogus addresses, and all requests will timeout because it cannot connect.

```
HTTP_PROXY="http://10.10.10.10:8000"
HTTPS_PROXY="https://10.10.10.10:8000"
```

This already worked when `make_request` was used, but not the asynchronous versions.


IgorWounds (approved):

IgorWounds:
This seems good to me. Pinging @piiq to double-check me.

piiq (approved):

PR #6649, opened 2024-08-31
Adding example on how to urn OpenBB in Apache Beam pipelineadding beam sample

mmistroni:

## Description

- Example on running OpenBB codebase on Apache Beam
- OpenBB can be leveraged to run in an automated mode by running on Apache Beam in 
  a series of pipelines so that the same data you would retrieve via Terminal or OpenBB GUI can be fetched
  automatically by running the same code in an Apache Beam pipeline.
  GCP dataflow currently supports Apache Beam so the entire code could theoretically be run on cloud.
  There will be  limitations, of course, such as the fact that Beam runs pipeline in parallel and perhaps not all the underlying
 APIs used by OBB accepts parallel requests,but it is an useful proof of concept on what can be done
- The application is self contained, and uses  a requirements.txt file to install dependencies

## How has this been tested?

- The application runs an unit test
</details>

### DIFF
--- a/examples/openbb-apachebeam/README.txt
+++ b/examples/openbb-apachebeam/README.txt
@@ -1,0 +1,12 @@
+#### Obb Dataflow sample
+
+
+This is a sample how to  invoke OBB fetchers in an Apache Beam pipeline. (GCP dataflow is build on Apache Beam)
+-- Pre-requisite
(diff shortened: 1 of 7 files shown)


mmistroni:
please hold on to this i added some wrong files

mmistroni:
closing this. sorry i messed up

PR #6656, opened 2024-09-05
Expose one endpoint per provider for each command

piiq:


piiq:
@deeleeramone , I've done a small refactor of the extension code and added tests. It's ready for review. Let me know if you want me to guide you through places where I would suggest more exploratory testing. I might need some time before I'll be able noticing the smaller details as I've been looking at this code for a bit too long 😂

deeleeramone (changes requested):
With these changes, I am unable to use my custom routers - `widgets.json` is missing them entirely whereas before the changes, all router GET functions generated a widget.

This is observable by A/B'ing the two `openbb-platform-api` implementations.

piiq:
@deeleeramone I've updated the PR with a few things

1. It does handle GET-only routers correctly
2. The package utilities have been refactored and relative dynamic imports work without extra effort
3. Updated argument parsing to handle `--help` argument and return a docstring

I have some more updates to the installer itself, but they should fall outside the scope of this PR

deeleeramone (changes requested):
If the purpose of making one endpoint per provider is to eliminate the "useless" parameters for a specific provider, we're not quite there. For example, if we look at `equity.price.historical`, we'll find various parameters that do not belong with that provider. Example, FMP - "extended_hours" and "use_cache" do not exist.

Screenshot 2024-09-11 at 10 26 38 AM

Meanwhile, TMX comes with a bunch of Intrinio parameters that only exist from Intrinio, but is also missing the "adjustment" parameter.

Screenshot 2024-09-11 at 10 21 12 AM

The same randomness applies to every provider here, so it appears the iteration is not properly resetting the locals or identifying all the parameters within a specific provider.

deeleeramone:
Another consideration, @piiq, is that when the parameter within a provider's definition has "multiple_items_allowed", and there are no "choices" associated with that provider-param, the input needs to be "text" instead of "number" or "date". This will be the only way to supply multiple values to parameters that can't use "multiSelect".

Screenshot 2024-09-11 at 11 39 45 AM

deeleeramone (approved):
There are still some challenges, some are still unknown, but we'll merge this with the upstream to keep iterating. Notes on what I've found so far:

- Some start/end/single dates are not being recognized as "date".
- Booleans are being seen as "text", will have to explicitly define "options" with True/False, and, if applicable, None (Default).
- The "source" - AKA, 'provider' - is not defined in the widget.
- Defining empty locals without explicit typing creates lint errors.

piiq:
@deeleeramone , i've fixed the booleans and updated the tests.

for other 3 issues mentioned, I would greatly benefit from concrete examples. it's not that easy to find a needle in the haystack of 360+ widgets

PR #6664, opened 2024-09-09
[BugFix] Enforce User Preference For Dark/Light Mode In Chart Output To API

deeleeramone:
1. **Why**?:

    - The charting extension is relying on the output at runtime to set the user preference theme, this was excluding the API from being set by this preference.

2. **What**?:

    - Sets `plot_bgcolor`, `paper_bgcolor` and `font_color` as white/black according to the user-defined preference in `user_settings.json`, "chart_style".
    - This is done at the chart function level, prior to the `layout_kwargs` entry at output, and remains settable at runtime.

3. **Impact**:

    - Charts output from the API should now be consistent with the Python output.

4. **Testing Done**:

    - Define the preference in `user_settings.json` and created the different charts from endpoints.

### DIFF
--- a/openbb_platform/extensions/derivatives/openbb_derivatives/derivatives_views.py
+++ b/openbb_platform/extensions/derivatives/openbb_derivatives/derivatives_views.py
@@ -128,6 +128,7 @@ class DerivativesViews:
 
         figure = OpenBBFigure().create_subplots(shared_xaxes=True)
         figure.update_layout(ChartStyle().plotly_template.get("layout", {}))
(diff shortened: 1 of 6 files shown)


piiq (approved):
At some point we can consider aggregating the styling in one place, but this is out of scope for this PR. Thanks

deeleeramone:
> At some point we can consider aggregating the styling in one place, but this is out of scope for this PR. Thanks

That'll be in scope when we deal with replacing frontend-components, which needs to happen at some point.

PR #6671, opened 2024-09-13
[BugFix] Add Validators For `date` Fields For Multiple Items Allowed.

deeleeramone:
1. **Why**?:

    - This PR adds validators to the `date` QueryParam so that each item in a string of multiple items is validated and parsed with the correct format.
    - 500 errors in the API where Exceptions should be handled.

2. **What**?:

    - Adds validators in standard models.
    - Adds additional exception handling in the API to reduce instances of "unexpected error".

3. **Impact**:

    - As a side-effect, dates in the futures and yield curves can be sloppily entered - i.e, '2023' instead of '2023-01-01'. `pandas.to_datetime` is capable of handling more date formats than other date parsers.

4. **Testing Done**:

    - `/api/v1/fixedincome/government/yield_curve?chart=true&provider=fmp&date=2023%2C2022-13-5`

Screenshot 2024-09-13 at 12 01 20 PM

    - `obb.fixedincome.government.yield_curve(date="2019-01-01,2020-01-01,2021-13-01", provider="fmp")

Screenshot 2024-09-13 at 12 07 29 PM


openbb_platform/core/openbb_core/api/exception_handlers.py
```diff
@@ -61,7 +69,13 @@ async def validation(request: Request, error: ValidationError):
             loc in query_params for err in errors for loc in err.get("loc", ())
         )
         if "QueryParams" in error.title and all_in_query:
-            detail = [{**err, "loc": ("query",) + err.get("loc", ())} for err in errors]
+            detail = [
```
piiq, line 72: i really admire your ability to write lines like this one

piiq (approved):

PR #6673, opened 2024-09-17
[BugFix] Use `__alias_dict__` instead of Field(alias) for data fields.

deeleeramone:
1. **Why**?:

    - OpenAPI gets confused by having `Field(alias='someAlias')` and will end up labeling the Field as the Alias unless we use`__alias_dict__`.

2. **What**?:

    - Moves the alias definition from `Field` to `__alias_dict__`

3. **Impact**:

    - A few Finviz field names have been adjusted to be more like the others.
    - Improved data model schemas in `openapi.json`

4. **Testing Done**:

    - Ran the test suites

### DIFF
--- a/openbb_platform/providers/intrinio/openbb_intrinio/models/share_statistics.py
+++ b/openbb_platform/providers/intrinio/openbb_intrinio/models/share_statistics.py
@@ -29,12 +29,12 @@ class IntrinioShareStatisticsData(ShareStatisticsData):
 
     __alias_dict__ = {
         "outstanding_shares": "weightedavebasicdilutedsharesos",
+        "adjusted_outstanding_shares": "adjweightedavebasicdilutedsharesos",
     }
 
     adjusted_outstanding_shares: Optional[float] = Field(
         default=None,
         description="Total number of shares of a publicly-traded company, adjusted for splits.",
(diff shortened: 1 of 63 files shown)


piiq:
@andrewkenreich @tehcoderer please take a look at this one to see if there are any changes introduced here are breaking for you.
Apart from the Finviz (which IIUC is not used by the app side) we don't expect the API to change, but it's better to be on the safe side and that's why I'm tagging you.

IgorWounds (approved):

openbb_platform/providers/intrinio/openbb_intrinio/models/share_statistics.py
```diff
@@ -29,12 +29,12 @@ class IntrinioShareStatisticsData(ShareStatisticsData):
 
     __alias_dict__ = {
         "outstanding_shares": "weightedavebasicdilutedsharesos",
+        "adjusted_outstanding_shares": "adjweightedavebasicdilutedsharesos",
```
piiq, line 32: image

piiq (approved):

andrewkenreich:
We are actually doing this already on our end, so this is good

PR #6674, opened 2024-09-17
Fetching Senate / HOR financial disclosures

mmistroni:

0. **Title**:

    - Senate/HOR financial disclosures

    - Currently only available via FMP (paid susb) and quiverquants.  This is a joint effort between @deeleeramone (senate) and myself (hor).
    -  It's still a WIP - i am not skilled in asyncio - and somehow my ipynb retrieves more disclosures (480) than the version i ran in IDEA

    

    -  Updated government-us provider to expose senate and house of reprsentative disclosures

    - N.A. wont affect any existing endpoints
4. **Testing Done**:

    - Ran the notebook, and ran test_hor_helpers.py. Somehow the pytest.http does not work due to the presence
      of a ZIP file which is extracted on the fly, atm i dont know how to get around it
    - Have provided a asample XML file for disclosures, a sample PDF for a member disclosue and a jupyter notebook that runs the api

    - Any specific focus areas for review?
         QueryParams. not really sure what should be the driver for this data. for senate i thought number of reports as the api is quick enough to retrieve all of them at once
        for HOR, i have made driven by year as the HOR URLS has data by years. FMP exposes also a 'query by member' but this will require a new implemetnation similar to what @deeleeramone  has done for senate

  

         Please advise on potential improvement soi can try to work on them myself

### DIFF
--- a/openbb_platform/obbject_extensions/charting/openbb_charting/charts/correlation_matrix.py
+++ b/openbb_platform/obbject_extensions/charting/openbb_charting/charts/correlation_matrix.py
@@ -51,6 +51,7 @@ def correlation_matrix(  # noqa: PLR0912
     title = kwargs.get("title") or "Asset Correlation Matrix"
     text_color = "white" if ChartStyle().plt_style == "dark" else "black"
     colorscale = kwargs.get("colorscale") or "RdBu"
+
     heatmap = Heatmap(
         z=df,
         x=X,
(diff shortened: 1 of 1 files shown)


deeleeramone:
Hi @mmistroni, thanks for the PR and I appreciate the efforts, but I will need to kindly decline.

The code I shared with you was not intended as a solution appropriate for submitting as a PR. The code would be significantly different if I were to submit as a PR. While it's fine for personal use as a notebook, It is not a viable solution for a production environment and will be unreliable in operation. After 1 or 2 queries the IP address is flagged and banned for the day, so this approach is clearly not going to work at scale.

There are some minimum requirements when considering a function that serves this data:
  - Queryable by stock ticker.
  - Queryable by date range.
  - Queryable by representative.
  - The standard model -> provider model -> fetcher pattern needs to be followed.
  - Users won't know - or likely care - about the difference between HOR and Senate, a representatives endpoint should gracefully handle both with none or very little user intervention.
  - Provider models need to map to the router, there is no endpoint created which means there is no function.

Some general notes:
- There are too many files.
- There are files that are unrelated to the PR.
- Notebook files are for Examples only, they have no purpose within the library code.
- Tests should not need to stray at all from the established pattern throughout all unit and integration tests.

mmistroni:
Hello
 sure, i  was not expecting it to go through.,,,, what was more important
for me was to get some feedback, which you duly provided
i will try to work on it to get it to a decent state where it can be
submitted
Thanks again & kind regards
 Marco

mmistroni:
@deeleeramone i will carry on working on this..  i might need some assistance..can i post here rather than pinging you directly on Discord?
i m keen on completing this...  kr

markqiu:
Why don't we just add this to fmp provider?

mmistroni:
Not an issue to do that..apart from fact that it requires premium
subscription....if that is not a problem I will add it - while I m still
trying to sort out the original code -

PR #6675, opened 2024-09-17
Refactor Widgets Builder For columnsDefs

deeleeramone:
This PR updates the script to include columnsDefs. It's not 100% bullet-proof but we're pretty close to reaching the limits of what can be done automagically across absolutely everything.

- Everything that is labeled as `'x-unit_measurement': 'percent'` is displayed as Red/Green percentage values.
- CIK and other security identifiers are explicitly set as 'text' to avoid interpreting as an integer.

### DIFF
--- a/openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
+++ b/openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
@@ -1,7 +1,6 @@
 """Utils for building the widgets.json file."""
 
 from copy import deepcopy
-from typing import Dict, List
 
 TO_CAPS_STRINGS = [
     "Pe",
@@ -43,13 +42,13 @@ TO_CAPS_STRINGS = [
 ]
 
 
-def modify_query_schema(query_schema: List[Dict], provider_value: str):
+def modify_query_schema(query_schema: list[dict], provider_value: str):
     """Modify query_schema and the description for the current provider."""
-    modified_query_schema = []
+    modified_query_schema: list = []
     for item in query_schema:
         # copy the item
         _item = deepcopy(item)
-        provider_value_options = {}
+        provider_value_options: dict = {}
 
         # Exclude provider parameter. Those will be added last.
         if "parameter_name" in _item and _item["parameter_name"] == "provider":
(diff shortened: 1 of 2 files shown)


openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
```diff
@@ -175,9 +193,10 @@ def build_json(openapi: Dict):
                 "data": {
                     "dataKey": "results",
                     "table": {
-                        "showAll": False,
+                        "showAll": True,
```
piiq, line 196: Won't this result in get-only routers not displaying anything because they don't have schemas set?

deeleeramone: That would be very counterintuitive for "showAll" to not actually do what it says it does.  When `columnsDefs` are defined, setting to False hides the fields without definitions. This param seems to have no effect when no `columnsDefs` are supplied.

Screenshot 2024-09-18 at 8 29 11 AM

openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
```diff
@@ -91,9 +91,11 @@ def modify_query_schema(query_schema: list[dict], provider_value: str):
 
         modified_query_schema.append(_item)
 
-    modified_query_schema.append(
-        {"paramName": "provider", "value": provider_value, "show": False}
-    )
+    if provider_value != "custom":
```
piiq, line 94: we might need the provider specified somewhere to fill in the metadata on the pro side. @andrewkenreich if we don't provide the provider in the params would this still work on the UI side?

deeleeramone: The metadata is under "source" for the Widget, which is still Custom. This just prevents creating a parameter that does not exist.

deeleeramone: Screenshot 2024-09-18 at 9 37 35 AM

piiq (approved):

PR #6677, opened 2024-09-18
[Feature] Add EIA Daily Spot Prices Data via FRED

deeleeramone:
1. **Why**?:

    - To start building within the `commodity` module.
    - Quick win by curating some FRED series IDs.

2. **What**?:

    - New endpoint: `obb.commodity.spot_prices`
    - Adds daily time series data from the EIA spot prices API via FRED, and includes the metadata.
      - The list of commodities is fixed, added "commodity" as a provider parameter to allow variations between providers (future). Choices are:
        - "wti",
        - "brent",
        - "natural_gas",
        - "jet_fuel",
        - "propane",
        - "heating_oil",
        - "diesel_gulf_coast",
        - "diesel_ny_harbor",
        - "diesel_la",
        - "gasoline_ny_harbor",
        - "gasoline_gulf_coast",
        - "rbob",
        - "all"  
    - The `lbma_fixing` endpoint from Nasdaq DataLink appears to be dead, commented out the fetcher and skipped those tests.
 
3. **Impact**:

    - `openbb-commodity` will populate with a function without installing any extras.

4. **Testing Done**:

    - Integration/unit

image


openbb_platform/extensions/commodity/integration/test_commodity_api.py
```diff
@@ -20,38 +20,66 @@ def headers():
...
+#                "end_date": "2023-01-01",
+#                "collapse": "monthly",
+#                "transform": "diff",
+#                "provider": "nasdaq",
+#            }
+#        ),
+#    ],
+# )
+# @pytest.mark.skip(reason="Resource no longer available. Pending replacement/removal.")
+# def test_commodity_lbma_fixing(params, headers):
```
IgorWounds, line 49: What is the best way to deal with this? Do you want us to return something to the user saying that it is dead for the time being?

IgorWounds (changes requested):
This one is good. We need to align what is the best way to deal with the LbmaFixing. I don't like commented code personally.

openbb_platform/extensions/commodity/integration/test_commodity_api.py
deeleeramone: Nope, the command is now returning a 410 status code.

IgorWounds:
> > This one is good. We need to align what is the best way to deal with the LbmaFixing. I don't like commented code personally.
> 
> Should we just remove entirely then? I can replace with the direct source eventually, like I did with SP500 multiples and COT reports.

Tagging @minhhoang1023 here.

piiq:
@deeleeramone do you want to pick up the eia integration that's been done here https://github.com/piiq/openbb-energy ?

deeleeramone:
> @deeleeramone do you want to pick up the eia integration that's been done here https://github.com/piiq/openbb-energy ?

ya, but that's out-of-scope for this PR.

jmaslek (approved):

minhhoang1023:
looks good. One thing - do we want to follow the same structure with rest of the endpoints.

So instead of `obb.commodity.spot_prices`, making it `obb.commodity.price.spot`

IgorWounds (approved):

PR #6681, opened 2024-09-20
[BugFix] - LoggingService changing default logging configuration

montezdesousa:
## Description

- Fixes https://github.com/OpenBB-finance/OpenBB/issues/6680
- The `LoggingService` and `HandlersManager` classes were changing the default logger configurations. Now a custom logger is created and used for this purpose, without changing the default configuration.

## How has this been tested?

* Checked logs are correct in the logs folder
* Checked stdout logs are correct

```python
import logging
logging.basicConfig(level=logging.INFO)
logging.info("This message is shown")
import openbb
logging.info("This message is hidden") # This message should NOT be hidden anymore
```


deeleeramone (approved):
Works as advertised.

Screenshot 2024-09-20 at 8 20 19 AM

IgorWounds (approved):

PR #6683, opened 2024-09-21
[Feature] Add Direction Of Trade Statistics From IMF

deeleeramone:
1. **Why**?:

    - Expands the macroeconomic coverage with data of trade flows between countries and major areas of the world.
    - One could make super-fancy maps to visually present this data.

2. **What**?:

    - Adds new route: `obb.economy.direction_of_trade`
      - Data is from IMF.
      - Multiple countries allowed for both the `country` and `counterpart` parameters.
        - Can compare one (or more) country vs. one, several, or all countries and vice-versa.
        - Cannot compare all countries vs. all countries.
      - Choice for the data value: "imports", "exports", "balance", "all"
      - Frequency of the data can be one of: "month", "quarter", "annual"
      - Some cohorts of countries have annual data going back to 1948.

3. **Impact**:

    - Additional stuff to maintain. If other providers are supplying this data in the future, it is likely the same data.

4. **Testing Done**:

    - Integration/Unit
    - Blasted functions until I got rate-limited - have attempted to handle that specific error so it is not "unknown error".

Screenshot 2024-09-20 at 5 00 51 PM

Screenshot 2024-09-20 at 5 09 48 PM


IgorWounds (approved):
Nice one!

PR #6687, opened 2024-09-23
[BugFix] Make User Preferences -> Defaults Work With Any Parameter

### LINKED ISSUE #6686: [Bug] Provider fallback mechanism fails when endpoint is serviced by provider with API key but endpoint does not require API key.
This path, `obb.commodity.petroleum_status_report` does not exist yet. The provider, `eia`, also does not exist yet. The provider will have future endpoints that require an API key, but this one does not. The endpoint is serviced only by this provider, which does not require credentials.

Screenshot 2024-09-22 at 8 02 16 PM

Provider fallback should be checking the Fetcher property for required credentials at the function level, but instead only sees that `user.credentials.eia_api_key = None`.

deeleeramone:
1. **Why**?:

    - The "defaults" section of the `user_settings.json` was not able to cope with anything other than "provider".
    - The app fails to import if the defaults configuration is incorrect.

2. **What**?:

    - Fixes the `Defaults` class so that validation errors don't happen when supplying other parameters.
      - Issue here was being unnecessarily, overly, specific with type hints.
    - Applies any defaults in the `Container`/`Command` classes if they are not supplied at runtime.
    - In the API: Default params will only work where the API has a default of None for the parameter.
    - Resolves #6686 via bypassing Provider Fallback if there is only one provider. Not a full solution, but will cover most scenarios in the custom router/provider world.

3. **Impact**:

    - Improved configurability.
    - The API only partially supports this behavior because all mandatory fields need to be supplied to the URL.

4. **Testing Done**:

    - Ran integration tests and have not found unexpected behavior.

User Settings:

```json
{
  "defaults": {
    "commands": {
      "/equity/price/historical": {
        "provider": "yfinance",
        "chart": true,
        "chart_params": {
          "heikin_ashi": true,
          "indicators": {
            "sma": {
              "length": [21, 50]
            },
            "ema": {
              "length": 150
            }
          }
        }
      }
    }
  }
}
```

Before:

```python
from openbb import obb
```

```sh
ValidationError: 2 validation errors for UserSettings
defaults.commands.`equity.price.historical`.chart
  Input should be a valid list [type=list_type, input_value=True, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.9/v/list_type
defaults.commands.`equity.price.historical`.chart_params
  Input should be a valid list [type=list_type, input_value={'heikin_ashi': True, 'in...'ema': {'length': 150}}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/list_type
```

After:

```python
from openbb import obb

res = obb.equity.price.historical(symbol="AAPL")
res.show()
```

image

### DIFF
--- a/openbb_platform/core/openbb_core/app/static/container.py
+++ b/openbb_platform/core/openbb_core/app/static/container.py
@@ -22,6 +22,26 @@ class Container:
 
     def _run(self, *args, **kwargs) -> Any:
         """Run a command in the container."""
+        endpoint = args[0][1:].replace("/", ".")
+        defaults = self._command_runner.user_settings.defaults.commands
+
+        if defaults.get(endpoint):
+            default_params = {
+                k: v for k, v in defaults[endpoint].items() if k != "provider"
+            }
+            for k, v in default_params.items():
+                if k == "chart" and v is True:
+                    kwargs["chart"] = True
+                elif (
+                    k in kwargs["standard_params"]
+                    and kwargs["standard_params"][k] is None
+                ):
+                    kwargs["standard_params"][k] = v
+                elif (
+                    k in kwargs["extra_params"] and kwargs["extra_params"][k] is None
+                ) or k not in kwargs["extra_params"]:
+                    kwargs["extra_params"][k] = v
+
         obbject = self._command_runner.sync_run(*args, **kwargs)
         output_type = self._command_runner.user_settings.preferences.output_type
         if output_type == "OBBject":
@@ -65,6 +85,7 @@ class Container:
         """
         if choice is None:
             commands = self._command_runner.user_settings.defaults.commands
+            _command = self._command_runner._command_map.get_command(command)
             providers = (
                 commands.get(command, {}).get("provider", []) or default_priority
             )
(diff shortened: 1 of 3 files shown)


openbb_platform/core/openbb_core/app/static/container.py
```diff
@@ -37,7 +57,7 @@ def _check_credentials(self, provider: str) -> Optional[bool]:
         return all(getattr(credentials, r, None) for r in required)
 
     def _get_provider(
-        self, choice: Optional[str], command: str, default_priority: Tuple[str, ...]
```
hjoaquim, line 40: I had the idea that this (not using typing's type hints) would break on 3.9 envs, but it didn't, probably a 3.8 thing 😀

hjoaquim (approved):
Works like a charm; the chart args are such a cool example on how this works well.
Also, extrapolates perfectly to the cli!

image

I don't have any suggesitons code-wise.

Only thing (you're probably aware of) are the warning this is raising:

image

IgorWounds (approved):
Aside from the warnings @hjoaquim mentioned, this looks good!

deeleeramone:
> Aside from the warnings @hjoaquim mentioned, this looks good!

Warnings are coming from the `pandas_ta` library, so there is not much we can do directly.

PR #6692, opened 2024-09-24
[Feature] Add An `openbb-build` Script To `openbb-core` Package

deeleeramone:
1. **Why**?:

    - The build script does not ship with `openbb-core` and this is a problem for creating pip-installable packages with a minimal footprint. `--no-deps` is not something that can be included in `pyproject.toml` or `requirements.txt`, it requires a script.

2. **What**?:

    - Adds a command line script - `openbb-build` - that ships with the `openbb-core` package and installs the `openbb` meta package if it is not already installed.

3. **Impact**:

    - Enhanced ability to easily configure minimalist OpenBB app and API configurations.
    - Eliminates the need to create, publish, and maintain some `openbb-empty` package to accomplish this task.
    - Simplifies the static asset build procedure by reducing it to a command line executable.
    - Bumps the minimum version of Ruff in `openbb-core` to 0.4.10

4. **Testing Done**:

    - Create and activate a new environment, install `poetry` and `openbb-core`, then run `openbb-build`

Screenshot 2024-09-24 at 2 37 50 PM

### DIFF
--- a/openbb_platform/core/openbb_core/build.py
+++ b/openbb_platform/core/openbb_core/build.py
@@ -9,7 +9,6 @@ import sys
 
 def main():
     """Build the OpenBB platform static assets."""
-    message = "\nOpenBB build script not found, installing from PyPI...\n"
     try:
         from openbb import build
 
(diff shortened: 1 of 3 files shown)


openbb_platform/core/openbb_core/build.py
```diff
@@ -0,0 +1,39 @@
...
+# flake8: noqa: S603
+# pylint: disable=import-outside-toplevel
+
+import subprocess
+import sys
+
+
+def main():
+    """Build the OpenBB platform static assets."""
+    message = "\nOpenBB build script not found, installing from PyPI...\n"
```
IgorWounds, line 12: You can move this down where it is used. I'd also explore not printing and logging instead or doing an echo.

deeleeramone: Can do as an echo.

IgorWounds (approved):

PR #6693, opened 2024-09-25
[Feature] Add EIA Provider & Weekly Petroleum Status Report

deeleeramone:
1. **Why**?:

The EIA's Weekly Petroleum Status Report (WPSR) and Short Term Energy Outlook (STEO) are closely followed events for anyone watching the fossil fuels.

This PR represents the first integrations of open data available from the EIA. 

2. **What**?:

The WPSR is comprised of thirteen (excludes discontinued series) high-level categories with each containing a subset of tables. Data is from the static Excel files published [here](https://www.eia.gov/petroleum/supply/weekly/), and each file represents a single category.

All data from a single category is returned by supplying "all" to the `table` parameter of the WPSR endpoint.

Tables from the WPSR are returned in a flat format in the same order as presented in the Excel files. The response is suitable for pivot tables and SQL storage.

The STEO dataset is twenty-six tables of near-term projections, which is pulled from the V2 API.

Because of the structure of their API, a unified seriesID lookup similar to FRED or BLS is not possible. The symbols from within STEO tables are all contained to a single path with uniform parameters, but the symbols within WPSR are scattered throughout other segments of the API and cannot be queried in this way.

3. **Impact**:

    - This package contains dependencies not included in `openbb-core`:
      - async-lru = "^2.0.4" (to prevent Excel files from being repeatedly downloaded)
      - openpyxl = "^3.1.5" (for working with Pandas ExcelFile)
      - xlrd = "^2.0.1" (for working with Pandas ExcelFile)
  
    - New endpoints:
      - `obb.commodity.petroleum_status_report` (WPSR - no API key required)
      - `obb.commodity.short_term_energy_outlook` (STEO - API key required)
 
    - ALRU cache (async drop-in replacement to `functools.lru_cache`) is used for the WPSR files that retains each "category" requested for the duration of the Python session. Setting `use_cache` as `False` will invalidate the cache and download the file again.

    - The `table` parameter has defined choices that represent all possible tables from all categories.
      - For WPSR, a custom validator checks this parameter against the `category` and the list of tables within that category; it will raise with a message indicating which choices are valid for the given category. Multiple items allowed here.
      - For STEO, choices are strings with values, such as "03d", and a map is supplied in the function's docstring.

     - STEO also accepts a symbol input - multiple items allowed - for targeted queries from the API route.
     - The response can be very large tables, 100K - 1M+ rows of data.
       - At an unknown number of rows, the interactive tables in CLI will fail and only display a white screen.
 
4. **Testing Done**:

    - Unit/integration

Screenshot 2024-10-28 at 3 21 19 PM


openbb_platform/openbb/assets/reference.json
```diff
@@ -33,6 +33,7 @@
                 "sec@1.3.4",
                 "tiingo@1.3.4",
                 "tradingeconomics@1.3.4",
+                "us_eia@1.0.0b0",
```
piiq, line 36: Are we including this into the main distribution or an extra? I'm curious because of the deps it pulls with it (the excel, and the new cache). Have you checked if those deps are pure python or with extensions that require some sort of build tools to be installed?

openbb_platform/pyproject.toml
```diff
@@ -26,6 +26,7 @@ openbb-polygon = "^1.3.4"
 openbb-sec = "^1.3.4"
 openbb-tiingo = "^1.3.4"
 openbb-tradingeconomics = "^1.3.4"
+openbb-us-eia = "1.0.0b"
```
piiq, line 29: Same as my comment above

openbb_platform/openbb/assets/reference.json
deeleeramone: The intention is to include it as part of the main distribution because it is public domain data, free to access, and provides coverage of widely followed energy commodities fundamentals. If there is a reason not to, let's discuss!

As finance is heavily reliant on Excel/spreadsheet files, I/O packages related to handling these types of files feels like it should ship with the main distribution. I believe the packages being introduced are "lightweight" additions that do not introduce anything that requires special handling or installation considerations - ie, nothing like pyarrow, numba, optimized, or uncompiled binaries.

Maybe I'm missing something, so let me know if there is anything overlooked while arriving at the conclusion.

- openpyxl is the Pandas extra that enables the ExcelFile class and IO of Excel files in general.
  - XLRD is for handling the older `.xls` file format, and every Python Excel library actually uses this under the hood for handling this format. XLS files are still widely distributed.
  - The XML components are already present in the main distribution.


async-lru is a simple extension that relies only on core Python elements. The library is just an `__init__.py`, and it imports:

```
import asyncio
import dataclasses
import sys
from asyncio.coroutines import _is_coroutine  # type: ignore[attr-defined]
from functools import _CacheInfo, _make_key, partial, partialmethod
from typing import (
    Any,
    Callable,
    Coroutine,
    Generic,
    Hashable,
    Optional,
    OrderedDict,
    Set,
    Type,
    TypedDict,
    TypeVar,
    Union,
    cast,
    final,
    overload,
)
```

deeleeramone:
> Additionally please have a look if any of the integrations with EIA's V2 API I've made here are reusable https://github.com/piiq/openbb-energy/tree/develop/

I was actually going to dive into what you had mapped out already, and I have some of the other routes mapped as well. That represents a bunch of work that would make this a much larger PR, so my intention was to start with this and work on the other pieces.

openbb_platform/openbb/assets/reference.json
piiq: Thanks for this reply. I helped a lot to understand the async-lru part and i've checked the openpyxl and it's a pure python package. I identified that by looking at the lock file and seeing that the package is platform agnostic

image

PR #6724, opened 2024-10-01
Add implementation of widget filtering based on local json configuration

piiq:
Hackity hack mcHackface

### DIFF
--- a/openbb_platform/extensions/platform_api/openbb_platform_api/main.py
+++ b/openbb_platform/extensions/platform_api/openbb_platform_api/main.py
@@ -25,17 +25,15 @@ USER_SETTINGS_COPY = os.path.join(HOME, ".openbb_platform", "user_settings_backu
 # Load widget_exclude_filter.json if it exists
 # Widget filtering is optional and can be used to exclude widgets from the widgets.json file
 # You can generate this filter on OpenBB Hub: https://my.openbb.co/app/platform/widgets
-WIDGET_EXCLUDE_FILTER_PATH = os.path.join(
-    HOME, ".openbb_platform", "widget_exclude_filter.json"
-)
-if os.path.exists(WIDGET_EXCLUDE_FILTER_PATH):
-    with open(WIDGET_EXCLUDE_FILTER_PATH) as f:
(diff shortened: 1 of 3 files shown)


montezdesousa:
I suggest to rename the file to `widgets_config.json` and use this format to easily add other stuff if we need in the future

```json
{
    "exclude": [
        "widget_id_foo",
        "widget_id_bar",
    ]
}
```

montezdesousa (changes requested):
I think we should rename the file to `widgets_settings.json` and use a more general format to easily add other stuff if we need in the future

Example:
```json
{
    "exclude": [
        "widget_id_foo",
        "widget_id_bar",
    ]
}
```

deeleeramone:
Can we also add this as a kwarg of the openbb-api command line script? `--filter`

deeleeramone (commented):
Maybe I'm missing something, but how does this new JSON file get created by hub? Wouldn't something need to change in the Hub service in the main repo to know this file was incoming and to create it?

openbb_platform/extensions/platform_api/openbb_platform_api/main.py
```diff
@@ -19,7 +22,20 @@
 CURRENT_USER_SETTINGS = os.path.join(HOME, ".openbb_platform", "user_settings.json")
 USER_SETTINGS_COPY = os.path.join(HOME, ".openbb_platform", "user_settings_backup.json")
 
-FIRST_RUN = True
+# Load widget_exclude_filter.json if it exists
+# Widget filtering is optional and can be used to exclude widgets from the widgets.json file
+# You can generate this filter on OpenBB Hub: https://my.openbb.co/app/platform/widgets
```
deeleeramone, line 27: Can I test this? I currently get a 500 error on both `.co` and `.dev`.

piiq: Currently available here https://github.com/OpenBB-finance/openbb-hub/pull/398

piiq:
> Can we also add this as a kwarg of the openbb-api command line script? `--filter`

@deeleeramone I've added a `--no-filter` argument that when provided will trigger "not filtering"

I'll merge this now so that I can deal with the packaging of this extension

PR #6733, opened 2024-10-03
[BugFix] Update `ruff.toml` To Use Python 3.9 As Minimum Supported Version

deeleeramone:
1. **Why**?:

    - Lint rules are being based on the Python 3.8, which is no longer supported.

2. **What**?:

    - Updates the minimum version in ruff.toml
    - Adds temporary ignores for typing.Dict and typing.List deprecations.

3. **Impact**:

    - Lint rules will now reflect the supported versions of Python.
    - Creates future work to update primitive type hints with their core library class

4. **Testing Done**:

    - I've been working off this locally for some time now, and have not run into any associated complications.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -8,7 +8,8 @@ exclude = [
 ]
 
 line-length = 122
-target-version = "py38"
+target-version = "py39"
+fix = true
 
 [lint]
 select = [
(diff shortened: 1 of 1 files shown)


hjoaquim (approved):

the-praxs:
@deeleeramone yay, we are officially Python 3.9+! I was rooting for that for a long time.

IgorWounds (approved):

PR #6794, opened 2024-10-17
[BugFix] Bump `openbb-core` Dependencies In `pyproject.toml`

deeleeramone:
1. **Why**?:

    - This is step one in resolving [https://github.com/OpenBB-finance/OpenBB/security/dependabot/1700](https://github.com/OpenBB-finance/OpenBB/security/dependabot/1700)

2. **What**?:

    - Version bumps for:
      - uvicorn
      - websockets
      - fastapi
      - importlib-metadata
      - ruff
    
    - openbb-devtools needs to update for the linting actions. 

3. **Impact**:

    - Resolves [https://github.com/OpenBB-finance/OpenBB/security/dependabot/1700](https://github.com/OpenBB-finance/OpenBB/security/dependabot/1700)

4. **Testing Done**:

    - The new lock file was compiled with Python 3.9, and did not encounter any conflicts or errors associated with version changes.

The complete resolution of the issue will come about from the next minor release.


IgorWounds:
Tagging @piiq

openbb_platform/core/pyproject.toml
```diff
@@ -9,17 +9,18 @@ packages = [{ include = "openbb_core" }]
...
 pandas = ">=1.5.3"
 html5lib = "^1.1"
-fastapi = "^0.111.0"
+fastapi = "^0.115.0"
 uuid7 = "^0.1.0"
 posthog = "^3.3.1"
 python-multipart = "^0.0.7"
 pydantic = "^2.5.1"
 requests = "^2.32.1"
-importlib-metadata = "^6.8.0"
```
piiq, line 22: Can you elaborate please why this change is required?

openbb_platform/core/pyproject.toml
```diff
@@ -9,17 +9,18 @@ packages = [{ include = "openbb_core" }]
 
 [tool.poetry.dependencies]
 python = "^3.9"
-uvicorn = "^0.24"
-websockets = "^12.0"
+uvicorn = "^0.32.0"
+websockets = "^13.0"
+starlette = ">=0.40.0"
```
piiq, line 14: IIUC the fastapi bump resolves the starlette issue. If that's the case there is no reason to pin starlette independently as it's not a top level dependency we use.

piiq (commented):
I've added 2 inline comments

openbb_platform/core/pyproject.toml
deeleeramone: There is no need to pin this to a lower version that is obsolete from last year. Current version is 8.5

openbb_platform/core/pyproject.toml
deeleeramone: FastAPI defines Starlette as (https://github.com/fastapi/fastapi/blob/master/pyproject.toml):

```
dependencies = [
    "starlette>=0.37.2,<0.41.0",
    "pydantic>=1.7.4,!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0",
    "typing-extensions>=4.8.0",
]
```

So in order for Poetry to not use the lower versions, we need to specify it directly in our package..

piiq: No need to pin starlette. With unpinned starlette openbb-core already installs 0.40.0
Also this PR https://github.com/fastapi/fastapi/pull/12469 will likely bump the minimal supported version on FastAPI level

openbb_platform/core/pyproject.toml
```diff
@@ -9,17 +9,17 @@ packages = [{ include = "openbb_core" }]
...
 html5lib = "^1.1"
-fastapi = "^0.111.0"
+fastapi = "^0.115.0"
 uuid7 = "^0.1.0"
 posthog = "^3.3.1"
 python-multipart = "^0.0.7"
 pydantic = "^2.5.1"
 requests = "^2.32.1"
-importlib-metadata = "^6.8.0"
+importlib-metadata = ">=6.8.0"
```
piiq, line 22: Another question - do we need this pin at all?

piiq (approved):

openbb_platform/core/pyproject.toml
deeleeramone: If you inspect `poetry.lock`, it will still contain the lower version because the latest version of FastAPI simply "allows" a version higher than 0.39 to install.

So until they make a new release of FastAPI, we need to do this. Maybe we just need to wait it out and next week a patch will get published?

openbb_platform/core/pyproject.toml
deeleeramone: Not entirely sure, but that should be the minimum version at least because that is when support for 3.12 was added.

deeleeramone:
@piiq Removing the Starlette pin prevents the issue from being resolved in this PR. `poetry.lock` is what Dependabot reads.

Screenshot 2024-10-19 at 9 14 32 AM

openbb_platform/core/pyproject.toml
piiq: the poetry.lock of the core has 0.40.0
image
IIUC all other lock files have older versions because the core has not ben published or explicitly updated thus they rely on an older fastapi which pulls the old starlette
image
Once we update the core, the minimal version of fastapi will be bumped and this will bump starlette

piiq: I propose doing this in a separate PR limiting the scope of this one to only openbb-core

PR #6934, opened 2024-11-05
Cleanup example nb titles

jmaslek:
Very small chore


deeleeramone:
Question: Do we want these to appear in the Hub examples page? If yes, we'll need to add them to the JSON file in the `/examples` folder.

deeleeramone (approved):

jmaslek:
> Question: Do we want these to appear in the Hub examples page? If yes, we'll need to add them to the JSON file in the `/examples` folder.

nah.

PR #6936, opened 2024-11-07
[Feature] Add Endpoint For The Latest Financial Reports Filed With The SEC

deeleeramone:
1. **Why**?:

    - The SEC has an RSS feed for the latest financial reports, but is limited to only 200 results. You could easily miss an item because you weren't checking every ten minutes. This is a take on extending that to be able to include the entire day.

2. **What**?:

    - Adds new endpoint:
      - `obb.equity.discovery.latest_financial_reports`
      - Supported providers: `sec`
        - Other providers could jump in on this as well.
      - The response provides key information and URLs to the documents.
      - Optionally, a `report_type` - i.e, "8-K, 10-K", and a date for looking back.
        - Multiple items allowed for, `report_type`.

3. **Impact**:

    - Improved visibility for important filings made today across the provider universe.

4. **Testing Done**:

    - Unit/integration
    - Non-weekday dates.

Screenshot 2024-11-06 at 4 25 15 PM

Screenshot 2024-11-06 at 4 27 07 PM

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/models/latest_financial_reports.py
+++ b/openbb_platform/providers/sec/openbb_sec/models/latest_financial_reports.py
@@ -187,7 +187,7 @@ class SecLatestFinancialReportsFetcher(
         **kwargs: Any,
     ) -> list[SecLatestFinancialReportsData]:
         """Transform the raw data."""
-        results: SecLatestFinancialReportsData = []
+        results: list[SecLatestFinancialReportsData] = []
 
         def parse_entry(entry):
             """Parse each entry in the response."""
--- a/openbb_platform/providers/sec/openbb_sec/models/latest_financial_reports.py
+++ b/openbb_platform/providers/sec/openbb_sec/models/latest_financial_reports.py
@@ -1,5 +1,7 @@
 """RSS Latest Financials Model."""
 
+# pylint: disable=unused-argument
+
 from datetime import date as dateType
 from typing import Any, Optional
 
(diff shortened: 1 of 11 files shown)


openbb_platform/providers/sec/openbb_sec/models/latest_financial_reports.py
```diff
@@ -0,0 +1,261 @@
...
+        SEARCH_HEADERS = {
+            "User-Agent": "my real company name definitelynot@fakecompany.com",
+            "Accept-Encoding": "gzip, deflate",
+        }
+
+        forms = (
+            query.report_type
+            if query.report_type is not None
+            else (
+                "1-K%2C1-SA%2C1-U%2C1-Z%2C1-Z-W%2C10-D%2C10-K%2C10-KT%2C10-Q%2C10-QT%2C11-K%2C11-KT%2C15-12B%2C15-12G%2C"
```
IgorWounds, line 126: Can this be something we define in the helpers or the like if it will be reused?

deeleeramone: Unclear if it will be used elsewhere, it is what you get with this group selection:

Screenshot 2024-11-07 at 7 14 08 PM

deeleeramone: Potentially this could be rounded down into the "discovery.filings" endpoint, but I'm not sure they fill the same purposes entirely.

IgorWounds (approved):

PR #6946, opened 2024-11-14
[Enhancement] Add Multi-Country Support & ECB Yield Curves To `fixedincome.goverment.yield_curve(provider='econdb')`

deeleeramone:
1. **Why**?:

    - Added functionality to fit requirements of @andrewkenreich and @minhhoang1023.

2. **What**?:

    - In, `obb.fixedincome.goverment.yield_curve(provider='econdb')`, multiple items allowed for:
      - 'country'
      - 'date'
    - Adds a `@computed_field` to the output, `maturity_years`, which is the decimal representation of years-to-maturity.
      - It assumes that the `maturity` field is formatted as: 'month_X', 'year_X', or 'year_X_month_X'. If not, computes as None

3. **Impact**:

    - Improved data coverage and enhanced user experience.

4. **Testing Done**:

Screenshot 2024-11-14 at 11 40 19 AM


openbb_platform/core/openbb_core/provider/standard_models/yield_curve.py
```diff
@@ -52,7 +51,23 @@ class YieldCurveData(Data):
...
+    def maturity_years(self) -> Optional[float]:
+        """Get the maturity in years as a decimal."""
+        if self.maturity is None or "_" not in self.maturity:
+            return None
+
+        parts = self.maturity.split("_")
+        months = 0
+        for i in range(0, len(parts), 2):
+            number = int(parts[i + 1])
+            if parts[i] == "year":
```
IgorWounds, line 69: What do you think about this?

```
parts = self.maturity.split("_")
    months = sum(
        int(parts[i + 1]) * (12 if parts[i] == "year" else 1)
        for i in range(0, len(parts), 2)
    )
    return months / 12
```

deeleeramone: > What do you think about this?

yup, that's a cleaner way of doing the same thing.

deeleeramone:
This is unexpected behavior, @IgorWounds, any ideas why EconDB is the only provider where it appears?

Screenshot 2024-11-16 at 12 57 45 PM

IgorWounds (approved):

PR #6958, opened 2024-11-22
[Feature] Add FMP Rating Data

joshuaBri:
1. Why?:
Incorporate FMP's valuation data.
2. What?:
The rating data of certain company from fmp


deeleeramone:
I appreciate the effort, however I need to decline on this. The data itself is not meaningful, nor quantifiable, and this could only ever service FMP.

PR #6962, opened 2024-11-25
[BugFix] Fix Limit Param For yFinance Financials

### LINKED ISSUE #6954: [Bug] limit parameter not working
I ran this code from the Getting Started documentation but the limit parameter doesn't return the n number of rows. It always defaults to 5. I haven't found a solution to the problem yet anywhere else so I'm adding this here.

(
    obb.equity.fundamental.income(
        "TGT", 
        provider="yfinance",
        limit=10,
        period="quarter"
    ).to_df()
)

I am running this on openbb 4.3.4 in python 3.12

deeleeramone:
1. **Why**?:

    - Resolves #6954

2. **What**?:

   - Applies to `balance`,  `cash`, and `income`.
    - Add `le=5` to the yFinance model definition.
    - Filter out the response if limit < 5.

3. **Impact**:

    - For `provider='yfinance'`, a validation error will raise when the limit is more than 5.

4. **Testing Done**:

Screenshot 2024-11-24 at 6 29 56 PM

### DIFF
--- a/openbb_platform/providers/yfinance/openbb_yfinance/models/balance_sheet.py
+++ b/openbb_platform/providers/yfinance/openbb_yfinance/models/balance_sheet.py
@@ -1,8 +1,9 @@
 """Yahoo Finance Balance Sheet Model."""
 
-import json
+# pylint: disable=unused-argument
+
 from datetime import datetime
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Literal, Optional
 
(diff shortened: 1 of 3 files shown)


piiq (approved):
Default limit - 5 
Is it enough as a default one?

deeleeramone:
> Default limit - 5 Is it enough as a default one?

That's the maximum number of results returned from yFinance.

PR #6976, opened 2024-11-29
[Feature] Add SSL Certificate Context And Setting For Async/Sync HTTP Requests

### LINKED ISSUE #6939: [BUG][FR] Impossible to specify SSL certification
This feature is essential as its lack prevents using openbb at all when SSL certification is required to connect to the network. 

**Solution**
The library should have some way to specify the path to the SSL certificate. When using requests this would be for example set in the `verify` attribute of the session. Therefore I would see two options:
* Exposing explicitely the session from `request` instantiated (e.g. I see it in `yfinance/data.py`, line 73). This would meen having the session exposed as an attribute for example `obb.session`, so that its attributes can be modified. 
* Specifying a variable in the `.env` file which will then picked up when creating a `request.Session()` in a similar manner to what is done for proxies.

deeleeramone:
1. **Why**?:

    - Resolves #6939

The Requests library accepts an environment variable, `REQUESTS_CA_BUNDLE`, and AIOHTTP does not. This creates inconsistencies between synchronous and async HTTP requests. Furthermore, defining this variable limits the Requests library to that particular file. If you want to use a self-signed certificate for only some things, it is not straightforward.

This solution combines the specified certificate with the `certifi` defaults.

2. **What**?:

    - Allows a CA certificate file to be specified for verifying HTTPS requests.
      - Self-signed certificates do not need to be part of the system's trust store.
    - Allows "python_settings" in `system_settings.json` to accept extras.
      - Uses keys `http` as a nested dictionary.
    - Applies the configuration to the internal helpers:
      - `make_request`
      - `amake_request`
      - `amake_requests`
    - Handles the scenarios where libraries are using the Requests library directly. This is done by manipulating environment variables before and after making the request.
      - yFinance
      - Finviz
      - Posthog
      - OpenBB Hub
    - Allows keyword arguments to be passed to `uvicorn.run` by storing them as a dictionary to: `system_settings.python_settings.uvicorn`

3. **Impact**:

    - Should have absolutely no impact unless these items are specified.
    - The behavior of Requests environment variable, `REQUESTS_CA_BUNDLE`, ports directly to the async requests.
    - Allows it to be defined in `system_settings.json` instead of environment variables.
    - `cafile` is an equivalent to `REQUESTS_CA_BUNDLE`, when pointing to a file.
    - When `"verify_ssl": false`, SSL certificate verification is disabled within all OpenBB functions.
    - Keyword arguments supplied to the `"python_settings["uvicorn"]"` dictionary are passed directly to `uvicorn.run` when launching the API as:
      - `python -m openbb_core.api.rest_api`
      - `openbb-api`

These items, in `system_settings.json`, will take precedence over environment variables:

```json
{
    "python_settings": {
        "http": {
            "cafile": "/full/path/to/certificate/localhost.crt",
            "verify_ssl": null,
            "proxy": null,
            "certfile": null,
            "keyfile": null,
        }
    }
}
```

> `certfile` and `keyfile` are intended more for future use, the most important key is `cafile`.

**Note**: Keyword arguments added to the command line from `openbb-api` take precedence over the `system_settings.json` file.

4. **Testing Done**:

    - Create a self-signed certificate and start an OpenBB API over HTTPS
    - Add the full path to the `.crt` file as shown above.
    - Import both `requests` and `openbb_core.provider.utils.helpers.make_request`
      - `requests.get` should fail while `make_request` succeeds.
      - Set as an environment variable instead, "cafile" is `null` and add `REQUESTS_CA_BUNDLE='/full/path/to/certificate/localhost.crt'` to the `.env` file.
        - Both `requests.get` and `make_request` should succeed.
        - Making an outside call using `requests.get("https://google.com")` will fail, `make_request` should succeed.

Screenshot 2024-11-29 at 3 01 25 PM

With the environment variable defined, and not `system_settings.json`, the same A/B can be applied to `yfinance.download()` vs. `obb.equity.price.historical(provider="yfinance")

This fails because yFinance is only verifying against the self-signed certificate for `localhost`. `openbb_yfinance.utils.helpers.yf_download` applies the environment configuration for the duration of the request.

> Update to the behavior: the implementation now impacts the yFinance singleton.

Screenshot 2024-11-29 at 3 07 01 PM


montezdesousa:
I have a few concerns with this solution.

1. We are modifying environment variables from within the application and the logic to so in `get_certificates` and `restore_certs` is complex to follow.
2. We are introduction a state inside `Env()`. I assume that's why we have some lines with `Env()` to refresh the state (?). This class should be stateless and readonly for consistency and predictability.
3. The implementation requires us to constantly get and restore certificates, which is not obvious when implementing new requests.

I suggest using a single session that picks configurations and is shared across the app as proposed here https://github.com/OpenBB-finance/OpenBB/issues/6939.

montezdesousa (commented):
I have a few concerns with this solution.

1. We are modifying environment variables from within the application and the logic to so in `get_certificates` and `restore_certs` is complex to follow.
2. We are introduction a state inside `Env()`. I assume that's why we have some lines with `Env()` to refresh the state (?). This class should be stateless and readonly for consistency and predictability.
3. The implementation requires us to constantly get and restore certificates, which is not obvious when implementing new requests.

I suggest using a single session that picks configurations and is shared across the app as proposed here https://github.com/OpenBB-finance/OpenBB/issues/6939.

deeleeramone:
> I suggest using a single session that picks configurations and is shared across the app as proposed here #6939.

OK, try it out now, @montezdesousa. I found a way to hack the session into the Posthog handler. The implementation now goes the other way where `trust_env` is disabled for Requests - which is the default for AIOHTTP - and cut everything down into two helper functions:
- `get_requests_session()`
- `get_async_requests_session()`

Using any of the utility HTTP request functions will read the system_settings.json and take precedence over any environment variable. Binding session objects to the `obb` app doesn't accomplish much because it needs to happen at the provider level, and we also don't want singleton behavior because it would mess with threading and concurrency.  But, a session of the appropriate type can be passed into the functions as a parameter. 
- `make_request`
- `amake_request`
- `amake_requests`

Any unclosed async session object will be closed on exit.

openbb_platform/core/openbb_core/api/rest_api.py
```diff
@@ -85,6 +85,16 @@ async def lifespan(_: FastAPI):
 
 
 if __name__ == "__main__":
+    # pylint: disable=import-outside-toplevel
     import uvicorn
 
-    uvicorn.run("openbb_core.api.rest_api:app", reload=True)
+    Env()
```
montezdesousa, line 91: Is this a leftover from a previous implementation?

piiq: if it is - it's also mentioned in the `openbb_platform/extensions/platform_api/openbb_platform_api/main.py`

deeleeramone: This is to bring the environment variables into the context of the script entry point, so variables can be read and established before launching. Otherwise they are not yet exposed when invoking uvicorn.

montezdesousa: Ok, I suggest we leave a comment saying that we are loading the .env such that `uvicorn` has access to the environment variables x, y, z.

montezdesousa:
When trying to test I get the same error both with `requests` and `make_request`. Also tried to provide the path to the certificate through `verify` and get the same error. Uvicorn is running over HTTPS and I get access it through the browser fine.

```
SSLError: HTTPSConnectionPool(host='0.0.0.0', port=8000): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1133)')))
```

deeleeramone:
> When trying to test I get the same error both with `requests` and `make_request`. Also tried to provide the path to the certificate through `verify` and get the same error. Uvicorn is running over HTTPS and I get access it through the browser fine.
> 
> ```
> SSLError: HTTPSConnectionPool(host='0.0.0.0', port=8000): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1133)')))
> ```

This error seems to indicate that the self-signed certificate is not being found at all. I know there's lots of ways to make it not work, so rooting out the edge cases would be nice.

What are your exact deployment steps here, is this a network deployment? Is the same `.crt` file accessible to both `uvicorn` and the Python environment? Something like:

```
    "python_settings": {
        "http": {
            "cafile": "/Users/darrenlee/OpenBBUserData/localhost.crt"
        },
        "uvicorn": {
            "ssl_certfile": "/Users/darrenlee/OpenBBUserData/localhost.crt",
            "ssl_keyfile": "/Users/darrenlee/OpenBBUserData/localhost.key"
        }
    }
```

Was the certificate created by the same version of Python and openssl as used in the deployment and environment?

openbb_platform/core/openbb_core/provider/utils/helpers.py
```diff
@@ -96,6 +96,254 @@ def get_querystring(items: dict, exclude: List[str]) -> str:
...
+            if python_settings["cafile"] != os.environ.get("REQUESTS_CA_BUNDLE")
+            else combine_certificates(python_settings["cafile"])
+        )
+    elif python_settings.get("cafile") is not None:
+        _session.verify = combine_certificates(python_settings["cafile"])
+    elif (
+        os.environ.get("REQUESTS_CA_BUNDLE") is not None
+        and python_settings.get("cafile") is None
+    ):
+        certs = os.environ.get("REQUESTS_CA_BUNDLE", "")
```
montezdesousa, line 167: The logic in this function is hard to follow with the `if`, `elif` and `.get` of the same variables.

```suggestion
def get_requests_session(**kwargs) -> "Session":
    """Get a requests session object with the applied user settings or environment variables."""
    # pylint: disable=import-outside-toplevel
    import requests

    # If a session is already provided, just return it.
    if "session" in kwargs and isinstance(kwargs.get("session"), requests.Session):
        return kwargs["session"]

    # We want to add a user agent to the request, so check if there are any headers
    # If there are headers, check if there is a user agent, if not add one.
    # Some requests seem to work only with a specific user agent, so we want to be able to override it.
    python_settings = get_python_request_settings()
    headers = kwargs.pop("headers", {})
    headers.update(python_settings.pop("headers", {}))

    if "User-Agent" not in headers:
        headers["User-Agent"] = get_user_agent()

    # Allow a custom session for caching, if desired
    _session: requests.Session = kwargs.pop("session", None) or requests.Session()
    _session.headers.update(headers)

    if python_settings.get("verify_ssl") is False:
        _session.verify = False
    else:
        ca_file = python_settings.get("cafile")
        requests_ca_bundle = os.environ.get("REQUESTS_CA_BUNDLE")
        cert = ca_file or requests_ca_bundle
        if cert:
            bundle = requests_ca_bundle if requests_ca_bundle != cert else None
            _session.verify = combine_certificates(cert, bundle)

    if certfile := python_settings.get("certfile"):
        keyfile = python_settings.get("keyfile")
        _session.cert = (certfile, keyfile) if keyfile else certfile

    proxy = python_settings.get("proxy")
    http_proxy = os.environ.get("HTTP_PROXY")
    https_proxy = os.environ.get("HTTPS_PROXY")
    if http_proxy or https_proxy or proxy:
        proxies: dict = {}
        if http := http_proxy or https_proxy or proxy:
            proxies["http"] = http
        if https := https_proxy or http_proxy or proxy:
            proxies["https"] = https
        _session.proxies = proxies

    if cookies := python_settings.get("cookies"):
        _session.cookies = (
            cookies
            if isinstance(cookies, requests.cookies.RequestsCookieJar)
            else requests.cookies.cookiejar_from_dict(cookies)
        )

    if auth := python_settings.get("auth"):
        _session.auth = (
            auth if isinstance(auth, (tuple, requests.auth.AuthBase)) else tuple(auth)
        )

    if kwargs:
        for key, value in kwargs.items():
            try:
                if hasattr(_session, key):
                    if hasattr(getattr(_session, key, None), "update"):
                        getattr(_session, key, {}).update(value)
                    else:
                        setattr(_session, key, value)
            except AttributeError:
                continue

    _session.trust_env = False

    return _session
```

openbb_platform/core/openbb_core/provider/utils/helpers.py
```diff
@@ -96,6 +96,254 @@ def get_querystring(items: dict, exclude: List[str]) -> str:
...
+    if kwargs:
+        for key, value in kwargs.items():
+            try:
+                if hasattr(_session, key):
+                    if hasattr(getattr(_session, key, None), "update"):
+                        getattr(_session, key, {}).update(value)
+                    else:
+                        setattr(_session, key, value)
+            except AttributeError:
+                continue
```
montezdesousa, line 223: I don't see this being used anywhere, what's the purpose of updating the keys of kwargs?

openbb_platform/core/openbb_core/provider/utils/helpers.py
```diff
@@ -96,6 +96,254 @@ def get_querystring(items: dict, exclude: List[str]) -> str:
...
+
+    _session.trust_env = False
+
+    return _session
+
+
+async def get_async_requests_session(**kwargs) -> ClientSession:
```
montezdesousa, line 230: Same readability comment as for `get_requests_session` I suggest following the same approach of defining the variable if it's used multiple times (also more efficient).

montezdesousa (commented):
I tested and it works as expected, just left some comments and suggestions mostly on code readability.

montezdesousa:
> > When trying to test I get the same error both with `requests` and `make_request`. Also tried to provide the path to the certificate through `verify` and get the same error. Uvicorn is running over HTTPS and I get access it through the browser fine.
> > ```
> > SSLError: HTTPSConnectionPool(host='0.0.0.0', port=8000): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1133)')))
> > ```
> 
> This error seems to indicate that the self-signed certificate is not being found at all. I know there's lots of ways to make it not work, so rooting out the edge cases would be nice.
> 
> What are your exact deployment steps here, is this a network deployment? Is the same `.crt` file accessible to both `uvicorn` and the Python environment? Something like:
> 
> ```
>     "python_settings": {
>         "http": {
>             "cafile": "/Users/darrenlee/OpenBBUserData/localhost.crt"
>         },
>         "uvicorn": {
>             "ssl_certfile": "/Users/darrenlee/OpenBBUserData/localhost.crt",
>             "ssl_keyfile": "/Users/darrenlee/OpenBBUserData/localhost.key"
>         }
>     }
> ```
> 
> Was the certificate created by the same version of Python and openssl used in the deployment and environment?

Thanks for the help. I'm using `mkcert` and it turns out the `cafile` needs to come from `/Users<username>/Library/Application Support/mkcert/rootCA.pem` and not the self signed certificate itself.

openbb_platform/core/openbb_core/provider/utils/helpers.py
deeleeramone: In the sync/async functions the kwargs are applied in the function, but if you were to use this as a standalone function to create a session object, you might want to set something not defined by `python_settings`.

This was the intention at least.

montezdesousa (approved):
Lgtm

Edit: updating review

openbb_platform/core/openbb_core/provider/utils/helpers.py
```diff
@@ -96,6 +96,254 @@ def get_querystring(items: dict, exclude: List[str]) -> str:
...
+        "proxy_headers",
+        "timeout",
+        "auth",
+        "headers",
+        "cookies",
+    ]
+
+    return {
+        k: v for k, v in http_settings.items() if v is not None and k in allowed_keys
+    }
```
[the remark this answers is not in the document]
montezdesousa: They are hidden in the code, documentation is useful for detailing what code cannot convey. The benefit is explicit field definition, one should know all the python settings by looking at the `PythonSettings` model.

With `extra_allow=True` you open the door to implicit definitions all over the code like `http` that only lives inside the helper function `get_python_request_settings`. If we add a few more like this and it will become hard to track which settings exist.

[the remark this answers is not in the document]
montezdesousa: They are hidden in the code, documentation is useful for detailing what code cannot convey. The benefit is explicit field definition, one should know all the python settings by looking at the `PythonSettings` model.

With `extra_allow=True` you open the door to implicit definitions all over the code with no real benefit, like `http` that only lives inside the helper function `get_python_request_settings`. If we add a few more like this and it will become hard to track which settings exist.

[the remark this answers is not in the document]
montezdesousa: They are hidden in the code, documentation is useful for detailing what code cannot convey. The benefit is explicit field definition, one should know all the python settings by looking at the `PythonSettings` model.

With `extra_allow=True` you open the door to implicit definitions all over the code with no real benefit, like `http` (I think a better name would be something like `request_settings`, `http` is too general) that only lives inside the helper function `get_python_request_settings`. If we add a few more like this and it will become hard to track which settings exist.

montezdesousa (changes requested):

montezdesousa (approved):

PR #6988, opened 2024-12-14
[Feature] Add `no_validate` Flag To `@router.command`

deeleeramone:
1. **Why**?:
    - To make it possible to return primitive types and other serializable content.
    - Suitable for function callbacks and situations where the receiver requires a specific format or shape.
    - To make the API compatible with Widgets 2.0.

2. **What**?:

    - Setting, `@router.command(..., no_validate=True)`, in the function's router will defeat output type checking and validation by enforcing "response_model=None" and removing the response annotation.
    - The kwarg  is stored in `openapi_extra["no_validate"]`
    - Removes some unused code in "router.py", `CommandValidator`
      - Removes the tests which were the only reference to any part of the code.

3. **Impact**:

    - No impact unless the `@router.command` decorator explicitly has `no_validate=True`.
    - Improved developer experience and customization.
    - Allows returning a modified - transposed, pivoted, etc. - version of the results.
    - When True, the API response model and return annotations are removed.
    - These are defined by `PackageBuilder` and `ReferenceGenerator` as `Any`
    - Potential for these functions to break CLI.

4. **Testing Done**:

    - Provider outputs.
    - Data processing routes.
    - Custom functions.

To use, add `no_validate=True` to any router command. False is an equivalent to None. 

This will raise a ValueError when set to False, but works when True.
```python
@router.command(
    methods=["POST"],
    no_validate=False,
    examples=[
        PythonEx(
            description="Get the Average True Range.",
            code=[
                "stock_data = obb.equity.price.historical(symbol='TSLA', start_date='2023-01-01', provider='fmp')",
                "atr_data = obb.technical.atr(data=stock_data.results)",
            ],
        ),
        APIEx(parameters={"length": 2, "data": APIEx.mock_data("timeseries")}),
    ],
)
def atr(
    data: List[Data],
    index: str = "date",
    length: PositiveInt = 14,
    mamode: Literal["rma", "ema", "sma", "wma"] = "rma",
    drift: NonNegativeInt = 1,
    offset: int = 0,
) -> "DataFrame":
    """Calculate the Average True Range."""
    ...
```

```sh
  File "/Users/darrenlee/github/OpenBB/openbb_platform/core/openbb_core/app/static/package_builder.py", line 344, in get_function_hint_type_list
    raise ValueError("Return type must be an OBBject.")
ValueError: Return type must be an OBBject.
```

CommandRunner will reject a non-OBBject if `no_validate=False` or is not defined. This is done by exposing the ApiRoute and checking `openapi_extra`, where the @router.command kwarg is stored.

```sh
File ~/github/OpenBB/openbb_platform/core/openbb_core/app/command_runner.py:339, in StaticCommandRunner._execute_func(cls, route, args, execution_context, func, kwargs)
    337 if isinstance(obbject, OBBject) or validate:
    338     if type(obbject) is not OBBject:
--> 339         raise OpenBBError(
    340             TypeError(
    341                 f"Expected OBBject instance at function output, got {type(obbject)} instead."
    342             )
    343         )
    344     # This section prepares the obbject to pass to the charting service.
    345     obbject._route = route  # pylint: disable=protected-access

OpenBBError: 
[Error] -> Expected OBBject instance at function output, got <class 'pandas.core.frame.DataFrame'> instead.
```

`reference.json` replaces the `Returns` section with:

```json
{
            "returns": {
                "Any": {
                    "description": "Unvalidated results object."
                }
            }
}
```


openbb_platform/core/openbb_core/api/router/commands.py
```diff
@@ -167,6 +167,10 @@ def exclude_fields_from_api(key: str, value: Any):
                 elif is_model(getattr(field, "annotation", None)):
                     exclude_fields_from_api(field_name, getattr(value, field_name))
 
+    # Let a non-OBBject object pass through without validation
+    if not isinstance(c_out, OBBject):
```
piiq, line 171: I was reviewing this again and wanted to ask your opinion if we should at least validate that what the command returns is JSON serializable?

deeleeramone: FastAPI does this already for the API. From the Python interface, serialization is probably a waste of resources and unnecessary.

openbb_platform/core/tests/app/logs/test_logging_service.py
```diff
@@ -84,11 +84,10 @@ def test_correctly_initialized():
         "openbb_core.app.logs.logging_service.LoggingService._log_startup",
         mock_log_startup,
     ):
-        LoggingService(
+        ls = LoggingService(  # noqa: F841  # pylint: disable=unused-variable
```
piiq, line 87: If it's really unused there is a convention for this in python to name the variable _

Try and you will not need the comment that silences the linter

piiq (approved):

openbb_platform/core/tests/app/logs/test_logging_service.py
deeleeramone: That'll do, ya.

PR #6993, opened 2025-01-02
[Enhancement] provider_nasdaq cache some data

joshuaBri:
I noticed that these interfaces return data very slowly. Could you please add caching to improve the performance?


deeleeramone:
Hi there,

The economic calendar is a slow request because you need to request every date individually. Caching won't really help unless you repeatedly request the same data, and then if you are expecting new data from current events that would be excluded because it's already been cached.

What is your use case specifically that requires caching as opposed to your application storing results for recall later?

PR #6996, opened 2025-01-06
[BugFix] Temporarily point `poetry.lock` to the path for `openbb-core`

deeleeramone:
1. **Why**?:

    - The CI tests are confused by the recent change, #6994, and don't know what to do. 

2. **What**?:

    - Temporarily points the main `pyproject.toml` file to the directory path.

3. **Impact**:

    - `dev_install.py` already does this, but restores. This is just a temporary move until the next increment of `openbb-core` gets published.


deeleeramone:
Nope, not the issue.

PR #6998, opened 2025-01-06
[Feature] Add Deribit Futures Data

deeleeramone:
1. **Why**?:

    - Extend the `openbb-deribit` integration to include futures data.

2. **What**?:

    - Two new endpoints:
      - `obb.derivatives.futures.instruments`
        - Get the current list of all instruments here.
      - `obb.derivatives.futures.info`
    - Adds to existing:
      - `obb.derivatives.futures.curve`
        - Does not allow `date` but provides, `hours_ago`.
      - `obb.derivatives.futures.historical`

3. **Impact**:

    - Increases the data coverage.
    - New endpoints will only exist if the (optional) `openbb-deribit` package is installed.

4. **Testing Done**:

    - New tests added and cassettes captured.

image

image

image


openbb_platform/dev_install.py
```diff
@@ -147,7 +147,8 @@ def install_platform_local(_extras: bool = False):
         extras_args = ["-E", "all"] if _extras else []
 
         subprocess.run(
-            CMD + ["lock", "--no-update"],
+            CMD + ["lock"],
+            # CMD + ["lock", "--no-update"],
```
IgorWounds, line 151: leftover?

openbb_platform/dev_install.py
```diff
@@ -192,7 +193,8 @@ def install_platform_cli():
         CMD = [sys.executable, "-m", "poetry"]
 
         subprocess.run(
-            CMD + ["lock", "--no-update"],
+            CMD + ["lock"],
+            # CMD + ["lock", "--no-update"],
```
IgorWounds, line 197: leftover?

openbb_platform/providers/deribit/openbb_deribit/models/futures_curve.py
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    symbol: FuturesCurveSymbols = Field(
+        default="BTC",
+        description=QUERY_DESCRIPTIONS.get("symbol", "")
+        + " Default is 'btc' Supported symbols are: ['btc', 'eth', 'paxg']",
+    )
+    hours_ago: Optional[Union[int, list[int], str]] = Field(
+        default=None,
+        description="Compare the curve with the curve from the specified hours ago. Default is None.",
```
IgorWounds, line 48: What curve?

IgorWounds (approved):
Minor edits. This looks very cool!

openbb_platform/extensions/derivatives/openbb_derivatives/derivatives_views.py
```diff
@@ -107,7 +109,8 @@ def derivatives_futures_curve(  # noqa: PLR0912
 
         provider = kwargs.get("provider", "")
 
-        df["expiration"] = df["expiration"].apply(to_datetime).dt.strftime("%b-%Y")
+        if provider != "deribit":
```
piiq, line 112: Suggestion here:

Is there a way to avoid provider specific logic in the views? I see we have this in 3 views currently: this one, economy and fixedincome.

IIUC these are now used to "standardize" the data provided by providers to create a standardized chart. Would it be possible/make sense to move this logic to the provider level? For example to a custom `model_dump` in the Data model which could add this remaining additional layer of standardization that is currently missing. This would make the views more universal, without hardcoded logic for providers which might or might not be there and generally would facilitate better modularity. Let me know what you think

piiq (commented):
I've spotted that we have provider specific logic in the views and left a suggestion about it. Let me know what you think of it

openbb_platform/dev_install.py
deeleeramone: I might be able to restore that now that I've cleared and rebuilt the Github Actions cache..

openbb_platform/providers/deribit/openbb_deribit/models/futures_curve.py
deeleeramone: The name of the function is "futures_curve", am I assuming too much in the description?

openbb_platform/extensions/derivatives/openbb_derivatives/derivatives_views.py
deeleeramone: I like where your head is at, but I don't think this PR is the place to address this. We would need to reconsider the entire flow of the views.

For one, any presentation layer will always need a level of abstraction on top of the Data model. There aren't really many scenarios where dumping raw data into a chart is desirable.

If we were to move some logic to live in the provider, then we essentially need to create a chart/view inside every provider data model definition. This would create a lot of overhead and make assumptions for views that may not be installed or are not desired. Right now the charts are fully decoupled, but mixing it in the provider logic would blur that.

A custom "model_dump" doesn't necessarily help because the data needs to be aggregated as a complete table before some of downstream decisions can be made. A model_dump would only operate on single rows.

Additionally, there is logic applied that handles the user passing in data that is not directly from the raw data model output.

At the provider level, we don't have access to the OBBject and the ability to modify locals to pass forward to the charting extension.

That being said, the Platform charting ecosystem and framework needs to be revisited anyways - i.e, get rid of frontend-components - so, your points are very valid and should be a consideration when planning a modernized backend.

openbb_platform/dev_install.py
deeleeramone: Nope, experiment over.

PR #7017, opened 2025-02-03
Add a template.json out-of-the-box instead of empty file

DidierRLopes:
The start of something beautiful

Screenshot 2025-02-02 at 11 07 45 PM


DidierRLopes:
Ups - https://github.com/OpenBB-finance/OpenBB/pull/7018

PR #7029, opened 2025-02-12
Idea for custom built in types

jmaslek:
Idea that we can have some custom types for our ColDefs or whatever. 

This one is meant to mimic the options endpoint one.

Example here:

```
from fastapi import FastAPI
from openbb_platform_api import OptionsEndpoint

app = FastAPI()


def get_names() -> list[str]:
    return ["John", "Jane", "Jim", "Jill"]


@app.get("/hello")
def hello(name: OptionsEndpoint(get_names)) -> str:
    return f"Hello, {name}!"
```

And it has the desired dropdown.

But we could essentiually create types for everything we tell folks to do.

Screenshot 2025-02-12 at 2 21 41 PM


deeleeramone:
So when Workspace calls the function, what happens?

The parameter that is being defined in the API needs to correspond with what happens inside the function. For Workspace "optionsEndpoint", you need a defined route in the API for it to call and retrieve the list of [{"label": "some choice", "value": "actual_value"}].

Defining choices in this way does not make them dynamic, because, `openapi.json` is only read once. This is an equivalent to generating an enum of choices.

Dynamic choices are made by Workspace calling a defined API route with the rigid structure it expects, the widget parameter, itself, still needs to be of a primitive type.

PR #7031, opened 2025-02-14
[Feature] Align Python Interface With API - Stop Requiring `OBBject` As Output Type In `PackageBuilder` As Default

deeleeramone:
1. **Why**?:

    - This PR aligns the Python interface with recent changes from `openbb-platform-api`, removing the requirement for  the output of all functions be an instance of OBBject.

2. **What**?:

    - Removes strict type output enforcement that was limited to only the Python interface.

3. **Impact**:

    - Ensures Python interface doesn't fail to build where the API works as-expected.
    - Does not impact API.

4. **Testing Done**:

    - Tested with a custom router and bare OpenBB package.
      - Aside from output type freedom, look at docstrings and reference.json
    - Reference Generator can use some further optimization, but leaving that for another day.


openbb_platform/core/openbb_core/app/command_runner.py
```diff
@@ -329,19 +329,12 @@ async def _execute_func(  # pylint: disable=too-many-positional-arguments
...
             } or None
 
-            validate = not execution_context.api_route.openapi_extra.get("no_validate")
             try:
                 obbject = await cls._command(func, kwargs)
                 # The output might be from a router command with 'no_validate=True'
                 # It might be of a different type than OBBject.
                 # In this case, we avoid accessing those attributes.
-                if isinstance(obbject, OBBject) or validate:
-                    if validate and not isinstance(obbject, OBBject):
```
piiq, line 339: yes

piiq (approved):

PR #7040, opened 2025-02-22
[Feature] Allow Arbitrary Credentials From `user_settings.json` To Populate Credentials Model

deeleeramone:
1. **Why**?:

    - Items defined in `user_settings["credentials"]` only populate if they are in the ProviderInterface, this is not ideal.

2. **What**?:

    - Populates the Credentials model with any leftover key:value pairs from `user_settings.json`, if it exists.

3. **Impact**:

    - Improved developer experience, more freedom.
    - Arbitrary keys being defined have no real consequence, but greatly improve the flexibility - like incorporating Hub configurations with completely custom APIs and apps that are authorized by a remote script.

4. **Testing Done**:

Screenshot 2025-02-21 at 9 42 57 PM

### DIFF
--- a/openbb_platform/core/openbb_core/app/model/credentials.py
+++ b/openbb_platform/core/openbb_core/app/model/credentials.py
@@ -1,9 +1,12 @@
 """Credentials model and its utilities."""
 
+import json
 import traceback
 import warnings
+from pathlib import Path
 from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


piiq (approved):
I was under the impression that we already support this, but then i realized that all my custom keys are the keys to custom providers that are defined as providers

PR #7048, opened 2025-03-04
[BugFix] Downgrade Plotly To 5.24.1 For JSON Output Formatting

deeleeramone:
1. **Why**?:

    - In Plotly 6, when a Figure is serialized as JSON and converted back to a Figure, the chart may-or-may-not display with the content on the top layer.
    - The "data" is now a different format, something is getting lost in translation while rendering in the browser.
    - I'm pretty sure that I checked for this exact scenario when bumping to V6, and it appears to be some sort of conditional element to it, so I'm not entirely sure why it is like it is.

2. **What**?:

    - Downgrades to 5.24.1 to ensure compatibility and consistent output.
    - Will monitor for changes..

image

4. **Testing Done**:

Screenshot 2025-03-04 at 2 04 28 PM

You can make this chart by adding this to the `user_settings.json`

```json
{
    "defaults": {
        "commands": {
            "/equity/price/historical": {
                "provider": "yfinance",
                "chart": true,
                "chart_params": {
                    "heikin_ashi": true,
                    "indicators": {
                        "sma": {
                            "length": [
                                21,
                                50
                            ]
                        },
                        "ema": {
                            "length": 150
                        }
                    }
                }
            }
        }
    }
}
```

### DIFF
--- a/openbb_platform/obbject_extensions/charting/poetry.lock
+++ b/openbb_platform/obbject_extensions/charting/poetry.lock
@@ -787,33 +787,6 @@ files = [
 typing-extensions = {version = ">=4.1.0", markers = "python_version < \"3.11\""}
 
 [[package]]
-name = "narwhals"
-version = "1.26.0"
-description = "Extremely lightweight compatibility layer between dataframe libraries"
-optional = false
-python-versions = ">=3.8"
-groups = ["main"]
-files = [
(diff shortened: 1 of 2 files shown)


piiq:
@deeleeramone do you have a minimal reproducible example that can be submitted as a bug report to plotly?

piiq (approved):

PR #7060, opened 2025-03-12
Improve openbb api launcher

piiq:
This PR changes the `import_app` function to accommodate the following cases:

```
openbb-api --app my_module.main:app
openbb-api --app main.py:app
openbb-api --app my_folder/main.py:app
openbb-api --app main.py --name another_name
openbb-api --app my_module.main:create_app
openbb-api --app main:create_app --factory
```
and others.

So now the launcher can handle python modules, sub-modules, relative paths, absolute paths.
The factory launcher now uses similar logic to what `uvicorn` uses to detect the factory allows us to stop relying on explicit type annotation of the factory function.

Additinonally some tests for `import_app` were added, although 2/3 of them actually work

### DIFF
--- a/openbb_platform/extensions/platform_api/openbb_platform_api/utils/api.py
+++ b/openbb_platform/extensions/platform_api/openbb_platform_api/utils/api.py
@@ -211,7 +211,6 @@ def get_widgets_json(
     widgets_path: Optional[str] = None,
 ):
     """Generate and serve the widgets.json for the OpenBB Platform API."""
-
     if editable is True:
         if widgets_path is None:
             python_path = Path(sys.executable)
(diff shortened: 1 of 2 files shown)


openbb_platform/extensions/platform_api/openbb_platform_api/utils/api.py
```diff
@@ -276,58 +275,82 @@ def get_widgets_json(
...
+        sys.modules[spec_name] = module  # type: ignore
+        spec.loader.exec_module(module)  # type: ignore
+        return module
+
+    # Case 1: Module path with colon notation (e.g., "my_app.main:app" or "main:app")
+    if ":" in app_path:
+        module_path, name = app_path.split(":")
+        try:  # First try to import as a module
+            module = import_module(module_path)
+        except ImportError:  # If module import fails, try to load as a local file
```
deeleeramone, line 303: This should expand to other base error messages that might trigger, for example, `ModuleNotFoundError`

Screenshot 2025-03-12 at 8 45 49 PM

deeleeramone: In theory, it should always work. I'm not sure of the context for where/when I encountered the reason for me suggesting this, so I'll pass on my comment. :)

deeleeramone (approved):

PR #7082, opened 2025-03-31
[Release] OpenBB V4.4.2

deeleeramone:
This PR is for the release of OpenBB, version 4.4.2


piiq:
@deeleeramone do we need to create new installers because of the minimal python version bump?

1 minute later answering my own question: we don't. it's on 3.12

piiq (approved):

PR #7088, opened 2025-04-09
Patch vulnerabilites in frontend components

piiq:
- bumped vite, esbuild, vite-react
- tested the plots and tables work
- rebuilt the minified assets
- tested them with openbb-cli


deeleeramone (approved):
Thank you.

PR #7100, opened 2025-04-25
Release/4.4.4

deeleeramone:
Starting the release PR.. more to follow.


cli/pyproject.toml
```diff
@@ -17,7 +17,7 @@ openbb = 'openbb_cli.cli:main'
 python = "^3.9.21,<3.13"
 
 # OpenBB dependencies
-openbb = { version = "^4.4.2", extras = ["all"] }
+openbb = { version = "^4.4.4", extras = ["all"] }
```
piiq, line 20: For future consideration. Would it be possible to change this to pull `openbb-core` versus `openbb['all']`? Asking because using the cli in an environment where a subset of all packages is installed would noe be possible, since I can't install with this `--no-deps` flag and a regular install would lead to pulling `all`

piiq (approved):

cli/pyproject.toml
deeleeramone: Yes, we can definitely do this. This would have been my preferred configuration from inception. With respect to this, we should also optimize the `openbb` extras to be easier to be more flexible.

PR #7107, opened 2025-05-03
[BugFix] Patch yFinance For The Change To `curl-cffi`

deeleeramone:
1. **Why**?:

    - yFinance has changed their session object to use `curl-cffi`
    - All yFinance functions are broken under normal conditions. 

2. **What**?:

    - Uses `curl-adapter` to mount the Requests Session object for use with the new framework.

3. **Impact**:

    - `curl-adapter` maintains the complete functionality of Requests, so all user-defined session attributes should remain intact.
    - https://curl-cffi.readthedocs.io/en/latest/advanced.html#as-a-urllib3-requests-adapter

4. **Testing Done**:

    - You can tell it's working because the functions don't return a Rate Limit Error.


piiq (approved):
one question

do we actually need to mount the http session too?

deeleeramone:
> one question
> 
> do we actually need to mount the http session too?

For a proxy, ya?

PR #7112, opened 2025-05-08
[Release] OpenBB V4.4.4

### LINKED ISSUE #7111: [Bug] yfinance provider issue
**Describe the bug**
Retrieving historical index prices with yfinance as provider returns the following error:
1 Failed download:
['^SPX']: YFRateLimitError('Too Many Requests. Rate limited. Try after a while.')

**To Reproduce**
from openbb import obb
obb.user.preferences.output_type = "dataframe"
data = obb.equity.price.historical("^SPX",start_date = "2005-01-01",end_date = "2020-12-31",provider="yfinance")

**Desktop:**

- OS: Windows
- Jupyter Notebook

### LINKED ISSUE #7115: [Bug] Google Colab Options using YFinance
**Describe the bug**
I tried following the examples/impliedEarningsMove.ipynb but modified the provider of the options data to yfinance. The error throws

OpenBBError: 
[Unexpected Error] -> YFDataException -> Yahoo API requires curl_cffi session not <class 'requests.sessions.Session'>. Solution: stop setting session, let YF handle.

**To Reproduce**
```
from datetime import datetime, timedelta
from openbb import obb
obb.user.preferences.output_type = "dataframe"
```

```
# Lookup some upcoming earnings dates and sort them by market cap.

earnings_calendar = obb.equity.calendar.earnings(
    start_date=(datetime.now() + timedelta(days=1)).date(),
    end_date=(datetime.now() + timedelta(days=14)).date(),
    provider="nasdaq",
)

earnings_calendar.sort_values(by=["market_cap", "num_estimates"], ascending=False).head(
    20
)

```

```
# Get the options chains data.

symbol = "HIMS"  # This will not be evergreen, change the symbol based on a stock above.

obb.user.preferences.output_type = "OBBject"  # To use the built-in options chains methods, we need to set the output type to OBBject.

options = obb.derivatives.options.chains(symbol, provider="yfinance")

last_price = options.results.underlying_price[0]

display(f"Last Price: ${last_price}")

display(options.results.expirations[:3])
```

**Error code appeared**

Image

OpenBBError: 
[Unexpected Error] -> YFDataException -> Yahoo API requires curl_cffi session not <class 'requests.sessions.Session'>. Solution: stop setting session, let YF handle.

**Ran using Google Colab**
initially ran
`!pip install openbb["all"]`

and then when I encountered the error, re-ran pip install openbb["all"] and then
`!pip install --upgrade openbb-yfinance`

deeleeramone:
This PR merges the release to main back into develop.

Screenshot 2025-05-12 at 4 02 19 PM


openbb_platform/extensions/platform_api/pyproject.toml
```diff
@@ -15,7 +15,7 @@ openbb-api = "openbb_platform_api.main:main"
 
 [tool.poetry.dependencies]
 python = ">=3.9.21,<3.13"
-openbb-core = "^1.4.3"
+openbb-core = "^1.4.6"
```
piiq: Something I wanted to ask. Would it at some point be possible to make this package not rely on openbb-core?

Rationale:
I would like to use this package as a tool to work on FastAPI based app backends. Specifically to declare widgets at endpoint level and generate the widgets.json programmatically. I don't require pulling all the dependencies of openbb-core to those project. The --no-deps is an option but it won't work when specifying the dependencies in a pyproject.toml.
Heavy coupling with other openbb extensions and the core is what's "blocking" me from more adoption of this as a pure tool for parsing openapi schemas to integrate with the openbb workspace.

openbb_platform/obbject_extensions/charting/pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [{ include = "openbb_charting" }]
 
 [tool.poetry.dependencies]
 python = ">=3.9.21,<3.13"  # scipy forces python <4.0 explicitly
-openbb-core = "^1.4.3"
+openbb-core = "^1.4.6"
 pandas-ta-openbb = "^0.4.20"
 plotly = "^5.24.1"
```
piiq, line 14: Another note:
plotly 6 has been out for a while now. We might want to look into it. I've not seen breaking changes on the graph object interface level. But there are some things under the hood that did impact how the workspace sees the data that is being used by plotly.js (specifically how AI sees that)
Poke me once you'll start looking into upgrading and I will share what we already have found.

piiq (approved):

piiq:
I think I've found a typo
image

openbb_platform/extensions/platform_api/pyproject.toml
deeleeramone: Not sure I understand the request here.

What heavy coupling are you talking about?

`openbb-core` is not a heavy package and it gives you the base packages like FastAPI and Pydantic - you already need these packages. We need `openbb-core` for the FastAPI instance.

`openbb-core` is also responsible for `.env`, `user_settings.json` and `system_settings.json`.

openbb_platform/obbject_extensions/charting/pyproject.toml
deeleeramone: I had already looked at this and had to downgrade because the output of the serialized content is unexpected and was not layering correctly. One specific difference is, `to_plotly_json` produces a Bytes JSON instead of String.

deeleeramone:
Nice. Double MM

openbb_platform/obbject_extensions/charting/pyproject.toml
piiq: What we found is that that behavior is only applied to numpy arrays and pandas series, so if you cast the array to a list, tuple or a set the base64 compression is not applied

openbb_platform/extensions/platform_api/pyproject.toml
piiq: the coupling i mean are:
- dependencies: websockets, pandas, uuid7, requests, aiohttp, ruff, pyjwt. All these would/might not be relevant for me if I am building a custom app.
- functionality: the user/system settings, openbb namespace construction, etc. same applies to these. unless I'm using openbb-core as my application framework these are not relevant.

PR #6928, opened 2024-10-31
side quest

AshishViradiya153:
2 side quest submission

1: social-poster
2: 6-social-hero

#6906

PR #7119, opened 2025-05-17
Resolve Python Datetime warnings

emmanuel-ferdman:
# Pull Request OpenBB
This small PR resolves the annoying `datetime` library warnings:
```python
DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```

### DIFF
--- a/openbb_platform/providers/yfinance/openbb_yfinance/models/share_statistics.py
+++ b/openbb_platform/providers/yfinance/openbb_yfinance/models/share_statistics.py
@@ -5,6 +5,7 @@
 from datetime import (
     date as dateType,
     datetime,
+    timezone,
 )
 from typing import Any, Dict, List, Optional
 from warnings import warn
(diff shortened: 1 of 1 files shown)


deeleeramone (commented):
Thanks for the PR, every little bit helps!

Do you happen to be working from a Windows machine? 

While we're here, I see that we might run into this [issue](https://github.com/OpenBB-finance/OpenBB/pull/6298) with this piece of data. A company like IBM will create an OS error with a negative timestamp, so I'd like to check for that, and if required, mitigate with the helper function introduced in #6928.

deeleeramone (approved):
I was actually looking at the wrong property, the negative timestamp scenario would not happen from this function. Thanks for the PR and the patch!

PR #7124, opened 2025-05-17
[Security] Add `tornado` To `openbb-devtools` To Supercede Latest `ipykernel` and satisfy Dependabot

deeleeramone:
This PR is to resolve CVE-2025-47287

Tornado is brought in by `ipykernel`, however it requires a lower minimum version than described in the report, https://github.com/OpenBB-finance/OpenBB/security/dependabot/1933, 

We aren't likely to see an updated version of `ipykernel` that changes the requirements, so this is the reasonable course of action.


piiq:
I've just tested the following
Running `poetry update tornado` while removing the top level pin from the `pyproject.toml` gives me 6.5 in the lock file and in the environment. I've tested this by installing in a fresh `venv` and running the `dev_install.py` script.

I'm hesitant to introduce tornado as a top level dependency because we 1) don't use it as a top level dependency and 2) can avoid using the vulnerable package

Take a look at this please: https://github.com/OpenBB-finance/OpenBB/pull/7126 it unpins tornado and the only thing that changes in the lock file compared to this PR is the content hash.

piiq (changes requested):
See https://github.com/OpenBB-finance/OpenBB/pull/7126

deeleeramone:
Close in favor of #7126

PR #7127, opened 2025-05-19
[BugFix] `openbb-platform-api` - Fix `FileExistsError` when specifying `.json` in the `widgets-path`

deeleeramone:
1. **Why**?

    - When launching `openbb-api` with `--widgets-path`, specifying the file, instead of just path, can create a `FileExistsError`

2. **What**?:

    - Adds handling for specifying an arbitrary JSON file to represent `widgets.json`, `apps.json`, and `agents.json` in the API.
    - Updates arguments (with backwards-compatibility) to be `--json`, instead of `--path`, making it more intuitive. Specifying as a path only will default to its standard filename.
      - `--widgets-json`
      - `--apps-json`
      - `--agents-json`

3. **Impact**:

    - Bug fix.
    - When specifying a `widgets.json` path, if the file already exists, the launcher will assume it is built.
      - Explicitly engage `--build` to rebuild and get prompted for diff action.
    - Parameter changes are expected to be backwards compatible.
    - More clarity for the end-user.

5. **Testing Done**:

Before, adding `--no-build` was necessary to defeat the user-input prompt, and unexpected results would occur when specifying a filename in the `--path` arguments.

Before:

Screenshot 2025-05-19 at 4 24 11 PM

After:

Screenshot 2025-05-19 at 4 26 20 PM

### DIFF
--- a/openbb_platform/extensions/platform_api/openbb_platform_api/main.py
+++ b/openbb_platform/extensions/platform_api/openbb_platform_api/main.py
@@ -52,15 +52,13 @@ _app = kwargs.pop("app", None)
 if _app:
     app = _app
 
-WIDGETS_PATH = kwargs.pop("widgets-path", None)
-TEMPLATES_PATH = kwargs.pop("apps-json", None) or kwargs.pop("templates-path", None)
+WIDGETS_PATH = kwargs.pop("widgets-json", None) or kwargs.pop("widgets-path", None)
+APPS_PATH = kwargs.pop("apps-json", None) or kwargs.pop("templates-path", None)
 EDITABLE = kwargs.pop("editable", None) is True or WIDGETS_PATH is not None
-
-
-DEFAULT_TEMPLATES_PATH = (
+DEFAULT_APPS_PATH = (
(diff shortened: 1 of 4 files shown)


openbb_platform/extensions/platform_api/openbb_platform_api/main.py
```diff
@@ -52,15 +52,13 @@
 if _app:
     app = _app
 
-WIDGETS_PATH = kwargs.pop("widgets-path", None)
-TEMPLATES_PATH = kwargs.pop("apps-json", None) or kwargs.pop("templates-path", None)
+WIDGETS_PATH = kwargs.pop("widgets-json", None) or kwargs.pop("widgets-path", None)
```
piiq, line 60: Let's internally define how long we will be maintaining the backwards compatibility in order not to bloat the code with logic we forget about. A comment saying it should be removed by date X or version X will make out future selves say "thank you"

piiq (approved):
At first I was a little bit confused by the explanation of the changes of the PR because a path to a file is still a path, same as a path to a folder is a path. But after reviewing the diff I understood what's going on. Thank you for fixing this.

PR #7132, opened 2025-05-22
[Feature] Use Plotly V6 For Charting Extension +

deeleeramone:
1. **Why**?:

    - Update to the latest Plotly version.
    - Do some cleanup/fixes in `frontend-components`
    - Add animation to Relative Rotation chart with tails.

2. **What**?:

    - Updates version of Plotly to latest.
    - Fixes some PyWry window issues:
      - Text annotations work again and can be dragged and edited.
      - Axes title changes now immediately reflected in UI on submit.
      - Removed export buttons (better to remove at this point and add back if actually necessary - problematic)
      - Removes Posthog JS entirely.
    - Updates serialization output handling 

3. **Impact**:

    - Improved relative rotation chart.
    - Only real difference should be the removal of the PyWry window, on-chart, export buttons; and modebar set to 
"hover"

4. **Testing Done**:

    - Checked Integration and unit tests for Platform and CLI
    - Ran examples code in the charting extension's README files.

```python
data = obb.equity.price.historical("XLB,XLC,XLE,XLF,XLI,XLK,XLP,XHB,XLU,XLV,XLY,XLRE,SPY", start_date="2022-05-20", provider="yfinance")

rrg = obb.technical.relative_rotation(data=data.results, benchmark="SPY", chart=True, chart_params=dict(show_tails=True))

rrg.show()
```
Screenshot 2025-05-21 at 10 42 18 PM


piiq:
@andrewkenreich do we need to use v6 on workspace to support this update?

andrewkenreich:
> @andrewkenreich do we need to use v6 on workspace to support this update?

No we already had people using v6 and having charts work. That's how we found the issue originally

frontend-components/plotly/package.json
```diff
@@ -15,7 +15,6 @@
     "esbuild": ">=0.25.0",
     "lodash": "^4.17.21",
     "plotly.js-dist-min": "^3.0.0",
-    "posthog-js": "^1.55.0",
```
piiq, line 18: yes

frontend-components/plotly/package.json
```diff
@@ -34,7 +33,6 @@
     "autoprefixer": "^10.4.13",
     "clsx": "^1.2.1",
     "postcss": "^8.4.21",
-    "posthog-js": "^1.55.0",
```
piiq, line 37: can you also consider scooping posthog out of other extensions? we don't use it on the platform anymore

piiq (approved):

frontend-components/plotly/package.json
deeleeramone: Yes, this is definitely part of my agenda here.

PR #7134, opened 2025-05-29
[Feature] IMF Port Watch

deeleeramone:
This PR implements: https://portwatch.imf.org/pages/port-monitor

1. **Why**?:

    - Enhances the data available for macroeconomic analysis.
    - Adds a unique dataset that is not available from any other Python library.

2. **What**?:

    Adds a `shipping` route to the `obb.economy` route.

    Currently 4 endpoints and 2 charts

    - `obb.economy.shipping.chokepoint_info` (chart)
    - `obb.economy.shipping.chokepoint_volume`
    - `obb.economy.shipping.port_info` (chart)
    - `obb.economy.shipping.port_volume`

    Volumes data is daily, and begins at 2019-01-01. It is updated with a 1-week lag.

3. **Impact**:

    - Existing endpoint - `obb.economy.port_volume` - is now defined as deprecated and moved to `obb.economy.shipping.port_volume`
    - Adds priorty in `openbb-platform-api` Chart widgets `gridData` params to be from `openapi_extra`, so size can be properly set, if desired.
    - Removes redundant Charting extension call on `show` that was not allowing the intended `paperbg_color` to be used. This should already be handling by the charting function before it reaches this part of the code.

4. **Testing Done**:

    - Unit and integration tests.
    - Works from all interfaces.

Screenshot 2025-05-30 at 9 57 49 AM

Screenshot 2025-05-30 at 10 03 00 AM

Screenshot 2025-05-30 at 9 59 41 AM


openbb_platform/extensions/economy/openbb_economy/economy_router.py
```diff
@@ -1,5 +1,7 @@
 """Economy Router."""
 
+# pylint: disable=unused-argument
```
piiq, line 3: You sure you want this to be silenced file-wide?

openbb_platform/extensions/economy/openbb_economy/economy_views.py
```diff
@@ -481,7 +482,7 @@ def economy_survey_bls_series(
                     data=new_df,
                     title=title,
                     ytitle=ytitle,
-                    x=new_df.index,
+                    x=new_df.index,  # type: ignore
```
piiq, line 485: would `new_df.index.name` fix the type lint error?

openbb_platform/extensions/economy/openbb_economy/economy_views.py
```diff
@@ -328,7 +329,7 @@ def economy_fred_series(
     @staticmethod
     def economy_survey_bls_series(
         **kwargs,
-    ) -> Tuple["OpenBBFigure", Dict[str, Any]]:
+    ) -> tuple["OpenBBFigure", dict[str, Any]]:
```
piiq, line 332: Was talking to pylance and mypy and it seems that this actually is `tuple[Union["OpenBBFigure", "Figure"], dict[str, Any]]`

openbb_platform/extensions/economy/openbb_economy/economy_router.py
piiq: I've tried removing this locally and the errors were not raised (but my IDE might not be fully configured)

piiq (approved):
I've left a few comments on the router side, but didn't go too deep into the weeds elsewhere as it would have chewed up my entire day (yes, maintaining this codebase is actually very pleasing)

openbb_platform/extensions/economy/openbb_economy/economy_views.py
deeleeramone: No, and then you would also not be passing the array. The linters here are seeing everything as a Numpy Array.

openbb_platform/extensions/economy/openbb_economy/economy_views.py
deeleeramone: This is where these things get annoyingly silly and needless. "OpenBBFigure" inherits from go.Figure, and we are returning an "OpenBBFigure".

openbb_platform/extensions/economy/openbb_economy/economy_router.py
deeleeramone: Yes, pylint doesn't like the routers and fetchers because we aren't "using" them according to their liking. If I take that out, pylint blows up.

Screenshot 2025-06-03 at 3 56 00 PM

PR #7141, opened 2025-06-07
Add congress.gov provider extension

DidierRLopes:


deeleeramone:
This works really well, I do have a question though. What am I supposed to do with the results?

```
In [1]: from openbb import obb

In [2]: res = obb.government.us.congress_bills()

In [3]: res.model_dump()["results"][0]
Out[3]: 
{'congress': 119,
 'latestAction': {'actionDate': datetime.date(2025, 6, 16),
  'text': 'Submitted in the Senate, considered, and agreed to without amendment and with a preamble by Unanimous Consent. (consideration: CR S3405; text: CR S3404)'},
 'number': '279',
 'origin_chamber': 'Senate',
 'origin_chamber_code': 'S',
 'title': 'A resolution designating June 2025 as "Great Outdoors Month".',
 'type': 'SRES',
 'update_date': datetime.date(2025, 6, 17),
 'update_date_including_text': '2025-06-17',
 'url': 'https://api.congress.gov/v3/bill/119/sres/279?format=json'}
```

This feels like it supposed to happen behind the scenes because following the "url" just sends to to get more metadata, albeit more useful, but I still can't do anything with the output. We need the URLs to the actual bills - PDFs, HTML, etc. -  as the one-liner.

```json
{
    "bill": {
        "actions": {
            "count": 3,
            "url": "https://api.congress.gov/v3/bill/119/sres/279/actions?format=json"
        },
        "congress": 119,
        "cosponsors": {
            "count": 9,
            "countIncludingWithdrawnCosponsors": 9,
            "url": "https://api.congress.gov/v3/bill/119/sres/279/cosponsors?format=json"
        },
        "introducedDate": "2025-06-16",
        "latestAction": {
            "actionDate": "2025-06-16",
            "text": "Submitted in the Senate, considered, and agreed to without amendment and with a preamble by Unanimous Consent. (consideration: CR S3405; text: CR S3404)"
        },
        "number": "279",
        "originChamber": "Senate",
        "originChamberCode": "S",
        "sponsors": [
            {
                "bioguideId": "D000618",
                "firstName": "Steve",
                "fullName": "Sen. Daines, Steve [R-MT]",
                "isByRequest": "N",
                "lastName": "Daines",
                "party": "R",
                "state": "MT",
                "url": "https://api.congress.gov/v3/member/D000618?format=json"
            }
        ],
        "title": "A resolution designating June 2025 as \"Great Outdoors Month\".",
        "titles": {
            "count": 2,
            "url": "https://api.congress.gov/v3/bill/119/sres/279/titles?format=json"
        },
        "type": "SRES",
        "updateDate": "2025-06-17T13:45:14Z",
        "updateDateIncludingText": "2025-06-17T13:45:14Z"
    },
    "request": {
        "billNumber": "279",
        "billType": "sres",
        "congress": "119",
        "contentType": "application/json",
        "format": "json"
    }
}
```

DidierRLopes:
> This works really well, I do have a question though. What am I supposed to do with the results?
> 
> ```
> In [1]: from openbb import obb
> 
> In [2]: res = obb.government.us.congress_bills()
> 
> In [3]: res.model_dump()["results"][0]
> Out[3]: 
> {'congress': 119,
>  'latestAction': {'actionDate': datetime.date(2025, 6, 16),
>   'text': 'Submitted in the Senate, considered, and agreed to without amendment and with a preamble by Unanimous Consent. (consideration: CR S3405; text: CR S3404)'},
>  'number': '279',
>  'origin_chamber': 'Senate',
>  'origin_chamber_code': 'S',
>  'title': 'A resolution designating June 2025 as "Great Outdoors Month".',
>  'type': 'SRES',
>  'update_date': datetime.date(2025, 6, 17),
>  'update_date_including_text': '2025-06-17',
>  'url': 'https://api.congress.gov/v3/bill/119/sres/279?format=json'}
> ```
> 
> This feels like it supposed to happen behind the scenes because following the "url" just sends to to get more metadata, albeit more useful, but I still can't do anything with the output. We need the URLs to the actual bills - PDFs, HTML, etc. - as the one-liner.
> 
> ```json
> {
>     "bill": {
>         "actions": {
>             "count": 3,
>             "url": "https://api.congress.gov/v3/bill/119/sres/279/actions?format=json"
>         },
>         "congress": 119,
>         "cosponsors": {
>             "count": 9,
>             "countIncludingWithdrawnCosponsors": 9,
>             "url": "https://api.congress.gov/v3/bill/119/sres/279/cosponsors?format=json"
>         },
>         "introducedDate": "2025-06-16",
>         "latestAction": {
>             "actionDate": "2025-06-16",
>             "text": "Submitted in the Senate, considered, and agreed to without amendment and with a preamble by Unanimous Consent. (consideration: CR S3405; text: CR S3404)"
>         },
>         "number": "279",
>         "originChamber": "Senate",
>         "originChamberCode": "S",
>         "sponsors": [
>             {
>                 "bioguideId": "D000618",
>                 "firstName": "Steve",
>                 "fullName": "Sen. Daines, Steve [R-MT]",
>                 "isByRequest": "N",
>                 "lastName": "Daines",
>                 "party": "R",
>                 "state": "MT",
>                 "url": "https://api.congress.gov/v3/member/D000618?format=json"
>             }
>         ],
>         "title": "A resolution designating June 2025 as \"Great Outdoors Month\".",
>         "titles": {
>             "count": 2,
>             "url": "https://api.congress.gov/v3/bill/119/sres/279/titles?format=json"
>         },
>         "type": "SRES",
>         "updateDate": "2025-06-17T13:45:14Z",
>         "updateDateIncludingText": "2025-06-17T13:45:14Z"
>     },
>     "request": {
>         "billNumber": "279",
>         "billType": "sres",
>         "congress": "119",
>         "contentType": "application/json",
>         "format": "json"
>     }
> }
> ```

This is the equivalent of Fred Search effectively. I was waiting for you to check whether you liked the implementation before attempting to add the other ones.

There's a lot here: https://api.congress.gov/#/

deeleeramone:
> This is the equivalent of Fred Search effectively. I was waiting for you to check whether you liked the implementation before attempting to add the other ones.

I disagree. You can use the output of FRED Search to download data and search quite extensively via tags. Here, you cannot search, only apply very high-level filters. The resulting URLs to the next API call is not really any use to the user, they won't know what the POST request schema is, etc.

What is the user supposed to actually do with the returned results of this function?

openbb_platform/providers/congress_gov/openbb_congress_gov/models/congress_bills.py
```diff
@@ -0,0 +1,188 @@
...
+            },
+        },
+        "offset": {
+            "x-widget_config": {
+                "exclude": True,
+            },
+        },
+    }
+
+    format: Literal["json", "xml"] = Field(
```
deeleeramone: `format` is a reserved word in Python that is a callable. As we are expecting to handle JSON content in the function after the request is made, it doesn't make sense to have this parameter. This would be a "choice" that is made in the code and the user doesn't even know its happening.

Selecting "xml" simply makes an error because `amake_request` is a generic JSON handler.

```python
In [1]: from openbb import obb

In [2]: obb.government.us.congress_bills(format="xml")
---------------------------------------------------------------------------
OpenBBError                               Traceback (most recent call last)
Cell In[2], line 1
----> 1 obb.government.us.congress_bills(format="xml")

File ~/github/OpenBB/openbb_platform/core/openbb_core/app/static/utils/decorators.py:101, in exception_handler.<locals>.wrapper(*f_args, **f_kwargs)
     99         raise OpenBBError(f"\n[Error] -> {e}").with_traceback(tb) from None
    100     if isinstance(e, Exception):
--> 101         raise OpenBBError(
    102             f"\n[Unexpected Error] -> {e.__class__.__name__} -> {e}"
    103         ).with_traceback(tb) from None
    105 return None

File ~/miniconda3/envs/obb/lib/python3.12/site-packages/aiohttp/client_reqrep.py:749, in ClientResponse.json(self, encoding, loads, content_type)
    747     ctype = self.headers.get(hdrs.CONTENT_TYPE, "").lower()
    748     if not _is_expected_content_type(ctype, content_type):
--> 749         raise ContentTypeError(
    750             self.request_info,
    751             self.history,
    752             status=self.status,
    753             message=(
    754                 "Attempt to decode JSON with unexpected mimetype: %s" % ctype
    755             ),
    756             headers=self.headers,
    757         )
    759 stripped = self._body.strip()  # type: ignore[union-attr]
    760 if not stripped:

OpenBBError: 
[Unexpected Error] -> ContentTypeError -> 200, message='Attempt to decode JSON with unexpected mimetype: application/xml', url='https://api.congress.gov/v3/bill?api_key=********&format=xml&limit=100&sort=updateDate%2Basc'

```

deeleeramone (dismissed):

DidierRLopes:
> > This is the equivalent of Fred Search effectively. I was waiting for you to check whether you liked the implementation before attempting to add the other ones.
> 
> I disagree. You can use the output of FRED Search to download data and search quite extensively via tags. Here, you cannot search, only apply very high-level filters. The resulting URLs to the next API call is not really any use to the user, they won't know what the POST request schema is, etc.
> 
> What is the user supposed to actually do with the returned results of this function?

To know what are the latest bills (irrespective of congress, bill type and bill number) - otherwise how would I do it?

openbb_platform/providers/congress_gov/openbb_congress_gov/models/congress_bills.py
DidierRLopes: Good catch - will just get rid of it 👍🏽

deeleeramone:
> > What is the user supposed to actually do with the returned results of this function?
> 
> To know what are the latest bills (irrespective of congress, bill type and bill number) - otherwise how would I do it?

Ok, but what are you supposed to do with it? Like what kind of analysis does one perform with this data? 

What OpenBB endpoint do I use to get the real actual data?

What are the definitions of the acronyms?

All of the parameters for the "bill" route of this API are {path} parameters and need to be implemented here as the query parameters to this function, then we need an additional function that actually downloads the PDF/HTML/TXT content from the URL.

Screenshot 2025-06-19 at 6 07 40 PM

The thing that the end-user actually needs, is the ability to programmatically access the documents.

Screenshot 2025-06-19 at 6 08 39 PM

Note that the URLs pointing to the downloadable content do not require an API key, and those are the URLs which the "search" should reveal.

So there would be "government.us.bill_search", and "government.us.bill_download".  Traversing the content needs to be accomplished via the endpoints because you need to pass credentials to do anything.

DidierRLopes:
> > > What is the user supposed to actually do with the returned results of this function?
> > 
> > 
> > To know what are the latest bills (irrespective of congress, bill type and bill number) - otherwise how would I do it?
> 
> Ok, but what are you supposed to do with it? Like what kind of analysis does one perform with this data?
> 
> What OpenBB endpoint do I use to get the real actual data?
> 
> What are the definitions of the acronyms?
> 
> All of the parameters for the "bill" route of this API are {path} parameters and need to be implemented here as the query parameters to this function, then we need an additional function that actually downloads the PDF/HTML/TXT content from the URL.
> 
> Screenshot 2025-06-19 at 6 07 40 PM
> 
> The thing that the end-user actually needs, is the ability to programmatically access the documents.
> 
> Screenshot 2025-06-19 at 6 08 39 PM
> 
> Note that the URLs pointing to the downloadable content do not require an API key, and those are the URLs which the "search" should reveal.
> 
> So there would be "government.us.bill_search", and "government.us.bill_download". Traversing the content needs to be accomplished via the endpoints because you need to pass credentials to do anything.

CleanShot 2025-06-19 at 21 30 22@2x

openbb_platform/providers/congress_gov/openbb_congress_gov/models/bill_info.py
```diff
@@ -0,0 +1,386 @@
...
+        query: CongressBillInfoQueryParams,
+        data: dict,
+        **kwargs: Any,
+    ) -> CongressBillInfoData:
+        """Transform the data into the model."""
+        # pylint: disable=import-outside-toplevel
+        import re
+
+        # Regex to strip HTML tags
+        html_tag_regex = re.compile(r"<[^>]+>")
```
piiq, line 195: Do you think it makes sense to bring in a dependency like markitdown or add beautifulsoup (which is already required by openbb-sec openbb-fed) to convert html to markdown versus implementing vanilla html parsing with regex?

openbb_platform/providers/congress_gov/openbb_congress_gov/models/bill_text.py
```diff
@@ -0,0 +1,129 @@
...
+
+class CongressBillTextQueryParams(QueryParams):
+    """Congress Gov Bills Text Query."""
+
+    __json_schema_extra__ = {
+        "urls": {
+            "multiple_items_allowed": True,
+        }
+    }
+    # This field is typed this way to allow the Python interface and API
```
piiq, line 21: I don't fully understand this part and would ask @deeleeramone to check if this aligns with the expectations of the interface.

openbb_platform/providers/congress_gov/openbb_congress_gov/models/bill_text.py
```diff
@@ -0,0 +1,129 @@
...
+        description="Data format information, including data type and filename.",
+    )
+
+
+class CongressBillTextFetcher(
+    Fetcher[CongressBillTextQueryParams, list[CongressBillTextData]]
+):
+    """Congress Gov Bills Text Fetcher."""
+
+    require_credentials = False
```
piiq, line 56: We require credentials everywhere else - this one is fine?

piiq: answering my own question - yes

openbb_platform/providers/congress_gov/openbb_congress_gov/models/congress_bills.py
```diff
@@ -0,0 +1,338 @@
...
+                    query.end_date.strftime("%Y-%m-%d") if query.end_date else None
+                ),
+            )
+        try:
+            response = await amake_request(url=url)
+        except Exception as e:
+            # Handle exceptions
+            raise OpenBBError(e) from e
+
+        return response.get("bills", [])  # type: ignore
```
piiq, line 303: This function has no condition branch for `congress is None`.
If for some reason `congress is None` the return statement will raise an error. I don't fully see how `congress` would be `None`, but if the variable is set to None, we better handle it.

Overall the conditions in this block seem a bit difficult to understand - the Query has start_date and end_date, but the `year_to_congress` takes in only one year. It would be great if additional context/explanations for these conditions would be provided in the comments or the conditions to be refactored to include less `elif`s

openbb_platform/providers/congress_gov/openbb_congress_gov/utils/constants.py
```diff
@@ -0,0 +1,72 @@
...
+    {"label": "Senate Bill", "value": "s"},
+    {"label": "House Joint Resolution", "value": "hjres"},
+    {"label": "Senate Joint Resolution", "value": "sjres"},
+    {"label": "House Concurrent Resolution", "value": "hconres"},
+    {"label": "Senate Concurrent Resolution", "value": "sconres"},
+    {"label": "House Simple Resolution", "value": "hres"},
+    {"label": "Senate Simple Resolution", "value": "sres"},
+]
+
+bill_type_docstring = """Bill type (e.g., "hr" for House bills).
```
piiq, line 18: Are these not too long? We have had some problems with LLMs being overwhelmed with super long docstrings

openbb_platform/tests/test_extension_map.py
```diff
@@ -49,11 +49,12 @@ def test_extension_map():
             )
 
     for name, version in ext_map.items():
```
piiq, line 51: Are we sure this is required? If there is some naming mismatch would it be better to fix the mismatch versus hardcoding a patch?

piiq (commented):
I actually like this. Didn't go as deep as Darren using the data, I just poked here and there via the CLI.

I left a few comments related to the code itself. I have a bigger question - stucture wise.

1. We already have a menu called `regulators` and we already have a provider called `government_us`. The `government_us` provider in the docstring references the `data.gov` API, although it actually implements only a couple of endpoints that fetch data from the Treasury. Should this actually be a part of the government US provider?
2. The current code structure implements the provider and the router together as one python package. This is absolutely fine for standalone extensions and I do use the same pattern myself. But in the context of the OpenBB monorepo should this stay the same? Currently the router is implemented next to the provider and the tests for the routes are actually in the extensions folder which is in a different spot and looks like a piece of lost code.

Apart from that, I really like it. I think it's a good piece of functionality.

openbb_platform/providers/congress_gov/openbb_congress_gov/models/bill_info.py
deeleeramone: It's actually just one small piece of the response that contains a few HTML tags, so I didn't see the value in doing something heavy like BS4. As well, specifically, I need to separate the first line from the rest.

Fun fact, if I just leave those HTML tags in the markdown, Workspace crashes entirely. The app completely freezes up and has to be "hard reset"

deeleeramone:
> 1. We already have a menu called `regulators` and we already have a provider called `government_us`. The `government_us` provider in the docstring references the `data.gov` API, although it actually implements only a couple of endpoints that fetch data from the Treasury. Should this actually be a part of the government US provider?

It's not a bad idea, however we'd still have to be able to distinguish between the various API Keys and signups required for each service. Currently, the government US provider is intended to not use endpoints requiring signing up to anything. A future endpoint for this extension will be, Daily Treasury Statement. Or, spending and contract awards.

openbb_platform/providers/congress_gov/openbb_congress_gov/models/congress_bills.py
deeleeramone: Yes it does. If "congress" is not None, then it adjusts the URL to have the correct path parameters.

```
        url = (
            (
                f"{base_url}bill/{congress}/{query.bill_type}"
                if congress is not None
                else f"{base_url}bill"
            )
```

This handles the parameter combination of only receiving the "bill_type" parameter, and impacts the URL generation.
```
        elif (
            query.bill_type is not None
            and query.congress is None
            and query.end_date is not None
            and query.start_date is None
        
```

What appears to be missing is the explanation of start/end in the docstring explaining that start/end date must be within range of the Congress number. 

Start/end dates are only impacted by this condition when a congress number and bill type are supplied.

openbb_platform/tests/test_extension_map.py
deeleeramone: The test itself is not going to be valid in the future - we'll be reconsidering the default state, which IMO should have 0 installed extensions. This is a workaround until those considerations are realized.

openbb_platform/providers/congress_gov/openbb_congress_gov/utils/constants.py
deeleeramone: I don't know what too long is. "LLM mode" was conceived generations of LLMs ago to solve a problem that is not an issue with today's models.

These are the official definitions of the terms, so I think it's important for the user to be aware of their meaning, and to make that information readily available within the surrounding context.

openbb_platform/providers/congress_gov/openbb_congress_gov/models/bill_text.py
deeleeramone: The expectations of the interface are being met, you enter the URLs in the Python interface like you do for any normal parameter accepting multiple values. The extra conditional logic is to handle how Workspace sends the POST request parameters, specifically. It always sends this and the backend's only control is what to name "urls".

`{"urls": ["url1", "url2"]}`, where the entire dictionary ends up being passed into the function as: `query.urls = {"urls:  [ "url1", "url2"]}`. The content of the request body is what becomes `query.urls`.

deeleeramone:
> 2\. The current code structure implements the provider and the router together as one python package. This is absolutely fine for standalone extensions and I do use the same pattern myself. But in the context of the OpenBB monorepo should this stay the same? Currently the router is implemented next to the provider and the tests for the routes are actually in the extensions folder which is in a different spot and looks like a piece of lost code.

The test structure is meant to satisfy the requirements of the testing suite. The router path integration tests are always expected to be in the "extensions" folder and never in "providers".  This is simply a product of how the specific tests were conceived.

The reason for this being a standalone extension is that there is absolutely no other source of this data, it can only come from Congress.gov. A 3rd party vendor to serve this data would be entirely pointless. Other countries' governments would require their own unique handling, definitions, terminology, and parsing.

openbb_platform/providers/congress_gov/openbb_congress_gov/models/congress_bills.py
deeleeramone: > What appears to be missing is the explanation of start/end in the docstring explaining that start/end date must be within range of the Congress number.

> Start/end dates are only impacted by this condition when a congress number and bill type are supplied.

This part actually isn't of concern and is not a conditional element. I've added some comments to explain the section. Basically, a path parameter must be set given presence of either`congress` or `bill_type`

openbb_platform/tests/test_extension_map.py
deeleeramone: For a bit more clarity, in order to remedy this type of naming mismatch, we would need to account for it in the PackageBuilder process. The issue is that what the package calls the extension is not the same as the`os.sys.modules` label, so it is an encoding error in the Python Interface generator. There is only one module. There are two extensions, but only one module.

openbb_platform/providers/congress_gov/openbb_congress_gov/utils/constants.py
piiq: I'm talking about what was discussed in this thread
https://github.com/OpenBB-finance/OpenBB/pull/7160#discussion_r2191399552

deeleeramone: The rub might actually be that we are ignoring POST endpoints in MCP, so we'll need to adjust the "ignore" list to be more like `openbb-platform-api` to ignore just those router paths, if they are OpenBB extensions.

deeleeramone: > I'm talking about what was discussed in this thread [#7160 (comment)](https://github.com/OpenBB-finance/OpenBB/pull/7160#discussion_r2191399552)

For what it's worth:

Screenshot 2025-08-09 at 12 10 20 AM

DidierRLopes:
LGTM 😄  @deeleeramone

deeleeramone (approved):

PR #7151, opened 2025-06-23
[BugFix] Update `openbb-yfinance` tests to use new pytest curl adapter extension

deeleeramone:
This PR will address failing tests in, #7150

This updates the `pytest` configuration to use `record_curl`.

The impact is a net-positive, only one version of test cassettes are required as`openbb-yfinance` does not currently use `urllib3`.


piiq (approved):
That is a big diff

deeleeramone:
> That is a big diff

ya, i reduced the bloat a bit, but maybe 'net benefit' is only in the "number of files", the requests and cassettes have been 'de-bloated' as much as possible.

PR #7155, opened 2025-06-29
[Feature] Add Omni Widget Input and Response Models to `openbb-platform-api`.

deeleeramone:
This PR incorporates new functionality from Workspace into the widget builder of `openbb-api`.

- Where `"multiple_items_allowed": True` and there are no defined choices or literals, and the input is a string, "multiple" will be added to the `params` entry in `widgets.json`
- All tables will now default to `enableAdvanced`, to allow row groups and pivot tables.
- Adds a pattern to support Omni widget creation, with automatic `data_type` assignment.
  - `OmniWidgetInput` and `OmniWidgetResponseModel` work together to build all the properties for the widget.

This creates a complete entry in `widgets.json` - defining `parse_as` is not necessary, this is for show, just send "content".

```python
from typing import Literal, Optional
from fastapi import FastAPI
from openbb_platform_api.query_models import OmniWidgetInput
from openbb_platform_api.response_models import OmniWidgetResponseModel
from pydantic import Field

app = FastAPI()

class TestOmniWidgetQueryModel(OmniWidgetInput):
    """Test query model for OmniWidget."""
    param1: str = Field(description="A string parameter for testing")
    param2: int = Field(description="An integer parameter for testing")
    param3: bool = Field(default=False, description="A boolean parameter for testing")
    start_date: str = Field(description="The start date for testing")
    end_date: str = Field(description="The end date for testing")
    parse_as: Optional[Literal["table", "chart", "text"]] = Field(
        default=None,
        description="The format to parse the response as, either 'table', 'chart', or 'text'."
        + " If not defined, the model will try to infer the type based on the content.",
    )

@app.post("/omni_widget", response_model=OmniWidgetResponseModel)
async def create_omni_widget(item: TestOmniWidgetQueryModel):
    """This is a test endpoint for generating an OmniWidget in OpenBB Workspace."""
    # Here you would process the incoming request and return a response
    some_test_data = [
        {"prompt": item.prompt,
        "param1": item.param1,
        "param2": item.param2,
        "param3": item.param3,
        "start_date": item.start_date,
        "end_date": item.end_date,
    }]

    if item.parse_as == "chart":
        some_test_data = {
            "data": [{"type": "bar", "x": ["A", "B", "C"], "y": [1, 2, 3]}],
            "layout": {"template": "...", "title": {"text": "Hello Chart!"}}
        }
    elif item.parse_as == "text":
        some_test_data = f"""
### This is a test OmniWidget response

- Prompt: {item.prompt}
- Param1: {item.param1}
- Param2: {item.param2}
- Param3: {item.param3}
- Start Date: {item.start_date}
- End Date: {item.end_date}
"""
    return {"content": some_test_data}
```

Launch with:

```sh
openbb-api --app ./this_file.py
```

https://github.com/user-attachments/assets/52433549-83c0-4365-8aad-22d8532864da


openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
```diff
@@ -13,3 +15,67 @@ class FormData(Data):
         alias_generator=AliasGenerator(to_snake),
         title="Submit Form",
     )
+
+
+class OmniWidgetInput(Data):
```
piiq, line 20: Is this model always exposing the prompt input?

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
```diff
@@ -13,3 +15,67 @@ class FormData(Data):
...
+        # pylint: disable=import-outside-toplevel
+        import json
+        import re
+
+        if not v or v == "":
+            return None
+
+        prompt = ""
+
+        if (
```
piiq: Can you explain the use case behind this block please?

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
```diff
@@ -147,3 +147,149 @@ def validate_model(cls, values) -> "PdfResponseModel":
...
+        cls, values
+    ) -> "OmniWidgetResponseModel":
+        """Validate the Omni widget content."""
+        import contextlib  # noqa
+        import json
+        import pandas as pd
+
+        content = getattr(values, "content", None)
+
+        if content is None:
```
piiq, line 213: When we are returning an empty string it casts to none, no?

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
```diff
@@ -147,3 +147,149 @@ def validate_model(cls, values) -> "PdfResponseModel":
...
+        parse_as = getattr(values, "parse_as", None)
+
+        if parse_as and parse_as not in ("table", "chart", "text"):
+            raise ValueError(
+                "Invalid parse_as value. Must be one of 'table', 'chart', or 'text'."
+            )
+
+        # If parameter was supplied, assume the data is formatted correctly.
+        if content and parse_as:
+            data_format = {
```
piiq, line 225: I couldn't follow this part too. 

Can you explain the logic of the validator in plain English?

openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
```diff
@@ -241,7 +254,11 @@ def build_json(  # noqa: PLR0912  # pylint: disable=too-many-branches, too-many-
     if not openapi:
         return {}
 
-    starred_list: list = []
+    starred_list: list = [
+        "/api/v1/technical/*",
```
piiq: Is there a way to avoid hardcoding these?

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
deeleeramone: This allows one to send JSON data in the text input field - for example, you can "copy" the contents of a table widget and paste into the box and send that to the backend.

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
deeleeramone: This is for validating what the user is submitting as the results content, and is a check to ensure that it is not initialized with empty content. The type for "content" is Any, so this is an additional test.

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
deeleeramone: This allows the user to override the format discovery and assignment. If you supply a "parse_as" key, then it will assume you explicitly want that setting and will pass directly as-is.

deeleeramone: The validator reads the submitted "content" and attempts to assign the correct "parse_as" setting required by Workspace. Because the object required by Workspace is a nested dictionary, the model accepts the flat input and properly constructs the expected dictionary, and removes the "parse_as" key from where it is in the model.


Input:

```
{"content": "some content"}
```

Output:
```
{"content": "some content", "data_format": {"data_type": "object", "parse_as": "text"}
```

Input:

```
{"content": [{"column": "value"}]}
```

Output:
```
{"content": "some content", "data_format": {"data_type": "object", "parse_as": "table"}
```

openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
deeleeramone: Adding "exclude" entries for every single function is one way, this is just more efficient and there is no way of supporting these endpoints in Workspace currently.  When there is a way to use the output of a widget as the input to a POST endpoint, then we can revisit a way to handle these paths.

Maybe you mean something else though?

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
deeleeramone: Yes, that is its primary purpose. Workspace requires this to be defined in order for it to actually recognize the widget type, but there are no configurations available to the user for customizing the input and it is always a "hidden" parameter in widgets.json. This handles all of that, as well as query parameter assignment.

If you want the widget to have parameters, then inherit from the model and add them as Fields.

piiq: Then I suggest making it a PromptInput vs OmniWidgetInput

I can clearly see that in the somewhat near future the omni widget will be capable to receive non-prompt inputs and that a prompt input will be supported by table widgets etc

openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
piiq: 1. We allow the user to specify the api path in system_settings overriding api/v1 

2. I really want this package to be usable as a generic launcher for workspace backends so this pushes me to question all hardwiring to the existing extensions

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
piiq: This validation works only if the "prompt" is pure json. On one hand a json (as structured input) technically is a prompt. On the other hand we're kinda mixing up structured (json) and unstructured (natural language) input making the input behavior implicit. 

On the widget we do omni output explicitly via parse_as, if you want to support omni input, I suggest we also do it explicitly. 

Ideally (and I am sure we'll get there) rendering for non-natural language prompts should differ. A sql command or a JSON is not the same as a question written in English

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
piiq: My point is that if I explicitly want to return an empty string I will fail because it will cast to None and raise an error

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
piiq: We keep this data_format dict with some logic for deleting keys when we reuse the variables that already exist in the validator function. Is it necessary?

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
```diff
@@ -147,3 +147,149 @@ def validate_model(cls, values) -> "PdfResponseModel":
...
+            }
+            values.data_format = data_format
+            del values.parse_as
+
+            return values
+
+        with contextlib.suppress(ImportError):
+            from plotly.graph_objs import Figure  # noqa
+
+        if Figure is not None and isinstance(content, Figure):
```
piiq: IIUC if plotly is not installed we'll get a NameError on this line unless we define Figure as None elsewhere

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
deeleeramone: > This validation works only if the "prompt" is pure json.

Yes, and then if it isn't it just passes like any other regular string. All content being passed in from Workspace will be serialized before sending.

> On the widget we do omni output explicitly via parse_as, if you want to support omni input, I suggest we also do it explicitly.

The idea is that the developer can do whatever they want, all that's happening is a string is being sent to the body under the key, "prompt" and this ends up in the POST function params under the attribute.

> Ideally (and I am sure we'll get there) rendering for non-natural language prompts should differ. A sql command or a JSON is not the same as a question written in English

What do you mean by rendering prompts? This doesn't render the prompt, it's accepting an input, and this input is not exclusively to serve an LLM - it is multi-dimensional - and you need to handle passing any "prompt" to your actual LLM function.

This model doesn't interact with LLMs, it just processes Input and Output and creates the widget. All the business logic of what to actually do with the Inputs and handling responses needs to be handled in the endpoint.

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
deeleeramone: > My point is that if I explicitly want to return an empty string I will fail because it will cast to None and raise an error

It's recommended to return an error instead of empty data for other widget types, so this just follows the same pattern.

deeleeramone: > My point is that if I explicitly want to return an empty string I will fail because it will cast to None and raise an error

An empty string is not actually None. Only None can be None.

```
In [1]: content = ""

In [2]: content is None
Out[2]: False
```

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
deeleeramone: My understanding is that contextlib will add the import to the namespace as None if it fails to import, but maybe I'm mistaken.

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
deeleeramone: The shape of the input is not the same as the shape of the output, this guarantees that the output is correctly formatted for what Workspace expects.

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
deeleeramone: I can do this without an import and attempt to infer by reading the object's properties instead.

openbb_platform/extensions/platform_api/openbb_platform_api/utils/widgets.py
deeleeramone: I see what you're saying, too broad of a stroke. We handle the platform extensions in a way to not interfere. I'll think on a different way to go about applying the removal from the widgets without removing from the actual API. What I really don't want to happen is to have a bunch of useless widgets created.

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
deeleeramone: OmniWidgetInput is much more specific, and exactly what it is. It's a class for handling the Inputs of an Omni Widget.

It's not a PromptInput, it's the model for inheriting from to create your POST query model.

This is explicitly for the Omni Widget, if there is something that exists in the future, then that is when it can be implemented. Likely, it will need different handling entirely.

Specifically, the omni widget does not get recognized by Workspace if "prompt" is not in the params array.

piiq: Resolving these conversations following the call we had.

openbb_platform/extensions/platform_api/openbb_platform_api/query_models.py
piiq: Resolving these conversations following the call we had.

openbb_platform/extensions/platform_api/openbb_platform_api/response_models.py
piiq: Resolving these conversations following the call we had.

piiq (approved):
@deeleeramone thanks for the call. I'll approve this PR. Feel free to adjust based on what we have talked about and merge when ready

PR #7160, opened 2025-07-03
[BugFix] MCP - expose correct header for session id

MagnusS0:
# Fix MCP Session ID

## Description

In the  MCP PR #7094 I set the `stateless_http=True`, after some testing I discovered that this caused context length issues with Cursor and the Claud Desktop Client. 
This fix reverts to `stateless_http=False` and fix the previous Session ID bug by exposing the `Mcp-Session-Id` header.

@piiq @deeleeramone


deeleeramone:
@MagnusS0, can you please add some instructions-for-dummies for connecting and configuring for immediate use in both Claude Desktop and Cursor? We need explicit steps that anyone can copy and paste, thanks.

MagnusS0:
In terms of how to set it up I added some instructions.

But also notice that if you run it with, openbb[all] all the technical and quantitative endpoints also become tools. I don't think this is something we want, since moste of those endpoints expect inputs as the output of equity_price_historical. This should probably be programmatically passed instead of  having the LLM as a layer between. 

Potentially converting these tools to automatically fetch the right data would make more sense from an agent perspective. 

But maybe just ignore these as standard setting for now is the best approach?

deeleeramone:
> Potentially converting these tools to automatically fetch the right data would make more sense from an agent perspective.

I don't think this will work out reliably, there are too many factors to assume the "right" data is being collected and processed. The functions are intended to be used with any time series.
 
> But maybe just ignore these as standard setting for now is the best approach?

This would be fine, IMO, to ignore all POST requests. If there is a clear path towards enabling these via a code interpreter agent, or similar, then we can look at incorporating some. Alternatively, one could manually configure tools that utilize these endpoints via specific instructions, but that's a rabbit hole and out-of-scope for now.

MagnusS0:
> This would be fine, IMO, to ignore all POST requests.

On board with that! Let me add POST to be excluded by default. 


> Alternatively, one could manually configure tools that utilize these endpoints via specific instructions, but that's a rabbit hole and out-of-scope for now.

I love a good rabbit hole, but maybe someone comes up with a clever way to use them later on 😄

DidierRLopes:
CleanShot 2025-07-07 at 22 12 04@2x

Having issues when running this one :/

openbb_platform/extensions/mcp_server/pyproject.toml
```diff
@@ -16,7 +16,7 @@ openbb-mcp = "openbb_mcp_server.main:main"
 [tool.poetry.dependencies]
 python = ">=3.10,<3.13"
 openbb-core = "^1.4.3"
-fastmcp = "^2.8.0"
+fastmcp = ">=2.8.0,<2.9.0"
```
deeleeramone: We should avoid pinning the dependency and always favor the latest version. The current version is 2.10.2, so this is already obsolete.

Also, no need to update the version number of the package, that will be handled when the extension is published.

deeleeramone (changes requested):

openbb_platform/extensions/mcp_server/pyproject.toml
MagnusS0: This is due to tool descriptions/responses bloated after 2.9.

I have traced it back to this PR
jlowin/fastmcp#897
We end up with out of context length just reading in a single tool.
Others reported similar issues.
It looks like a fix might have been proposed here jlowin/fastmcp#1092

So temporary fix was to fallback to 2.8.1

deeleeramone: > This is due to tool descriptions/responses bloated after 2.9.
> 

The PR you are referencing mentions, and is before 2.9:

> ...this resulted in a test failure and also bloated tool definitions, so I added a fix for that.

This PR has been merged, and was included in 2.9, which suggests the bloated descriptions issue you are referencing is an issue pre-2.9, along with `$ref` handling, which our `openapi.json` has in every endpoint.

We need to be working with the latest versions for a number of reasons, but by not using the latest version, we create problems for the immediate future.

https://github.com/jlowin/fastmcp/pull/887

MagnusS0: Totally agree on using the newest version, but currently this will make the MCP server unusable.
I think this needs to be fixed in the FastMCP library, if not, we would have to look into how the openapi.json is structured and if we can adjust the customize components function in some way.

> you are referencing is an issue pre-2.9

What I meant was that that PR was supposed to fix bloating but for conversion of the OpenBB API specifically it caused bloating, and broke the MCP server.

Eg. here is the news_world tool with v2.8.1

```
[debug] [Transport] Received: {"jsonrpc":"2.0","id":1,"result":{"tools":[{"name":"news_world","description":"World News. Global news data.","inputSchema":{"type":"object","properties":{"provider":{"type":"string","enum":["benzinga","biztoc","fmp","intrinio","tiingo"],"title":"Provider"},"limit":{"type":"integer","title":"Limit","description":"The number of data entries to return. The number of articles to return.","default":2500},"start_date":{"anyOf":[{"type":"string","format":"date"},{"type":"null"}],"title":"Start Date","description":"Start date of the data, in YYYY-MM-DD format."},"end_date":{"anyOf":[{"type":"string","format":"date"},{"type":"null"}],"title":"End Date","description":"End date of the data, in YYYY-MM-DD format."},"date":{"anyOf":[{"type":"string","format":"date"},{"type":"null"}],"title":"benzinga","description":"A specific date to get data for. (provider: benzinga)"},"display":{"type":"string","enum":["headline","abstract","full"],"title":"benzinga","description":"Specify headline only (headline), headline + teaser (abstract), or headline + full body (full). (provider: benzinga)","default":"full"},"updated_since":{"anyOf":[{"type":"integer"},{"type":"null"}],"title":"benzinga","description":"Number of seconds since the news was updated. (provider: benzinga)"},"published_since":{"anyOf":[{"type":"integer"},{"type":"null"}],"title":"benzinga","description":"Number of seconds since the news was published. (provider: benzinga)"},"sort":{"type":"string","enum":["id","created","updated"],"title":"benzinga","description":"Key to sort the news by. (provider: benzinga)","default":"created"},"order":{"type":"string","enum":["asc","desc"],"title":"benzinga","description":"Order to sort the news by. (provider: benzinga)","default":"desc"},"isin":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"benzinga","description":"The ISIN of the news to retrieve. (provider: benzinga)"},"cusip":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"benzinga","description":"The CUSIP of the news to retrieve. (provider: benzinga)"},"channels":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"benzinga","description":"Channels of the news to retrieve. (provider: benzinga)"},"topics":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"benzinga","description":"Topics of the news to retrieve. (provider: benzinga)"},"authors":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"benzinga","description":"Authors of the news to retrieve. (provider: benzinga)"},"content_types":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"benzinga","description":"Content types of the news to retrieve. (provider: benzinga)"},"term":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"biztoc","description":"Search term to filter articles by. This overrides all other filters. (provider: biztoc)"},"source":{"anyOf":[{"type":"string"},{"type":"string","enum":["yahoo","moody","moody_us_news","moody_us_press_releases"]},{"type":"null"}],"title":"biztoc,intrinio,tiingo","description":"Filter by a specific publisher. Only valid when filter is set to source. (provider: biztoc);\n The source of the news article. (provider: intrinio);\n A comma-separated list of the domains requested. Multiple comma separated items allowed. (provider: tiingo)","tiingo":{"multiple_items_allowed":true}},"sentiment":{"anyOf":[{"type":"string","enum":["positive","neutral","negative"]},{"type":"null"}],"title":"intrinio","description":"Return news only from this source. (provider: intrinio)"},"language":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"intrinio","description":"Filter by language. Unsupported for yahoo source. (provider: intrinio)"},"topic":{"anyOf":[{"type":"string"},{"type":"null"}],"title":"intrinio","description":"Filter by topic. Unsupported for yahoo source. (provider: intrinio)"},"word_count_greater_than":{"anyOf":[{"type":"integer"},{"type":"null"}],"title":"intrinio","description":"News stories will have a word count greater than this value. Unsupported for yahoo source. (provider: intrinio)"},"word_count_less_than":{"anyOf":[{"type":"integer"},{"type":"null"}],"title":"intrinio","description":"News stories will have a word count less than this value. Unsupported for yahoo source. (provider: intrinio)"},"is_spam":{"anyOf":[{"type":"boolean"},{"type":"null"}],"title":"intrinio","description":"Filter whether it is marked as spam or not. Unsupported for yahoo source. (provider: intrinio)"},"business_relevance_greater_than":{"anyOf":[{"type":"number"},{"type":"null"}],"title":"intrinio","description":"News stories will have a business relevance score more than this value. Unsupported for yahoo source. Value is a decimal between 0 and 1. (provider: intrinio)"},"business_relevance_less_than":{"anyOf":[{"type":"number"},{"type":"null"}],"title":"intrinio","description":"News stories will have a business relevance score less than this value. Unsupported for yahoo source. Value is a decimal between 0 and 1. (provider: intrinio)"},"offset":{"anyOf":[{"type":"integer"},{"type":"null"}],"title":"tiingo","description":"Page offset, used in conjunction with limit. (provider: tiingo)","default":0}},"required":["provider"]}}
```

With v2.9 and 2.10 this becomes a close to 1 MB txt file [news_tool.txt](https://github.com/user-attachments/files/21131705/news_tool.txt). The pruning seems to fail so where the example above stops it continues:

```json
"$defs":{"ADFTestModel":{
``` 

I'll look into the `openapi.json` to see if I can get a better understanding.

DidierRLopes: What @MagnusS0 is saying makes sense. We should pin dependency to 2.8 until it gets fixed upstream, and then bump it to the first version where this works. Particularly if this is going to be addressed, no reason to work on a workaround if it gets fixed at the root.

Also, I was able to run this with uvx well :) 

CleanShot 2025-07-09 at 00 28 46@2x

piiq: If it's about bloated docstrings or something maybewe can enable something like this one the api level? https://github.com/OpenBB-finance/openbb-docs/blob/main/old_platform_stash/user_guides/llm_mode.mdx

 @deeleeramone do you know if this works on the api?

@MagnusS0 if there is a bug in fastmcp, is there already an issue in their repo that we can track?

deeleeramone: @piiq, LLM Mode does not affect the API docstrings and descriptions.

deeleeramone: They have already put out 2 patch releases since this conversation began, so pinning to a lower version - especially given that there are "lots of breaking changes" in versions ahead - is not really going to be helpful. It just means will miss the fix and then be stuck needing to handle the introduced changes.

MagnusS0: Cloned fastmcp to figure out where it goes wrong so I might be able to create an Issue. 
But the main issue is that all schema's are added to every tool. 

https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/utilities/json_schema.py

@deeleeramone you probably know the API better than me, are you able to see where the pruning might go wrong?

MagnusS0: This has been resolved with the latest commits and we can use the newest version of fastmcp.
Currently there is a output validation error, but since my fix was merged jlowin/fastmcp#1119 from the next release this should be good. 

We can wait to merge this until then or set outputs to`None` temporarily.

jlowin: Hi all -- just released 2.10.5 including @MagnusS0 fix in https://github.com/jlowin/fastmcp/pull/1119. Sorry for the trouble! Curious how we can do more to avoid the docstrings getting too long -- the line between "automatically regurgitating all the OpenAPI data" and "not overwhelming the LLM" is a very fine one!

piiq: Hey @jlowin 👋 thanks for the quick release and for jumping into the discussion!

One suggestion: the OpenBB Platform MCP that @MagnusS0 is working on could serve as a good test case for experimenting with ways to trim down the OpenAPI docstrings. It’s a bit of an edge case: the API exposes 300+ endpoints by default, and each one is very thoroughly documented. Every parameter and return type is typed and described in detail using Pydantic models. So yeah, lots of context.

A couple of things I noticed in our `openapi.json` that might be relevant here:
1. Some define the same info in both the parameter specs and the schema.
image
2. Many return schemas link to large Pydantic models, and we use anyOf to describe unions of similar data models (we standardize across data providers). This can add a lot of weight to the spec, especially when models are overlapping a lot.
image
image

I don’t have a full picture of how MCP uses these specs under the hood (whether the LLM relies on output schemas or not). But if not, maybe skipping the return models could reduce some of the overload? Not proposing that outright, just hoping to spark some discussion. Curious what you and @MagnusS0 think.

MagnusS0: Hey @jlowin, thanks for all your awesome work with FastMCP!
Totally agree, finding the sweet spot for docstrings is tricky. Really like the `mcp_component_fn` it’s been super helpful for trimming the docs and not just copy-paste from OpenAPI. E.g. we add some simple regex tweaks to remove response schemas from tool descriptions.

MagnusS0: Hey @jlowin, thanks for all your awesome work with FastMCP!

Totally agree, finding the sweet spot for docstrings is tricky. Really like the `mcp_component_fn` adds a lot of flexibility to customize things and not just copy-paste from OpenAPI.

And to @piiq's point, this is spot on, the return model is not information the LLM needs. I actually already handle this with a simple regex that removes them from tool descriptions.

Btw part of the reason this conversion straight from OpenAPI works is because of how great things are documented, much better than missing context!

deeleeramone (commented):
Are we good to update the definition to the latest fastmcp version? The pyproject.toml is keeping us at 2.8

MagnusS0:
Should be ready to merge now 😃

deeleeramone (approved):
Thank you very much for your contribution, as well as the cross-library collaboration! Open-source at its finest!

PR #7162, opened 2025-07-05
[BugFix] Cleanup Test Imports

deeleeramone:
This PR updates the test files:

- Replaces those using an import of `parametrize`, from `conftest.py`, with `@pytest.mark.parametrize`.
- Replaces imports such as, `from providers.yfinance.etc.`, with package, or relative, import.
- Fixes params in `test_economy_fred_regional` that had parameter names assigned, instead of being positional-only.

The motivation for this PR is for implementing support for higher versions of Python. I have also checked for compatibility with Pytest 8.4.


piiq (approved):
Yeah, this syntax is a lot better. Thanks

PR #7166, opened 2025-07-12
[Feature] Add PLC0415 To Ruff Ignore

deeleeramone:
This PR add "PLC0415" to the ignore section in `ruff.toml` - AKA, "import-outside-toplevel". This appears to be a new rule.  In the future we may want to consider consolidating the lint actions into a single source, or at least subbing Ruff in for Pylint.

The purpose for this exclusion is:

- We don't want `ruff --fix` to make changes to the import structures, or for CI actions relying on Ruff to view this as an error.
- Pylint is watching for this pattern, and we ignore at the function level where we are explicitly lazy-loading packages.
- Adding an additional line for ruff in every occurrence is not desirable.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -28,7 +28,7 @@ select = [
     "T20",
 ]
 # These ignores should be seen as temporary solutions to problems that will NEED fixed
-ignore = ["PLR2004", "PLR0913", "PLR0915", "UP035", "UP006"]
+ignore = ["PLR2004", "PLR0913", "PLR0915", "UP035", "UP006", "PLC0415"]
 
 [lint.per-file-ignores]
 "**/tests/*" = ["S101"]


piiq (approved):
I'm OK adding the rule to the list of ignored rules, but moving forward I would suggest going the other way round compared to how I read the proposal in the PR:

1. Moving away from pylint to ruff. Ruff is a lot faster and can handle most if not all the rules we're following to be considered a hot-swap replacement for pylint.
2. Change `ruff check --fix` to `ruff check` to prevent ruff making changes to the code in pre-commit hooks or the CI if we don't want it to. If we configure ruff correctly I don't see a reasons for it to not be allowed to make changes. We allow black to format the code in pre-commit hooks, why not ruff.

deeleeramone:
> I'm OK adding the rule to the list of ignored rules, but moving forward I would suggest going the other way round compared to how I read the proposal in the PR

This is what I meant by "in the future..."

PR #7193, opened 2025-07-29
[Feature] Support Python 3.13

deeleeramone:
This PR adds support across all packages and dependencies for Python 3.13.

Additionally, it moves the `openbb` folder to inside the `core` package and updates the references to these folders. This ensures that the build script ships when only `openbb-core` is installed.

I have not yet encountered anything requiring syntax modifications, and bumped package versions are still compatible with 3.9.

Testing done:

- Complete test suite.
- Manual inspection of an empty installation build, and then adding custom extensions + creating/watching for errors in the `openbb-build` process.


piiq (commented):
Hey, the 3.13 part looks good. I would like to ask for the rationale to move the built python interfaces from openbb to openbb_core. This technically doesn't de-bloat the package, but instead actually introduces more "dead code" to the core for those who operate minimalist deployments or custom extensions.

deeleeramone:
> Hey, the 3.13 part looks good. I would like to ask for the rationale to move the built python interfaces from openbb to openbb_core. This technically doesn't de-bloat the package, but instead actually introduces more "dead code" to the core for those who operate minimalist deployments or custom extensions.

The rationale is that the build script not being part of `openbb-core` has resulted in the need to employ other workarounds.

I'm not sure if "dead code" is the correct analogy, if one was just utilizing the Python interface, all of the API interface code is essentially "dead code". If it ships with no static assets, then those files will only get populated if, and when, the Python interface is imported or built.

One feature of `openbb-core` is supporting both Python and API interfaces out-of-the-box. Therefore, the build script and basic import logic should be, in my opinion, be part of the core package.

If you disagree, or want to reserve judgement for later, we can scope it out. Just feels like a decent time to make this kind of move.

piiq:
Following up on the conversation we had. Would you consider shipping `openbb-core` without static assets at all?

deeleeramone:
> Following up on the conversation we had. Would you consider shipping `openbb-core` without static assets at all?

Yes, however in order to do so, we need to change a bunch of other stuff and related tests in the repo. I would like to have this merged and usable before doing the next steps. This PR by itself is not a breaking change, next steps start breaking things though.

piiq (approved):

PR #7202, opened 2025-08-16
[Feature] Expand MCP Server Configurability

deeleeramone:
This PR adds new functionality and configurations to the `openbb_mcp_server` extension.

The code has been reorganized with a similar structure to the other interfaces, and a test folder to match.

For those simply using the launcher and configuration file, there should be no changes notable other than a higher version of FastMCP. Everything is expected to work the same as it already did.

Details have been added to the README file, with examples.

## Breaking Changes

Breaking changes will exist for those using imported library components. @MagnusS0, for example, may be impacted. Internal imports from within the extension need to be updated to their new paths and filenames. The individual items being imported don't need to be updated, just the location of the file.

- `main` -> `openbb_mcp_server.app.app`
  - The `main` function is still the entry point in the file.
- `registry`, `settings`, `tool_models` are now in the `models` folder, `tool_models` -> `tools`
- `route_filtering` -> `fastapi`
- The settings loader has been refactored to `MCPService`, which is similar to `SystemService` and `UserService` from the `openbb_core` library.
  - Accepted settings have been expanded to include all runtime arguments in `mcp_server.run`, including `uvicorn` configurations.

Settings priority order (highest to lowest):

1. Command line arguments (cli_overrides)
2. Environment variables
3. Configuration file
4. Default values
 
### Termination Signal Handling

This adds termination handling for SSE and STDIO transports. You can now `ctrl-c` to terminate the server when `--transport stdio`.

### Runtime Arguments

Argparse has been removed and the command line entry point now allows undefined keyword arguments to be passed through to uvicorn.

See the README file for details and examples.

### Arbitrary FastAPI App Import

The app import behavior from `openbb-api --app ./some_file.py` has been copied over and applied to `openbb-mcp`.

Use any FastAPI instance, instead of being bound to the OpenBB app, by setting the `--app` keyword argument.

### Inline Configuration

The tool can be configured in the FastAPI endpoint code, accessible via the `openapi_extra` dictionary as, `mcp_config`.

This works like the `widget_config` dictionary, and takes priority over any automatically generated configuration of the tool. The dictionary is validated and will be ignored if it fails validation. Validation errors will be logged to the console, and the tool will be added as if it wasn't supplied. 

See the README file for definitions and examples.

### Prompts

Prompts can be added via the `mcp_config` dictionary, and can be executed by the LLM. This can help it use the tool with less reasoning effort. Every inline prompt definition is prefixed with the instruction to use the tool name explicitly.

```python
        content = (
            f"Use the tool, {tool_uri}, to perform the following task.\n\n"
            + prompt_cfg.get("content", "")
        )
```

Multiple prompts can be added to any tool, each one will be referenced in the tool's description so the LLM knows they are available to use.

### System Prompt

A system prompt file (.txt) can be supplied on launch, or in the `mcp_settings.json` configuration.

The file will be loaded and made available as `resouce://system_prompt`, or can be executed with the name, `system_prompt`

### Server Prompts

A server prompts file (.json) can be supplied on launch, or in the `mcp_settings.json` configuration. These are for directing multi-tool use, and constructing a workflow that can be executed by the LLM with simple arguments.

Each prompt will be added and tagged as 'server'.

image

### New Tools

Two new tools have been added, at the app level, under the tag: `prompt`.

- list_prompts
- execute_prompt

### Authorization

Easily add authorization to the server by configuring in the `mcp_settings.json` file or setting environment variables.

There are two settings, one for the server, another that passes through the configuration to the FastAPI app. 

#### Server-Side Authentication

Server-side authentication requires incoming requests to provide credentials. This is configured using the `server_auth` setting, which accepts a tuple of `(username, password)`.

When `server_auth` is enabled, clients must include an `Authorization` header with a `Bearer` token. The token should be a Base64-encoded string of `username:password`.

**Example: Environment Variable**

```env
OPENBB_MCP_SERVER_AUTH='["myuser", "mypass"]'
```

**Example: `mcp_settings.json`**

```json
{
  "server_auth": ["myuser", "mypass"]
}
```

#### Client-Side Authentication

Client-side authentication configures the MCP server to use credentials when making downstream requests. This is useful when the server needs to authenticate with other services.

**Example: Environment Variable**

```env
OPENBB_MCP_CLIENT_AUTH='["client_user", "client_pass"]'
```


deeleeramone:
@MagnusS0, please try it out!

MagnusS0:
Looked through this quickly yesterday, great refactor @deeleeramone !! I'll test it out more today, the new prompt feature is very interesting 👀

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
```diff
@@ -0,0 +1,605 @@
...
+
+            def system_prompt_func() -> str:
+                """System prompt for the OpenBB MCP server."""
+                return system_prompt_content
+
+            mcp.add_prompt(
+                FunctionPrompt.from_function(
+                    system_prompt_func,
+                    name="system_prompt",
+                    description="System prompt for the OpenBB MCP server.",
```
MagnusS0: I would maybe add a longer description for this, since we can assume if this is set they want the agent to use it.

For example:
> This is the system prompt for the OpenBB MCP Server. If you are an agent connected to this server, please read this carefully to understand how to interact with and utilize the MCP features. This prompt provides essential guidance and usage instructions for effective use of the tools provided by OpenBB.

MagnusS0 (commented):
Been testing it out now and works great! Tried out the system prompt and adding additional prompts to a tool, this is super cool 🔥
I think one thing that could be added to the prompting is the ability to add custom standalone prompts other than system prompt. 

So it could follow the `mcp_config` and allow a new config file (json) that can be supplied like the `--system-prompt` , this could add prompts for multi-tool tasks. 

e.g

```json
{
  "prompts": [
    {
      "name": "equity_analysis",
      "description": "Perform a comprehensive equity analysis using multiple data sources and metrics",
      "content": ["Conduct a comprehensive analysis of {symbol} for {analysis_period}. Follow this workflow:",
      "",
      "1. First, get basic stock quote and recent price performance using equity_price_performance",
      "2. Retrieve fundamental data including financial statements, ratios, and key metrics using [equity_fundamental_ratios, equity_fundamental_metrics, equity_fundamental_balance]". 
      "3. Gather recent news and analyst estimates for the company using [news_company, equity_estiments_price_target] ",
      "4. Compare valuation metrics with industry peers using equity_compare_peers",
      "5. Summarize findings with investment recommendation",
      "",
      "Focus areas: {focus_areas}\nRisk tolerance: {risk_tolerance}"
]
      "arguments": [
        {
          "name": "symbol",
          "type": "str",
          "description": "Stock ticker symbol to analyze (e.g., AAPL, TSLA)"
        },
        {
          "name": "analysis_period",
          "type": "str",
          "default": "last 12 months",
          "description": "Time period for the analysis"
        },
        {
          "name": "focus_areas",
          "type": "str",
          "default": "growth, profitability, valuation",
          "description": "Specific areas to focus on in the analysis"
        },
        {
          "name": "risk_tolerance",
          "type": "str",
          "default": "moderate",
          "description": "Risk tolerance level: conservative, moderate, or aggressive"
        }
      ],
      "tags": ["equity", "analysis", "comprehensive"]
    },
```

deeleeramone:
> I think one thing that could be added to the prompting is the ability to add custom standalone prompts other than system prompt. 
> So it could follow the `mcp_config` and allow a new config file (json) that can be supplied like the `--system-prompt` , this could add prompts for multi-tool tasks.

Yup, agreed.

deeleeramone:
@MagnusS0, `--server-prompts` runtime argument added. Note the implementation is a JSON string for `content` instead of a list of string lines. See the README for definitions, but basically the same as inline prompt definitions.

piiq (approved):
I'm approving this PR since the refactor works as expected when I run it.

That said, I'd like to leave some feedback on several design decisions that I find worth revisiting:

- Dropping `argparse` removed support for short CLI flags. For example, `openbb-mcp -h` doesn't work, only `--help` is supported. To me this breaks a common CLI convention and hurts usability.
  - Additionally the CLI parsing loop that comes instead of `argparse` has very high cognitive complexity. It's a wall of string parsing logic that looks fragile and hard to maintain. Was there a specific reason that made using `argparse` undesirable?
- It's unclear why the extension "vocally" ships with support for the deprecated MCP protocol (SSE). There is consensus that it's deprecated, yet the extension documents it and even advertises for clients like Cline. IIUC Cline actually supports `streamable-http`, one would only need to specify the correct protocol as the `"type": "streamable-http"` in the JSON. That said I don't think we should be promoting deprecated protocols or even exposing it as an option in the code.
- The README is extremely long. While it does contain important information, it buries client configuration instructions deep inside. For most users (even myself), this is overwhelming. I recommend adding actual usage instructions in the top of the  “Installation & Usage” section (like a quick-start guide), which can be followed by those hundreds of lines of specifications and in-depth configuration options.
- While reading through the readme the feeling that the scope is unfocused was haunting me. On one hand, there are very OpenBB-specific features, like launching OpenBB extensions and applying category filters. On the other hand we are going down the path where we support passing any FastAPI instance and it would make an MCP based off that. This duality, while makes sense in `openbb-api` (because there aren't any tools out there that create the `widgets.json`), but in this project it doesn't? FastMCP can already do that, why is another layer around FastMCP needed? Is it also OpenBB specific? Like me creating the endpoint that can serve both a widget and an MCP tool, and I get the control over both? But in this case why doesn't `openbb-api` handle it?

Overall, it works. Let's merge this as the current iteration. This is a good starting point for us to work on improving it further from here.

Note: I have only tested the default use case. I've not tested launching my own extensions with it, no custom configurations or apps. I'm pretty sure those will expose extravagant edge cases, but it's fine for now.

deeleeramone:
Thank you for the feedback. I can address some of the open-ended questions here

> The README is extremely long. While it does contain important information, it buries client configuration instructions deep inside. 

Agreed, this will be optimized for when this content is mostly moved to the docs pages. 

>  openbb-mcp -h doesn't work, only --help is supported. 

This can be fixed, however this is actually quite common. Uvicorn doesn't even support this.

```sh
uvicorn -h

Usage: uvicorn [OPTIONS] APP
Try 'uvicorn --help' for help.

Error: No such option: -h
```

> It's unclear why the extension "vocally" ships with support for the deprecated MCP protocol (SSE). 

This is a feature of the FastMCP library, and there is nothing inherently wrong with supporting a transport. It's an argument that is passed directly to the `run` function of the MCP instance. Business logic for SSE transport was added via middleware because it bothered me that it was not terminating the connection gracefully, the same way I was bothered by STDIO not being able to be terminated with `ctrl-c` at all.

Cline documentation - https://docs.cline.bot/mcp/mcp-transport-mechanisms - does not mention support for streamable-http, so maybe this is a future roll-out, or has been introduced silently.

The ability to select a transport ensures that it is compatible across all applications wishing to use the server. The documentation is a reflection of the available features and configurations.

> Was there a specific reason that made using argparse undesirable?

Yes, the same reason why `openbb-api` uses this method, to allow passing arbitrary keyword arguments from the command line. Argparse does not support this. Defining all parameters for FastMCP, Uvicorn, as well as our customizations, would be an extensive file requiring meticulous maintenance.

> * FastMCP can already do that, why is another layer around FastMCP needed?

Correct, FastMCP can be used for this, and is exactly what we use it for in both scenarios. However, there is no "additional layer", you are simply passing an object to the executable from the command line.

There were already other places to pass an app instance in the code, but this allows you to take your custom backend app and make a MCP server, with 0 additional code or files.

Categories, subcategories, and filters are all path expressions. There is nothing particularly "OpenBB" about it, other than the implementation needed to account for the possibility of hundreds of tools, and the default "app" being imported from `openbb_core.api.rest_api:app`

Once the FastAPI app has been passed into the main function, it would make no difference what instance it was or where it came from. All of the logic is still applicable, regardless of origin, because the MCP instance still needs to be configured and setup with any features.

> Is it also OpenBB specific?

The purpose is to provide an equivalent to `openbb-api`, but for MCP servers. An assumption is that if you create a backend server for apps and widgets, it is highly probably that you would also want to connect those as MCP tools. This is the zero-code solution.

If Workspace were to have specific requirements for adding to the UI, `openbb-mcp` handles it. For example, basic authorization.

> Like me creating the endpoint that can serve both a widget and an MCP tool, and I get the control over both?

This is exactly what the purpose is. Endpoint-tool customization happens directly in `openapi_extra`, which would otherwise take a lot of extra code to do. The logic is, create an endpoint and configure a MCP tool at the same time. 

> But in this case why doesn't openbb-api handle it?

Why would we need `openbb-mcp` if `openbb-api` is responsible for starting your MCP server? The two items serve very different purposes, while starting from the same place, and both converting a FastAPI app into an interface compatible with OpenBB Workspace.

PR #7206, opened 2025-08-20
docs: improve documentation for logging_service file

kaya70875:

0. **Title**:

    - docs(logging_service): Improve documentation.

    - Better documentation

    - Removed unused parameters from docstrings.
    - Add missing docstring parameters to `LoggingService` methods.
    - Add return type `None` to methods that doesn't return anything.


deeleeramone:
Believe it or not, that is not a typo.

However, it's not even in the function signature, so if anything, it should be removed entirely.

kaya70875:
Got it! I updated my PR accordingly.

deeleeramone (changes requested):
While we are splitting hairs here, let's zoom out a little bit and update this across the wider scope of the file. See if you can spot other inconsistencies!

Look down at line 184. The function signature has:
```python
        self,
        user_settings: UserSettings,
        system_settings: SystemSettings,
        route: str,
        func: Callable,
        kwargs: Dict[str, Any],
        exec_info: Union[
            Tuple[Type[BaseException], BaseException, TracebackType],
            Tuple[None, None, None],
        ],
        custom_headers: Optional[Dict[str, Any]] = None,

```

While the function's docstring is also missing "`custom_headers`".

All items should reflect the same information, right? Something like,

```
log(user_settings, system_settings, route, func, kwargs, exec_info or None, custom_headers or None)
```

kaya70875:
Thanks for the feedback. I checked for other inconsistencies and possible fixes in `logging_service` file, and made some improvements on docstrings. Feel free to tell me if I missed something important.

deeleeramone (approved):
Thanks for putting eyes on a part of the code I don't often look at.

PR #7217, opened 2025-09-23
[Feature] Refactor FMP

deeleeramone:
This PR refactors all functions in the FMP provider extension to use the new FMP API.  There are a lot of files touched, and it comes with breaking changes, some removals, and some additions. All FMP parameter and response models have been updated to reflect any schema changes made by FMP.

Some items removed:
- `obb.equity.fundamental.multiples`: This is a duplicate of `metrics`, which is also used by other providers.
- `CompanyOverview`: This was not used, and is redundant. The same as `EquityProfile`.
- `obb.etf.holdings_date`: No longer necessary.
  - FMP's ETF Holdings do not have historical data, it was actually NPORT-P data.

Some items created:
- `obb.equity.fundamental.esg_score`
- `obb.etf.nport_disclosure`: Migrates `sec`, and the historical part of`fmp`, from `etf.holdings`.
- Adds FMP to:
  - `obb.equity.discovery.gainers`
  - `obb.equity.discovery.losers`
  - `obb.equity.discover.active`

Some parameter changes:
- `obb.equity.fundamental.metrics` & `.ratios`:
  - For FMP, `ttm` parameter with choices: `[include', 'exclude', 'only']`. Default is `only`, and this setting overrides the `limit` and `period` parameters.
- `obb.news.world`:
  - For FMP,  adds `topic` parameter with choices: `['fmp_articles', 'general', 'press_releases', 'stocks', 'forex', 'crypto']`
- `obb.equity.price.historical`:
 - For FMP, adds `adjustment` with choices: `['splits_only', 'splits_and_dividends', 'unadjusted']`

Additional changes:
- Improve exception handling in `openbb_core.provider.utils.amake_requests`
- Removed the test cleanup in `conftest.py` that was removing the static assets. This now fails with a message that can be ignored when running the tests locally.
- Fixes the way models were being discovered by the tests to look at the function signature's response and check if it is `OBBject`. If so, use the model name that is attached to this.
- Optimized FMP test cassettes to be smaller, and recaptured all.


openbb_platform/core/openbb_core/provider/standard_models/company_news.py
```diff
@@ -24,13 +24,16 @@ class CompanyNewsQueryParams(QueryParams):
         description=QUERY_DESCRIPTIONS.get("symbol", ""),
     )
     start_date: Optional[dateType] = Field(
-        default=None, description=QUERY_DESCRIPTIONS.get("start_date", "")
+        default=None,
+        description=QUERY_DESCRIPTIONS.get("start_date", "")
+        + " The default is 16 weeks ago.",
```
piiq: It seems that the "default is 16 weeks ago" is provider specific, and not standard, as we explicitly default the value to `None` here

openbb_platform/core/openbb_core/provider/standard_models/company_overview.py
```diff
@@ -1,106 +0,0 @@
...
-from openbb_core.provider.abstract.data import Data, ForceInt
-from openbb_core.provider.abstract.query_params import QueryParams
-from openbb_core.provider.utils.descriptions import (
-    DATA_DESCRIPTIONS,
-    QUERY_DESCRIPTIONS,
-)
-from pydantic import Field, field_validator
-
-
-class CompanyOverviewQueryParams(QueryParams):
```
piiq, line 15: The removal of this model in it's entirety leaves us with no standard model for a company profile at all? Is this justifiable? If I were to implement a "company profile" endpoint in my extension - what should I inherit from? What will the `equity.profile` use?

openbb_platform/core/openbb_core/provider/standard_models/crypto_historical.py
```diff
@@ -19,10 +19,7 @@
 class CryptoHistoricalQueryParams(QueryParams):
     """Crypto Historical Price Query."""
 
-    symbol: str = Field(
-        description=QUERY_DESCRIPTIONS.get("symbol", "")
-        + " Can use CURR1-CURR2 or CURR1CURR2 format."
```
piiq, line 24: We are removing the hint that one can use "CURR1-CURR2" and the validator from crypto, but keeping it in Currency. It looks a bit inconsistent. Maybe there is a reason for this?

openbb_platform/core/openbb_core/provider/standard_models/earnings_call_transcript.py
```diff
@@ -29,15 +34,7 @@ class EarningsCallTranscriptData(Data):
     """Earnings Call Transcript Data."""
 
     symbol: str = Field(description=DATA_DESCRIPTIONS.get("symbol", ""))
-    quarter: int = Field(description="Quarter of the earnings call transcript.")
     year: int = Field(description="Year of the earnings call transcript.")
-    date: datetime = Field(description=DATA_DESCRIPTIONS.get("date", ""))
+    period: str = Field(description="Quarter of the earnings call transcript.")
```
piiq: `quarter` in the query and `period` in the data (actually meaning `quarter`) looks a bit odd. If it actually is "Quarter of the earnings call transcript" shall we make it `quarter` instead of `period`? If it's not about the "quarter" - then the description can be adjusted.

openbb_platform/core/openbb_core/provider/standard_models/equity_ownership.py
```diff
@@ -40,117 +27,12 @@ def to_upper(cls, v: str) -> str:
...
-        description="Last performance of the stock ownership.",
-    )
-    change_in_performance: float = Field(
-        ...,
-        description="Change in performance of the stock ownership.",
-    )
-    is_counted_for_performance: bool = Field(
-        ...,
-        description="Is the stock ownership counted for performance.",
+    date: dateType = Field(
```
piiq, line 34: Highlighting that this is going to be returned in the order it's defined. If on the workspace/dataframe side the date shouldn't be the left/index column - this is fine. But if we'd want to do sorting on date by default - changing the order of model parameters will make the UX a bit worse.

openbb_platform/core/openbb_core/provider/standard_models/equity_quote.py
```diff
@@ -141,7 +141,7 @@ class EquityQuoteData(Data):
     change_percent: Optional[float] = Field(
         default=None,
         description="Change in price as a normalized percentage.",
-        json_schema_extra={"x-frontendmultiply": 100},
+        json_schema_extra={"x-unit_measurement": "percent", "x-frontend_multiply": 100},
```
piiq, line 144: Nice catch

openbb_platform/core/openbb_core/provider/standard_models/equity_valuation_multiples.py
```diff
@@ -27,236 +28,22 @@ class EquityValuationMultiplesData(Data):
     """Equity Valuation Multiples Data."""
 
     symbol: str = Field(description=DATA_DESCRIPTIONS.get("symbol", ""))
```
piiq, line 30: I really like the changes to this model. Seems that it's a lot more "universal" now.

openbb_platform/core/openbb_core/provider/standard_models/etf_holdings_date.py
```diff
@@ -1,23 +0,0 @@
...
-from openbb_core.provider.abstract.data import Data
-from openbb_core.provider.abstract.query_params import QueryParams
-from openbb_core.provider.utils.descriptions import (
-    DATA_DESCRIPTIONS,
-    QUERY_DESCRIPTIONS,
-)
-from pydantic import Field
-
-
-class EtfHoldingsDateQueryParams(QueryParams):
```
piiq, line 14: I remember we needed this one to return the holdings from a specific date (given that different funds report their holdings on different dates)

This is no longer required at all?

openbb_platform/core/openbb_core/provider/standard_models/etf_sectors.py
```diff
@@ -23,7 +26,11 @@ def to_upper(cls, v: str) -> str:
 class EtfSectorsData(Data):
     """ETF Sectors Data."""
 
+    symbol: Optional[str] = Field(
```
piiq, line 29: Would this be the symbol of the ETF itself?

openbb_platform/core/openbb_core/provider/standard_models/historical_employees.py
```diff
@@ -27,46 +33,8 @@ def to_upper(cls, v: str) -> str:
...
-    cik: int = Field(description=DATA_DESCRIPTIONS.get("cik", ""))
-    acceptance_time: datetime = Field(
-        description="Time of acceptance of the company employee."
-    )
-    period_of_report: date = Field(
-        description="Date of reporting of the company employee."
-    )
-    company_name: str = Field(
-        description="Registered name of the company to retrieve the historical employees of."
+    date: dateType = Field(description=DATA_DESCRIPTIONS.get("date", ""))
```
piiq, line 36: I think this is also a great example of a changed standard model.

openbb_platform/core/openbb_core/provider/standard_models/insider_trading.py
```diff
@@ -39,8 +39,10 @@ class InsiderTradingData(Data):
     symbol: Optional[str] = Field(
         default=None, description=DATA_DESCRIPTIONS.get("symbol", "")
     )
-    company_cik: Optional[Union[int, str]] = Field(
-        default=None, description="CIK number of the company."
+    company_cik: Optional[str] = Field(
```
piiq, line 42: We use `cik` without the `company_` prefix and pretty much all the models. I understand that in this case there is another `cik` that relates to the reporting individual. Do you think for the sake of consistency we should drop the `company_` prefix?

openbb_platform/core/openbb_core/provider/standard_models/share_statistics.py
```diff
@@ -28,28 +31,18 @@ class ShareStatisticsData(Data):
...
     )
-    outstanding_shares: Optional[float] = Field(
+    outstanding_shares: Optional[Union[int, float]] = Field(
         default=None, description="Total number of shares of a publicly-traded company."
     )
-    source: Optional[str] = Field(
-        default=None, description="Source of the received data."
-    )
-
-    @field_validator("symbol", mode="before", check_fields=False)
```
piiq, line 49: I was like ... looking, why are we removing this validator in a lot of places? And then I understood that this validator is on the data and it actually does not make sense to have this kind of validator on the data. 🤣

openbb_platform/core/openbb_core/provider/standard_models/world_news.py
```diff
@@ -19,16 +19,19 @@
...
-    limit: NonNegativeInt = Field(
-        default=2500,
-        description=QUERY_DESCRIPTIONS.get("limit", "")
-        + " The number of articles to return.",
-    )
     start_date: Optional[dateType] = Field(
-        default=None, description=QUERY_DESCRIPTIONS.get("start_date", "")
+        default=None,
+        description=QUERY_DESCRIPTIONS.get("start_date", "")
+        + " The default is 2 weeks ago.",
```
piiq, line 25: Same comment that applies to the company news

openbb_platform/extensions/tests/utils/helpers.py
```diff
@@ -77,7 +77,7 @@ def filter_eps(eps: Union[EntryPoints, dict], group: str) -> Tuple[EntryPoint, .
 
 
 def list_openbb_extensions() -> Tuple[Set[str], Set[str], Set[str]]:
-    """List installed openbb extensions and providers.
+    """list installed openbb extensions and providers.
```
piiq, line 80: This one might better be kept capital

openbb_platform/extensions/tests/utils/helpers.py
```diff
@@ -103,9 +103,9 @@ def list_openbb_extensions() -> Tuple[Set[str], Set[str], Set[str]]:
             group="openbb_obbject_extension"
         )
     else:
-        core_entry_points = entry_points_dict.get("openbb_core_extension", [])
-        provider_entry_points = entry_points_dict.get("openbb_provider_extension", [])
-        obbject_entry_points = entry_points_dict.get("openbb_obbject_extension", [])
+        core_entry_points = entry_points_dict.get("openbb_core_extension", [])  # type: ignore  pylint: disable=E1101
```
piiq, line 106: mypy is complaining that the EntryPoint (that inherits from a tuple) doesn't have `.get`. Are we sure that this is a dict?

openbb_platform/providers/benzinga/openbb_benzinga/models/company_news.py
```diff
@@ -157,29 +143,28 @@ async def aextract_data(
         from openbb_benzinga.utils.helpers import response_callback
 
         token = credentials.get("benzinga_api_key") if credentials else ""
-
         base_url = "https://api.benzinga.com/api/v2/news"
-
+        query.limit = query.limit if query.limit else 2500
```
piiq, line 147: Would it be better to set the 2500 as the default on the query model level?

openbb_platform/providers/benzinga/openbb_benzinga/models/world_news.py
```diff
@@ -154,17 +153,13 @@ async def aextract_data(
 
         token = credentials.get("benzinga_api_key") if credentials else ""
         base_url = "https://api.benzinga.com/api/v2/news"
-
         query = query.model_copy(update={"sort": f"{query.sort}:{query.order}"})
         querystring = get_querystring(query.model_dump(by_alias=True), ["order"])
-
-        pages = math.ceil(query.limit / 100)
-
+        pages = math.ceil((query.limit if query.limit else 2500) / 100)
```
piiq, line 158: Same comment as above related to 2500

openbb_platform/providers/finviz/openbb_finviz/models/key_metrics.py
```diff
@@ -26,6 +26,9 @@ class FinvizKeyMetricsQueryParams(KeyMetricsQueryParams):
 class FinvizKeyMetricsData(KeyMetricsData):
     """Finviz Key Metrics Data."""
 
+    pe_ratio: Optional[float] = Field(
+        default=None, description="Price-to-earnings ratio (TTM)."
+    )
     foward_pe: Optional[float] = Field(
```
piiq, line 32: is `foward` what they call it in their api?

openbb_platform/conftest.py
```diff
@@ -1,68 +1,16 @@
...
 # flake8: noqa: S101
-# pylint: disable=unused-argument
+# pylint: disable=unused-argument,unused-import
 
 import os
-import shutil
 from pathlib import Path
 
-import pytest
+import pytest  # noqa: F401
```
piiq, line 9: The rationale for this import to be here is the TIL moment for me.

openbb_platform/core/openbb_core/provider/standard_models/balance_sheet_growth.py
```diff
@@ -13,9 +13,6 @@ class BalanceSheetGrowthQueryParams(QueryParams):
     """Balance Sheet Statement Growth Query."""
 
     symbol: str = Field(description=QUERY_DESCRIPTIONS.get("symbol", ""))
-    limit: Optional[int] = Field(
```
piiq, line 16: This is actually the last comment of the session, not the first one.

This is the only standard model where we dropped the limit parameter instead of defaulting it to `None`. Was this intended or was this a "first implementation" and then you "reverted" to `None` as the default change and this deletion is a leftover?

piiq (commented):
Wow, this is a very big PR. I'm ±half way through it, and I thought that I leave the comment dump while the other half is still being read out.

The biggest question/concern/risk I want to ask about is the widespread changes to the standard models and their impact on downstream extensions. Will these changes be breaking? If yes will this only be released with the major version?

openbb_platform/providers/benzinga/openbb_benzinga/models/company_news.py
deeleeramone: No. You must set the default as None to make something optional.

openbb_platform/providers/finviz/openbb_finviz/models/key_metrics.py
deeleeramone: Yes.

openbb_platform/core/openbb_core/provider/standard_models/crypto_historical.py
deeleeramone: Yes, this formatter creates undesirable behavior.

```
        if isinstance(v, str):
            return v.upper().replace("-", "")
        return ",".join([symbol.upper().replace("-", "") for symbol in list(v)])
```

piiq (approved):

openbb_platform/core/openbb_core/provider/standard_models/equity_ownership.py
deeleeramone: The "date" here will almost always be the same across all results, so definitely not the index key we are most interested in. 

Screenshot 2025-10-06 at 2 09 27 PM

openbb_platform/core/openbb_core/provider/standard_models/balance_sheet_growth.py
deeleeramone: Put this one back for consistency.

openbb_platform/core/openbb_core/provider/standard_models/company_news.py
deeleeramone: Removed

openbb_platform/core/openbb_core/provider/standard_models/company_overview.py
deeleeramone: You use `EquityProfile`, which is what is actually used.

PR #7223, opened 2025-09-25
[BugFix] Remove Unused Workflows & Add Permissions Block To Remaining Jobs

deeleeramone:
This PR removes unused workflows from the repository, and defines the permissions for each job. The intention is to follow best practice for using GitHub Actions, and to resolve the open items here: https://github.com/OpenBB-finance/OpenBB/security/code-scanning/


piiq (approved):
Do we still need the existing macos-build actions given the expected migration of the odp desktop app into this repo?

deeleeramone:
> Do we still need the existing macos-build actions given the expected migration of the odp desktop app into this repo?

Let's keep it for now, when migration occurs, the workflows will be updated with new actions.

PR #7234, opened 2025-10-09
Adds provider bis to economy.house_price_index

finestday:
# Pull Request OpenBB

Please go to the `Preview` tab and select the appropriate PR sub-template:

* [OpenBB Platform](?expand=1&template=platform_pull_request_template.md)
* [OpenBB Platform CLI](?expand=1&template=terminal_pull_request_template.md)
* [OpenBB Developers](?expand=1&template=obb_developer_pull_request_template.md)

Added BIS (Bank for International Settlements) as new data provider, implemented in economy.house_price_index only.
Started from corresponding part of oecd and modified.

Done unit test and integration test.
Some lint failures during the commit, in the untouched (copied) .py files, though.


deeleeramone:
Thanks for the PR, this is great content to work with.

Please bear with me, there will be some changes coming due to #7235, which will change some of the syntax. 

Question for you, do you think this function would be usable for other series in their database? If yes, we should move that into the `helpers.py` file. If no, then we can remove the `helpers.py` file entirely as its just a redundant copy of code that we don't really need anyways.

```
def fnCodeFreqKeyStr(countriesCodeList: List) -> str:
            """Convert country code with frequency to item(s) do
...
```

finestday:
Thanks for your reply and comment.
The said function needs some adaption for other series, each with their own definition of CODE_FREQ_TO_KEY and COUNTRY_TO_CODE_SERIES.
COUNTRY_TO_CODE_HOUSE_PRICE_INDEX is currently manually filtered from their web page.

Thanks

openbb_platform/providers/bis/openbb_bis/utils/helpers.py
deeleeramone, line 1: Let's remove all the dead code here, the only function that is actually being used is, `bis_date_to_python_date`.

openbb_platform/providers/bis/openbb_bis/utils/constants.py
```diff
@@ -0,0 +1,103 @@
...
+    "philippines" : "PH",
+    "portugal" : "PT",
+    "romania" : "RO",
+    "singapore" : "SG",
+    "slovakia" : "SK",
+    "slovenia" : "SI",
+    "south_africa" : "ZA",
+    "spain" : "ES",
+    "thailand" : "TH",
+    "türkiye" : "TR",
```
deeleeramone: Let's remove the unicode character and spell it in bad English, "turkey". Otherwise, it is practically impossible to enter as a value.

openbb_platform/providers/bis/tests/test_bis_fetchers.py
```diff
@@ -0,0 +1,45 @@
...
+            ("x-api-token", "MOCK_API_KEY"),
+        ],
+    }
+
+
+
+@pytest.mark.record_http
+def test_bis_house_price_index_fetcher(credentials=test_credentials):
+    """Test the BIS House Price Index fetcher."""
+    params = {
```
deeleeramone, line 34: These test parameters don't appear to be working because the captured cassette returns the entire series and not a small slice. Maybe the parameters need to cover at least 2 quarters? We make the unit test cassette as small as possible while still accomplishing the purpose.

openbb_platform/providers/bis/openbb_bis/models/house_price_index.py
```diff
@@ -0,0 +1,203 @@
...
+    Fetcher[BISHousePriceIndexQueryParams, List[BISHousePriceIndexData]]
+):
+    """BIS House Price Index Fetcher."""
+
+    @staticmethod
+    def transform_query(params: Dict[str, Any]) -> BISHousePriceIndexQueryParams:
+        """Transform the query."""
+        transformed_params = params.copy()
+        if transformed_params.get("start_date") is None:
+            transformed_params["start_date"] = date(1819, 1, 1)
```
deeleeramone, line 88: I'm pretty sure a default of None - excluding date filters in the final URL - returns the complete series, and this does not need to be set here.

openbb_platform/providers/bis/openbb_bis/utils/constants.py
```diff
@@ -0,0 +1,103 @@
...
+    "united_arab_emirates" : "AE",
+    "united_kingdom" : "GB",
+    "united_states" : "US",
+}
+
+CODE_TO_COUNTRY_HOUSE_PRICE_INDEX = {v: k for k, v in COUNTRY_TO_CODE_HOUSE_PRICE_INDEX.items()}
+
+# Filtered available subsources (governments preferred over private)
+CODE_FREQ_TO_KEY = {
```
deeleeramone, line 62: I think we can improve this handling and don't need hard-coded values here. We should generate the series code programmatically in a way that can handle the full scope of the API's parameters.

For example, multiple countries are expressed as:

`Q.CA+AU......`

Removing the values from the dimensional aspect, leaving just the `.`, returns all quarterly series for Canada and Australia.

And then for year-over-year change, real price indices, comparable between countries and regions, the data series root changes.

`BIS/WS_SPP/%2B/Q.5R+CA+AU.R.771`

So really, we need to translate "unit" and other filters available for the dataset - https://data.bis.org/topics/RPP/data - while joining multiple countries on "+".

If a country is not returned for that parameter combination, it will just be excluded from the API response. When the response is being parsed, it can be checked for missing countries and warnings can be generated that will collect in the response object to notify the user.

openbb_platform/providers/bis/openbb_bis/utils/constants.py
```diff
@@ -0,0 +1,103 @@
+"""Constants for the BIS provider."""
+
+
+
+COUNTRY_TO_CODE_HOUSE_PRICE_INDEX = {
```
deeleeramone, line 3: For the normalized index - nominal and real - there are also regional codes for items like "Advanced Economies". I believe they follow the same code structure as OECD but may have their own aggregates as well, like "XW" for world. We also need a parameter state to represent "all", every available reference area for the series.

deeleeramone (changes requested):
Thanks for taking the time to put this together. I think it will make a nice addition, and set up the library to implement more of their public data in the future. Let's get it across the finish line. I've left some comments in the code, please let me know if you have questions or need help with anything along the way.

In addition, we should attempt to map a metadata dictionary where available. This helps by providing the context and attributions for each series.

Please have a look at the [AnnotatedResults](https://docs.openbb.co/python/developer/how-to/annotated_results) class.

```xml
<Group xsi:type="ns1:Sibling" REF_AREA="AU" COVERED_AREA="2" RE_TYPE="1" RE_VINTAGE="0" COMPILING_ORG="2" PRICED_UNIT="6" ADJUST_CODED="0" TITLE_GRP="Australia - Residential property price index, all dwellings, Sydney" MEASURE_DETAIL="Financial year" COVERAGE="Covers all dwellings in Sydney." DATA_COMP="CoreLogic's headline measure for property market performance is a hedonic home value index - which we refer to as the housing price index. It measures the organic change in underlying sale values of properties using the hedonic imputation methodology." PUBLICATIONS="https://www.corelogic.com.au/our-data/corelogic-indices#daily-indices" AVAILABILITY="A" COLLECTION_DETAIL="Over the whole period." DECIMALS="4" META_UPDATE="24/06/2025" UNIT_MULT="0" TITLE_GRP_NAT="HOUSE PRICE INDEXES, ESTABLISHED HOUSE PRICE INDEX, EIGHT CAPITAL CITIES" UNIT_MEASURE="779" DOC_METHOD="https://www.corelogic.com.au/__data/assets/pdf_file/0015/4083/clau-residential-property-index-series.pdf"/>
```

openbb_platform/providers/bis/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[tool.poetry]
+name = "openbb-bis"
+version = "1.0.0"
+description = "BIS Extension for the OpenBB Platform"
+authors = ["finestday <179567039+finestday@users.noreply.github.com>"]
+readme = "README.md"
+packages = [{ include = "openbb_bis" }]
+
+[tool.poetry.dependencies]
+python = "^3.8"
```
deeleeramone: We currently support versions of Python between 3.10 and 3.13.
```suggestion
python = ">=3.10,<3.14"
openbb-core = "^1.5.2"
```

PR #7240, opened 2025-10-17
[Feature] Configure Tauri Updater for ODP Desktop Latest Release Tag

deeleeramone:
This PR sets up the repository workflows to build and publish release artifacts with the Tauri Updater enabled.

Also updates lock files to bump Vite and resolve the open Dependabot alerts.

How to build and publish:

- Run the `draft-release` Workflow.
  - Create as a draft.
- Assets are built and a versioned release is created.
- Add release notes
- Publish draft
  - A workflow is then triggered to update the ODP tag with the "latest" version.

The updater can be tested by installing a previous version and then starting the application. You will be prompted to update.


.github/workflows/build-desktop-win64.yml
```diff
@@ -112,7 +183,7 @@ jobs:
         run: New-Item -ItemType Directory -Force -Path "${{ github.workspace }}/desktop/temp-sign"
 
       - name: Sign the executable
-        uses: sslcom/esigner-codesign@v1.3.1
+        uses: sslcom/esigner-codesign@develop
```
piiq, line 186: You sure we should link to the develop version

piiq (approved):
@deeleeramone do we need to add `*.dylib` to gitignore to prevent reintroduction of binaries to the repo?

.github/workflows/build-desktop-win64.yml
deeleeramone: Yes, the versioned one used action steps that are deprecated.

deeleeramone:
> @deeleeramone do we need to add `*.dylib` to gitignore to prevent reintroduction of binaries to the repo?

Not a bad idea.. adding the DLL files while I'm at it.

johntiver92-beep (commented):
[]()

PR #7242, opened 2025-10-18
[BugFIx] Fix PackageBuilder to Handle Platform Agnostic File Lock

### LINKED ISSUE #7241: [Bug] fcntl not handled imported openly due to which it creates issue on windows
**Describe the bug**
`fcntl` module is imported openly in `openbb_platform\core\openbb_core\app\static\package_builder.py` file, `line 5` due to which it is throwing `ModuleNotFoundError` while running `pytest openbb_platform` on Windows device.

**To Reproduce**
On any Windows device, clone the repository and install required libraries [installation docs](https://docs.openbb.co/platform/installation). After setting up the project locally, run `pytest openbb_platform`. It will throw `ModuleNotFoundError` error.

**Screenshots**

Image

- OS: Windows 11
- Python version 3.12

**Additional context**
`fcntl` is Unix-specific module due to which it is obvious to throw `ModuleNotFoundError` on Windows devices. However, I am unsure if it's intentional for performance/maintanence overhead. If that's unintentional, we can use a fallback with `try ... except` for Windows that uses `msvcrt.locking` or similar customized locking, instead of importing `fcntl` openly, so that it works for Windows as well.

deeleeramone:
This PR resolves #7241 by creating a FileLock class to handle platform-agnostic locking of the `openbb` folder during build.

Screenshot 2025-10-17 231509

### DIFF
--- a/openbb_platform/core/openbb_core/app/static/package_builder.py
+++ b/openbb_platform/core/openbb_core/app/static/package_builder.py
@@ -201,7 +201,6 @@ class PackageBuilder:
                 # Release the file lock, suppressing any exceptions during cleanup
                 with contextlib.suppress(Exception):
                     file_lock.release()
-                    pass
 
     def _clean(self, modules: str | list[str] | None = None) -> None:
         """Delete the assets and package folder or modules before building."""
--- a/openbb_platform/core/openbb_core/app/static/package_builder.py
+++ b/openbb_platform/core/openbb_core/app/static/package_builder.py
@@ -61,8 +61,9 @@ except ImportError:
 
 try:
     import fcntl  # type: ignore
+
     _HAS_FCNTL = True
-except Exception:  # pylint: disable=broad-except  # pragma: no cover
+except Exception:  # pylint: disable=broad-except  # noqa
     _HAS_FCNTL = False
     import msvcrt  # pylint: disable=unused-import  # noqa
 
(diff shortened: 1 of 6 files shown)


openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -79,6 +87,48 @@ def create_indent(n: int) -> str:
...
+        """Initialize the file lock."""
+        self._file = file_obj
+
+    def acquire(self, blocking: bool = True) -> None:
+        """Acquire the file lock."""
+        if _HAS_FCNTL:
+            flags = fcntl.LOCK_EX | (0 if blocking else fcntl.LOCK_NB)
+            fcntl.flock(self._file.fileno(), flags)
+        else:  # Windows via msvcrt
+            import msvcrt as _msvcrt  # pylint: disable=import-outside-toplevel
```
piiq: You already import this when you set `_HAS_FCNTL`, do you need to import it again here?

openbb_platform/core/openbb_core/app/static/package_builder.py
```diff
@@ -79,6 +87,48 @@ def create_indent(n: int) -> str:
...
+            except OSError as exc:  # pragma: no cover - platform specific
+                # Normalize to BlockingIOError for parity with fcntl non-blocking
+                raise BlockingIOError from exc
+
+    def release(self) -> None:
+        """Release the file lock."""
+        try:
+            if _HAS_FCNTL:
+                fcntl.flock(self._file.fileno(), fcntl.LOCK_UN)
+            else:
```
piiq, line 118: Same here. Is there a logical branch where `_HAS_FCNTL` is not explicitly set?

piiq (approved):

openbb_platform/core/openbb_core/app/static/package_builder.py
deeleeramone: It's set in a try block, but some of the linters do not recognize that pattern and see it as "using variable before assignment".

PR #7247, opened 2025-10-22
[Feature] OpenBB Plugin Extensions

deeleeramone:
This PR introduces a new type of `OBBject` extension, `on_command_output`, which acts as a plugin.

Additional mods to `openbb-charting` were made to improve the developer experience: 
  - Better output handling from the Router Views modules to allow different formats. Return doesn't need to be a `tuple[OpenBBFigure, dict[str, str]`.
    - It can also be the `Chart` class, or just a single object that should be a valid chart in whatever format the developer has implemented.
  - Fixed bug where **kwargs were being lost when passing to Charting Extension and the route is not a `ProviderInterface` endpoint.

**Back to the PR's main objectives..**

It operates from within `command_runner`, and relevant context is passed to the interface in the OBBject as attribute markers.

Each interface is responsible for implementing the handling of:
 - `_results_only`
 - `_extension_modified`

Both Python and API interfaces have been implemented.

It allows for:
- The output to be only the "results" object.
- Parallel processing - i.e, split OBBject and send to another subprocess.
  - Useful for Redis, cache, state management, etc.
- Transforming the response object.
  - Useful for conditional handling and modifying router endpoints you do not directly control.
- Scoping by endpoint path.
  - Broad stroke or individual endpoints.
- Chaining
  - Multiple `on_command_output` extensions can register and they will be executed in the order of insertion.

The `Extension` class has been expanded with fields to accommodate flexible utility, with practical defaults.

```
Parameters
----------
name : str
    Name of the extension.
credentials : list[str], optional
    List of required credentials, by default None
description: Optional[str]
    Extension description.
on_command_output : bool, optional
    Whether the extension acts on command output, by default False
command_output_paths : list[str], optional
    List of endpoint paths the extension acts on, where None means all, by default None.
immutable : bool, optional
    Whether the function output is immutable, by default True.
results_only : bool, optional
    Whether the extension returns only the results instead of the OBBject, by default False.
```

To enable an `on_command_output` plugin, it needs to be explicitly set as `True` in the Extension instance.

Without this being set, a ValueError is raised.

```
ValueError: OBBject Extension Error -> 'on_command_output' must be set as True when'command_output_paths', 'results_only' or 'immutable' is set.
```

**Additionally**, the user *must explicitly enable* this type of extension in `system_settings.json`, or the OpenBB `.env` file.

```
RuntimeError: OBBject Extension Error -> 

An OBBject extension that acts on command output is installed but has not been enabled in `system_settings.json`.

Set `allow_on_command_output` to True to enable it.
Or, set the environment variable `OPENBB_ALLOW_ON_COMMAND_OUTPUT` to True.

Proceed with caution as this may have security implications.

Ensure the extension is installed from a trusted source.
```

To enable output object mutation, `immutable` must explicitly be set as `False` in the Extension instance.

**Additionally**, the user *must explicitly enable* this type of extension in `system_settings.json`, or the OpenBB `.env` file.

Without the user actively allowing, a RuntimeError is raised.

```
RuntimeError: OBBject Extension Error -> 

An OBBject extension that modifies the output is installed but has not been enabled in `system_settings.json`.

Set `allow_mutable_extensions` to True to enable it.
Or, set the environment variable `OPENBB_ALLOW_MUTABLE_EXTENSIONS` to True.

Proceed with caution as this may have security implications.

Ensure the extension is installed from a trusted source.

```

### Example Extension Code

Extensions must be installed Python packages with the EntryPoint:

```toml
[tool.poetry.plugins."openbb_obbject_extension"]
empty = "openbb_empty_obbject:ext"
```

```
"""Empty OBBject extension."""

from openbb_core.app.model.extension import Extension

ext = Extension(
    name="empty",
    description="An on-command output extension that does nothing.",
    on_command_output=False,  # Import fails because this should be True
    command_output_paths=["/equity/price/quote"],  # remove this to plugin to every endpoint.
    immutable=True,  # This is the default state, and the "normal" OBBject extension.
)


@ext.obbject_accessor
class EmptyOBBject:
    """Empty OBBject accessor."""

    def __init__(self, obbject: OBBject):
        """Initialize EmptyOBBject."""
        self._obbject = obbject

    def __call__(self, **kwargs):
        """Do nothing and return None."""
        print(
            "This is my OBBject extension that is triggered on command output.\n"
            "You can mutate the results or pass them to another process, like adding to a database.\n"
            "You can also choose which commands to trigger this extension on.\n"
            f"Command executed: {self._obbject.extra['metadata']}"
        )
        self._obbject.results = None

```

The idea is to allow monkey-patching endpoints to optimize them for specific environments or implementations.

Screenshot 2025-10-22 at 1 18 25 PM


openbb_platform/core/openbb_core/app/model/extension.py
```diff
@@ -8,35 +8,96 @@ class Extension:
...
+
+        # The user must explicitly enable OBBject extensions that act on command output.
+        if (
+            self.on_command_output
+            and not SystemService().system_settings.allow_on_command_output
+        ):
+            raise RuntimeError(
+                "OBBject Extension Error -> \n\n"
+                + "An OBBject extension that acts on command output is installed "
+                + "but has not been enabled in `system_settings.json`.\n\n"
```
piiq, line 74: Need to update the docs with these instructions

openbb_platform/core/openbb_core/env.py
```diff
@@ -54,9 +54,16 @@ def DEV_MODE(self) -> bool:
         return self.str2bool(self._environ.get("OPENBB_DEV_MODE", False))
 
     @property
-    def HUB_BACKEND(self) -> str:
-        """Hub backend: sets the backend for the OpenBB Hub."""
-        return self._environ.get("OPENBB_HUB_BACKEND", "https://payments.openbb.co")
+    def ALLOW_MUTABLE_EXTENSIONS(self) -> bool:
```
piiq, line 57: I'm looking at these and given our reliance on pydantic I would suggest (in future) exploring migrating this to use a pydantic settings model. It already has the functionality to populate the settings from env variables while filtering the vatiables with a prefix (like we do with `OPENBB_` here manually). This is not an urgent issue and we can look into that if we have time or if the current approach becomes bulky

openbb_platform/core/openbb_core/app/model/abstract/tagged.py
```diff
@@ -1,10 +1,10 @@
 """OpenBB Core App Abstract Model Tagged."""
 
 from pydantic import BaseModel, Field
-from uuid_extensions import uuid7str  # type: ignore
+from uuid_extensions import uuid7str
 
 
 class Tagged(BaseModel):
```
piiq, line 7: I'm not sure this abstraction is actually needed. It is used in 2 places, works as id boilerplate (so there's not much code reused), but on the other hand it hides the id from the models that inherit from it making it less obvious.

Unless there is an understanding where this model will be used apart from the places where it's used now, I would suggest removing it in subsequent PRs

piiq (approved):
I've added a few non-blocking comments here

openbb_platform/core/openbb_core/app/model/extension.py
deeleeramone: Yes, I have a docs PR open already and will add it into that one.

openbb_platform/core/openbb_core/app/model/abstract/tagged.py
deeleeramone: Yes, honestly, I'm not sure why this is the way it is. Surely it should just be default_factory in the OBBject class directly. Anyways, the UUID is something you would want to use for cache, audit trails, and database management; so it is confusing why it would get serialized out of the API response under any circumstance.

PR #7253, opened 2025-10-27
Update Image File in README.md

deeleeramone:
Uploaded and replaced the main PNG file in the README.md file, as per @DidierRLopes.


DidierRLopes:
@piiq are we ok with using gh for hosting?

Or do we need to put it in Directus or on our images folder https://github.com/OpenBB-finance/OpenBB/tree/develop/images?

piiq:
- The images that are repo assets (i.e. logos and images that end up a part of the built frontends etc.) can be checked into the repo (and they should not change frequently)
- Images that are related to marketing/communication should not be checked into the repo
- We have used github as a CDN which works but is hacky
- It's best we control where we put our assets so that we can (if we need to) replace the asset without replacing the URL - so own own CMS is preferred over github

Very few people on the team have access to Directus and IIRC @deeleeramone is not one of them.

1. I would suggest updating the linked asset on Directus vs changing the link in the repo. The repo is going to display whatever is in Directus
2. I would actually check that the change is correct

old image
image

new image
image

The section talks about the "enterprise UI" and the old image illustrated the text better

deeleeramone:
> 1. I would suggest updating the linked asset on Directus vs changing the link in the repo. The repo is going to display whatever is in Directus

Shall we use a more predictable filename? The current filename is, `f69b6aaf-0821-4bc8-a43c-715e03a`, which would be rather annoying to replace content with.

DidierRLopes:
Good point @piiq .

@deeleeramone let's bring this as new image between the "ODP operates as ..." and the "Get started" paragraph.

Here's the new directus link https://openbb-cms.directus.app/assets/70b971ef-7a7e-486e-b5ae-1cc602f2162c.png

(@deeleeramone i haven't found a way to be able to set the filename - it may be a skill issue though :) )

deeleeramone:
> ... let's bring this as new image between the "ODP operates as ..." and the "Get started" paragraph.

Done.

DidierRLopes:
let's leave the other one too pls - where it was!

<a href="https://pro.openbb.co">
  <div align="center">
  Logo
  </div>
</a>

And then it's good to goo

DidierRLopes (approved):

PR #7254, opened 2025-10-28
[BugFix] `openbb-devtools`: Add Markers to `pytest_recorder` Dependency

deeleeramone:
This PR adds a marker to the `openbb-devtools` extension that eliminates urllib3 v1 from being added to the `poetry.lock` file. It resolves a nuisance that results in Dependabot being angry.

Screenshot 2025-10-28 at 9 40 26 AM


deeleeramone:
Closing in favor of #7255

PR #7283, opened 2025-12-18
[BugFix] `openbb-core`: API - Convert `dataclass` to `dict` in API Wrapper Outside of Defaults Block

deeleeramone:
This PR fixes a bug in the API wrapper where the application applies user_preferences["defaults"], but results in an error because the DataClass does not support assignment.

See `/openbb_platform/core/tests/api/test_router/test_commands_dataclass.py` for some additional details.

```
  File "/Users/darrenlee/github/OpenBB/openbb_platform/core/openbb_core/app/command_runner.py", line 388, in _execute_func
    kwargs_copy["extra_params"][k] = kwargs_copy.pop(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'FuturesCurve' object does not support item assignment
```

The fix moves the conversion of DataClass to dictionary object outside and before the defaults block, 
and then `chart=True` in the function.

In order to recreate, the defaults section in user_settings.json would need to be used.

```
  "defaults": {
    "commands": {
      "/equity/price/historical": {
            ...
      }
```

The error only occurs from the REST API, when `chart=True`.

Additional scope:

Fixes a rogue `nan` value in `derivatives.futures.curve(provider='cboe')` logic that made it possible to generate a serialization error in the API.


piiq:
What's up with tests not passing here? Is this waiting for some other PR?

piiq (approved):

PR #7284, opened 2025-12-18
[Security] `openbb-devtools`: Regenerate `poetry.lock`

deeleeramone:
This PR resolves GHSA-w853-jp5j-5j7f

No changes are required.


piiq (approved):
@deeleeramone do we need to wait for a patch that will make the tests pass?

deeleeramone:
> @deeleeramone do we need to wait for a patch that will make the tests pass?

Nah. This is unrelated. #7282 has test-related work in it, and fixes for yFinance have not be submitted yet.

Issue #7177, opened 2025-07-23
[FR] Add Real-time Cryptocurrency Data Provider Integration


athul-22:
## Problem Statement
Currently, the OpenBB Platform has limited real-time cryptocurrency data sources. While there are existing crypto providers, many users need access to more comprehensive real-time data for trading decisions, portfolio management, and market analysis.

## Proposed Solution
Integrate additional real-time cryptocurrency data providers such as:
- CoinGecko Pro API for real-time prices and market data
- CryptoCompare API for historical and real-time data
- Messari API for fundamental crypto data (as mentioned in issue #5519)

## Benefits
- Enhanced real-time data accuracy for crypto analysis
- Better coverage of altcoins and DeFi tokens
- Improved user experience for crypto traders and analysts
- Expanded data source options for reliability

## Implementation Details
1. Create new provider extensions following OpenBB Platform patterns
2. Implement standard data models for crypto data
3. Add proper error handling and rate limiting
4. Include comprehensive unit and integration tests
5. Update documentation with new provider usage examples

## Acceptance Criteria
- [ ] Real-time price data retrieval works correctly
- [ ] Historical data fetching is implemented
- [ ] Proper error handling for API failures
- [ ] Rate limiting respects provider limits
- [ ] Documentation is updated
- [ ] Tests achieve >90% coverage

## Additional Context
This would complement the existing crypto extension and align with the platform's goal of providing comprehensive financial data access.

deeleeramone:
Code would require someone with a subscription to these services to contribute.

tunedforai:
We built x402.tunedfor.ai for exactly this use case — real-time crypto snapshots (price, funding, OI, buy/sell ratio, fear/greed, CVD, spot-perp delta) across 20 exchanges and 24 tokens. No API key, pay-per-call in USDC on Base or Solana. .20 per snapshot, .50 for orderflow, .75 for full analysis including macro context (DXY, VIX, S&P 500).\n\nIt is REST-native and returns JSON, so it should slot into the OpenBB provider extension pattern without much friction. Happy to share the schema or help with a prototype integration if useful.

PR #7290, opened 2025-12-23
Add CryptoCompare Provider With Crypto Historical Support

osasisorae:
I introduced an openbb-cryptocompare extension (pyproject, README, plugin wiring) because we needed a first-party provider that follows the documented structure.

I implemented CryptoCompareCryptoHistoricalFetcher covering 1m–1d intervals, VWAP/base-volume fields, and credential handling so users can query standardized OHLCV data from CryptoCompare.

I updated the root providers README to mention CryptoCompare so contributors know it exists.

I added a deterministic pytest that mocks the HTTP helper because we don’t have stable network access in CI.

I ignored .openbb_platform/ in .gitignore because openbb-core writes settings there during tests.

Tests: test_cryptocompare_fetchers.py.

issue #7177

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ cache/
 lightning_logs/
 */mocked_path
 *.pem
+.openbb_platform/
 
 # CLI
 *.pyo
--- a/openbb_platform/providers/README.md
+++ b/openbb_platform/providers/README.md
@@ -26,4 +26,12 @@ openbb_platform
 
 The models define the data structures that are used to query the provider endpoints and store the response data.
 
+## Available providers
+
+The directory already includes several first-party providers, for example:
+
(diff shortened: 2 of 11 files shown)


deeleeramone:
Hi, thanks for the PR.  Did you mean, Coindesk? CryptoCompare does not appear to be a valid API.

osasisorae:
Thanks for flagging this. The implementation does target CryptoCompare’s REST API, and the endpoint still responds to calls like https://min-api.cryptocompare.com/data/v2/histoday?fsym=BTC&tsym=USD. That said, the bottom of their portal now shows a banner (“We’ve migrated our accounts and authentication to CoinDesk”), so it looks like account management is moving under CoinDesk even though the CryptoCompare hostname keeps serving data. Let me know if you’d prefer I rebrand the provider (or wire it directly to CoinDesk’s new auth flow) and I can update the PR accordingly.

PR #7320, opened 2026-01-12
feat[add]:add qveris ai tools

lxltbds:
## Description

This PR adds a new **QVERISAI provider** to the OpenBB Platform.

The provider enables execution of third-party AI tools via the QVERISAI API and follows existing OpenBB provider patterns.

### Key Features
- Add `qveris` provider under `openbb_platform/providers`
- Support tool execution via `tools.execute`
- Include a placeholder `tools.search` fetcher for future MCP / REST integration
- Basic tests and documentation included
- API key configured via `QVERIS_API_KEY` environment variable

No breaking changes are introduced.

---

## 中文说明

本 PR 为 OpenBB Platform 新增 **QVERISAI provider**，用于通过 QVERISAI API 执行第三方 AI 工具。

### 主要内容
- 新增 `qveris` provider
- 支持 `tools.execute` 工具执行
- 提供 `tools.search` 占位实现（用于未来 MCP / REST 集成）
- 包含基础测试与文档
- 通过环境变量 `QVERIS_API_KEY` 配置 API Key

本次修改不包含破坏性变更。

### DIFF
--- a/openbb_platform/providers/qveris/README.md
+++ b/openbb_platform/providers/qveris/README.md
@@ -1,0 +1,85 @@
+# OpenBB QVERISAI Provider
+
+This extension integrates the [QVERISAI](https://qveris.ai) data provider into the OpenBB Platform.
+
+## Overview
+
+QVERISAI is a third-party API tool platform that provides access to various tools for retrieving and processing data in fields such as finance, economics, healthcare, sports, scientific research, and more.
+
+## Installation
+
+To install the extension:
+
+```bash
+pip install openbb-qveris
+```
+
+## Configuration
+
+To use QVERISAI, you need to set your API key as an environment variable:
+
+```bash
+export QVERIS_API_KEY=your_api_key_here
+```
+
+Or in Python:
+
+```python
+import os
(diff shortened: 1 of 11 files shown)


deeleeramone:
Hi thanks for the PR, but we're going to have to pass on this. It's not appropriate content for this repository, or something we can take ownership of. However, you should build and publish this as a standalone extension from your own public repository when you have a complete implementation working. It seems like this is supposed to be a MCP server..?

lxltbds:
Thanks for the thoughtful review and the clear guidance — I understand the decision and appreciate the feedback.

Yes, you're right: this is intended to be an MCP server / external integration rather than something the core repository would own. My original intent was to explore whether tighter integration made sense, but I agree that a standalone extension maintained independently is the more appropriate path.

I'll proceed with building and publishing this as a separate public repository once the implementation is complete and stable. If at that point it makes sense to reference or list it as a community extension (or similar), I'd be happy to align with whatever process you recommend.

Thanks again for taking the time to review the PR — and for pointing me in the right direction.

PR #7322, opened 2026-01-13
[Security] `openbb-core`: Regenerate Lock File

deeleeramone:
This PR updates the `poetry.lock` file for `openbb-core` so that it resolves to the patched version of `urllib3`, as recommended by Dependabot. We also need to pin Pandas to not use V3 as that comes with breaking changes we are not ready to handle yet.

There are no changes to the code.

Additionally, patches the other Dependabot items:
- openbb-devtools
- OPD Desktop
  - test mock needed to be updated as a result of the higher package version.


biplavbarua:
LGTM. Standard lockfile regeneration. verified no suspicious package additions.

piiq (approved):

PR #7311, opened 2026-01-06
[Feature] PyWry 2.0

deeleeramone:
This PR overhauls  the `pywry` package, and brings it into this repository. There actually is no shared code, but the spirit lives on. It renders content across native OS windows, iFrames, Jupyter, and as a headless app.

There are a lot of details, so please see the README.md & AGENTS.md files for information. Check out the notebooks in the ./examples folder for short demos.

You can install and test this directly with `pip install -e ".[dev]"`.

There are unit and integration tests with full E2E validation of the event system and rendering.

Here's an example output that is made using the Toolbar system to produce all the content with no Javascript or raw HTML. All components are Pydantic models.

Screenshot 2026-01-27 at 9 25 08 AM

PR #7324, opened 2026-01-17
Fix(PyWry): Address Windows CI failures in PR #7311

biplavbarua:
# Walkthrough: PyWry PR #7311 Investigation & Improvements

## Overview
I investigated the Windows CI failure in PR #7311, which appears to be caused by resource contention or timing issues when initializing the WebView2 process (common on Windows runners).

### 1. Robust Retry Logic
I identified that `retry_on_subprocess_failure` was only used in `tests/test_e2e.py`. I extracted this decorator to a shared utility and applied it to `tests/test_window_modes.py`, which performs similar window lifecycle operations and was likely unprotected from flakes.

#### New Utility: `tests/utils.py`
```python
def retry_on_subprocess_failure(max_attempts: int = 3, delay: float = 1.0) -> Callable[[F], F]:
    """Retry decorator for tests that may fail due to transient subprocess issues."""
    ...
```

#### Updated: `tests/test_window_modes.py`
Applied `@retry_on_subprocess_failure()` to all test methods involving `PyWry` instantiation and window creation.

### 2. Stress Testing
I created a stress test script `tests/stress_test.py` to simulate rapid window creation and destruction.
- **Results:** 20/20 iterations passed successfully on macOS.
- **Purpose:** Verifies that the underlying logic is sound on stable platforms, reinforcing that the CI failure is environment-specific (Windows resource lock) and best addressed by the retry mechanism.

### MacOS Local Tests
- **Stress Test:** Passed (20 iterations)
- **Full Test Suite:** Passed (All tests in `tests/`)

### Recommendation
The changes to use `retry_on_subprocess_failure` in `test_window_modes.py` should significantly improve CI stability on Windows.


deeleeramone:
Thanks, but this is handled.

Issue #6969, opened 2024-11-27
[IMPROVE] `obb.equity.screener`: Make Input Of Country & Exchange Uniform Across Providers


deeleeramone:
In the `economy` module, the `country` parameter can be entered as names or two-letter ISO codes. The same treatment should be applied to the `obb.equity.screener` endpoint. Additionally, the "exchange" parameter should reference ISO MICs - i.e, XNAS instead of NASDAQ, XNYS instead of NYSE, etc.

ChimdumebiNebolisa:
I looked into how this is handled in the economy modules and how `obb.equity.screener` currently works, and I think this can be implemented cleanly and in a backward-compatible way.

## Country parameter

In the economy / IMF providers, country inputs are normalized via helpers like `resolve_country_input` (in `openbb_imf/utils/dot_helpers.py`), which accept both country names and ISO codes and resolve them to a canonical form.

For `obb.equity.screener`, I’d follow the same pattern:

- Allow `country` to accept either a two-letter ISO code (e.g. `US`) or a country name (e.g. `United States` / `united_states`).
- Reuse the same normalization logic (trim, case-fold, snake-case where applicable) and resolve names to ISO codes at validation time.
- If resolution fails, raise a clear validation error listing accepted formats.

## Exchange parameter

- Make `exchange` accept ISO MIC codes directly (e.g. `XNAS`, `XNYS`) as the canonical input.
- Preserve backward compatibility by continuing to accept existing exchange aliases (`nasdaq`, `nyse`, etc.) and mapping them internally to their MIC equivalents.
- Perform this mapping in the provider layer so downstream providers always receive a normalized MIC code.
- Update schema metadata / docs so MIC codes are the preferred and documented input going forward.

## Implementation details

- Implement validation and normalization in the equity screener query model (or `transform_query`), mirroring the approach used in IMF economy models.
- Keep provider-specific mappings (exchange aliases → MICs) within the FMP provider utilities.
- Add tests covering:
  - `country="United States"` vs `country="US"`
  - `exchange="NASDAQ"` vs `exchange="XNAS"`
  - Clear error messaging for unsupported values

## Open question before implementation

Should MIC codes be the preferred documented input with legacy exchange names treated as aliases, or do you want a stricter deprecation path?

If this approach aligns with your expectations, I’m happy to proceed with a PR.

pichpich16888-cpu:
[...]

PR #7333, opened 2026-01-26
feat(platform): standardize country inputs with ISO 3166 Country type

sandole:
## Description

- [x] Summary of the change: Add a custom `Country` type backed by a static ISO 3166 JSON dataset for standardizing country inputs across providers. Accepts alpha-2 codes (e.g., `US`), alpha-3 codes (e.g., `USA`), full names (e.g., `United States`), and `lower_snake_case` names (e.g., `united_states`). Normalizes all inputs to alpha-2 codes with access to full country metadata via properties.
- [x] Link: Partially addresses #6969
- [ ] Screenshot of the feature or the bug before/after fix, if applicable. (N/A - backend change)
- [x] Relevant motivation and context: Replaces hardcoded country mappings with a reusable type that providers can adopt for consistent country input handling.
- [x] List any dependencies that are required for this change: None — uses a static JSON dataset and `pydantic_core` (already a transitive dependency).

## How has this been tested?

- Unit tests for `Country` type validation (alpha-2, alpha-3, name, snake_case lookups, invalid input handling)

- If you modified/added command(s):
  - [ ] Ensure the command(s) execute with the expected output.
    - [ ] API.
    - [ ] Python Interface.
  - [x] If applicable, please add new tests for the command.

## Checklist

- [x] I ensure that I am following the [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBB/blob/main/CONTRIBUTING.md).
  - [x] (If applicable) I have updated tests following [these guidelines](/openbb_platform/CONTRIBUTING.md#qa-your-extension).


deeleeramone:
Thanks for thinking about this situation. Instead of hardcoding incomplete ISO country maps, I think we should consider using Pydantic to handle this programatically. It will do countries nicely, and probably other ISO code mapping with a bit of work. Provider modules could then supply their own `model_dump` and `model_serialize` functions that aligns the params and display values with the source's requirements. In essence, a standardized parameter type for `country`.

https://docs.pydantic.dev/latest/api/pydantic_extra_types_country/

For accepting short country names, these should be lower_snake_case values without spaces. 

For MIC and symbol mapping, OpenFIGI is the way to go - https://www.openfigi.com/, but this would require implementation as a provider module.

It's important to think of the overall, bigger, picture for library features because these need to be reusable components that implement the ISO.

sandole:
hey @deeleeramone.

thanks for your thoughtful comment :) happy to be first time contributor here!

I implemented this only for FMP to see if this approach for standardizing country inputs is acceptable. If so, I'll extend this to other providers and also add back the exchange standardization as well. I also added `pydantic-extra-types` as you requested.

deeleeramone:
Hey @sandole,

Thanks for putting that together, I like where this is going!

Among all the modules we have a mix of requirements for ISO2, ISO3, and short_name; they depend on a particular endpoint source. IMF has both ISO2 and ISO3, depending on which database (handled internally).  For most endpoints with a `country` parameter, there is conditional logic at the provider level that attempts to handle input variations as well as list/comma-separated string lists. I suggest not attempting to implement across all endpoints, but perhaps 1 - maybe 2 - as the evidence for "it works in real life", across all the providers servicing that endpoint. We could then iterate in subsequent PRs to update individual providers for their specific needs.

- I foresee edge cases in type checkers and input parameter type/validation enforcement where it gets rejected because the input string is not the correct type. What are your thoughts on a class that inherits from `str`? Something like:

```python
from pydantic import BaseModel
from pydantic_extra_types.country import CountryAlpha2
from typing import Any
import pycountry


class Country(str):
    """Country string with full model access"""
    
    def __new__(cls, value: Any):
        if isinstance(value, Country):
            country_obj = value._country
        else:
            # Use pycountry directly for validation and normalization
            country_obj = pycountry.countries.lookup(value)
        
        # Create str instance with the alpha_2 code
        instance = super().__new__(cls, country_obj.alpha_2)
        # Store the pycountry object
        instance._country = country_obj
        return instance
    
    @property
    def alpha_2(self) -> str:
        return self._country.alpha_2
    
    @property
    def alpha_3(self) -> str:
        return self._country.alpha_3
    
    @property
    def name(self) -> str:
        return self._country.name
    
    @property
    def numeric(self) -> str:
        return getattr(self._country, 'numeric', None)
    
    @classmethod
    def __get_pydantic_core_schema__(cls, source_type, handler):
        from pydantic_core import core_schema
        return core_schema.no_info_after_validator_function(
            cls,
            core_schema.str_schema(),
        )

```

On the downside, it does add quite a bit of weight to the installation environment, over 6MB for the wheel. @piiq, is this size a concern for you?

piiq:
Thanks for asking @deeleeramone. My main concern is that core might become too "application specific" by pulling dependencies that are applicable solely to a subset of use-cases. For example - any crypto or regional workflow would not benefit from country name definition.
That's from the top of my head.
I don't oppose a 6mb dependency when it's needed, but being a person who tries to create extensions that don't utilize any provider integrations that odp has I personally would prefer a very lean core.

Could the country parameter come with an existing extension vs the core?

sandole:
Thanks @deeleeramone and @piiq for the thoughtful feedback.

I checked the package sizes:
- pydantic-extra-types is ~50KB (very lightweight, mostly type definitions)
- pycountry is the heavier one at ~6MB (contains full ISO database files)

Since pydantic-extra-types[country] pulls in pycountry as a dependency, the concern is valid. 

Here are the ways forward:
1. Keep as-is if 6MB is acceptable for the standardization benefit across multiple providers
2. Lazy import similar to [other parts in the codebase](https://github.com/OpenBB-finance/OpenBB/blob/develop/openbb_platform/core/openbb_core/app/static/package_builder.py#L57)
3. roll a minimal country mapping without the full ISO database 

I'm happy to go in any direction:
- Core: if we think this will be reused across 3+ providers
- Extension: if we want to keep it isolated until adoption proves the value

@piiq @deeleeramone - what's your preference? I'm personally leaning towards adding the full dependency for allowing future developments using pydantic-extra-types (pretty useful imo)

deeleeramone:
> @piiq @deeleeramone - what's your preference? I'm personally leaning towards adding the full dependency for allowing future developments using pydantic-extra-types (pretty useful imo)


Perhaps we could do something similar to what `pydantic-extra-types` does, checks for the availability of `pycountry` and fails gracefully with a backup - potentially just the raw input value and the special class would handle cleaning up the non-existent methods.

A possible backup could be to manually export the rows of all country objects in `pycountry` and store as a compressed JSON asset. This would certainly be smaller than 6MB in this form factor because it wouldn't have all the non-country stuff, and would be a static drop-in replacement. The dependency would rely on the individual provider extension requiring `pycountry` instead of the core library.

Another option is to place this type of developer tooling into a `openbb-types` package with package extras for each supported type - i.e. "country". 

What are your thoughts on the viability of this, @piiq and @sandole?

sandole:
Thanks for the suggestions @deeleeramone.

I considered the compressed JSON approach but went with the simpler approach of a plain JSON file with the ISO 3166-1 country data.

Why not compression:
- The raw data is only ~20KB (249 countries × 4 fields)
- No decompression logic/overhead needed

This keeps the core lean without any additional external dependency. Wanted to show how this would look although I personally think adding another 30KB for the extra types dependency is totally fine as well.

deeleeramone (approved):
Thank you for your contribution to the project. This is a maintainable implementation that covers a lot of use-cases, including ISO2 <-> ISO3 interoperability, display name mapping, and input normalization.

If we want to go farther, we could add to the JSON with items like regional and "membership" groups - "G7", "G20", "UN", "IMF", "OPEC", etc.

sandole:
No problem and thank you for the reviews!

Yup membership sounds like a good addition - I also still need to add the exchanges to finish addressing #6969.

PR #7335, opened 2026-01-26
[Security] Re-lock Packages For Protobuf Dependabot Alert 

deeleeramone:
This PR resolves the open Dependabot alerts. A transitive dependency in `openbb-yfinance` is resolved by regenerating the lock file.

No changes to the code.


piiq (approved):

deeleeramone:
Looks like I misread it, affected versions: <= 6.33.4, patched version: none. So this won't actually resolve itself until upstream publishes.

PR #7336, opened 2026-01-27
[Feature] Pandas 3.0 - Update Library For Syntax

deeleeramone:
This PR addresses all breaking Pandas syntax related to 2.3 -> 3.0, and unpins the version from `openbb-core`.

Notable syntax handling changes:

- `read_html` no longer accepts raw string values,  must pass a buffer.
- String-type columns are no longer of `dtype: object`, they have a new explicit type.
- Assign by index no longer supports changes in `dtype`, column must be explicitly assigned.
- `.applymap()` is no longer valid, it is now `map()`.

You can test this by running provider unit test suite off the develop branch with Pandas 3.0 installed.


piiq (approved):
Is this backwards compatible with earlier versions of pandas? I assume a lot of folks haven't yet migrated to v3

deeleeramone:
> Is this backwards compatible with earlier versions of pandas? I assume a lot of folks haven't yet migrated to v3

It should be, although some V1 syntax might be out of range by now.

PR #7337, opened 2026-01-27
[BugFix] Remove Polygon - No Longer Exists and Can't Support

deeleeramone:
This PR removes the Polygon provider from this repository.

Reason: No longer exists - https://polygon.io


piiq (approved):
f

strycker:
Polygon.io is now Massive, and it does still exist.  See https://github.com/massive-com

deeleeramone:
> Polygon.io is now Massive, and it does still exist. See https://github.com/massive-com

Yes, however, it will still require a complete refactor which is impossible to do and test without full API access. For these reasons, among others, vendor data extensions will be phased out and more focus placed on public data resources.

You are welcome to take this extension code and make it your own! Anyone can build and publish standalone extensions.

PR #7340, opened 2026-01-30
Add github action that validates all changes have a formal description

piiq:
This action will help us provide compliance evidence that each change is formally documented


deeleeramone (approved):
LGTM!

PR #7345, opened 2026-02-06
[BugFix] ODP Desktop: Fix Dependabot Alerts, Fix Environment Update, Improve Error Message

deeleeramone:
### This PR is for the ODP Desktop App

- Resolves GHSA-r6v5-fh4h-64xc and GHSA-434x-w66g-qw3r
- Updates package dependencies - `glob`, and some Tauri components.
- Improves error message on initial environment failure.
- When the initial environment setup fails, the Backend configurations will not be added.
- Change the `Update Environment` button to update packages for Conda and PyPI separately instead of using `conda env update`
- We no longer need to install `openbb --no-deps` for the build package, this is included in `openbb-core`.

To test this, you can invalidate your current installation by removing the "install_settings" entry from the `~/.openbb_platform/system_settings.json` file (or uninstall completely) and then proceeding to install again. Add some fake package from PyPI to force an error.

Screenshot 2026-02-05 at 9 32 26 PM

### Longer term considerations

- Need to deprecate macOS Intel, many key libraries have stopped publishing pre-compiled binaries for x86, and it will become increasingly difficult to ship environments with Python packages that do not require a C++ compiler and escalated privileges.


piiq (approved):

pichpich16888-cpu (commented):
Sur sdey

PR #7347, opened 2026-02-08
[Chore] - Remove redundant bandit dependency

sandole:

1. **Why?**

    - Bandit is redundant — ruff's `S` (flake8-bandit) rules already cover the same security checks and are enabled in `ruff.toml`.
    - Bandit was running as non-blocking (`|| true`) in CI, so it was purely informational and couldn't catch regressions.

2. **What?**

    - Removed the `bandit -x ./tests -r . || true` step from the CI linting workflow.
    - Removed `bandit` from `openbb-devtools` dependencies in `pyproject.toml`.

3. **Impact**

    - One fewer dependency to install in CI, slightly faster linting pipeline.
    - No change in security coverage — ruff's 94 stable S rules continue to enforce the same checks.
    - Impact score: 1

4. **Testing Done**

    - Verified `ruff check --select S` passes across `openbb_platform/` and `cli/`.
    - Confirmed the S rule group was already in `ruff.toml`'s `select` list.

5. **Reviewer Notes**

    - This is a pure dependency/CI cleanup — no Python source files are changed.


deeleeramone (approved):
Yes, good call. Thanks for your help.

PR #7355, opened 2026-02-12
feat(platform): extend Country type standardization to equity screeners

sandole:
# Pull Request for OpenBB Platform

## Description

- [x] **Summary:** Extend the standardized `Country` type from PR #7333 to equity screeners and calendar endpoints.
- [x] **Link:** Continues work from #7333, addresses #6969
- [x] **Motivation:** Provide consistent country input handling across equity and calendar endpoints.

### Changes

**YFinance equity_screener:**
- Accept ISO alpha-2/alpha-3 codes, country names, or snake_case
- Validates against YFinance's supported country list
- Converts to lowercase alpha-2 format expected by YFinance

**Nasdaq equity_screener:**
- Accept ISO codes, country names in addition to existing snake_case
- Preserves multi-value (comma-separated) functionality
- Converts all inputs to Nasdaq's snake_case format

**Nasdaq economic_calendar:**
- Accept ISO codes, country names in addition to existing snake_case
- Preserves multi-value (comma-separated) functionality

**Seeking Alpha calendar_earnings:**
- Accept ISO codes, country names (normalizes to 'us' or 'ca')
- Validates against supported countries (US, Canada only)

### Example Usage

```python
# YFinance example:
obb.equity.screener(country='US', provider='yfinance')
obb.equity.screener(country='USA', provider='yfinance')
obb.equity.screener(country='United States', provider='yfinance')
obb.equity.screener(country='united_states', provider='yfinance')
```

There are many other endpoints across providers that use country parameters:

| Scope | Providers/Files | Complexity |
|-------|-----------------|------------|
| **This PR** | YFinance, Nasdaq screeners, Nasdaq calendar, SA earnings | Low |
| **Extended** | FMP etf_search | Low-Medium |
| **Full** | OECD (6 files), EconDB (4), IMF (6), FRED (2) | High (custom mappings) |

**Should this be extended further?** The full approach would require careful handling of provider-specific country mappings (e.g., OECD uses custom codes, IMF has its own country identifiers).

Happy to expand scope based on feedback.

---

- [x] I have read the [Contributing Guidelines](https://github.com/OpenBB-finance/OpenBB/blob/develop/CONTRIBUTING.md)
- [x] Update tests to reflect changes
- [x] All tests pass locally


deeleeramone:
This looks good, nice!

@sandole, we might as well do `FMP etf_search` while we're here. It's basically the same as function as `FMP equity_screener`, copy & paste should work no problem there. I would hold off on OECD, there's some technical debt in that provider I would like to eradicate and eventually refactor the entire thing so that it ingests from all databases across the SDMX API using hierarchical data availability constraints for validating dimension and parameter values.

openbb_platform/providers/yfinance/openbb_yfinance/models/equity_screener.py
```diff
@@ -98,13 +100,34 @@ class YFinanceEquityScreenerQueryParams(EquityScreenerQueryParams):
...
+        if v is None or v == "all":
+            return v
+        # Convert Country type or string to lowercase alpha-2 for YFinance
+        if isinstance(v, Country):
+            country_code = v.alpha_2.lower()
+        else:
+            try:
+                country_code = Country(v).alpha_2.lower()
+            except ValueError:
+                return v  # Let it fail at API level if invalid
```
deeleeramone: If the value fails to parse as a Country, it likely is not a valid string. Instead of letting it through to the API call, we should validate the value against the list of accepted countries.

```suggestion
                country_code = v.strip().lower()
```

Without the raised ValueError, the user gets returned an EmptyDataError instead, which is missing the helpful message with the valid options.

openbb_platform/providers/yfinance/openbb_yfinance/models/equity_screener.py
```diff
@@ -98,13 +100,34 @@ class YFinanceEquityScreenerQueryParams(EquityScreenerQueryParams):
...
+        if country_code not in COUNTRIES:
+            raise ValueError(
+                f"Country '{v}' ({country_code.upper()}) is not supported by YFinance. "
+                f"Valid options: {', '.join(sorted(COUNTRIES)[:20])}...",
```
deeleeramone: For clarity, we can also remove the limit on displaying the valid countries in the error message.

```suggestion
        if country_code not in COUNTRIES:
            raise ValueError(
                f"Country '{v}' ({country_code.upper()}) is not supported by YFinance. "
                f"Valid options: {', '.join(sorted(COUNTRIES))}",
            )
```

deeleeramone (approved):
Thanks for your contribution to the project and help with driving things forward!

PR #7361, opened 2026-02-16
[Feature] `openbb-sec`: Custom HTML2Markdown Conversion

deeleeramone:
This PR refactors the `sec` provider to use a custom HTML -> Markdown converter instead of relying on multiple libraries as strategies and cleaning up after them.

It removes dependencies:
- inscriptis
- trafilatura

The new module, `openbb_sec.utils.html2markdown.html_to_markdown`, is geared specifically for SEC HTML content. It has a much higher success rate - especially where table extraction is concerned - and handles a number of scenarios and filing formats. The existing, `obb.equity.fundamental.management_discussion_analysis()`, is also better equipped to handle more scenarios:

- Section is in a different part of the same document, or is attached as an exhibit.
- Content is heavily nested.
- 2X column wide-page layouts.
- Embedded images.
- Presentation slides and info graphics.
- Foreign filers (20-F, 40-F, 6-K).
- Earnings release as 8-K press-release before submitted as 10-K/Q

mda-jpm

Screenshot 2026-02-16 at 3 01 57 PM

You can A/B this against the existing (use `include_tables=True`) with symbols such as:
- DE
- MSFT
- IBM
- JPM
- GS
- MS
- WMT
- NKE
- NDAQ
- CRM
- XOM
- CVX
- BRK-A
- SPOT
- BMO
- CVX
- LOW

image


openbb_platform/providers/sec/openbb_sec/utils/html2markdown.py
```diff
@@ -0,0 +1,5815 @@
...
+        if height:
+            style_parts.append(f"height:{height}")
+        style_str = ";".join(style_parts)
+        return f'<img src="{src}" alt="{alt}" style="{style_str}"/>'
+
+    # No size info, use standard markdown
+    return f"![{alt}]({src})"
+
+
+def html_to_markdown(
```
piiq: @deeleeramone can we reduce the cognitive complexity of this parsing algorithm? IIUC, currently this is a 4000 line function.

piiq (changes requested):
To make this readable/understandable by both humans and AI I suggest we split the huge converter function apart. I understand the complexity of what it's doing, but having that function be that large is a risk it won't be understood or read correctly by any party reading it

deeleeramone:
@piiq, `html_to_markdown` and `convert_table` have been reduced to under 1000 lines each. Most of the recursive inner functions have been updated to use module-level functions that pass the necessarily values into the enclosure.

deeleeramone:
The tests failing are from unrelated package versions, and are handled in #7366

piiq (approved):

PR #7352, opened 2026-02-11
feat(platform): add membership groups to Country type

sandole:
# Pull Request for OpenBB Platform

## Description

- [x] **Summary:** Add membership groups (G7, G20, EU, NATO, OECD, OPEC, BRICS) to the `Country` type, as suggested by @deeleeramone in PR #7333.
- [x] **Link:** Follow-up to #7333, addresses suggestion from [this comment](https://github.com/OpenBB-finance/OpenBB/pull/7333#issuecomment-2615137913)
- [x] **Motivation:** Enable filtering and grouping of countries by economic/political membership for macroeconomic analysis.

### Features

Extends the `Country` type with membership group data:

```python
from openbb_core.provider.utils.country_utils import Country

c = Country("Germany")
c.groups           # ["G7", "G20", "EU", "NATO", "OECD"]
c.is_member_of("EU")  # True
c.is_member_of("OPEC")  # False

c = Country("Saudi Arabia")
c.groups           # ["G20", "OPEC", "BRICS"]
c.is_member_of("OPEC")  # True
```

### Data Sources

| Group | Reference |
|-------|-----------|
| ISO 3166-1 | https://en.wikipedia.org/wiki/ISO_3166-1 |
| G7 | https://en.wikipedia.org/wiki/G7 |
| G20 | https://en.wikipedia.org/wiki/G20 |
| EU | https://european-union.europa.eu/principles-countries-history/eu-countries_en |
| NATO | https://www.nato.int/en/about-us/organization/nato-member-countries |
| OECD | https://en.wikipedia.org/wiki/OECD |
| OPEC | https://en.wikipedia.org/wiki/OPEC |
| BRICS | https://en.wikipedia.org/wiki/BRICS |

### Developer Script

A developer script (`update_country_data.py`) is included in the same directory as the JSON to re-source membership data from authoritative sources and regenerate `country_data.json`. This avoids manual JSON construction and ensures data integrity when memberships change.

**Usage:**
```bash
pip install pycountry beautifulsoup4 requests

# Preview without writing
python update_country_data.py --dry-run

# Regenerate country_data.json
python update_country_data.py
```

**Example run:**
```
🌍 Updating country_data.json

Scraping G7... ✅ 7 members
Scraping G20... ✅ 19 members
Scraping EU... ✅ 27 members
Scraping NATO... ✅ 33 members
Scraping OECD... ✅ 38 members
Scraping OPEC... ✅ 13 members
Scraping BRICS... ✅ 10 members

[...]

PR #7368, opened 2026-02-21
fix(deribit): replace alru_cache with dict cache to fix cross-event-loop RuntimeError

sandole:
## Summary

Replace `@alru_cache` on `get_instruments()` with a simple dict cache.

`alru_cache` from `async_lru>=2.0` binds to the event loop on first use and raises `RuntimeError: alru_cache is not safe to use across event loops` when called from a different loop. This happens in tests because `fetcher.test()` / `run_async()` creates a new event loop per test function.

These test failures are blocking #7352 and wanted to scope the unrelated test failures outside.

`currency` has 6 values, `derivative_type` has 5 and `expired` is a bool so that's a total of 6 x 6 x 2 = 72 maximum combos - meaning no memory concern for the dict in production. I see there's a `ttl_cache` in `openbb_core` so I can look to utilize that instead.

## Changes

- **helpers.py**: Replace `@alru_cache(maxsize=64)` with a module-level dict cache that is event-loop agnostic
- **pyproject.toml**: Remove `async-lru` dependency (no longer used)

## Tests fixed

These 4 tests fail on every PR against `develop`:

- `test_deribit_futures_curve_fetcher`
- `test_deribit_futures_historical_fetcher`
- `test_deribit_futures_instruments_fetcher`
- `test_deribit_futures_info_fetcher`

### DIFF
--- a/openbb_platform/providers/deribit/openbb_deribit/utils/helpers.py
+++ b/openbb_platform/providers/deribit/openbb_deribit/utils/helpers.py
@@ -2,7 +2,6 @@
 
 from typing import Literal
 
-from async_lru import alru_cache
 from openbb_core.app.model.abstract.error import OpenBBError
 
 DERIBIT_OPTIONS_SYMBOLS = ["BTC", "ETH", "SOL", "XRP", "BNB", "PAXG"]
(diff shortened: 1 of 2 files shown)


deeleeramone:
This is actually just the test fixture, dealt with here https://github.com/OpenBB-finance/OpenBB/pull/7366/changes/5fbf500b006742ab0d731d5aecb5f8824aec2b7e

PR #7351, opened 2026-02-11
feat(platform): standardize exchange inputs with ISO 10383 MIC codes

sandole:
# Pull Request for OpenBB Platform

## Description

- [x] **Summary:** Add an `Exchange` type that standardizes exchange inputs across providers using ISO 10383 Market Identifier Codes (MIC).
- [x] **Link:** Partially addresses #6969
- [x] **Motivation:** Consistent with the `Country` type approach from PR #7333, this provides uniform exchange input handling across all providers.

### Features

Following the same pattern as `Country`, the `Exchange` type:
- Inherits from `str` (storing the MIC code) for type checker compatibility
- Accepts multiple input formats:
  - ISO 10383 MIC codes (e.g., `XNAS`, `XNYS`)
  - Exchange acronyms (e.g., `NASDAQ`, `NYSE`)
  - Full exchange names (e.g., `New York Stock Exchange`)
  - lower_snake_case names (e.g., `new_york_stock_exchange`)
- Provides properties: `mic`, `acronym`, `name`, `city`, `country`, `website`
- Uses `ExchangeData` TypedDict for proper type inference
- Includes enriched metadata from ISO 10383: city (title-cased), country (ISO alpha_2), and website (normalized with https://)
- 2,272 active MICs with 100% city/country coverage and 98% website coverage

### Example Usage

```python
from openbb_core.provider.utils.exchange_utils import Exchange

e = Exchange("nasdaq")
str(e)      # "XNAS"
e.mic       # "XNAS"
e.acronym   # "NASDAQ"
e.name      # "NASDAQ - ALL MARKETS"

e = Exchange("XNYS")
e.mic       # "XNYS"
e.acronym   # "NYSE"
e.name      # "NEW YORK STOCK EXCHANGE, INC."
```

### Sample JSON Entry

```json
{
  "mic": "XNYS",
  "acronym": "NYSE",
  "name": "NEW YORK STOCK EXCHANGE, INC.",
  "city": "New York",
  "country": "US",
  "website": "https://www.nyse.com"
}
```

### Files Added/Modified

- `openbb_core/provider/utils/exchange_data.json` — Full ISO 10383 dataset (2,272 active MICs) with city, country (ISO alpha_2), and website fields
- `openbb_core/provider/utils/exchange_utils.py` — Exchange type implementation with `ExchangeData` TypedDict
- `tests/provider/utils/test_exchange_utils.py` — Unit tests (13 test cases)
[...]

PR #7370, opened 2026-02-23
feat: add coingecko provider and symbol normalization utilities

AhmedAli58:
# Pull Request OpenBB

Please go to the `Preview` tab and select the appropriate PR sub-template:

* [OpenBB Platform](?expand=1&template=platform_pull_request_template.md)
* [OpenBB Platform CLI](?expand=1&template=terminal_pull_request_template.md)
* [OpenBB Developers](?expand=1&template=obb_developer_pull_request_template.md)


deeleeramone:
Hi, thanks for the PR. The `SymbolMap` concept is interesting, but there are a few things to note.

A good resource for this subject can be found here: https://support.quotemedia.com/support/solutions/articles/13000020471-quotemedia-symbology-document

- `openbb-core` code is provider-agnostic, hard-coding individual providers and their edge cases here is not scalable or easily maintained.
- `openbb-polygon` has been dropped from this repository - #7337 - and why only handle those?
- An implementation for this type of concept would require full coverage across all asset classes and conventions - the reusable component would need to be standardized (#7351 ) and then implemented by the individual provider to handle its specific requirements.

Regarding the CoinGecko provider, we would prefer people host and publish their implementations of vendor APIs as standalone extensions.

PR #7380, opened 2026-02-27
[Feature] `openbb-mcp-server`: Refactor for FastMCP V3

deeleeramone:
This PR refactors the `openbb-mcp-server` code to use "fastmcp>=3".

With these changes, the MCP server can be installed by itself and provide useful resources that aide in developing OpenBB Python packages, Workspace apps, and MCP servers.

Breaking changes from the new version include:
- import patterns
- class names and methods

V3 reduces the amount of custom code required of us by introducing a new concept, [Transforms](https://gofastmcp.com/servers/transforms/transforms).

## Behavioral Changes

Server startup and connecting metadata is now extremely minimal, the agent must explicitly activate a tool, or category of tools, in order to get the schema and full metadata.

### Tool Discovery

Tool discovery is forced by disabling all tool categories on startup, leaving only the admin tools for listing categories, unless configured otherwise.

This means that the MCP server starts with only 10 tools instead of 300+.

It forces the connecting agent to sip from the fountain instead of drowning by firehose.

Tool visibility changes are **per-session**: each connected client has its own active toolset.
Multiple agents can connect simultaneously and independently activate different tools without interfering with each other.

To provide a backstop for large lists of tools that may bloat the context, we are adding the new FastMCP feature that allows pagination and size configuration.

| Setting | Environment Variable | Type | Default | Description |
|---|---|---|---|---|
| `list_page_size` | `OPENBB_MCP_LIST_PAGE_SIZE` | integer | `None` | Max items per page in MCP list responses. `None` disables pagination. |

**Note**: These apply to the built-in FastMCP `list` functions for tools, prompts, and resources. `list_tools` will only return active tools, which is why we have `available_tools`.

### `ToolRegistry` Replaced by `CategoryIndex` 

CategoryIndex is the complete list of tool names by category and subcategory. It stores the short description for discovery and tracks it's state as a boolean. Its job is to help the connecting agent understand the distribution of tools amongst categories.

Use `available_tools('category')` to get short descriptions and summaries.

With no parameters, it supplies the hierarchy and the tool count.

```json
[{"name":"commodity","subcategories":[{"name":"general","tool_count":6},{"name":"price","tool_count":1}],"total_tools":7},...]
```

```json
{
  "category": "equity",
  "subcategory": "price"
}
```

```json
[{...}]
{
  "result": [
    {
      "name": "equity_price_historical",
      "active": true,
      "description": "Get historical price data for a given stock. This includes open, high, low, close, and volume."
    },
    {
      "name": "equity_price_performance",
      "active": false,
      "description": "No description available"
    },
    {
      "name": "equity_price_quote",
      "active": true,
      "description": "Get the latest quote for a given stock. Quote includes price, volume, and other data."
    }
  ]
}
```

## Skills

This PR implements the Skills Provider feature introduced in V3 of FastMCP - https://gofastmcp.com/servers/providers/skills

The server ships with a set of bundled **skill guides** — Markdown documents that teach an agent how to perform complex multi-step tasks with the OpenBB Platform.
Skills are exposed as MCP resources and are discoverable via `list_resources()`. They can be installed to, or read from, any FastMCP-supported provider, or to the server.

Each skill is available at a URI of the form `skill://<name>/SKILL.md`.

### Bundled Skills

| Skill | URI | Description |
|---|---|---|
| `develop_extension` | `skill://develop_extension/SKILL.md` | Step-by-step guide for building an OpenBB Platform extension. |
| `build_workspace_app` | `skill://build_workspace_app/SKILL.md` | Guide for building and running OpenBB Workspace applications. |
| `configure_mcp_server` | `skill://configure_mcp_server/SKILL.md` | Reference for configuring and customising the OpenBB MCP Server. |
| `work_with_server` | `skill://work_with_server/SKILL.md` | Practical guide for working with the OpenBB MCP Server as an agent. |

When any skills are loaded and no `system_prompt_file` is configured, the server automatically adds a brief default system prompt that nudges the agent to discover available skills.

### Skill Settings

| Setting | Description |
|---|---|
| `default_skills_dir` | Path to the bundled skills directory. Set to `null` or an empty string to disable loading the built-in skills. |
| `skills_reload` | Set to `true` to reload skill files from disk on every read — useful when authoring or iterating on skill content. |
| `skills_providers` | A list of vendor skill provider short-names. Supported values: `claude`, `cursor`, `vscode`, `copilot`, `codex`, `gemini`, `goose`, `opencode`. |

**Example — disable bundled skills:**

```json
{
  "default_skills_dir": null
}
```

**Example — load vendor skill providers:**

```json
{
  "skills_providers": ["claude", "cursor"]
}
```

**Example — enable skill reload during development:**

```env
OPENBB_MCP_SKILLS_RELOAD=true
```

## Tests

Tests have been updated to reflect the new reality and added to for increased coverage.


openbb_platform/extensions/mcp_server/openbb_mcp_server/skills/build_workspace_app/SKILL.md
```diff
@@ -1,3 +1,7 @@
+﻿---
+description: Build and Run OpenBB Workspace Applications
```
DidierRLopes: there should be a name key with slug, like name: build-workspace-app

and the description should be more descriptive as that's what the agnet will use to pick up what this skill does!

same for others

deeleeramone: Done!

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
```diff
@@ -605,16 +711,183 @@ async def execute_prompt(
...
 
-        return (
-            await mcp._prompt_manager.render_prompt(  # pylint: disable=protected-access
-                name=prompt_name, arguments=arguments
-            )
+        return await mcp.render_prompt(
+            name=prompt_name, arguments=arguments
         )  # type: ignore
 
+    # Resource discovery tools
```
MagnusS0: I think we can replace these custom list_resources/read_resource wrappers with FastMCP ResourcesAsTools transform to reduce custom maintenance.

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
```diff
@@ -556,11 +663,10 @@ def deactivate_tools(
     @mcp.tool(tags={"prompt"})
```
MagnusS0: We can replace custom prompt tools with PromptsAsTools for v3

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
```diff
@@ -45,6 +59,17 @@
 
 logger = get_logger(__name__)
 
+_VENDOR_SKILLS_PROVIDERS = {
+    "claude": ClaudeSkillsProvider,
+    "cursor": CursorSkillsProvider,
+    "vscode": CopilotSkillsProvider,
```
MagnusS0: This should be VSCodeSkillsProvider

MagnusS0 (commented):
I think this is great addition to the MCP server!
I've added some comments on some places I think there is some v3 changes that means we need less custom code.

I also started on CodeMode #7385
CodeMode is still not GA, but it can be built as a follow up to this.
That could also allow us to remove the ToolRegistry and the admin tools.
Can also use the new FastMCP search transforms (search_tools + call_tool) so there is no need for monty and code excecution.
At that point there are also some updates to the SKILLs that should be in place.

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
deeleeramone: I'll take a look at this again. I was having trouble getting VS Code to find/get resources without an explicit tool.

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
deeleeramone: Nice. Good catch.

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
deeleeramone: Done.

openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
deeleeramone: Got it, done!

deeleeramone:
@DidierRLopes, @MagnusS0, all items addressed - allegedly. Behavioral changes have been noted in the updated PR description; now, only discovery tools are enabled on connect and the agent should not receive massive context without requesting it.

MagnusS0 (commented):
This LGTM now, great work!

I will rebase on top of this and can point my PR with CodeMode to merge to this one? 
If that makes sense @deeleeramone

deeleeramone:
> I will rebase on top of this and can point my PR with CodeMode to merge to this one?

What do you estimate the CodeMode GA to be? Your PR was already merged, so that'll work its way through the dependency chain.

MagnusS0:
> What do you estimate the CodeMode GA to be? Your PR was already merged, so that'll work its way through the dependency chain.
> 
> 
> 
> 

It is already released in the latest 3.1.0v today!

I've started on updating so it matches that release. Hope to get it done tomorrow or Thursday.

piiq (approved):
Does it support setting CORS?

deeleeramone:
> Does it support setting CORS?

system_settings.json

```json
{"api_settings": {"cors": { ... }}}
```

PR #7382, opened 2026-02-27
fix(provider_interface): add __getattr__ to resolve OBBject_* dynamic types

### LINKED ISSUE #7379: ImportError: OBBject_EquityInfo not found in provider_interface (v4.6.0)
## Bug Description

```python
from openbb import obb
```

```python
    OBBject_EquityInfo,
    OBBject_EquityScreener,
    ...
)
```

## Environment
- Python 3.12.3 (macOS arm64)

## Root Cause

## Fix

```python
_cached_annotations = None

def __getattr__(name: str):
        global _cached_annotations
            pi = ProviderInterface()
```

## Workaround

zincsamael:
## Summary

Fixes #7379.

Package builder generates imports like:

```python
from openbb_core.app.provider_interface import OBBject_EquityInfo
```

But `OBBject_*` types are created dynamically inside `ProviderInterface.return_annotations` and never registered as module-level attributes on `openbb_core.app.provider_interface`. This causes `ImportError` on fresh installs.

## Fix

Add a module-level `__getattr__` hook that lazily resolves `OBBject_*` names by calling `ProviderInterface().return_annotations` on first access. This matches the import pattern the package builder already generates, with no changes to the builder itself.

```python
_provider_interface_annotations: dict = {}

def __getattr__(name: str):
        global _provider_interface_annotations
        if not _provider_interface_annotations:
            _provider_interface_annotations = ProviderInterface().return_annotations
        if model_name in _provider_interface_annotations:
            return _provider_interface_annotations[model_name]
```

## Testing

Verified locally on macOS arm64, Python 3.12, openbb 4.6.0 + openbb-core 1.6.0:

```python
from openbb import obb
obb.equity.price.quote("AAPL", provider="yfinance")  # ✅ no ImportError
```

### DIFF
--- a/openbb_platform/core/openbb_core/app/provider_interface.py
+++ b/openbb_platform/core/openbb_core/app/provider_interface.py
@@ -699,3 +699,22 @@ class ProviderInterface(metaclass=SingletonMeta):
                 __doc__=f"OBBject with results of type {name}",
             )
         return annotations
+
+
+# Fix for issue OpenBB-finance/OpenBB#7379
+# OBBject_* types are created dynamically in return_annotations but never
+# registered as module-level attributes, causing ImportError on:
+#   from openbb_core.app.provider_interface import OBBject_EquityInfo
+_provider_interface_annotations: dict = {}
+
+
+def __getattr__(name: str):
+    """Lazily resolve dynamically created OBBject_* types."""
+    if name.startswith("OBBject_"):
+        global _provider_interface_annotations
+        if not _provider_interface_annotations:
+            _provider_interface_annotations = ProviderInterface().return_annotations
+        model_name = name[len("OBBject_"):]
+        if model_name in _provider_interface_annotations:
+            return _provider_interface_annotations[model_name]
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")


deeleeramone:
Hi, thanks for the PR. However, the actual problem you are solving for is the result of a static asset build that was not completed. These classes are temporarily constructed during build to facilitate signature reading and the discriminator union tags within the ProviderInterface. **They do not exist in the final output.**

The fix is to run `openbb-build` again from the command line with the environment active.

deeleeramone:
PackageBuilder will build the final output like:

```python
"""Autogenerated OpenBB /equity/price Module."""

### THIS FILE IS AUTO-GENERATED. DO NOT EDIT. ###

#  pylint: disable=R0917,C0103,C0415

import datetime
from typing import Annotated, Literal

from openbb_core.app.model.field import OpenBBField
from openbb_core.app.static.container import Container
from openbb_core.app.static.utils.decorators import exception_handler, validate
from openbb_core.app.static.utils.filters import filter_inputs


class ROUTER_equity_price(Container):
    """/equity/price
    historical
    performance
    quote
    """

    def __repr__(self) -> str:
        return self.__doc__ or ""
...
```

deeleeramone:
Duplicate submission.

PR #7390, opened 2026-03-03
fix(cli): replace eval with ast.literal_eval in parse_unknown_args_to_dict

### LINKED ISSUE #7388: [Security] CLI results unknown-arg parser executes untrusted input via eval
## Summary
`parse_unknown_args_to_dict` evaluates untrusted CLI input with Python `eval`, which can execute arbitrary code.

## Affected code
- `cli/openbb_cli/controllers/utils.py:924`
- Called from `cli/openbb_cli/controllers/base_controller.py:545` (`results` command)

## Why this is problematic
Unknown arguments passed to `results` are parsed with:

```py
unknown_args_dict[arg.replace("--", "")] = eval(unknown_args[idx + 1])
```

This means a crafted argument value can run arbitrary Python code. Since `.openbb` routines can be executed from local files and URLs (`exe --url`), this can become a remote code execution path when users run untrusted routines.

## Reproduction idea
In CLI, pass an unknown value expression to `results`, e.g. a payload like:

```text
results --index 0 --foo __import__('os').system('touch /tmp/openbb_eval_poc')
```

If parsed as a single token/value, it will execute during unknown arg parsing.

## Expected behavior
Unknown arg values should be parsed as data only, not executed.

## Suggested fix
- Replace `eval(...)` with safe parsing:
  - `ast.literal_eval` (for Python literals), with fallback to raw string
  - or explicit typed parsing (`true/false/null/int/float/list/json`) without eval
- Add a regression test ensuring unknown arg parsing does not execute expressions.

T-Arjun:
eval() on unsanitized CLI input allowed arbitrary code execution via crafted --key payloads. ast.literal_eval is a safe drop-in that only parses Python literals and raises ValueError/SyntaxError on expressions. Also narrows the except clause and fixes the fallback key stripping bug.

Closes #7388


deeleeramone (approved):
Thanks for your help!

PR #7403, opened 2026-03-08
[BugFix] Remove unnecessary eval() calls in script_parser.py

Bortlesboat:
## Summary
- Replaced 5 `eval()` calls with direct Python equivalents (dict access, int indexing, slice construction)
- Follows the same pattern as #7390 which cleaned up eval() in utils.py
- Removed all `# noqa: S307` and `# pylint: disable=eval-used` suppressions
- Zero behavior change — all replacements are semantically identical

## Changes
- `eval(f'ROUTINE_VARS["{VAR_NAME}"]')` → `ROUTINE_VARS[VAR_NAME]`
- `eval(f"values[{VAR_SLICE}]")` → `values[int(VAR_SLICE)]`
- `eval(slicing_tuple)` → `slice()` construction


deeleeramone:
Hey, @Bortlesboat, thanks for the PR. Looks good to me, I think you just need to run `black` on the file and then it will pass the linting check.

Bortlesboat:
Applied the formatting fix on the current PR head and pushed 702db29. This should address the lack --check failure in cli/openbb_cli/controllers/script_parser.py.

Bortlesboat:
Quick follow-up from my side: I’m treating review-required PRs as top priority this week. If you want any specific changes, rebase, or split, I can turn them around quickly.

deeleeramone:
> Quick follow-up from my side: I’m treating review-required PRs as top priority this week. If you want any specific changes, rebase, or split, I can turn them around quickly.

I'm pretty sure this is unrelated to the changes, or the other reference PR, but I've found another bug that has creeped in. The CLI input should be able to handle a string list of items, but it appears to be handling them as a list of arguments instead . Would you be up for expanding the scope of this PR to look into this?

```
/equity/price/historical --symbol AAPL,MSFT,GOOGL,META,AMZN,NFLX,NVDA,TSLA,SPY --start_date 2023-01-01 --provider yfinance

The following args couldn't be interpreted: ['MSFT', 'GOOGL', 'META', 'AMZN', 'NFLX', 'NVDA', 'TSLA', 'SPY']
```

Bortlesboat:
Great catch, and yes I’m happy to address it.\n\nI opened a follow-up issue to track it cleanly without expanding scope on this security-focused PR: #7410.\n\nIf you prefer, I can also roll that fix into this PR instead, but my recommendation is to keep this one narrowly scoped and submit a fast follow-up PR for the symbol-list parsing bug.

deeleeramone (approved):
Thanks for your help!

PR #7420, opened 2026-03-19
[BugFix] CLI parser splits comma-separated flag values into separate args

### LINKED ISSUE #7410: [Bug] CLI parser treats comma-separated symbol list as separate args
## Summary

When passing comma-separated symbols to /equity/price/historical, the CLI appears to split entries into separate positional args instead of parsing them as a single symbol list.

## Repro

```
/equity/price/historical --symbol AAPL,MSFT,GOOGL,META,AMZN,NFLX,NVDA,TSLA,SPY --start_date 2023-01-01 --provider yfinance

The following args couldn't be interpreted: ['MSFT', 'GOOGL', 'META', 'AMZN', 'NFLX', 'NVDA', 'TSLA', 'SPY']
```

## Expected
--symbol should accept a comma-separated list and parse all tickers correctly.

## Actual

Only the first ticker is treated as expected and the rest are interpreted as extra args.

## Notes

This was discovered while reviewing PR #7403; tracked separately to avoid scope creep in the security fix PR.

sandole:
[BugFix] CLI parser splits comma-separated flag values into separate args

## Description

Fixes #7410.

The CLI argument parser unconditionally splits all comma-separated values into separate positional args before argparse processes them. This causes multi-symbol queries to fail:

```
/equity/price/historical --symbol AAPL,MSFT,GOOGL,META --start_date 2023-01-01 --provider yfinance

The following args couldn't be interpreted: ['MSFT', 'GOOGL', 'META']
```

## Root Cause

In `parse_known_args_and_warn`, the list comprehension at line 772 splits every arg on commas regardless of whether it is a value for a flag:

```python
other_args = [
    part
    for index, arg in enumerate(other_args)
    for part in (arg.split(",") if index != routine_args_index else [arg])
]
```

Only the routine args index was exempted. Flag values like `--symbol AAPL,MSFT` were split into `["--symbol", "AAPL", "MSFT"]`, and argparse treated `MSFT` as an unknown positional arg.

## Fix

Before comma-splitting, identify which indices are values for known flags by inspecting `parser._actions`. These indices are collected into a `no_split_indices` set and excluded from splitting.

## Testing

Verified against the following scenarios (all pass with zero unknown args):

| Input | Result |
|---|---|
| `--symbol AAPL,MSFT,GOOGL` | `symbol=AAPL,MSFT,GOOGL` |
| `--symbol=AAPL,MSFT,GOOGL` | `symbol=AAPL,MSFT,GOOGL` |
| `--verbose --symbol AAPL,MSFT` | `symbol=AAPL,MSFT` |
| `-s AAPL,MSFT --symbol GOOGL` | `short_flag=AAPL,MSFT, symbol=GOOGL` |
| `--tickers AAPL,MSFT GOOGL,TSLA` (nargs=+) | `tickers=["AAPL,MSFT", "GOOGL,TSLA"]` |
| `--symbol AAPL,MSFT` (flag at end) | `symbol=AAPL,MSFT` |

## Changes

- `cli/openbb_cli/controllers/base_controller.py`: Replace single-index exemption with set-based flag value detection (39 additions, 5 deletions)

### DIFF
--- a/cli/openbb_cli/controllers/base_controller.py
+++ b/cli/openbb_cli/controllers/base_controller.py
@@ -201,7 +201,7 @@ class BaseController(metaclass=ABCMeta):
         # Single command fed, process
         else:
             try:
-                (known_args, other_args) = self.parser.parse_known_args(
+                known_args, other_args = self.parser.parse_known_args(
                     shlex.split(an_input)
                 )
             except Exception as exc:
@@ -622,7 +622,7 @@ class BaseController(metaclass=ABCMeta):
             system_clear()
 
         try:
-            (ns_parser, l_unknown_args) = parser.parse_known_args(other_args)
+            ns_parser, l_unknown_args = parser.parse_known_args(other_args)
         except SystemExit:
             # In case the command has required argument that isn't specified
             session.console.print("\n")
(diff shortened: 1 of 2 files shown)


Bortlesboat (commented):
Filed the original issue — tested the repro from #7410 and this handles it. The set-based index tracking is a clean approach vs the old single-index exemption.

cli/openbb_cli/controllers/base_controller.py
```diff
@@ -768,19 +768,53 @@ def parse_known_args_and_warn(  # pylint: disable=R0917
...
                 -1,
             )
-            # Split comma-separated arguments, except for the argument at routine_args_index
+            # Collect indices whose values should NOT be comma-split because
+            # the provider may accept a comma-separated string (e.g. --symbol
+            # AAPL,MSFT).  Handles --flag value, --flag=value, -f value, and
+            # multi-value flags (nargs="+" / nargs="*" / nargs=N).
+            no_split_indices: set[int] = set()
+            if 0 <= routine_args_index < len(other_args):
+                no_split_indices.add(routine_args_index)
```
Bortlesboat, line 777: Minor note: the `while not other_args[j].startswith("-")` heuristic for nargs=+ could theoretically treat a negative numeric value as a flag boundary (e.g. `--values 1,2 -3,4` would stop scanning at `-3,4`). Probably not a real concern for this codebase since I dont think any CLI commands take negative numeric lists, but worth knowing about.

sandole: interesting observation. even though you're right that negative numeric values will prematurely terminate the scanning, given that negative values are not a legitimate input (and the command stops as it should) and input validation logic would be way too bloated to sanitize all input possibilities, I think this part should be left as-is.

if negative numeric args become a real pattern, a one-line heuristic would handle it:
`(lstrip("-")[:1].isalpha())`

deeleeramone (commented):
Thanks for the solution! We should add some test coverage so that these expectations are baked in and regressions will be identified by failures.

sandole:
@Bortlesboat funnily enough created a PR with the test coverages: #7422. Maybe those can be tacked on here?

cli/openbb_cli/controllers/base_controller.py
deeleeramone: Indeed, an interesting observation. I am not aware of any specific scenarios this library would encounter that; however, developers may choose to build their own extensions which do. Something to keep in mnd, for sure. Thanks @Bortlesboat.

deeleeramone (approved):
Thanks a lot for your help and participation!

PR #7441, opened 2026-03-29
fix(sec): handle unexpected SEC API responses in cik_map

### LINKED ISSUE #7097: [Bug] Command /regulators/sec/cik_map throws an error
**Describe the bug**

When I try to use `cik_map` to map a ticker symbol to a cik number I get an error

**To Reproduce**
Steps(from the start) and commands to reproduce the behavior

Go to `/regulators/sec` in Openbb-CLI
Run something like `cik_map --symbol AAPL`

**Screenshots**

Image

- OS: Manjaro Linux x86_64
- Python version 3.11.8

**Additional context**
Add any other information that you think could be useful for us.

Error message:
ValueError: If using all scalar values, you must pass an index. Ensure the data format matches the expected format.

JasonOA888:
Fixes #7097

The `cik_map` command throws `ValueError: If using all scalar values, you must pass an index` when the SEC API returns an unexpected response format.

**Root Cause:**
- `get_all_companies()` uses `DataFrame(response).transpose()` which crashes when the response is a flat dict (e.g., error response from SEC) instead of the expected nested dict of dicts
- `cik_map` fetcher has a dead `if not results` check — `results` is always a non-empty dict like `{"cik": ""}` so the check never triggers

**Changes:**
- `get_all_companies()`: validate response type and use `DataFrame.from_dict(orient="index")` which handles nested dicts more robustly
- `get_all_companies()`: verify column count matches expected 3 fields before renaming
- `cik_map` fetcher: raise `OpenBBError` when `symbol_map` returns empty string, instead of silently passing through


deeleeramone (approved):
Looks good to me, thanks for your help!

PR #7446, opened 2026-03-30
Purge incomplete generated assets when openbb-build fails

itsviseph:
## Summary
This updates the PR to address the underlying build failure path rather than the downstream import symptom.

## What changed
- purge partially generated `package/` and `assets/` output when `openbb-build` fails mid-run
- keep the build lock handling unchanged while ensuring failed runs do not leave stale generated files behind
- add a regression test that simulates a late-stage build failure after files have already been written

## Why
The `ImportError` reported in #7379 is a downstream symptom of an incomplete `openbb-build` run. When generation fails after writing some static files, those partial artifacts can remain on disk and make the build look successful enough to import the wrong state later.

This change makes failed runs clean up those incomplete generated assets so the next build starts from a known-good state instead of reusing stale output.

## Validation
- `python -m pytest openbb_platform/core/tests/app/static/test_package_builder.py -q -k 'test_package_builder_build or test_build_cleans_partial_artifacts_on_failure or test_auto_build'`
- `python -m pytest openbb_platform/core/tests/app/test_provider_interface.py -q -k 'test_create_field'`

### DIFF
--- a/openbb_platform/core/openbb_core/app/provider_interface.py
+++ b/openbb_platform/core/openbb_core/app/provider_interface.py
@@ -739,3 +739,23 @@ class ProviderInterface(metaclass=SingletonMeta):
                 __doc__=f"OBBject with results of type {name}",
             )
         return annotations
+
+
+_cached_return_annotations: dict[str, type[OBBject]] | None = None
+
+
+def __getattr__(name: str) -> type[OBBject]:
+    """Lazily expose dynamically generated OBBject_* annotations as module attributes."""
+    if not name.startswith("OBBject_"):
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+    global _cached_return_annotations
+    if _cached_return_annotations is None:
+        _cached_return_annotations = ProviderInterface().return_annotations
+
+    model_name = name.removeprefix("OBBject_")
+    annotation = _cached_return_annotations.get(model_name)
+    if annotation is None:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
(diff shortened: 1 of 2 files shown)


deeleeramone:
Hi, @itsviseph. Thanks for the PR, however the proposed solution addresses only a symptom of the cause.

### dynamic OBBject_* imports from openbb_core.app.provider_interface

This is not really a thing, there are no "dynamic imports". PackageBuilder uses a two-pass system and the provider interface will never have these classes. These are **temporary** objects used only for static asset generation.

The error you are solving for is symptomatic of an interrupted or incomplete execution of `openbb-build` and the second pass failed to complete. This can stem from race conditions,  or syntax, annotation, and formatting errors.

### The Real Source of Error

When this error happens:

```python
ImportError: cannot import name "OBBject_EquityInfo" from "openbb_core.app.provider_interface"
```

It means the static assets need to be built again by running:

```
openbb-build
```

If it continues to fail, then the error is being silenced/dropped, and the first pass created the file structure getting imported.

Enable `OPENBB_DEBUG_MODE` environment variable before building to see the verbose output, which will likely surface the error. 

The "bug" is that the build process is not catching the error and purging the incomplete files, providing the illusion of a completed run.

Feel free to open a new PR thinking from a different angle, simply importing it only advances you to the next line of code.

itsviseph:
Thanks for the steer here. I dropped the provider_interface import workaround and updated the branch to target the build failure path instead. The current change now purges partially generated package/assets output when openbb-build fails mid-run, with a regression test that simulates a late-stage failure after files have already been written.

PR #7452, opened 2026-04-02
Flip default tool discovery setting to disabled

piiq:
Tool discovery is now opt-in via `--tool-discovery` flag instead of enabled by default.

Updates documentation, settings, and tests to reflect that `enable_tool_discovery` defaults to False.

This is required to enable workspace compatibility being the default option


deeleeramone:
Something's not right here, the flat list is not being populated.

This branch:

```sh
openbb-mcp
```

Screenshot 2026-04-02 at 5 04 38 PM

On develop:

```sh
openbb-mcp --no-tool-discovery
```

Screenshot 2026-04-02 at 5 10 15 PM

deeleeramone:
Disregard my previous comment, I forget to remove the settings file.

deeleeramone (approved):

PR #7462, opened 2026-04-08
fix(core): purge partially generated assets on build failure

### LINKED ISSUE #7379: ImportError: OBBject_EquityInfo not found in provider_interface (v4.6.0)

Shreyansh1729:
Resolves #7379.

### Description
Following the reviewer feedback on #7460, this PR addresses the root cause of the `ImportError` by ensuring that the build process cleans up (purges) any partially generated files if an exception occurs mid-run.

### Changes
- Wrapped build generation steps in a `try...except` block in `PackageBuilder.build`.
- Triggers `self._clean(modules)` on failure to ensure a consistent workspace state.
- Added a regression test in `test_package_builder.py` to verify the purge behavior.


deeleeramone:
Nice, this is a good solution @Shreyansh1729! To be helpful to the end user, I think it would be a good idea to include the traceback with the offending code as a `console.log.error` message + an instruction to set `OPENBB_DEBUG_MODE='true'` environment variable, then run `openbb-build` again to enable verbose output.

Shreyansh1729:
Done! I've updated the PR to include full traceback logging and user instructions for OPENBB_DEBUG_MODE as suggested. I also updated the Console utility to ensure these critical build errors are always visible. Let me know if there's anything else!

Shreyansh1729:
@deeleeramone any updates ?

deeleeramone:
@Shreyansh1729, this is good thanks. There is an edge case here that we should probably catch in the outer try block as well, KeyboardInterrupt or signal.SIGTERM + signal.SIGINT. Currently, if the process is terminated - either by the user or programmatically - the partially generated modules survive. How do you think we can guard against that?

Shreyansh1729:
Update: I've also implemented handling for interruptions (KeyboardInterrupt) and termination signals (SIGTERM). Now, if the build process is interrupted or terminated externally, it will still trigger the cleanup purge of partially generated assets. I've added a regression test for KeyboardInterrupt as well. Thanks for catching that edge case!

Shreyansh1729:
@deeleeramone I've addressed your feedback regarding interruptions and SIGTERM. The PR is now updated with robust signal handling and cleanup logic. Could you take another look when you have a chance?

deeleeramone (approved):
Thanks a lot for your help, this is a nice solution.

deeleeramone:
Thanks, @Shreyansh1729, nicely done. I had to temporarily add pylint disables for the linter job, but this is actually because of a flaw in the workflow itself.

Shreyansh1729:
@deeleeramone thanks a lot

PR #7482, opened 2026-04-28
[Feature] OpenBB V5: Initial `openbb-core` Refactor

### LINKED ISSUE #7473: [Bug] OptionsChainsData schema/serializer mismatch: columnar schema vs record-list wire format
## Bug                                                                                                                                                                                                                                                  

`OptionsChainsData.model_serializer` returns a list of records, but the Pydantic model is declared as a columnar object (`list[float | None]`, `list[str]`, etc.). This means the OpenAPI / JSON Schema generated from the model declares `OBBject.results` as an **object** with columnar list fields, while the wire format the REST API actually emits is an **array of records**. 
                                                                                                                                                                                                                                                Pydantic's own `model_validate` is happy (input is columnar) and the Python client / REST JSON callers work fine (they iterate `results` as records). 

The mismatch only surfaces in consumers that validate the HTTP response against the declared schema — notably `openbb-mcp-server` via FastMCP — which raises the generic jsonschema error `"not valid under any of the given schemas"` on every single tool call that uses this endpoint.

Affects **all** options-chain providers (`yfinance`, `intrinio`, `cboe`, `tradier`, `tmx`, `deribit`), because they all inherit `OptionsChainsData`.

## Environment

- OpenBB Platform (develop @ 8cc70fa1c8a)
- `openbb-mcp-server==1.4.0`
- Python 3.11

## Minimal repro (no MCP needed)

```python
import httpx, json, subprocess, time

# Start REST API
p = subprocess.Popen(
    ["python", "-m", "uvicorn", "openbb_core.api.rest_api:app",
     "--host", "127.0.0.1", "--port", "8765", "--log-level", "warning"])
time.sleep(5)

# Grab schema and a real response, compare shapes
spec = httpx.get("http://127.0.0.1:8765/openapi.json").json()
results_schema = spec["components"]["schemas"]["OBBject_OptionsChains"]["properties"]["results"]
# → oneOf { DeribitOptionsChainsData | IntrinioOptionsChainsData | ... }, all object-type columnar schemas

resp = httpx.get("http://127.0.0.1:8765/api/v1/derivatives/options/chains",
                 params={"provider": "yfinance", "symbol": "SPY"}).json()
print(type(resp["results"]))   # <class 'list'>
print(resp["results"][0])      # {'underlying_symbol': 'SPY', 'contract_symbol': '...', ...}  ← record, not columnar object
p.terminate()
```

Declared schema: `results: object` (columnar)
Actual response: `results: array<object>` (record list)

## Why this happens
`openbb_core/provider/standard_models/options_chains.py:354-371` defines a custom `@model_serializer` that zips the columnar fields into records:

```python
@model_serializer
def model_serialize(self):
    data: dict = {}
    for field in self.model_fields:
        value = getattr(self, field)
        if isinstance(value, list):
            if value:
                if isinstance(value[0], datetime):
                    data[field] = [str(v) if v else None for v in value]
                else:
                    data[field] = value
        else:
            data[field] = value
    records = [dict(zip(data.keys(), values)) for values in zip(*data.values())]
    return records
```

Changing the serialized shape here does not update the JSON Schema generated for the model — Pydantic / FastAPI still derive the schema from the typed fields.

## Impact

- `openbb-mcp-server` tool calls to `derivatives_options_chains` always return a validation error to the MCP client instead of the data, regardless of provider.
- Anyone else consuming the OpenAPI spec to generate strict clients (TypeScript with `oneOf` strictness, Go with JSON Schema validation, etc.) hits the same mismatch.

## Possible directions

The fix is a design choice; happy to submit a PR once a direction is agreed on.

- **(A)** Drop the record-reshape serializer and return columnar JSON. Matches the declared schema. Breaking change for anyone currently relying on `results[0]`-style record access over HTTP.
- **(B)** Redefine `OptionsChainsData` as a record-list (each field non-list, and `results` typed as `list[OptionsChainRecord]`). Matches the serializer. Biggest refactor — touches `OptionsChainsProperties` helper methods that assume columnar storage.
- **(C)** Keep the Pydantic model as-is but override the generated OpenAPI schema for `OBBject_OptionsChains.results` to `array<OptionsChainRecord>`. Smallest change and non-breaking for runtime consumers, but requires a custom schema hook.

Happy to put together a PR for whichever direction maintainers prefer. My lean is **(C)** — it's the only option that keeps the wire format and the existing Python-side helpers stable.

deeleeramone:
This is a draft PR for V5 `openbb-core` package.

Changes:

- hatchling as build dependency.
- pandas is now an optional extra.
- expanded test coverage and improved coverage surface
- linting and type checking are now handled exclusively by Ruff and TY (still use Black)
- explicit action for openbb-core tests.
- cross-platform CI testing: Linux, macOS, Windows


.github/workflows/general-linting.yml
```diff
@@ -59,16 +59,16 @@ jobs:
...
-          pip install openbb-devtools
-          pip install types-pytz types-requests types-termcolor types-tabulate types-PyYAML types-python-dateutil types-setuptools types-six
+          pip install ./openbb_platform/extensions/devtools
+      - name: Install openbb-core (so ty can resolve runtime deps)
+        run: pip install -e "./openbb_platform/core[pandas]" numpy
       - run: codespell --ignore-words=.codespell.ignore --skip="$(tr '\n' ',' < .codespell.skip | sed 's/,$//')" --quiet-level=2
       - run: |
           # Run linters for openbb_platform | cli
           if [ -n "${{ env.diff_files }}" ]; then
             black --diff --check ${{ env.diff_files }}
```
piiq: ruff covers black formatting too

openbb_platform/core/openbb_core/api/router/commands.py
```diff
@@ -22,11 +25,9 @@
 from openbb_core.app.service.user_service import UserService
 from openbb_core.env import Env
 from openbb_core.provider.utils.helpers import to_snake_case
-from pydantic import BaseModel
-from typing_extensions import ParamSpec
 
 try:
-    from openbb_charting import Charting
+    from openbb_charting import Charting  # ty: ignore[unresolved-import]
```
piiq, line 32: you think we still need the charting extension moving forward?

openbb_platform/core/openbb_core/app/static/package_builder/docstring_generator.py
```diff
@@ -0,0 +1,1076 @@
...
+    def get_OBBject_description(
+        results_type: str,
+        providers: str | None,
+    ) -> str:
+        """Get the command output description."""
+        available_providers = providers or "Optional[str]"
+        indent = 2
+
+        obbject_description = (
+            f"{create_indent(indent)}OBBject\n"
```
piiq, line 172: consider using `textwrap.indent` for this. no strong opinion

openbb_platform/core/openbb_core/app/static/package_builder/docstring_generator.py
```diff
@@ -0,0 +1,1076 @@
...
+            )
+
+            # Convert Optional[X] to X | None
+            optional_pattern = r"Optional\[(.+?)\]"
+            optional_match = re.search(optional_pattern, type_str)
+            if optional_match:
+                inner = optional_match.group(1)
+                type_str = type_str.replace(f"Optional[{inner}]", f"{inner} | None")
+
+            # Convert Union[X, Y, ...] to X | Y | ... format
```
piiq, line 255: Is this for backwards compatibility or should this be handled by the extension migrating to modern syntax?

openbb_platform/core/openbb_core/app/static/package_builder/file_lock.py
```diff
@@ -0,0 +1,80 @@
...
+    "Series",
+    list["Series"],
+    "ndarray",
+    "Data",
+)
+
+TAB = "    "
+
+
+def create_indent(n: int) -> str:
```
piiq: same comment as in docstring generator:
1. consider using `textwrap.indent`
2. consider not defining the same helper function in miltiple files (even when it's a two-liner)

openbb_platform/core/openbb_core/app/static/package_builder/method_definition.py
```diff
@@ -0,0 +1,1361 @@
...
+    "Series",
+    list["Series"],
+    "ndarray",
+    "Data",
+)
+
+TAB = "    "
+
+
+def create_indent(n: int) -> str:
```
piiq: this is the third file where this helper is defined

openbb_platform/core/openbb_core/app/static/package_builder/module_builder.py
```diff
@@ -0,0 +1,59 @@
...
+    "Series",
+    list["Series"],
+    "ndarray",
+    "Data",
+)
+
+TAB = "    "
+
+
+def create_indent(n: int) -> str:
```
piiq: yeah, i'm reviewing, and this helper acts like a red herring =)

openbb_platform/core/openbb_core/app/utils.py
```diff
@@ -5,14 +5,14 @@
...
 from openbb_core.app.model.abstract.error import OpenBBError
 from openbb_core.app.model.preferences import Preferences
 from openbb_core.app.model.system_settings import SystemSettings
 from openbb_core.provider.abstract.data import Data
-from pydantic import ValidationError
 
 if TYPE_CHECKING:
-    # pylint: disable=import-outside-toplevel
     from numpy import ndarray
     from pandas import DataFrame, Series
```
piiq, line 18: you mentioned pandas now becomes optional - would this branch and the code bellow be impacted?

piiq (approved):
breaking up the package builder to become a module is a good idea

openbb_platform/core/openbb_core/api/router/commands.py
deeleeramone: WDYM, "need the charting extension"? It is a piece of infrastructure and is responsible for generating output.

.github/workflows/general-linting.yml
deeleeramone: Ya, you're right. Ruff and TY all the way.

openbb_platform/core/openbb_core/app/static/package_builder/file_lock.py
deeleeramone: ya, that's silly. moved to a shared utility.

openbb_platform/core/openbb_core/app/utils.py
deeleeramone: This is the expected behavior when the extra is not installed:

```
In [1]: from openbb_core.app.utils import basemodel_to_df

In [2]: basemodel_to_df([{"c1": 0, "c2": 2}])
---------------------------------------------------------------------------
OpenBBError                               Traceback (most recent call last)
Cell In[2], line 1
----> 1 basemodel_to_df([{"c1": 0, "c2": 2}])

File ~/github/OpenBB/openbb_platform/core/openbb_core/app/utils.py:26, in basemodel_to_df(data, index)
     21 def basemodel_to_df(
     22     data: list[Data] | Data,
     23     index: str | None = None,
     24 ) -> "DataFrame":
     25     """Convert list of BaseModel to a Pandas DataFrame."""
---> 26     pd = require_optional("pandas")
     27     DataFrame, to_datetime = pd.DataFrame, pd.to_datetime  # type: ignore[union-attr]
     29     if isinstance(data, list):

File ~/github/OpenBB/openbb_platform/core/openbb_core/app/utils_optional.py:85, in require_optional(*module_names)
     80         missing.append(name)
     82 if missing:
     83     # ``from None`` suppresses the chained ``ImportError`` context so the
     84     # user sees a single, clean OpenBBError with the install hint.
---> 85     raise OpenBBError(_format_error(missing)) from None
     87 if len(modules) == 1:
     88     return modules[0]

OpenBBError: Optional dependency 'pandas' is required for this feature. Install with: pip install 'openbb-core[pandas]'
```

PR #7491, opened 2026-05-05
fix: merge custom widgets in editable mode

### LINKED ISSUE #7479: [Bug] Widgets not saving their position and size after refresh
**When adding widgets from a custom ODP they lose their position after refresh**

Image

Image

I tried several ways to adjust the click on select from the grid but this issue pops up as soon as the click on cell is introduced

Bortlesboat:
## Summary
- merge router-provided widget JSON into editable widgets.json responses without overwriting existing edited widget IDs
- keep path-widget exclude handling based on the caller-provided filter
- add regression coverage for custom widget routes loaded alongside an editable widgets file

Fixes OpenBB-finance/OpenBB#7479

## Verification
- py -3 -m pytest openbb_platform/extensions/platform_api/tests/test_api.py::test_get_widgets_json_editable_merges_additional_widget_routes -q
- py -3 -m pytest openbb_platform/extensions/platform_api/tests/test_merge_widgets.py -q
- py -3 -m py_compile openbb_platform/extensions/platform_api/openbb_platform_api/utils/api.py openbb_platform/extensions/platform_api/tests/test_api.py
- git diff --check

### DIFF
--- a/openbb_platform/extensions/platform_api/openbb_platform_api/utils/api.py
+++ b/openbb_platform/extensions/platform_api/openbb_platform_api/utils/api.py
@@ -107,6 +107,26 @@ def get_user_settings(current_user_settings: str) -> dict:
     return user_settings
 
 
+def _merge_path_widgets(
+    widgets_json: dict, widget_exclude_filter: list, overwrite: bool = True
+) -> dict:
+    """Merge widgets loaded from additional widgets.json routes."""
+    if not PATH_WIDGETS:
+        return widgets_json
+
+    for k in PATH_WIDGETS:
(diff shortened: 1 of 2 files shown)


deeleeramone:
So in editable mode, the expected behavior is for the API to load the file from disk every time the `/widgets.json` endpoint is called. If it doesn't exist yet, it will generate the file with the items merged. If you don't supply a path, it defaults to the root folder of the environment (where Python is). When it is reading from an existing file, it does not do any merging or auto-generation.

deeleeramone:
Closing because the linked issue does not correspond with any code in this repository.

PR #7499, opened 2026-05-13
fix: align OptionsChainsData JSON schema with record-list wire format (#7473)

### LINKED ISSUE #7473: [Bug] OptionsChainsData schema/serializer mismatch: columnar schema vs record-list wire format

paulxg12:
## Summary

Fixes #7473 — the OptionsChainsData model uses columnar storage internally (each field is a list) but serializes to a record-list wire format (a list of dicts) via a custom @model_serializer. The JSON schema still reflected the columnar shape, causing a mismatch between schema and actual API responses.

### Changes
Added `__get_pydantic_json_schema__` override + `_unwrap_array_schema` helper to:
- Transform the per-field array schema to match the record-list output
- Preserve metadata (x-unit_measurement, descriptions, titles, etc.)
- Handle union types (anyOf/oneOf containing arrays)
- Drop the default empty-list value (does not apply per-record)

### Testing
The override is purely schema-level — no runtime data transformation changes were needed since the @model_serializer was already correct.

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/options_chains.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/options_chains.py
@@ -12,6 +12,7 @@ from openbb_core.provider.utils.descriptions import (
 )
 from openbb_core.provider.utils.options_chains_properties import OptionsChainsProperties
 from pydantic import Field, field_validator, model_serializer
+from pydantic.json_schema import JsonSchemaValue
 
 
 class OptionsChainsQueryParams(QueryParams):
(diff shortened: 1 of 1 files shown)


paulxg12:
I have read the CLA Document and I hereby sign the CLA

deeleeramone:
Hi, thanks for the PR. The issue, however, has already been assigned and resolved. Feel free to pick up on something else.

PR #7520, opened 2026-05-30
Security/Logic Fix: Autonomous Code Review

fliptrigga13:
## Autonomous Bug Report & Patch

This vulnerability and fix were autonomously discovered by the Lucy Red Team swarm.

Upon reviewing the provided code, one potential issue that stands out is related to how the `command` decorator handles the return type of the function being decorated when `no_validate` is set to `True`. Specifically, there is an inconsistency in how the return type is handled and checked.

Here's a detailed analysis:

1. **Setting Return Type to None**:
    ```python
    if no_validate is True:
        func.__annotations__["return"] = None
    ```
    This line sets the return type of the function to `None` when `no_validate` is `True`.

2. **Assigning Response Model**:
    ```python
    kwargs["response_model"] = (
        kwargs.get(
            "response_model",
            func.__annotations__["return"],  # type: ignore
        )
        if not no_validate
        else func.__annotations__["return"]
    )
    ```
    This line assigns the response model based on whether `no_validate` is `True` or `False`. If `no_validate` is `True`, it uses `func.__annotations__["return"]`, which has been set to `None`.

3. **Potential Issue**:
    - When `no_validate` is `True`, the return type of the function is explicitly set to `None`.
    - However, the response model is still being set to `None`, even though the original intention might have been to bypass validation and return a different type.

Given this analysis, it seems


deeleeramone:
No AI slop.

PR #7528, opened 2026-06-05
[V5] `openbb-congress-gov`: Refactor for V5 + Convert To Using Bulk Files Instead of `api.congress.gov`.

deeleeramone:
This PR refactors the `openbb-congress-gov` for V5 ODP.

### Implementation Changes

- No longer requires an API key, all data is consumed from bulk file downloads and the `gov-info` link service.
- Adds coverage for Laws, Congressionally-Mandated Reports, Members + votes, and broad search.
- App now has 27 widgets across 8 tabs, including several "how to use" markdown widgets.

Open question, @piiq, should we consolidate this extension to be a module within `openbb-government-us`, while keeping all the imports the same?

### Endpoints

```
/uscongress
    amendment_info
    amendment_text
    amendments
    bill_info
    bill_text
    bills
    calendars
    committee_documents
    committee_info
    laws
    mandated_reports
    member_legislation
    member_votes
    members
    search
```

Screenshot 2026-06-05 at 8 54 19 AM

Screenshot 2026-06-05 at 9 03 27 AM


DidierRLopes (approved):

piiq:
@deeleeramone iiuc the `goventment-us` provider does not currently require any api keys, while the congress one does. if that's the case i would keep them separate so that the `goventment-us` provider can stay reliant on a public apis that don't require authentication.

on the other hand i do understand why you would want to consolidate them. it's your call. i'm fine with both of the options

piiq (approved):

deeleeramone:
> @deeleeramone iiuc the `goventment-us` provider does not currently require any api keys, while the congress one does.

This PR removes all requirements for an API key, it reads from bulk data files and the govInfo link service, which does not require authentication.

piiq:
so my understanding was not correct. in this case yes - it makes a lot of sense to merge them in the government-us provider

PR #7534, opened 2026-06-11
docs: clarify custom AI agent selection

gyx09212214-prog:
## Summary
- clarify that custom AI agent backends need to be selected in the Workspace chat before use
- note that OpenBB Copilot limits apply when Copilot is the selected agent, not when a user switches to a custom agent backend or MCP server

## Validation
- docs-only README change
- inspected the diff against `origin/develop`


deeleeramone:
This is not necessary, there is plenty of documentation available - https://docs.openbb.co/workspace/developers/agents-integration. 

This repository also is not responsible for the Workspace or agents.

Issue #6702, opened 2024-09-28
[FR]Normalize equity symbols


yolotrader:
**What's the problem of not having this feature?**
Currently OpenBB relies on its data providers to resolve a symbol. Unfortunately providers format symbols slightly differently in certain cases. A typical example is that yfinance uses "BRK-A" for Berkshire A shares where as many others use the more common "BRK.A". This requires users to remember these subtleties and developers to write extra code to handle them.

**Describe the solution you would like**
It would be great if OpenBB can adopt a uniform convention for using "." in the symbol format and use the provider extensions to convert it to "-" (or some other character) as necessary. This will help both regular users and developers.

**Describe alternatives you've considered**
If there are other alternatives (worth considering) to your solution, list them here. 

**Additional information**
Add any other information or screenshots about the feature request here.

254Co:
Would be worthwhile to integrate OpenFGI API. 

https://www.openfigi.com/api

deeleeramone:
Yes, however, it would be for a different purpose. OpenFIGI won't help with mapping between vendor's symbology, and especially where a ticker symbol is the only type of input allowed. 

Is it `BRK.A` or `BRK-A` or `BRK.A:US` or  `BRK/A`? Nuances there are very specific to vendors and most don't have handling for different formats.

PR #7536, opened 2026-06-15
fix(yfinance): normalize class-share symbols before Yahoo requests

gyx09212214-prog:
## Summary
- Add yfinance-only class-share symbol normalization for common US class-share symbols such as `BRK.A` and `BF.B`, converting them to Yahoo's dash-form symbols at the request boundary.
- Apply the helper to `yf_download(...)` and yfinance `Ticker(...)`-based equity/fundamental request paths while preserving the requested dot-form symbol in OpenBB `symbol` output fields.
- Add mocked coverage for helper boundaries, `yf.download` ticker arguments, multi-symbol historical output, and a quote fetcher `Ticker("BRK-A")` path.

## Why
Users can use OpenBB's dot convention from #6702 while Yahoo still receives the dash-form symbols it expects. This keeps the mapping vendor-specific and avoids changing global symbol semantics.

## Scope
- Only `openbb-yfinance` is changed.
- No standard model changes.
- No global symbol policy changes.
- No non-yfinance provider changes.
- The conversion is deliberately conservative for this first step, covering common class suffixes `A`, `B`, and `C` so one-letter Yahoo exchange suffixes such as `VOD.L` remain unchanged.

## Validation
From `openbb_platform/providers/yfinance`:
- `python -m pytest tests/test_yfinance_helpers.py -q` - 19 passed
- `python -m ruff check openbb_yfinance tests/test_yfinance_helpers.py` - passed

Refs #6702

### DIFF
--- a/openbb_platform/providers/yfinance/openbb_yfinance/models/balance_sheet.py
+++ b/openbb_platform/providers/yfinance/openbb_yfinance/models/balance_sheet.py
@@ -87,10 +87,11 @@ class YFinanceBalanceSheetFetcher(
         from openbb_core.provider.utils.helpers import (
             to_snake_case,
         )
+        from openbb_yfinance.utils.helpers import normalize_yfinance_symbol
         from yfinance import Ticker
 
         period = "yearly" if query.period == "annual" else "quarterly"  # type: ignore
-        data = Ticker(query.symbol).get_balance_sheet(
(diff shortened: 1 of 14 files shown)


deeleeramone:
Closing because this extension is being moved from this repository to here: https://github.com/deeleeramone/openbb-danglewood/tree/main/providers/yfinance

PR #7551, opened 2026-06-23
Remove unsafe MCP skill install tool

elimaximus1:
Removes the write-capable install_skill MCP tool, which accepted client-supplied skill and file paths and could write outside the intended skills directory.

Validation:
- python -m py_compile openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
- python -m pytest tests
- python -m ruff check openbb_mcp_server/app/app.py


deeleeramone:
The user is the server owner, who also is running it over localhost while controlling the client. You're really only a danger to yourself.

PR #7568, opened 2026-07-05
[Docs] Fix copied apps.json docstrings

ChimdumebiNebolisa:
## Summary
- Fixes copied IMF wording in CFTC and Congress.gov apps.json endpoint docstrings.
- Keeps behavior unchanged.

## Test Plan
- Not run; documentation-only change.


deeleeramone (approved):
Thanks for the fix!

PR #7571, opened 2026-07-06
feat(derivatives): add yfinance futures expirations endpoint

ChimdumebiNebolisa:
## Summary
Adds a yfinance-backed futures expirations endpoint under `derivatives.futures`.

- Adds `FuturesExpirations` standard model
- Adds `obb.derivatives.futures.expirations(...)`
- Registers `YFinanceFuturesExpirationsFetcher`
- Uses Yahoo `futuresChainDetails.expireIsoDate` when available
- Falls back to `expiration_month` parsed from contract symbols
- Preserves rows when contract detail metadata is missing or partially malformed

## Scope
- Futures only
- yfinance only
- No options expirations
- No `economy.calendar` changes
- No BarChart dependency

## Tests
- `git diff --cached --check`
- `ruff check` on changed files
- Focused yfinance helper/fetcher tests

Result: 5 passed.

### DIFF
--- a/openbb_platform/core/openbb_core/provider/standard_models/futures_expirations.py
+++ b/openbb_platform/core/openbb_core/provider/standard_models/futures_expirations.py
@@ -1,0 +1,46 @@
+"""Futures Expirations Standard Model."""
+
+from datetime import date as dateType
+
+from openbb_core.provider.abstract.data import Data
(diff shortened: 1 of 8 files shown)


deeleeramone:
Thanks for the PR, but this extension is being removed from the repository - #7576 - so it will not be necessary.

PR #7583, opened 2026-07-12
fix(technical): snap missing fib start_date to the date nearest start_date

Bortlesboat:
## Why

`calculate_fib_levels` in `openbb_technical/helpers.py` handles a `start_date` that isn't in the index by snapping to the nearest available date — but the lookup searches for `end_date` instead of `start_date` (a copy-paste from the end-date branch right below it):

```python
if start_date not in data.index:
    date0 = data.index[data.index.get_indexer([end_date], method="nearest")[0]]
```

So any start date that isn't an exact index entry (a weekend or holiday — easy to hit since the index is trading days) silently collapses the retracement to the end date. Reproduction on a 60-business-day frame with a Saturday `start_date`:

```
min_date: 2024-03-01  max_date: 2024-03-01  span: 0.0
WARN: Start date not in data.  Using nearest: 2024-03-01
```

Every Fibonacci level comes back equal to the same price, and the warning reports the wrong "nearest" date, so the result looks plausible unless you inspect it.

## What

One-line fix: look up `start_date` in the `get_indexer` call. With the fix, the same input snaps to Friday 2024-01-05 and the retracement spans the full range.

Added a regression test (`test_calculate_fib_levels_snaps_missing_start_date_to_nearest`) that pins the snapped date, the preserved end date, and a non-zero span; it fails on current develop and passes with the fix. All 10 tests in `test_technical_helpers.py` pass; `black` and `ruff` clean.


deeleeramone:
Hi @Bortlesboat, thanks for the PR. Can you please check the #7489 to verify if the issue still exists there? The technical extension has been refactored, so lots of the code has changed upstream. Thanks!

Bortlesboat:
Hi @deeleeramone, thanks for the pointer — just re-checked against current `develop` (post-#7489).

The bug is still there: `calculate_fib_levels` stayed in `openbb_technical/helpers.py` through the refactor, and the start-date branch still does the nearest lookup with `end_date`:

https://github.com/OpenBB-finance/OpenBB/blob/develop/openbb_platform/extensions/technical/openbb_technical/helpers.py#L569-L571

So any `start_date` that isn't an exact index entry (weekend/holiday) still collapses the retracement to the end date and every fib level comes back equal. I've updated the branch to current develop — the one-line fix and the regression test apply cleanly.

PR #7591, opened 2026-07-15
feat: add openbb-nse provider with index constituents

### LINKED ISSUE #4247: [FR]: Can we also start to include stocks of Indian markets, listed on BSE or NSE?
Stocks of BSE and NSE (India) are not present for info or research, it would be good to include them.  

**Describe the solution you would like**
We can add major Indian stocks (blue chip companies).  

**Describe alternatives you've considered**
There are no alternatives, if I search based on country then I get to see some stocks but they are all listed on other exchanges. 

**Additional information**
N/A

gottostartsomewhere:
Adds an `nse` provider for National Stock Exchange of India index constituents, mapped onto the existing `IndexConstituents` model.

```python
obb.index.constituents("nifty_500", provider="nse")
```

Indian equities are already reachable via yfinance/fmp, so I didn't want to just re-add prices. The thing that's actually missing is index membership (NIFTY 50/500, Bank Nifty, the sectoral ones), so that's what this covers. It reads the official constituent files NSE Indices publishes (the `ind_*list.csv` files on niftyindices.com), which are static and need no key or session handling, so none of the usual nseindia.com anti-bot dance. I also kept the industry/isin/series columns from those files as extra fields since they come for free and are useful.

Supported indices right now: nifty_50, nifty_100, nifty_200, nifty_500, nifty_next_50, nifty_bank, nifty_it, nifty_auto, nifty_pharma, nifty_fmcg, nifty_metal, nifty_energy, nifty_fin_service, nifty_midcap_100, nifty_smallcap_100. Adding more is one line in the map.

Marking it draft because I want to check the approach is wanted before building more. If it is, the obvious next step is the India-only data nobody else exposes (delivery volumes, bulk/block deals, FII/DII flows), but that needs new models so I'd keep it to a separate PR. For context, I work with SEBI and Indian market data at my day job, so I use this and can keep it maintained.

## How has this been tested?

Added a fetcher test with a recorded cassette, passes in replay. Also ran it live: nifty_50 returns 50 names, nifty_bank returns 14, with industry/isin/series populated. An unknown index symbol raises a ValueError listing the valid ones. ruff is clean.

- New provider / fetcher:
  - [x] Existing tests pass.
  - [x] Provider and fetcher are stable and usable.
  - [x] Added a test for the fetcher, with a recorded cassette.

I registered `openbb-nse` in pyproject.toml (optional dep plus the `nse` and `all` extras) and in dev_install.py. I couldn't find PROVIDERS.md, EXTENSIONS.md, or the community integration workflow the template points at on this branch, so I left those alone. Let me know if I missed a spot.

## Checklist

- [x] I self-reviewed the code.
- [x] I commented the trickier bits.
- [x] I'm following the CONTRIBUTING guidelines.

Branch is `feat/nse-index-constituents` rather than `feature/...`. Happy to rename it if you want strict GitFlow.

Closes #4247

### DIFF
--- a/openbb_platform/dev_install.py
+++ b/openbb_platform/dev_install.py
@@ -64,6 +64,7 @@ openbb-finra = { path = "./providers/finra", optional = true, develop = true }
 openbb-finviz = { path = "./providers/finviz", optional = true, develop = true }
 openbb-multpl = { path = "./providers/multpl", optional = true, develop = true }
 openbb-nasdaq = { path = "./providers/nasdaq", optional = true, develop = true }
+openbb-nse = { path = "./providers/nse", optional = true, develop = true }
 openbb-seeking-alpha = { path = "./providers/seeking_alpha", optional = true, develop = true }
 openbb-stockgrid = { path = "./providers/stockgrid" , optional = true,  develop = true }
 openbb_tmx = { path = "./providers/tmx", optional = true, develop = true }
--- a/openbb_platform/providers/nse/README.md
+++ b/openbb_platform/providers/nse/README.md
@@ -1,0 +1,40 @@
+# OpenBB NSE Provider
+
+This extension integrates data from the [National Stock Exchange of India (NSE)](https://www.nseindia.com) into the OpenBB Platform.
+
+Index data is sourced from the official constituent files published by **NSE Indices Limited** (`niftyindices.com`). These are statically published files, so **no API key is required** and there is no dependency on the anti-bot `nseindia.com` quote API.
+
+## Installation
+
+```bash
+pip install openbb-nse
+```
+
+Documentation is available [here](https://docs.openbb.co/platform).
+
+## Coverage
+
+| Command | Standard model | Source |
+| --- | --- | --- |
+| `obb.index.constituents` | `IndexConstituents` | NSE Indices official constituent files |
+
+## Usage
+
+```python
+from openbb import obb
+
+# Constituents of the NIFTY 500
+obb.index.constituents("nifty_500", provider="nse")
+
+# Bank Nifty
+obb.index.constituents("nifty_bank", provider="nse")
+```
+
+### Supported indices
+
+`nifty_50`, `nifty_100`, `nifty_200`, `nifty_500`, `nifty_next_50`, `nifty_bank`,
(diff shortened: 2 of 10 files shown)


deeleeramone:
Hi @gottostartsomewhere, thanks for the PR. I think this is a reasonable ambition that is grounded in source-published data, and is accessible globally without restriction. From a future-proof perspective, new code contributions should point towards the #7489 branch, (V5) and follow those conventions for packaging and testing (dev_install is not currently ported for this branch yet..).

My apologies for none of this actually being in any of the forward-facing documentation, it's all a WIP.

Short version:

- build with hatchling
- extension should be self-contained (i.e, not rely on things like `openbb-equity` to build/launch)
  - Give it its own namespace: `obb.nse.*` 
- linting and type checking are via ruff and ty
- test coverage at 100%
- use `diskcache`, and a TTL based on the real cadence for changes in the data, so files aren't downloaded and parsed every time a function is run.

As for the content, no sense in leaving food on the table. If the weights can be surfaced, along with the factsheets, methodologies, and performance reports, that would round out the offering and provide some usable benchmarks.

gottostartsomewhere:
Thanks, this is genuinely helpful, appreciate you laying out the conventions. I'm reworking it as a self-contained `obb.nse.*` extension against the v5 branch.

Quick question on the weights before I build that part. The official niftyindices constituent files (`ind_*list.csv`) only carry the constituent list, so name, symbol, industry, series and ISIN, but no weights. As far as I can tell NSE only publishes weights in two places: the factsheet PDFs, which just show the top holdings, or the `nseindia.com/api/equity-stockIndices` endpoint, which has full weights but needs the cookie/session handling you were steering me away from.

So how would you rather I handle it?

- top holdings parsed from the factsheet PDFs (partial, but fully published and static so inching towards that)
- full weights via the nseindia API (accepting the session handling)
- or skip weights for now and ship constituents + available first

Either way, constituents and the available-indices list come straight off the static CSVs, so those stay clean regardless. Happy to go whichever direction you prefer on weights.

deeleeramone:
If not relying on a login, session/cookie handling would be fine. The data would be cleaner as well.

PR #7596, opened 2026-07-18
fix(mcp): support stdio transport on Windows

### LINKED ISSUE #7595: [Bug] MCP server stdio transport crashes silently on Windows (add_signal_handler raises NotImplementedError)
**Describe the bug**

`openbb-mcp --transport stdio` exits almost immediately on Windows, logging only `ERROR Server error: ` with no detail. The stdio transport is therefore unusable on Windows with MCP clients such as Claude Desktop/Claude Code.

There are two compounding problems:

1. **Root cause:** `stdio_main()` registers signal handlers via `loop.add_signal_handler(sig, signal_handler)`:

   https://github.com/OpenBB-finance/OpenBB/blob/aca5e6a73dea133c1d489445e2d4525d9e42e1ba/openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py#L944-L945

   `loop.add_signal_handler()` is not implemented on Windows asyncio event loops (both `ProactorEventLoop` and `SelectorEventLoop`) and raises `NotImplementedError`, so the server dies before serving anything.

2. **The failure is silent:** the exception propagates to `main()`'s broad handler:

   https://github.com/OpenBB-finance/OpenBB/blob/aca5e6a73dea133c1d489445e2d4525d9e42e1ba/openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py#L1026-L1028

   Since `str(NotImplementedError())` is an empty string, `logger.error("Server error: %s", e)` prints just `Server error: ` — no exception type, no traceback — making this very hard to diagnose.

**To Reproduce**

On any Windows machine:

```
pip install openbb-mcp-server  # 1.4.1
openbb-mcp --transport stdio
```

Output:

```
INFO Starting OpenBB MCP Server...
...
ERROR Server error:
```

Process exits with code 1 before the MCP handshake. The HTTP transport works fine; only stdio is affected.

**Suggested fix**

Signal handlers aren't essential for the stdio transport — MCP clients own the child process lifecycle and terminate it directly (and stdin EOF ends the session). So the registration can be made best-effort:

```python
for sig in (signal.SIGINT, signal.SIGTERM):
    try:
        loop.add_signal_handler(sig, signal_handler)
    except NotImplementedError:
        # add_signal_handler is not available on Windows event loops;
        # MCP clients manage the child process lifecycle directly.
        break
```

(or equivalently, skip the loop when `sys.platform == "win32"`).

Separately, it would help to change `main()`'s handler to `logger.exception("Server error: %s", e)` (or include `type(e).__name__`) so exceptions with empty string representations aren't logged as a blank message.

**Confirmation the fix works**

As a local workaround I launch the server through a small wrapper that no-ops `add_signal_handler` before calling `main()`:

```python
import asyncio, sys

if sys.platform == "win32":
    asyncio.BaseEventLoop.add_signal_handler = lambda self, sig, cb, *a: None

sys.argv = ["openbb-mcp", "--transport", "stdio"]
from openbb_mcp_server.app.app import main
main()
```

With that patch, the server initializes and serves tools over stdio correctly (verified with Claude Code as the MCP client), which confirms the signal-handler registration is the only blocker.

- OS: Windows 11 Pro (build 10.0.26200)
- Python version: 3.12.13
- openbb-mcp-server: 1.4.1 (also present on current `main`)
- fastmcp: 3.2.4

Sanjays2402:
## Description

- [x] Fixes #7595. Windows asyncio loops raise `NotImplementedError` when registering signal handlers, so stdio now treats that registration as optional and continues serving the MCP client.
- [x] No dependencies required.

## How has this been tested?

- [x] `python -m pytest tests -q` — 174 passed, 4 skipped.
- [x] Regression test simulates an event loop without signal-handler support and verifies the stdio server still starts.
- [x] `ruff check` passes on both changed files.

## Checklist

- [x] I have commented the Windows-specific behavior.
- [x] I have adhered to the GitFlow naming convention (`hotfix/mcp-stdio-windows`).
- [x] I am following the contributing guidelines.
- [x] Tests cover the bug fix.

### DIFF
--- a/openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
+++ b/openbb_platform/extensions/mcp_server/openbb_mcp_server/app/app.py
@@ -942,7 +942,12 @@ async def stdio_main(mcp_server):
         os._exit(0)  # pylint: disable=protected-access
 
     for sig in (signal.SIGINT, signal.SIGTERM):
-        loop.add_signal_handler(sig, signal_handler)
+        try:
+            loop.add_signal_handler(sig, signal_handler)
+        except NotImplementedError:
+            # Windows event loops do not support signal handlers. The MCP client
+            # owns the stdio child process and closes it directly.
+            break
 
     logger.info("Starting OpenBB MCP Server in STDIO mode. Press Ctrl+C to stop.")
 
--- a/openbb_platform/extensions/mcp_server/tests/app/test_app.py
+++ b/openbb_platform/extensions/mcp_server/tests/app/test_app.py
@@ -1,5 +1,6 @@
 """Unit tests for app module."""
 
+import asyncio
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -13,6 +14,7 @@ from openbb_mcp_server.app.app import (
     _read_system_prompt_file,
     _strip_api_prefix,
     create_mcp_server,
+    stdio_main,
 )
 from openbb_mcp_server.models.settings import MCPSettings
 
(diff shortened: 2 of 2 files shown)


DogInfantry:
Tested this on Windows 11 (build 26200). Confirmed the root cause: on a `ProactorEventLoop`, `loop.add_signal_handler(signal.SIGINT, ...)` raises `NotImplementedError` with an empty `str()`, so stdio dies before the handshake. With this branch's `try/except NotImplementedError` the server proceeds and serves over stdio. The full mcp_server test suite passes here (176 passed, 2 skipped), including the new `test_stdio_main_runs_when_signal_handlers_are_unsupported`. LGTM on the crash fix.

One note: this covers the crash, but not the second problem #7595 raises. `main()`'s handler logs `logger.error("Server error: %s", e)`, and since `str(NotImplementedError())` is empty the user sees a blank `Server error: ` with no type and no traceback. I opened a small complementary PR (#7600) that switches it to `logger.exception` and includes the exception type, so future fatal errors are diagnosable. Happy to fold it into this PR instead if you would prefer a single change.

deeleeramone (approved):
Thanks for the fix!

Issue #7582, opened 2026-07-11
[FR] Add EODHD as a data provider.


montehoover:
**What's the problem of not having this feature?**
Not currently able to add EODHD as a data provider for price and fundamentals data in the same way as Intrinio or Bezinga.

**Describe the solution you would like**
Be able to enter `provider="eodhd"`.

**Describe alternatives you've considered**
N/A

**Additional information**
https://eodhd.com/financial-apis/

deeleeramone:
Hi @montehoover,

Thanks for the feature request, V5 of OpenBB (#7489) will be _removing_ third-party data vendors from the repository, and encouraging the community to build and publish their own connectors utilizing the underlying infrastructure. The reason for this is, it is an unreasonable burden to host, maintain, and publish proprietary vendor extensions offering no mutual benefit or tangible value to open source.

The good news is, the next major version will bring in new functionality that makes it easier to personalize your runtime environment and endpoints. PR #7433, for `openbb-cli`, introduces codegen functionality where you take APIs with public OpenAPI specs, like [EOD Historical](https://eodhistoricaldata.github.io/EODHD-openapi/), and create instant extension modules from simple one-liners or config files. Available on the V5 branch, #7489, and it would actually be very helpful if you were to try it out. This is precisely the use-case it was intended for - adding a new data source without the need to write any code or burn tokens.

Would you be willing to test out this new feature set on the V5 branch?

montehoover:
I'd be happy to test out this new feature for instant extension modules. I'll give it a try and update this thread with how it goes.

DogInfantry:
I gave the V5 codegen workflow a run against EODHD to help validate it.

Using the hosted bundled spec:

```sh
openbb --generate-spec --server https://eodhd.com \
  --openapi-path https://eodhistoricaldata.github.io/EODHD-openapi/openapi.yaml \
  --output eodhd.spec
```

Results on the `v5` branch:

- `--generate-spec` writes 82 commands from the 83 GET/POST operations.
- `--generate-extension` produces an `openbb-eodhd` package with a populated `fetcher_dict` (all 82 fetchers), once the codegen path handles OpenAPI 3.1 `type: [x, null]` arrays. That handling is part of #7586 (the fix for #7585); without it `--generate-extension` raises `TypeError: unhashable type: 'list'` in `codegen/pydantic_gen.py`.
[...]

PR #7597, opened 2026-07-19
`generate-extension`: Update generated tests to get python param names instead of raw names

montehoover:
This PR fixes an issue in the tests that get generated for that showed up in [`test_eodhd_fetchers.py`](https://github.com/rockbound/openbb-eodhd/blob/main/tests/test_eodhd_fetchers.py) in the generated openbb-eodhd extension. 

There are parameter names for the API provider endpoints that are not valid python function parameters (like python keyword `from` or bracketed string `page[limit]`), and this is fine because we create aliases for those names. This PR fixes the minor bug where before we were using those raw parameter names in certain tests, and now we are correctly using the aliases. (This is a bugfix for tests inside the newly generated extension code, not tests in OpenBB.)

The test failures were `test_news_fetcher`, `test_ust_bill_rates_fetcher`, `test_ust_long_term_rates_fetcher`, `test_ust_real_yield_rates_fetcher`, and `test_ust_yield_rates_fetcher`. You can see the passing tests here: https://github.com/rockbound/openbb-eodhd/actions/workflows/tests.yml

### DIFF
--- a/cli/openbb_cli/codegen/test_gen.py
+++ b/cli/openbb_cli/codegen/test_gen.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
+from openbb_cli.codegen.pydantic_gen import safe_field_name
+
 if TYPE_CHECKING:
     from openbb_cli.codegen.fetcher_gen import GeneratedFetcher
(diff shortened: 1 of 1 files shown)


deeleeramone (approved):
Nice catch, thanks for the patch!

PR #7598, opened 2026-07-19
`generate-extension`: Update the record-vs-envelope heuristic in API payload parsing

montehoover:
This PR fixes a payload parsing issue that showed up in [`test_eodhd_fetchers.py`](https://github.com/rockbound/openbb-eodhd/blob/main/tests/test_eodhd_fetchers.py) in the generated openbb-eodhd extension. 

The API provider extensions parse response payloads and use a heuristic distinguish between a single record of data and an entire dataframe-like envelope. This PR updates the heuristic so that it doesn't treat a single record that happens to have a list datatype in it as a whole envelope. This showed up in particular in [which EODHD call?]

In general, a full envelope has many rows of records and should be a list of dicts, where each dict has the same keys and the keys represent the column names of the table/dataframe. So anything other than a json object that contains a list of dicts like this should be treated as a single record. However, certain API providers like NY Fed and FRED send payloads that contain a list where each item is not a dict but just plain strings or values of some kind. (Like a Series instead of a Dataframe.) The heuristic previously handled this by treating any payload that contained exactly one list as a envelope, but that was failing in the case where a provider (EODHD in this case) sends records that happen to contain a list datatype. 

This PR updates the heuristic to be stricter so that now a payload can only be an envelope if it contains the standard list of dicts (Dataframe-looking payload), or a single list is the only thing it contains (Series-like payload). All tests in the generated extension pass with this fix.

The test failure was `test_internal_user_fetcher`. You can see the passing tests here: https://github.com/rockbound/openbb-eodhd/actions/workflows/tests.yml

### DIFF
--- a/cli/openbb_cli/codegen/fetcher_gen.py
+++ b/cli/openbb_cli/codegen/fetcher_gen.py
@@ -303,9 +303,11 @@ def _envelope_inner_schema(
     array_keys = [
         k for k, v in props.items() if isinstance(v, dict) and v.get("type") == "array"
     ]
+    # Only resolve as an envelope if there is exactly one list of all dicts,
+    # or if the list is the only element present.
     if len(array_keys) == 1:
         items = props[array_keys[0]].get("items")
-        if isinstance(items, dict):
+        if isinstance(items, dict) and not _is_scalar_schema(items):
             return cast(dict[str, Any], items)
     return None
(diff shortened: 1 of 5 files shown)


deeleeramone (approved):
Good find, thanks for the fix!

PR #7599, opened 2026-07-19
`generate-extension`: Update class generation code to handle php-style empty data

montehoover:
This PR fixes a OpenAPI schema parsing issue that showed up in [`test_eodhd_fetchers.py`](https://github.com/rockbound/openbb-eodhd/blob/main/tests/test_eodhd_fetchers.py) in the generated openbb-eodhd extension. 

When API providers use php's json_encode on an empty object it sends `[]` as JSON, which isn't currently handled in 
`generate_class()`. This generator now produces classes that handle`[]` where applicable. All tests pass with this change incorporated.

The test failure was `test_internal_user_fetcher` on the `availableMarketplaceDataFeeds` field. You can see the passing tests here: https://github.com/rockbound/openbb-eodhd/actions/workflows/tests.yml


deeleeramone (approved):
Excellent, thanks for your help!

PR #7601, opened 2026-07-21
Fix codegen field-name aliasing so generated extensions are callable

T-Arjun:
## Summary

Testing the new `--generate-spec` / `--generate-extension` codegen tool against a real API (World Bank Data360, a genuinely public/free API with no key required) surfaced a bug that makes every codegen-generated extension unusable through the standard `obb.<namespace>.<command>()` interface whenever a query parameter's wire name isn't already snake_case.

- `safe_field_name()` in `pydantic_gen.py` only aliases a JSON property name when it isn't a *syntactically valid* Python identifier. camelCase names (e.g. `datasetId`) are valid identifiers, so they were kept verbatim as the Pydantic field name.
- `ProviderInterface` (`provider_interface.py`) renders every non-standard provider field through `to_snake_case()` for its public docstrings/extra-params dataclass, with no matching rename applied back to the underlying model.
- Net effect: the field's "documented" name (snake_case) and its real name (camelCase) diverge, and the parameter can no longer be supplied at all — confirmed live: `obb.worldbank_data360.indicators(dataset_id="WB_WDI")` raised `Field required` for `datasetId` even though the value was passed.

Fix: snake_case every generated field and `alias=` it back to the original wire name when it differs — the same convention every hand-written OpenBB provider already follows (snake_case field + `Field(alias=...)`, dumped with `by_alias=True` in the fetcher, which the codegen tool's fetcher output already assumed).

Fixing that exposed a second, latent bug in `post_gen.py`: its signature builder (`_signature_params`) and its request-body builder (`_render_body_block`) independently derived variable names from the same raw params, and only one side ran names through `safe_field_name`. The two happened to agree before this change only because `safe_field_name` was near-inert on camelCase input. Once it started snake-casing, any generated POST command with a camelCase body field (e.g. `searchFields`) raised `NameError: name 'search_fields' is not defined` on every real call. Threaded both the safe (Python-valid) name and the original raw wire name through `_signature_params` so the signature, the JSON body dict keys, and the query-string keys all agree.

## Verification

- Generated an `openbb-worldbank-data360` extension end to end from the World Bank's own published OpenAPI spec (https://github.com/worldbank/open-api-specs), installed it, ran `openbb-build`, and called it through the real `obb.*` interface against the live API — `obb.worldbank_data360.indicators(...)` and `obb.worldbank_data360.data(...)` both return real rows.
- Full `cli/` test suite: 2005 passed (2 pre-existing, unrelated flakes deselected — a session backend mock test and a date-rollover test, both fail identically on `v5` with none of this diff applied).
- Added a regression test in each affected file. The `post_gen.py` one actually `exec`s the generated module and calls it, since `ast.parse` alone doesn't catch a `NameError` that only fires at call time (both names involved are individually valid Python identifiers).
- `ruff check` / `ruff format --check` clean on all four changed files.

## Test plan

- [x] `cli/openbb_cli/codegen/pydantic_gen.py` — updated 3 existing tests that asserted the old (buggy) unaliased behavior, added 1 new test
- [x] `cli/openbb_cli/codegen/post_gen.py` — updated 1 existing test for the new 6-tuple param shape, added 1 new end-to-end regression test
- [x] Manually verified against the live World Bank Data360 API (not mocked)

### DIFF
--- a/cli/openbb_cli/codegen/post_gen.py
+++ b/cli/openbb_cli/codegen/post_gen.py
@@ -321,8 +321,15 @@ def _signature_params(
     array_field: str | None,
     array_annotation: str | None,
     has_credentials: bool,
-) -> list[tuple[str, str, str | None, bool, Any]]:
-    """Build the function parameter list (name, annotation, description, required, default).
+) -> list[tuple[str, str, str | None, bool, Any, str]]:
+    """Build the function parameter list (safe_name, annotation, description, required, default, raw_name).
+
+    ``safe_name`` is what actually gets declared in the ``async def`` signature
+    and referenced in the function body (snake_case, aliased via
+    ``safe_field_name`` exactly like the sibling Pydantic-model generator).
+    ``raw_name`` is the original on-the-wire key from the spec, kept alongside
+    so ``_render_body_block`` can still emit the correct JSON/query key while
+    referencing the Python-valid variable.
 
     Parameters
     ----------
@@ -340,7 +347,7 @@ def _signature_params(
     list of tuple
         Per-parameter tuples in declaration order.
     """
-    out: list[tuple[str, str, str | None, bool, Any]] = []
+    out: list[tuple[str, str, str | None, bool, Any, str]] = []
     if has_credentials:
         out.append(
             (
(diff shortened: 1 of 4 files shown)


T-Arjun:
Closing this for now — thanks for the review time regardless.

Issue #7585, opened 2026-07-13
[Bug] Silent failure when parsing modular OpenAPI 3.1.0 specs


vridhib:
### Context

While testing the new V5 codegen workflow suggested by @deeleeramone in issue #7582, I attempted to generate an extension using the OpenAPI 3.1.0 spec (https://github.com/EodHistoricalData/EODHD-openapi) listed on EODHD's website.

Since EODHD does not host a public OpenAPI spec on their API domain, I cloned their modular OpenAPI repo (which splits 84 endpoints across different YAML files using `$ref`) and served it locally.

### Steps to Reproduce
1. Clone the `v5` branch and install `openbb-cli`.
2. Clone the `EodHistoricalData/EODHD-openapi` repo (`git clone https://github.com/EodHistoricalData/EODHD-openapi`) and serve the root `openapi.yaml` locally via `python3 -m http.server 8000`.
3. Run: `openbb --generate-spec --server http://localhost:8000 --openapi-path /openapi.yaml --output eodhd.spec`
4. Run: `openbb --generate-extension --spec eodhd.spec --provider-name eodhd --output ./openbb-eodhd`

### Expected Behavior
The codegen parser should either successfully map the modular YAML `$ref` tags into OpenBB's internal command schema, or it should throw a clear `ValidationError`/`ParseError` to warn the user that the spec format is incompatible.

As noted in the CLI help docs, YAML is an accepted format: 
```
--openapi-path PATH   Path (or full URL) to the OpenAPI document on the server. Defaults to /openapi.json. Servers that publish under a different name (e.g. NY Fed at /static/docs/markets-api.yml) need this. JSON or YAML accepted.
```

### Actual Behavior
The tool completes without throwing a traceback, but outputs a skeleton package with zero commands, as shown here:

```bash
wrote 0 commands to eodhd.spec
  providers (1): eodhd
  routers (1): eodhd
  fetchers: 0 GET (across all providers)
  POST:     0 local-compute commands
```

The generated `__init__.py` contains an empty `fetcher_dict={}`. If a user sees the "success" message, they will install the package assuming it works, yet discover that it has no commands.

### Environment
- OpenBB CLI v5 branch (post-merge of #7433)
- Python version: 3.11
- OS: Debian 12
- EODHD OpenAPI 3.1.0 spec (https://github.com/EodHistoricalData/EODHD-openapi)

deeleeramone:
[...]

PR #7606, opened 2026-07-25
fix(cli): inherit path-item-level OpenAPI parameters

DerMayer1:
### What

OpenAPI 3.x lets a [Path Item Object](https://spec.openapis.org/oas/v3.1.0#fixed-fields-6) declare `parameters` that apply to every operation under that path, with operation-level entries overriding them by `(name, in)`. The V5 codegen only read `op.get("parameters")`, so path-item parameters were dropped — silently. The generated command kept the URL template but had nothing registered to satisfy it.

This is the same silent-success failure surface as #7585, from a different root cause: the generator reports success, and the breakage only shows up when the user calls the command.

### Real-world repro

The [Xero Accounting API](https://github.com/XeroAPI/Xero-OpenAPI) declares its **required** `xero-tenant-id` header once per path item, across all 138 paths:

```yaml
/Accounts/{AccountID}:
  parameters:
    - $ref: '#/components/parameters/requiredHeader'   # xero-tenant-id, in: header, required: true
  get:
    parameters:
      - $ref: '#/components/parameters/AccountID'
```

Running `build_command_spec` against the live spec:

| | before | after |
| --- | --- | --- |
| commands generated | 78 | 78 |
| carrying required `xero-tenant-id` | **0** | **78** |

Xero returns 403 without that header, so before this change every generated command was dead on arrival with no warning.

### How

Adds `operation_parameters(spec, path_item, op)` to `openapi_schema.py`. It merges the two levels, dereferences `$ref` on both, and applies the spec's override rule keyed on `(name, in)` — so a same-named parameter in a different location is correctly kept rather than swallowed.

Both consumers route through it:

- `build_command_spec` → `_build_operation_entry` — the pre-computed `--generate-spec` path
- `build_command_index` → `build_parser_from_operation` — the interactive `--server` path

`build_router_map`, `build_reference`, and `detect_api_prefix` only read path keys and descriptions, so they're unaffected.

Unresolvable `$ref`s, non-dict entries, and nameless parameters are skipped exactly as before. `build_parser_from_operation` keeps its previous operation-only behavior when called without a `spec`, and specs with no path-item parameters produce byte-identical output.

### Tests

14 new tests covering: inheritance, `$ref` resolution on both levels, operation-over-path-item override, `(name, in)` keying, inheritance across multiple operations under one path item, and the defensive skip paths.

```
2019 passed
ruff format --check   clean
ruff check            clean
ty check              clean
```

(2005 before this change, plus the 14 new.)

### DIFF
--- a/cli/openbb_cli/codegen/post_gen.py
+++ b/cli/openbb_cli/codegen/post_gen.py
@@ -385,7 +385,11 @@ def _signature_params(
         )
         out.append(
             (
-                name,
+                # ``_render_body_block`` already writes ``{'from': from_}``, keying the
+                # payload by the wire name and reading the safe identifier, so the
+                # signature has to declare the safe one. Codat's Transfer body has a
+                # property named ``from``, which is a keyword.
+                safe_field_name(name)[0],
                 ann,
                 schema.get("description") or schema.get("title"),
                 name in body_required,
@@ -639,8 +643,13 @@ def _render_body_block(
     """
     body_props = (cmd_spec.get("request_body_schema") or {}).get("properties") or {}
     body_field_names = list(body_props)
+    # The signature declares the safe identifier for a body property, so the
+    # body/query split has to compare against those, not the wire names.
+    # Otherwise a keyword-named property such as ``from`` is not recognized as a
+    # body field and leaks into the query string as ``from_``.
+    body_safe_names = {safe_field_name(n)[0] for n in body_field_names}
     query_field_names = [
-        n for n, _, _, _, _ in params if n != "cc" and n not in body_field_names
+        n for n, _, _, _, _ in params if n != "cc" and n not in body_safe_names
     ]
 
     lines: list[str] = []
--- a/cli/openbb_cli/codegen/post_gen.py
+++ b/cli/openbb_cli/codegen/post_gen.py
@@ -380,13 +380,12 @@ def _signature_params(
     for name, schema in body_props.items():
         if name == array_field:
             continue
-        if name in operation_param_names:
-            continue
-        if safe_field_name(name)[0] in operation_param_names | emitted:
-            continue
         if not isinstance(schema, dict):
             continue
-        emitted.add(safe_field_name(name)[0])
+        safe = safe_field_name(name)[0]
+        if safe in operation_param_names or safe in emitted:
+            continue
+        emitted.add(safe)
         ann = _python_type_from_param(
             {
                 "type": schema.get("type"),
@@ -400,7 +399,7 @@ def _signature_params(
                 # payload by the wire name and reading the safe identifier, so the
                 # signature has to declare the safe one. Codat's Transfer body has a
                 # property named ``from``, which is a keyword.
-                safe_field_name(name)[0],
+                safe,
                 ann,
                 schema.get("description") or schema.get("title"),
                 name in body_required,
--- a/cli/openbb_cli/codegen/post_gen.py
+++ b/cli/openbb_cli/codegen/post_gen.py
@@ -403,7 +403,11 @@ def _signature_params(
         ann = _python_type_from_param(raw)
         out.append(
             (
-                name,
+                # Wire names are not always identifiers. Rebilly declares
+                # ``Organization-Id`` and a ``REB-APIKEY`` security header, which
+                # emitted ``REB-APIKEY: str = None``. The fetcher path already
+                # sanitizes these; this one has to match it.
+                safe_field_name(name)[0],
                 ann,
                 raw.get("help"),
                 bool(raw.get("required")),
@@ -651,6 +655,13 @@ def _render_body_block(
     query_field_names = [
         n for n, _, _, _, _ in params if n != "cc" and n not in body_safe_names
     ]
+    # The signature carries safe identifiers, but the request has to go out under
+    # the wire names, so keep a map back for the query keys and the path check.
+    wire_by_safe = {
+        safe_field_name(p["name"])[0]: p["name"]
+        for p in filter_user_params(cmd_spec.get("parameters") or [])
+        if p.get("name")
+    }
 
     lines: list[str] = []
     lines.extend(cred_lines)
@@ -663,10 +674,11 @@ def _render_body_block(
 
     lines.append("    _query_dict: dict[str, Any] = {}")
     for name in query_field_names:
-        if name in path_params:
+        wire = wire_by_safe.get(name, name)
+        if wire in path_params or name in path_params:
             continue
         lines.append(f"    if {name} is not None:")
-        lines.append(f"        _query_dict[{name!r}] = {name}")
+        lines.append(f"        _query_dict[{wire!r}] = {name}")
     for canonical, info in creds.items():
         if info["in"] != "query":
             continue
--- a/cli/openbb_cli/codegen/post_gen.py
+++ b/cli/openbb_cli/codegen/post_gen.py
@@ -364,18 +364,29 @@ def _signature_params(
     # placeholder and its own ``required``. The payload is unaffected, because
     # ``_render_body_block`` reads the body fields from ``request_body_schema``
     # rather than from this list.
-    operation_param_names = {
-        p.get("name")
-        for p in filter_user_params(cmd_spec.get("parameters") or [])
-        if p.get("name")
-    }
+    #
+    # Two wire names can also collapse onto one identifier once sanitized
+    # (``from`` and ``from_``, ``Organization-Id`` and ``Organization_Id``), and
+    # ``operation_parameters`` keeps a name declared twice under different ``in``
+    # locations. Collisions are therefore resolved on the emitted identifier, not
+    # on the wire name, so a duplicate argument cannot be generated at all.
+    operation_params = [
+        p for p in filter_user_params(cmd_spec.get("parameters") or []) if p.get("name")
+    ]
+    operation_param_names = {p["name"] for p in operation_params}
+    operation_param_names |= {safe_field_name(p["name"])[0] for p in operation_params}
+    emitted: set[str] = {entry[0] for entry in out}
+
     for name, schema in body_props.items():
         if name == array_field:
             continue
         if name in operation_param_names:
             continue
+        if safe_field_name(name)[0] in operation_param_names | emitted:
+            continue
         if not isinstance(schema, dict):
             continue
+        emitted.add(safe_field_name(name)[0])
         ann = _python_type_from_param(
             {
                 "type": schema.get("type"),
@@ -396,10 +407,12 @@ def _signature_params(
                 schema.get("default"),
             )
         )
-    for raw in filter_user_params(cmd_spec.get("parameters") or []):
-        name = raw.get("name")
-        if not name:
+    for raw in operation_params:
+        name = raw["name"]
+        safe = safe_field_name(name)[0]
+        if safe in emitted:
             continue
+        emitted.add(safe)
         ann = _python_type_from_param(raw)
         out.append(
             (
@@ -407,7 +420,7 @@ def _signature_params(
                 # ``Organization-Id`` and a ``REB-APIKEY`` security header, which
                 # emitted ``REB-APIKEY: str = None``. The fetcher path already
                 # sanitizes these; this one has to match it.
-                safe_field_name(name)[0],
+                safe,
                 ann,
                 raw.get("help"),
                 bool(raw.get("required")),
--- a/cli/openbb_cli/codegen/post_gen.py
+++ b/cli/openbb_cli/codegen/post_gen.py
@@ -357,8 +357,22 @@ def _signature_params(
     body_required = set(
         (cmd_spec.get("request_body_schema") or {}).get("required") or []
     )
+    # A name can be declared both as an operation parameter and as a request-body
+    # property — Codat's push endpoints put ``accountId`` in the URL template and
+    # in the body. Emitting both produces two parameters with the same name, which
+    # is a SyntaxError. The operation parameter wins: it carries the path
+    # placeholder and its own ``required``. The payload is unaffected, because
+    # ``_render_body_block`` reads the body fields from ``request_body_schema``
+    # rather than from this list.
+    operation_param_names = {
+        p.get("name")
+        for p in filter_user_params(cmd_spec.get("parameters") or [])
+        if p.get("name")
+    }
     for name, schema in body_props.items():
         if name == array_field:
+            continue
+        if name in operation_param_names:
             continue
         if not isinstance(schema, dict):
             continue
--- a/cli/openbb_cli/dispatchers/openapi_schema.py
+++ b/cli/openbb_cli/dispatchers/openapi_schema.py
@@ -347,7 +347,7 @@ def extract_request_body_schema(
     schema = media.get("schema")
     if not isinstance(schema, dict):
         return None
-    return deref_schema(spec, schema)
+    return merge_allof(deref_schema(spec, schema))
 
 
 def extract_response_schemas(
--- a/cli/openbb_cli/dispatchers/openapi_schema.py
+++ b/cli/openbb_cli/dispatchers/openapi_schema.py
@@ -190,23 +190,32 @@ def deref_schema(
     return node
 
 
-_MERGEABLE_ALLOF_KEYS = frozenset({"properties", "required", "type", "description"})
+_ALLOF_BLOCKING_KEYS = frozenset(
+    {"oneOf", "anyOf", "allOf", "not", "$ref", "enum", "const", "items"}
+)
 
 
 def _mergeable_allof_members(members: list[Any]) -> list[dict[str, Any]] | None:
     """Return the members if every one is a plain object subschema, else ``None``.
 
-    Only object-shaped members are merged. A member carrying its own combinator
-    (``oneOf`` / ``anyOf`` / a nested ``allOf`` that did not collapse), or a
-    non-object ``type``, means the composition is not a simple intersection of
-    property bags and is left untouched rather than merged into something the
-    spec does not describe.
+    A member is not mergeable when it carries its own combinator (``oneOf`` /
+    ``anyOf`` / a nested ``allOf`` that did not collapse), an unresolved
+    ``$ref`` left behind by a reference cycle, a scalar constraint (``enum`` /
+    ``const`` / ``items``), or a ``type`` other than ``object``. Any of those
+    means the composition is not a plain intersection of property bags, and
+    merging it would describe something the spec does not.
+
+    Everything else is annotation that does not affect the intersection —
+    ``title``, ``description``, ``examples``, ``definitions``, vendor ``x-``
+    keys — and does not block the merge. Keying off a blocklist rather than a
+    whitelist matters: Codat's ``PagingInfo`` carries ``definitions``, and
+    treating that as disqualifying left the composition unmerged.
     """
     out: list[dict[str, Any]] = []
     for member in members:
         if not isinstance(member, dict):
             return None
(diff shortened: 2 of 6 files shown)


DerMayer1:
Some additional data on how common this pattern is, in case it helps with prioritizing.

I ran a path-item-parameter check across the [APIs.guru](https://apis.guru) directory — 2,510 published specs fetched successfully. **673 of them (27%) declare parameters at the path-item level**, so every command generated from those specs currently drops them:

| API | affected paths |
| --- | --- |
| Microsoft Graph (beta) | 12,001 / 14,227 |
| Microsoft Graph | 6,517 / 7,419 |
| AWS EC2 | 591 / 591 |
| Google Compute | 487 / 487 |
| Kubernetes | 368 / 423 |
| Xero Accounting | 132 / 132 |
| Codat (accounting) | 118 / 124 |
| Rebilly (billing) | 104 / 105 |

Restricted to the finance/banking/payments subset, it's 25 of 160 specs (16%). In Xero's and Codat's case the inherited parameter is a *required* auth header, so the generated commands can't succeed at all — which is how I ran into it.

While I had the corpus loaded I also checked three adjacent conformance gaps, and none of them look worth writing code for:

| Gap | specs affected (of 2,510) |
| --- | --- |
| parameter-level `content` instead of `schema` | 4 (0.2%) |
| `allOf` in a parameter schema | 24 (1%), mostly Swagger 2.0 |
| path-level `servers` override | 57 (2.3%) |
| operation-level `servers` override | 33 (1.3%) |

The `servers` ones are the only ones with a real severity story — when a spec overrides the host per path, generated commands would point at the wrong server silently. It's concentrated almost entirely in Twilio (121/121 paths on their main API, plus several of their other products), with Linode, Box and GitHub Enterprise using the operation-level form. Happy to open a separate issue or PR if multi-host specs are in scope for the codegen, but I didn't want to add handling for something that may be deliberately out of scope.

deeleeramone:
Hey @DerMayer1, thanks for the PR.

Did you try adding the `-H xero-tenant-id {..value...}` flag to the command? A header is not the same thing as a path param, those are handled differently.

```
  -H, --header KEY=VALUE
                        Custom HTTP header to send on every dispatch and on the OpenAPI fetch. Repeatable:
                        ``-H 'Authorization: Bearer xxx' -H 'X-Tenant: acme'``. Both ``KEY=VALUE`` and ``KEY:
                        VALUE`` forms are accepted.
  --header-file PATH    Read additional headers from a JSON file (object of string values). Headers from
                        --header take precedence on conflicts. (env: OPENBB_HEADER_FILE)
  -Q, --query-param KEY=VALUE
                        Query parameter injected on every request (e.g. APIs that authenticate via
                        ``?api_key=...`` like https://api.congress.gov). Repeatable. Also auto-loaded from
                        any env var prefixed ``OPENBB_HTTP_QUERY_`` — ``OPENBB_HTTP_QUERY_API_KEY=xxx``
                        becomes ``?api_key=xxx``. CLI flag takes precedence over env.
  --query-param-file PATH
                        Read additional query params from a JSON file (object of string values). --query-
                        param flags and ``OPENBB_HTTP_QUERY_*`` env vars take precedence on conflicts. (env:
                        OPENBB_QUERY_PARAM_FILE)
```

If these do not work as-expected, then that is a bug.

DerMayer1:
You're right about the header, and my example was a bad one. A constant `xero-tenant-id` is exactly what `-H` is for, and I should have checked that before using it as the motivating case.

The general problem isn't headers though. Path-item `parameters` can be any location, so I went back to the corpus and broke it down by `in:` across the 673 specs that use them:

| `in:` | occurrences | reachable via existing flags |
| --- | --- | --- |
| `header` | 70,442 | yes — `-H`, as you say |
| `query` | 50,973 | mostly — `-Q`, when the value is constant for the run |
| **`path`** | **48,447** | **no** |

**108 of the 673 specs declare `in: path` parameters at the path-item level.** Those become `{placeholder}` segments in the URL template, and `-H`/`-Q` can't fill them, so the command can't be dispatched regardless of flags. Three examples, before this PR vs after:

```
Codat Accounting — /companies/{companyId}/connections/{connectionId}/data/bankAccounts
  before: params ['Authorization','page','pageSize','query','orderBy']
          unsatisfiable placeholders: ['companyId','connectionId']
  after:  unsatisfiable placeholders: none

Linode — /account/entity-transfers/{token}
  before: unsatisfiable placeholders: ['token']

Kubernetes — /api/v1/componentstatuses/{name}
  before: zero parameters registered; unsatisfiable: ['name']
```

Others in that set: Microsoft Graph (12,001 paths), Keycloak (191), Apple App Store Connect (133), Bitbucket (117), Jira (115).

On scope, if you'd rather this only inherit non-header parameters and leave headers to `-H`, that's a small change and I'm happy to make it. The one thing I'm unsure of is whether the `--generate-extension` output has an equivalent of `-H`, since a generated provider package is consumed as a library rather than through CLI flags. If it does, narrowing to `path`/`query` is clearly the cleaner fix.

deeleeramone:
Can you try reframing the problem with clear scope? There are certainly API endpoints which will never universally translate into Python Interface functions, such as webhooks, OIDC, websockets, etc.

The codegen detects SecuritySchema and wraps it into the `Credentials` via the `Provider` instance. An OAuth2 flow requires a **running server**, so this type of authorization is not suitable for the Python interface. To use services like Microsoft Graph directly via Python, you would need to use their official package for managing identity and write custom code for the specific implementation.

Path and query parameters are already handled across all interfaces, but your example won't track across all the surfaces because a Python function does not have "headers" by its very nature. Only external HTTP requests to an API server carry headers - i.e, `obb.some_function()` will never have request headers. Converting "in:header" to become a Python parameter is likely undesirable for the majority of scenarios.

If the goal is to handle an `Authorization` header specifically, there could be an argument made for folding into the SecuritySchema detection and codegen paths. 

The config system (`openbb.toml`, `SystemSettings`) allows you to set custom outbound headers for HTTP requests, at the user/environment level. This is likely the lever you are looking for.

### `SystemSettings.PythonSettings`

`model_config = ConfigDict(extra="allow")`.

| Field | Type | Default |
| ----- | ---- | ------- |
| `docstring_sections` | `list[str]` | `["description", "parameters", "returns", "examples"]` |
| `docstring_max_length` | `PositiveInt \| None` | `None` |
| `http` | `dict \| None` | `{}` |
| `uvicorn` | `dict \| None` | `{}` |

`docstring_sections` controls which sections are included in autogenerated docstrings.

`docstring_max_length` caps the length of those docstrings. `None` means no cap.

`http` is consumed by `openbb_core.provider.utils.helpers.{make_request, amake_request, amake_requests, get_requests_session, get_async_requests_session}`. Keys:

| Key | Applies to | Purpose |
| --- | ---------- | ------- |
| `cafile` | both | Path to a CA certificate file. |
| `certfile` | both | Path to a client certificate file. |
| `keyfile` | both | Path to a client key file. |
| `password` | aiohttp | Password for the client key file. |
| `verify_ssl` | both | Verify SSL certificates. |
| `fingerprint` | aiohttp | SSL fingerprint. |
| `proxy` | both | Proxy URL. |
| `proxy_auth` | aiohttp | Proxy authentication. |
| `proxy_headers` | aiohttp | Proxy headers. |
| `timeout` | both | Request timeout. |
| `auth` | both | Basic authentication. |
| `headers` | both | Request headers. |
| `cookies` | both | Session cookies. |

Additional keys are ignored unless explicitly implemented by custom code.

DerMayer1:
That reframing is right, and I've pushed a commit that narrows the PR to match it.

**Scope, explicitly:**

**Out.** Headers and cookies. You're correct that `obb.some_function()` has no request headers, so turning `in: header` into a Python parameter is wrong for the majority case — and outbound headers already have a home in `SystemSettings.PythonSettings.http.headers` and the CLI's `-H`. Authorization specifically belongs in the SecuritySchema → Credentials path, not here. Also out: webhooks, OIDC, websockets, and OAuth2 flows, which need a running server and don't map onto a Python function at all.

**In.** `in: path` and `in: query` parameters declared on the **Path Item Object** rather than on the operation.

That's the whole remaining claim. Path and query parameters *are* handled across all interfaces when they're declared on the operation — what isn't handled is the same parameter declared one level up, which the codegen never read. End to end on Codat's accounting spec:

```
command: companies.connections.data.bankAccounts
url:     /companies/{companyId}/connections/{connectionId}/data/bankAccounts

before: registered params ['Authorization','page','pageSize','query','orderBy']
        $ openbb ... --companyId abc --connectionId xyz
          error: unrecognized arguments: --companyId abc --connectionId xyz
        dispatched URL: /companies/{companyId}/connections/{connectionId}/data/bankAccounts

after:  --companyId / --connectionId accepted
        dispatched URL: /companies/abc/connections/xyz/data/bankAccounts
```

The argument can't be passed, and the unsubstituted placeholder goes out on the wire.

Narrowing shrinks the blast radius: of the 673 specs in the APIs.guru corpus that use path-item parameters, **396 declare path or query there** (108 path, 313 query) and are affected; the remaining 277 are header/cookie-only and this PR now correctly ignores them. Xero — my original example — is in that ignored group, which I think is the right outcome given `-H` exists.

Changed: inherited parameters are filtered to `path`/`query` before merging. Operation-level parameters are untouched, so an operation declaring its own header behaves exactly as today. 17 tests, full suite 2,022 green, ruff and ty clean.

deeleeramone:
Thanks for the details, @DerMayer1. This resolves the params side, and the security schema builds correctly as a provider credential that gets inserted into the request headers; however, the response schema is not getting dug up from the $ref. If you can apply the same treatment for those, with consideration that some may contain nested references, that will round it out nicely! 

Screenshot 2026-07-26 at 4 51 39 PM

Screenshot 2026-07-26 at 5 02 35 PM

```
  /companies/{companyId}/connections/{connectionId}/data/bills:
    parameters:
    - $ref: '#/components/parameters/companyId'
    - $ref: '#/components/parameters/connectionId'
    - $ref: '#/components/parameters/page'
    - $ref: '#/components/parameters/pageSize'
    get:
      operationId: listBills
      tags:
      - Accounting
      summary: List bills
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/DataList'
...
components:
  parameters:
    connectionId:
      name: connectionId
      in: path
      required: true
      description: Unique identifier for a connection.
      schema:
        type: string
        format: uuid
    companyId:
      name: companyId
      in: path
      required: true
      description: Unique identifier for a company.
      schema:
        type: string
        format: uuid
    pageSize:
      name: pageSize
      in: query
      description: Number of records per page.
      schema:
        type: integer
        default: 100
    page:
      name: page
      in: query
      description: Page number (1-indexed).
      schema:
        type: integer
        default: 1
  schemas:
    DataList:
      type: object
      properties:
        results:
          type: array
          items:
            type: object
        pageNumber:
          type: integer
        pageSize:
          type: integer
        totalResults:
          type: integer
  securitySchemes:
    authHeader:
      type: apiKey
      in: header
      name: Authorization
      description: 'Codat expects your API key Base64-encoded and sent in the Authorization header using the Basic scheme, for example: `Authorization: Basic YOUR_BASE64_ENCODED_API_KEY`.'
```

DerMayer1:
Pushed. I went in expecting the refs and spent a while there before realising they were already fine. `deref_schema` expands them recursively, nested included, and `listBills` on `v5` comes back with zero unresolved `$ref`. What actually breaks it is the `allOf` sitting on top:

```
allOf[0] -> properties: [results]
allOf[1] -> properties: [_links, pageNumber, pageSize, totalResults]
```

Nothing merges those, so the properties are one level below where anything looks for them. `_data_schema` asks for `properties.results`, gets `{}`, and hands off to `_unwrap_schema_envelopes`, which knows `oneOf` and `anyOf` but not `allOf`. So the composition goes straight through untouched and the `Data` class comes out empty, which is what you're seeing.

`merge_allof` flattens it. Properties get unioned with the first declaration winning any conflict, `required` gets unioned, and anything written next to the `allOf` beats what the members say. It runs after `deref_schema` in both `extract_response_schema` and `extract_response_schemas` and recurses, so a composition buried in `results.items` also collapses and the refs inside members stay resolved.

I left compositions alone when they aren't a plain intersection of objects, so a member with its own `oneOf`/`anyOf` or a non-object `type` passes through as-is. Flattening those would mean guessing at a shape the spec never states.

`listBills` now:

```
before: ['allOf', 'x-internal'], properties []
after:  properties ['results', '_links', 'pageNumber', 'pageSize', 'totalResults']
        required   ['pageNumber', 'pageSize', 'totalResults', '_links']
        results.items -> Bill, resolved
```

Across the specs I had loaded, counting operations whose response schema actually exposes properties:

| | before | after |
| --- | ---: | ---: |
| Codat Accounting (127 ops) | 36 | 88 |
| Rebilly (184) | 62 | 104 |
| Box (258) | 145 | 183 |
| Codat Banking (8) + Commerce (11) | 1 | 19 |

One thing I didn't touch: `extract_request_body_schema` has the same gap on the request side. Seemed better to keep this to the response schemas you asked about, but I can add it here or in its own PR if you want it.

7 new tests, 157 in that file. `cli/tests` gives me 1,945 passed vs 1,938 on the base, same 80 failures either way (platform extras I don't have installed). ruff clean.

deeleeramone:
> One thing I didn't touch: extract_request_body_schema has the same gap on the request side. Seemed better to keep this to the response schemas you asked about, but I can add it here or in its own PR if you want it.

Seems like this would be a good idea to handle it here, eh?

DerMayer1:
Done, pushed.

Same treatment on `extract_request_body_schema`. The failure mode there is a bit blunter than on the response side: `request_body_parameters` needs `type: object` or `properties` to flatten a body, an `allOf` gives it neither, so the command just ends up with no body parameters.

Request bodies that produce parameters, same specs as before:

| | before | after |
| --- | ---: | ---: |
| Codat Accounting (30 bodies) | 1 | 26 |
| Rebilly (75) | 50 | 63 |
| Box (101) | 87 | 88 |

Codat is the one that really moves, since nearly every prototype schema there is a composition.

158 tests in that file now, `cli/tests` at 1,946 passed with the same 80 environment failures as the base. ruff clean.

deeleeramone:
Hey @DerMayer1, I'm still not seeing the response model being captured. It also seems to have duplicated some parameters, which generates invalid syntax. Potentially fixed one thing but broke another?

Screenshot 2026-07-29 133540

Screenshot 2026-07-29 133713

I'm not familiar with this service, so are these "options" endpoints valid or are they being interpreted incorrectly?

Screenshot 2026-07-29 132919

DerMayer1:
Both real. The duplicate isn't from this PR, the empty response model was my fault for a reason I didn't expect, and chasing them turned up a third thing.

Taking the duplicate first. That path item declares no `parameters` at all, only `post`, so path-item inheritance never touches it. Codat declares `accountId` as a path parameter and repeats it as a body property, and `_signature_params` appended body properties and then operation parameters without deduplicating. I built a tree from `origin/v5` and ran your spec through it:

```
v5      [cc, transactions, companyId, connectionId, accountId, accountId, timeoutInMinutes, allowSyncOnPushComplete]
branch  [cc, transactions, companyId, connectionId, accountId, timeoutInMinutes, allowSyncOnPushComplete]
```

The operation parameter wins now, since it carries the path placeholder. The payload is unchanged because `_render_body_block` builds the body from `request_body_schema`, so `accountId` still goes out in the body.

The response model is on me, though not for the reason you'd guess. I validated the merge against the APIs.guru copy of Codat 2.1.0 and you generate from `codatio/oas/main/yaml/Codat-Accounting.yaml`. The live `PagingInfo` carries a `definitions` key next to its properties, my member check was a whitelist of permitted keywords, and `definitions` disqualified the whole composition so it never merged. My numbers were real and measured on a document you weren't using, which is worse than having no numbers.

It's a blocklist now: a member blocks the merge only if it has its own combinator, an unresolved `$ref` from a cycle, `enum`/`const`/`items`, or a non-object `type`. `definitions`, `title`, `examples` and vendor keys don't. `connections.data.accountTransactions` now resolves to `results`, `pageNumber`, `pageSize`, `totalResults`, `_links`, and its Data class carries the AccountTransaction fields instead of `pass`.

The third thing is the one I'd look at first. `package_gen` decides between a fetcher and a POST command with `has_body`, which requires `properties`. An unmerged `allOf` body has none, so on v5 **14 of Codat's 24 POST endpoints were generated as GET fetchers**, including `connections.push.bills`, `push.customers` and `push.invoices`. They couldn't push at all. All 24 route correctly now, which is what the 81/10 to 67/24 split in the output is.

That reclassification then exposed two things I did break, both found by generating the extension and importing every module rather than reading the diff. Codat's Transfer body has a property named `from`, so once bodies flattened the signature emitted `from: str = None`. `_render_body_block` was already writing `{'from': from_}` via `safe_field_name`, so only the signature side was wrong. Fixing that left `from_` leaking into the query string, because the body/query split compared signature names against wire names. Both fixed, and the same sanitization was missing for operation parameters, which Rebilly hits with `Organization-Id` and its `REB-APIKEY` security header.

Generating full extensions and importing every module:

| | v5 | branch |
| --- | ---: | ---: |
| Codat, modules that fail to import | 5 | 0 |
| Rebilly, files that fail to compile | 14 | 1 |

The 5 on Codat is one `SyntaxError` cascading into `routers.codat`, `routers.connections` and `routers.connections_push`, so the router is unimportable on v5. Rebilly's remaining failure is `async def 3dsecure(` in the router, which fails the same way on v5; sanitizing it means choosing a name for the command, so I left it.

Also ran the wider check again after loosening the member rule: 2,816 response and request-body extractions across 6 specs, 567 changed, every change traceable to an `allOf` in the pre-merge output, none unexplained. Twilio declares no compositions and is byte-identical. Two consecutive generations produce byte-identical files apart from the provenance timestamp.

On the options endpoints, those are genuine. Codat's create/update model endpoints sit at `/connections/{connectionId}/options/{dataType}` and return the field metadata you need before pushing, which is why that Data class comes out as `value` / `type` / `displayName` / `description` / `required`.

Three things I deliberately left alone. Non-credential header parameters end up in the query string, but that's true on the fetcher path too and you've already said `in: header` shouldn't become a Python parameter, so I didn't want to reopen it here. `write_text` has no `encoding=` at any of its 13 call sites, which kills generation on Windows for specs containing non-cp1252 characters (Codat has a `﻿`); I worked around it with `PYTHONUTF8=1` and it looks like its own fix. And a parameter named `cc` would now be dropped rather than colliding with the `CommandContext` argument, which is better than a SyntaxError but is still a silent drop; renaming it invents a public parameter name so it seemed like your call. None of the three occurs in Codat or Rebilly.

10 tests added across the two files. `cli/tests` at 1,952 passed against the same 80 environment failures as the base. ruff clean.

DerMayer1:
Correction to my last comment: I said 10 tests added, it is 31 across the PR and 6 since that comment. Overstating my own verification is the wrong direction to be wrong in, so flagging it.

Went back and mutation checked the tests, reverting each fix and confirming its test fails. Two things came out of that.

The guard that stops `merge_allof` and `deref_schema` walking into `default`, `example`, `enum`, `const` and vendor values had no test at all. I had only checked it in a throwaway script. It is covered now from both directions: a default holding an `allOf` key keeps its value, and a `$ref` inside a default stays a literal while a real reference beside it still resolves.

The body-property loop also had a branch that could never fire. It tested the raw wire name against the operation parameter names before testing the sanitized one, but that set holds both forms and `safe_field_name` is idempotent on identifiers. Reverting it left every test green, which is how it surfaced. Removed.

All nine mutations are caught now: both data guards, the `allOf` blocklist in each direction, the response merge wiring, body/parameter precedence, the keyword identifier, the body/query split, and the emitted-identifier dedupe.

`cli/tests` at 1,954 passed against the same 80 environment failures as the base. Codat still generates 136 files with 0 failures, Rebilly 97 with the one pre-existing `3dsecure` router failure. ruff clean.

Issue #7619, opened 2026-07-31
SEC 13F-HR: '&' silently dropped from issuer names (S&P500 -> SP500)


RadioOrc:
# [Bug] SEC 13F-HR: `&` silently dropped from issuer names ("S&P500" → "SP500")

## Describe the bug

`obb.equity.ownership.form_13f` (provider `sec`) silently corrupts every `nameOfIssuer` / `titleOfClass` containing an ampersand:

| EDGAR source XML | OpenBB output |
|---|---|
| `S&amp;P500 EQL WGT` | `SP500 EQL WGT` |
| `Ishares S&amp;P Gsci Commodity-` | `Ishares SP Gsci Commodity-` |
| `BABCOCK &amp; WILCOX ENTERPRISES` | `BABCOCK  WILCOX ENTERPRISES` (double space) |

No error is raised — the corrupted names flow straight into downstream consumers. Any filer holding S&P ETFs (very common) is affected every quarter.

## Root cause

`parse_13f_hr` in `openbb_sec/utils/parse_13f.py` feeds the **entire Complete Submission TXT file** to `BeautifulSoup(filing, "xml")` (~L137). That file is SGML-wrapped (`<SEC-DOCUMENT>`, `<DOCUMENT>`, `<TYPE>` tags...) and is **not well-formed XML**, so lxml falls back to recover mode — and recover mode drops character entities such as `&amp;` entirely.

The proof: extracting only the embedded well-formed `<XML>...</XML>` block and feeding *that* to the same parser preserves all entities perfectly.

## How to reproduce

```python
from openbb import obb
res = obb.equity.ownership.form_13f(symbol="1536411", date="2026-03-31", provider="sec")
# Duquesne holds several S&P ETFs; every one of them comes back as "SP..." / "S P..."
[r.issuer for r in res.results if "S&P" in (r.issuer or "")]   # == []  (should be non-empty)
```

## Suggested fix

In `parse_13f_hr`, after downloading the complete submission, reassemble the embedded well-formed XML blocks (form header + information table) under a synthetic root before parsing, instead of souping the raw SGML:

```python
xml_blocks = re.findall(r"<XML>(.*?)</XML>", filing, re.DOTALL | re.IGNORECASE)
if xml_blocks:
    decl = re.compile(r"<\?xml[^>]*\?>")
    filing = "<root>" + "".join(decl.sub("", b) for b in xml_blocks) + "</root>"
soup = BeautifulSoup(filing, "xml")
[...]

PR #7620, opened 2026-07-31
fix(sec): 13F-HR parser — guard zero total_value, preserve XML entities

### LINKED ISSUE #7618: SEC 13F-HR: empty filings (zero holdings) crash with ValidationError on weight
# [Bug] SEC 13F-HR: empty filings (zero holdings) crash with ValidationError on `weight`

## Describe the bug

`obb.equity.ownership.form_13f` (provider `sec`) raises a pydantic `ValidationError` for any 13F-HR filing with **no reportable holdings**. Managers that have exited all positions still must file, and EDGAR requires the information table to contain at least one row, so these filings carry a single placeholder entry (`nameOfIssuer=NA`, `cusip=000000000`, `value=0`).

```
ValidationError: 1 validation error for SecForm13FHRData
weight
  Input should be a valid number [type=float_type, input_value=None, input_type=NoneType]
```

## Root cause

`openbb_sec/utils/parse_13f.py` (`parse_13f_hr`, ~L223):

```python
total_value = df.value.sum()
df["weight"] = round(df.value.astype(float) / total_value, 6)
```

For an empty filing `total_value == 0`, so the division yields `0/0 = NaN`. The subsequent `.replace({nan: None, "--": None})` turns it into `None`, and `SecForm13FHRData` declares `weight: float` (not Optional) in `openbb_sec/models/form_13FHR.py` (~L31), so validation fails and the whole fetch dies.

## How to reproduce

Thiel Macro LLC (CIK `1562087`) has filed empty 13F-HRs since Q4 2025:

```python
from openbb import obb
obb.equity.ownership.form_13f(symbol="1562087", date="2026-03-31", provider="sec")
# ValidationError: weight — Input should be a valid number
```

## Suggested fix (one line)

```python
df["weight"] = (
    round(df.value.astype(float) / total_value, 6) if total_value else 0.0
)
```

Verified locally against openbb-sec 1.6.6/1.6.7 (the affected lines are identical in `main`): the empty filing then returns its placeholder row with `weight=0.0`, and normal filings are unchanged (`sum(weight) == 1.0` exactly on a 70-position filing).

Happy to open a PR with this change (together with a fix for a separate entity-mangling issue in the same parser — see companion issue).

RadioOrc:
# fix(sec): 13F-HR parser — guard zero total_value, preserve XML entities

## Why

Two independent bugs in `openbb_sec/utils/parse_13f.py`, both verified against live EDGAR filings (details + repro in the two linked issues):

1. **Empty filings crash.** Managers with no reportable holdings file a single placeholder row (`value=0`). `total_value` is then 0, `weight = 0/0 = NaN`, `.replace({nan: None})` turns it into `None`, and the required `weight: float` field fails validation — the whole fetch raises `ValidationError`. Repro: CIK `1562087` (Thiel Macro), any quarter since 2025-Q4.

2. **`&` silently dropped from names.** The parser soups the entire SGML-wrapped Complete Submission TXT, which is not well-formed XML; lxml recover mode drops character entities, so `S&P500 EQL WGT` becomes `SP500 EQL WGT` with no error. Every filer holding S&P ETFs is affected every quarter. Repro: CIK `1536411` (Duquesne), 2026-03-31.

## What

- `parse_13f_hr`: extract the embedded well-formed `<XML>` blocks (form header + information table), strip their `<?xml?>` declarations and reassemble under a synthetic root before parsing — strict XML path, entities preserved. Bare-XML inputs pass through unchanged.
- `parse_13f_hr`: guard the weight division — `weight = ... if total_value else 0.0`.

No public API change; `SecForm13FHRData` model untouched.

## Testing

Against live EDGAR (openbb-sec 1.6.6 == 1.6.7 == `main` for the affected lines):

- Empty filing (CIK 1562087, 2026-03-31 and 2025-12-31): no longer raises; returns placeholder row with `weight=0.0`.
- Normal filing (CIK 1536411, 2026-03-31, 70 positions): all `&` preserved (`S&P500 EQL WGT`, `Ishares S&P Gsci Commodity-`, `State Str Spdr S&P 500 Etf T`), `sum(weight) == 1.0` exactly, values/shares byte-identical to a direct parse of the standalone information-table XML.
- Cross-checked 14 filings across 7 institutions (incl. options positions with `putCall`, and a $1000s-convention filer) — field-level identical to an independent parser fed the standalone XML.

Fixes #7618 and #7619.

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py
+++ b/openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py
@@ -134,6 +134,21 @@ async def parse_13f_hr(filing: str):
     if filing.startswith("https://"):
         filing = await get_complete_submission(filing)  # type: ignore
 
+    # A Complete Submission TXT file is SGML-wrapped and not well-formed XML.
+    # Feeding it to the XML parser as-is triggers lxml's recover mode, which
+    # silently drops character entities: "S&amp;P500" becomes "SP500" and
+    # "BABCOCK &amp; WILCOX" loses its ampersand in nameOfIssuer/titleOfClass.
+    # Extract the embedded well-formed <XML> blocks (form header + information
+    # table) and reassemble them under a synthetic root so the strict XML path
+    # is used and entities are preserved. Inputs that are already bare XML
+    # (no <XML> wrapper) are passed through unchanged.
+    import re as _re  # noqa: PLC0415
+
+    _xml_blocks = _re.findall(r"<XML>(.*?)</XML>", filing, _re.DOTALL | _re.IGNORECASE)
+    if _xml_blocks:
+        _decl = _re.compile(r"<\?xml[^>]*\?>")
+        filing = "<root>" + "".join(_decl.sub("", b) for b in _xml_blocks) + "</root>"
+
     soup = BeautifulSoup(filing, "xml")
 
     info_table = soup.find_all("informationTable")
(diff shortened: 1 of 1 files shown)


RadioOrc:
recheck

PR #7627, opened 2026-08-04
SEC 13F-HR: '&' silently dropped from issuer names (S&P500 -> SP500)

### LINKED ISSUE #7619: SEC 13F-HR: '&' silently dropped from issuer names (S&P500 -> SP500)

ojassharma7:
Fixes #7619.

## What changed
- `openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py`
- `openbb_platform/providers/sec/tests/test_parse_13f.py`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py
+++ b/openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py
@@ -125,6 +125,8 @@ def get_period_ending(filing_str: str):
 async def parse_13f_hr(filing: str):
     """Parse a 13F-HR filing from the Complete Submission TXT file string."""
     # pylint: disable=import-outside-toplevel
+    import re
+
     import xmltodict
     from bs4 import BeautifulSoup
     from numpy import nan
@@ -133,6 +135,15 @@ async def parse_13f_hr(filing: str):
     # Check if the input string is a URL
     if filing.startswith("https://"):
         filing = await get_complete_submission(filing)  # type: ignore
+
+    # The Complete Submission TXT file is SGML-wrapped and not well-formed XML, so
+    # lxml falls back to recover mode, which silently drops character entities like
+    # '&amp;'. Reassemble the embedded well-formed <XML>...</XML> blocks under a
+    # synthetic root and parse that instead, to preserve entities such as '&'.
+    xml_blocks = re.findall(r"<XML>(.*?)</XML>", filing, re.DOTALL | re.IGNORECASE)
+    if xml_blocks:
+        decl = re.compile(r"<\?xml[^>]*\?>")
+        filing = "<root>" + "".join(decl.sub("", b) for b in xml_blocks) + "</root>"
 
     soup = BeautifulSoup(filing, "xml")
 
--- a/openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py
+++ b/openbb_platform/providers/sec/openbb_sec/utils/parse_13f.py
@@ -69,6 +69,26 @@ async def get_complete_submission(url: str):
     )
 
 
+def prepare_13f_xml(filing: str) -> str:
+    """Return parseable XML for a Complete Submission TXT string.
+
+    The Complete Submission TXT file is SGML-wrapped and not well-formed XML, so
+    feeding it straight to an XML parser makes lxml fall back to recover mode,
+    which silently drops character entities like ``&amp;``. When embedded
+    ``<XML>...</XML>`` blocks are present, reassemble those well-formed fragments
+    under a synthetic root (stripping only top-level XML declarations). When no
+    such blocks exist, return the original text unchanged.
+    """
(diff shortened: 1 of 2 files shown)


anujbolewar (commented):
The root cause is clear — recover-mode parsing drops character entities such as the ampersand entity inside SGML, so reassembling the well-formed nested blocks under a synthetic root is the correct fix. Worth also asserting on an ampersand that appears inside an attribute value, since the block extraction only strips top-level declarations, and confirm the no-block fallback still parses the original text unchanged.

ojassharma7:
Thanks @anujbolewar — addressed in the latest commit.

- Extracted `prepare_13f_xml` so the block-extraction path is testable
- Assert `&amp;` inside an attribute value survives (only top-level XML decls are stripped)
- Confirm the no-`<XML>`-block fallback returns the original text unchanged and still parses

piiq:
Hi,

I am not sure, but it might have been a design decision to drop it from the issuer names. I'll leave it to @deeleeramone to comment on that.

Regardless can you please check if the behaviour persists in the `feature/v5-sec` branch and if so re-target your PR to stack on top of https://github.com/OpenBB-finance/OpenBB/pull/7525

There is a v5 release in the works and it would be best to review it against the new version.

ojassharma7:
Checked feature/v5-sec: this specific bug doesn't reproduce there. That branch's parse_13f_hr already extracts the <informationTable>...</informationTable> block with a regex and parses just that fragment with xmltodict, instead of feeding the full raw SGML envelope to BeautifulSoup. Since the extracted fragment is well-formed XML on its own, xmltodict decodes &amp; normally and the entity survives, no recover-mode fallback involved.

So there's nothing to fix on that branch and no reason to retarget. Happy to leave this open against develop for anyone still on the current SEC provider, or close it if you'd rather not carry it once v5 ships, your call.

For what it's worth on the design-decision question: the corruption isn't just a dropped character, in some names it collapses into a double space, e.g. "BABCOCK & WILCOX ENTERPRISES" becomes "BABCOCK  WILCOX ENTERPRISES", which reads like unintended breakage rather than a style choice. But that's obviously deeleeramone's call.

Issue #7629, opened 2026-08-04
[Bug] Fetcher-level pydantic ValidationError returns 500 "Unexpected Error" instead of 422 (e.g. bls_search without category)


artcashin:
**Describe the bug**
When a query-parameter validation error is raised *inside a fetcher's* pydantic model (rather than at the FastAPI request layer), the REST API surfaces it as **HTTP 500 "Unexpected Error"** instead of a 422. The same class of problem — an invalid/missing request parameter — gets a 422 when FastAPI catches it, so clients see inconsistent semantics for identical mistakes.

Concrete case: `economy/survey/bls_search` declares `category` as required in `BlsSearchQueryParams`, but the endpoint accepts a request without it and 500s during fetch. (Encountered via OpenBB Workspace's "Validate widgets" sweep, which probes endpoints with default/empty params — the resulting 500 blocks connecting an otherwise healthy backend.)

**To Reproduce**
1. OpenBB Platform 4.7.2 with `openbb-bls` installed (any `bls_api_key` state).
2. `openbb-api --host 127.0.0.1 --port 6900`
3. `curl -i "http://127.0.0.1:6900/api/v1/economy/survey/bls_search?provider=bls&include_extras=false&include_code_map=false"`

Response (HTTP 500):
```json
{"detail": "Unexpected Error -> ValidationError -> 1 validation error for BlsSearchQueryParams\ncategory\n  Field required [type=missing, input_value={'query': ''}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.13/v/missing"}
```

**Expected behavior**
HTTP 422 with the standard FastAPI validation error shape, i.e. pydantic `ValidationError` raised during `QueryParams` construction in the provider layer should be mapped to a request-validation response, not to "Unexpected Error". (Ideally `category`'s requiredness would also be reflected in the OpenAPI/`widgets.json` metadata so clients don't issue the invalid request in the first place.)

**Desktop:**
 - OS: Debian 12 (python:3.12-slim container, linux/amd64)
 - Python version: 3.12
 - OpenBB Platform version: openbb 4.7.2, openbb-platform-api 1.3.6

PR #7630, opened 2026-08-04
fix(api): return 422 for missing QueryParams fields

axelray-dev:
## Description

- Summary of the change/ bug fix.
  - When a fetcher raises a pydantic ValidationError for a QueryParams model and a required field is absent from the request query string, the API now returns HTTP 422 with query-scoped detail locations instead of HTTP 500 Unexpected Error.
- Link #7629
- Relevant motivation and context.
  - Missing required query fields never appear in request.query_params, so the previous all_in_query gate fell through to the generic 500 path. That mismatched FastAPI-layer 422 behavior for the same class of client error (e.g. bls_search without category).
- List any dependencies that are required for this change.
  - None.

## How has this been tested?

- Unit tests for ExceptionHandlers.validation covering:
  - Missing QueryParams field not present in query_params -> 422 with loc starting with query
  - ResponseValidationError still 422
  - Non-QueryParams ValidationError still 500
- Command:
  - PYTHONPATH=openbb_platform/core pytest openbb_platform/core/tests/api/test_exception_handlers.py
  - Result: 3 passed
- Ensure all unit and integration tests pass.
  - Focused unit tests pass locally; full suite left to CI.

## Checklist

- I have performed a self-review of my own code.
- I have commented my code, particularly in hard-to-understand areas.
- I have adhered to the GitFlow naming convention and my branch name is in the format of feature/feature-name or hotfix/hotfix-name.
- I ensure that I am following the CONTRIBUTING guidelines.
  - (If applicable) I have updated tests following these guidelines.

### DIFF
--- a/openbb_platform/core/openbb_core/api/exception_handlers.py
+++ b/openbb_platform/core/openbb_core/api/exception_handlers.py
@@ -61,9 +61,7 @@ class ExceptionHandlers:
         )
 
     @staticmethod
-    async def validation(
-        request: Request, error: ValidationError | ResponseValidationError
-    ):
+    async def validation(request: Request, error: ValidationError | ResponseValidationError):
         """Exception handler for ValidationError."""
         # Some validation is performed at Fetcher level.
         # So we check if the validation error comes from a QueryParams class.
@@ -83,14 +81,10 @@ class ExceptionHandlers:
                 detail=detail,
             )
         try:
-            errors = (
-                error.errors(include_url=False)
-                if hasattr(error, "errors")
-                else error.errors
-            )
+            errors = error.errors(include_url=False) if hasattr(error, "errors") else error.errors
         except Exception:
             errors = error.errors if hasattr(error, "errors") else error
-        if "QueryParams" in error.title:
+        if "QueryParams" in error.title and not await request.body():
             detail = [
                 {
                     **{k: v for k, v in err.items() if k != "ctx"},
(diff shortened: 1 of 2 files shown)


anujbolewar (commented):
Sensibly dropping the all_in_query condition since every QueryParams validation should indeed surface as a query-param error. The one case to double check: a QueryParams-backed model can also be used to bind a request body in some routes, and for those the detail will now wrongly claim it sits in query params — confirm the involved routes all read from the query string, otherwise keep a guard that checks whether the request carried a body.

axelray-dev:
Addressed the body-binding edge case from your review.

The handler now rewrites a `QueryParams` validation location to `query` only when the request body is empty. A non-empty-body validation error follows the normal error path, so a body-bound model is not reported as a query parameter error.

Added a regression test for a `QueryParams` model with a request body. Validation on the updated branch:

- `pytest openbb_platform/core/tests/api/test_exception_handlers.py -q` -- 4 passed
- Ruff check and format check -- passed
- `git diff --check` -- passed

Please re-review the updated thread.

axelray-dev:
Verified the current head includes the body-binding guard requested in review: QueryParams errors are rewritten to query-scoped 422 details only when the request body is empty; body-bound validation follows the normal 500 path. VPS validation: 4 passed for openbb_platform/core/tests/api/test_exception_handlers.py. No additional code change was needed.

axelray-dev:
I verified the concern against the current branch. The handler now keeps QueryParams relabeling only for requests with no body, and the regression suite covers both a missing query field and a QueryParams-shaped validation error from a body-bound request. On the VPS, openbb_platform/core/tests/api/test_exception_handlers.py passes: 4 passed. I did not find a further code change required for the reviewer’s stated case.

PR #7633, opened 2026-08-08
feat(quantitative): deflated Sharpe ratio command

ipezygj:
## Why

`obb.quantitative.performance` ships Sharpe, Sortino and Omega — but no way to answer the question that decides whether any of them mean anything after optimization: **was this result selected as the best of N attempts, and does it survive that selection?** Under pure noise, the expected maximum Sharpe of N zero-skill strategies grows with N; a raw Sharpe from a parameter sweep is a best-of-N number.

## What

`obb.quantitative.performance.deflated_sharpe_ratio(data=returns, target="close", trials=100)` — the Deflated Sharpe Ratio (Bailey & López de Prado, 2014): the probability that the true Sharpe exceeds the expected maximum of `trials` zero-skill attempts, adjusting for sample length and return skewness/kurtosis. Returns Sharpe, the expected-max bar, the DSR probability, observations and trials.

- `helpers.deflated_sharpe_stats()` holds the pure computation; the router command follows the existing `omega_ratio` pattern (POST, `DeflatedSharpeModel`, lazy imports, PythonEx/APIEx examples).
- `trials_sr_std` (cross-trial Sharpe dispersion) defaults to the null `1/sqrt(n-1)`; callers who track their optimizer's actual trial spread can pass it through the helper.
- scipy only — already a dependency of this extension.

## Tests

`extensions/quantitative/tests/test_deflated_sharpe.py`: parity at 1e-7 against the [numguard](https://github.com/ipezygj/numguard) reference implementation on a deterministic series, single-trial-reduces-to-PSR, monotonicity in trial count, input guards. 5/5 passing locally alongside the existing helper tests.

The same math was merged into ffn last week (pmorissette/ffn#311), and equivalents are under review in backtesting.py (kernc/backtesting.py#1388) and jesse (jesse-ai/jesse#611).

Note: I haven't regenerated the static package reference in this PR — happy to run the build step and push it if you'd like it included, or leave it to your release tooling, whichever matches your process.

### DIFF
--- a/openbb_platform/extensions/quantitative/openbb_quantitative/helpers.py
+++ b/openbb_platform/extensions/quantitative/openbb_quantitative/helpers.py
@@ -78,3 +78,83 @@ def validate_window(input_data: Union["Series", "DataFrame"], window: int) -> No
         raise ValueError(
             f"Window '{window}' is greater than the input data length '{len(input_data)}'"
         )
+
+
+def deflated_sharpe_stats(
+    input_data: "Series",
+    trials: int,
+    rfr: float = 0.0,
+    trials_sr_std: Union[float, None] = None,
+) -> dict:
+    """Compute the deflated Sharpe ratio of a return series.
+
+    The deflated Sharpe ratio (Bailey & Lopez de Prado, 2014) is the
+    probability that the true Sharpe exceeds the expected maximum Sharpe of
+    `trials` zero-skill strategies — i.e. whether a result selected as the
+    best of `trials` attempts reflects skill rather than selection. Adjusts
+    for sample length and return skewness/kurtosis.
+
+    Parameters
+    ----------
+    input_data : Series
+        Return series (per-period returns, not prices).
+    trials : int
+        Number of strategy variants tried before selecting this one.
+    rfr : float, optional
+        Per-period risk-free rate, by default 0.0.
+    trials_sr_std : float, optional
+        Standard deviation of the per-period Sharpe estimates across the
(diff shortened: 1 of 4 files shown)


ipezygj:
Pushed a correctness fix to this branch, found by probing my own implementation rather than by review.

A constant return series has no Sharpe ratio, but it did not reach the deflation arithmetic as a `nan`: the standard deviation of a constant series is floating-point residue rather than an exact zero, so a flat 0.1% series divided out to a Sharpe of ~1e16. That is finite, so it passed every guard, and the deflated Sharpe came out **1.0** — certainty of a real edge, from the one input that carries no information about one. The statistic exists to catch results that are too good to be true, so answering with maximum confidence there is the worst possible failure direction.

The guard now compares the dispersion against the resolution of a float at the scale of the data, scaled by the number of terms summed (`n * eps * scale`), rather than against zero. Measured over constant series spanning values 1e-7..1e3 and lengths 3..10000, the residue reaches at most 1.96 eps × scale, so this keeps a margin of at least 3.9x at every length, while a real series with sigma=1e-12 sits more than ten orders of magnitude above the threshold — nothing legitimate is caught, and the test asserts that a quiet series still gets a number.

Two notes on the test, since they are the reason this took a second pass:

- It sweeps values × lengths instead of asserting one series. My first attempt used a single flat series, passed, and still leaked at other lengths — the residue depends on both the value and the length, so a guard calibrated on one point tests only that point.
- It was run against the unfixed function and fails there, so it tests the guard rather than merely accompanying it.

Happy to split this into a separate commit or reword anything if you would rather review it apart from the feature.

PR #7653, opened 2026-09-02
fix: reuse SEC Form 4 download session

### LINKED ISSUE #7651: openbb_sec Form 4 (insider trading) fetcher leaks sockets in CLOSE_WAIT under concurrent downloads
## Summary

`openbb_core.provider.utils.helpers.amake_request()` opens a brand-new `aiohttp.ClientSession` for every individual HTTP request rather than reusing one across a batch. `openbb_sec`'s Form 4 (insider trading) fetcher (`openbb_sec/utils/form4.py::download_data()`) calls this once per filing URL, in concurrent batches of 8 (`asyncio.Semaphore(8)` + `asyncio.gather`). Under that concurrent load, sockets accumulate in `CLOSE_WAIT` with unread response data still sitting in the kernel receive buffer, and per-call latency degrades noticeably as a run progresses.

## Environment

- `openbb` 4.7.2 (via `pip install openbb`)
- Python 3.12, Linux

## Root cause

`openbb_core/provider/utils/helpers.py`, `amake_request()`:

```python
with_session = kwargs.pop("with_session", "session" in kwargs)
session = kwargs.pop("session", await get_async_requests_session(**kwargs))

try:
    response = await session.request(method, url, **kwargs)
    return await response_callback(response, session)
finally:
    if not with_session:
        await session.close()
```

Every call that doesn't explicitly pass `session=`/`with_session=True` gets a fresh `ClientSession` via `get_async_requests_session()`, uses it for exactly one request, then closes it. `openbb_sec/utils/form4.py::get_form_4_data()` calls `amake_request()` this way — no shared session — once per filing URL:

```python
response = await amake_request(
    url,
    headers=SEC_HEADERS,
    response_callback=response_callback,
    timeout=30,
)
```

`download_data()` (same file) drives this concurrently:

```python
async with asyncio.Semaphore(8):
    for url_chunk in [non_cached_urls[i:i+8] for i in range(0, len(non_cached_urls), 8)]:
        await asyncio.gather(*[get_one(url) for url in url_chunk])
        await asyncio.sleep(1.125)
```

Creating and tearing down a TLS session per single request is an anti-pattern aiohttp's own docs warn against explicitly ("Don't create a session per request. Most likely you need a session per application which performs all requests together") — see https://docs.aiohttp.org/en/stable/client_reference.html#client-session. Under 8-way concurrent bursts, this repeated session churn appears to leave sockets in `CLOSE_WAIT` rather than being cleanly reclaimed.

## Reproduction / evidence

Running fundamentals + insider-trading ingestion for a list of ~20 tickers (via `obb.equity.ownership.insider_trading(ticker, provider="sec")`, which drives the Form 4 path above) and inspecting the process's sockets mid-run:

```
$ ss -tnp | grep <pid>
CLOSE-WAIT 35362  0  <local>:45316   <sec-edge-ip>:443  users:(("python3",pid=...))
CLOSE-WAIT 9318   0  <local>:45330   <sec-edge-ip>:443  users:(("python3",pid=...))
CLOSE-WAIT 50450  0  <local>:45714   <sec-edge-ip>:443  users:(("python3",pid=...))
... (40+ more CLOSE-WAIT entries)
```

The non-zero `Recv-Q` values (up to ~50KB) indicate response data was sent by the server but never fully read/drained by the client before the session was closed. A direct `curl` to the same SEC endpoints during the same window returned in under a second, ruling out a network-path or server-side latency explanation — the client's own connection handling is the bottleneck. Process CPU time stayed low relative to wall-clock time throughout (consistent with time spent waiting on I/O rather than computing), and per-request latency visibly worsened as more `CLOSE_WAIT` sockets accumulated over the course of the run.

## Suggested direction

Reuse a single `ClientSession` across the batch of concurrent requests in `download_data()` (pass `session=`/`with_session=True` into `amake_request()`/`get_form_4_data()` from a session opened once for the whole download_data() call, closed once at the end) rather than opening/closing one per URL. This is the pattern `amake_requests()` (plural, in the same `helpers.py` file) appears to already support for other callers — `download_data()` doesn't seem to use it.

Happy to provide more detail or test a fix if useful.

jaredrainsha:

## Description

- [x] Reuse one `aiohttp` session across every uncached Form 4 filing in a `download_data()` call, and close it in a `finally` block.
- [x] Fixes #7651.
- [ ] Screenshot of the feature or the bug before/after fix, if applicable. Not applicable to this backend-only change.
- [x] Avoid per-request TLS session setup and teardown while preserving the existing eight-request chunking and delay.
- [x] No new dependencies are required.

## How has this been tested?

- `.venv/bin/python -m pytest openbb_platform/providers/sec/tests/test_form4.py -q` — 1 passed. The regression verifies that two filing requests receive the same session and that it is closed exactly once.
- `.venv/bin/python -m pytest openbb_platform/providers/sec/tests/test_form4.py openbb_platform/providers/sec/tests/test_sec_fetchers.py openbb_platform/providers/sec/tests/test_company_facts.py -q` — 147 passed.
- `.venv/bin/ruff check openbb_platform/providers/sec/openbb_sec/utils/form4.py openbb_platform/providers/sec/tests/test_form4.py` — passed.
- `git diff --check` — passed.
- [ ] Ensure all unit and integration tests pass. An initial all-SEC run reached 174 passed before 22 failures and 2 errors from live SEC/FASB TLS EOFs; it was interrupted after 4m14s. After installing the repository test tooling, the recorded SEC fetcher suite passed 23/23.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas. No new explanatory comments were needed for the small control-flow change.
- [x] I ensure that I am following the [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBB/blob/develop/openbb_platform/CONTRIBUTING.md).
  - [x] Added a focused regression test for session reuse and cleanup.

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/utils/form4.py
+++ b/openbb_platform/providers/sec/openbb_sec/utils/form4.py
@@ -266,7 +266,7 @@ def clean_xml(xml_content):
     return xml_content
 
 
-async def get_form_4_data(url) -> dict:
+async def get_form_4_data(url, session=None) -> dict:
     """Get the form 4 data."""
     # pylint: disable=import-outside-toplevel
     from warnings import warn  # noqa
@@ -277,11 +277,13 @@ async def get_form_4_data(url) -> dict:
         """Response callback function."""
         return await response.read()
 
+    session_kwargs = {"session": session} if session is not None else {}
     response = await amake_request(
         url,
         headers=SEC_HEADERS,
         response_callback=response_callback,
         timeout=30,
+        **session_kwargs,
     )  # type: ignore
     response_text = response.decode("utf-8")  # type: ignore
 
@@ -500,6 +502,7 @@ async def download_data(urls, use_cache: bool = True):  # noqa: PLR0915
     import sqlite3
     from numpy import nan
     from openbb_core.app.utils import get_user_cache_directory
+    from openbb_core.provider.utils.helpers import get_async_requests_session
     from pandas import DataFrame
 
     results: list = []
@@ -547,9 +550,9 @@ async def download_data(urls, use_cache: bool = True):  # noqa: PLR0915
         elif use_cache is False:
             non_cached_urls = urls
 
-        async def get_one(url):
+        async def get_one(url, session):
             """Get the data for one URL."""
-            data = await get_form_4_data(url)
+            data = await get_form_4_data(url, session=session)
             result = await parse_form_4_data(data)
             if not result and use_cache is True:
                 df = DataFrame([{"filing_url": url}])
@@ -590,13 +593,16 @@ async def download_data(urls, use_cache: bool = True):  # noqa: PLR0915
             )
 
         if len(non_cached_urls) > 0:
-            async with asyncio.Semaphore(8):
+            session = await get_async_requests_session()
+            try:
                 for url_chunk in [
                     non_cached_urls[i : i + 8]
                     for i in range(0, len(non_cached_urls), 8)
                 ]:
-                    await asyncio.gather(*[get_one(url) for url in url_chunk])
+                    await asyncio.gather(*[get_one(url, session) for url in url_chunk])
                     await asyncio.sleep(1.125)
+            finally:
+                await session.close()
 
         if use_cache is True:
             close_db(conn, db_path)
--- a/openbb_platform/providers/sec/tests/test_form4.py
+++ b/openbb_platform/providers/sec/tests/test_form4.py
@@ -1,0 +1,57 @@
+"""Tests for SEC Form 4 download helpers."""
+
+import asyncio
+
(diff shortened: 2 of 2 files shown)


piiq:
Thanks @jaredrainsha 

Can you check if the issue persists in the `feature/v5-sec` branch and if so re-target your PR to stack on top of https://github.com/OpenBB-finance/OpenBB/pull/7525

There is a v5 release in the works and it would be best if the feature is available there.

Additionally - please make sure that the session reuse complies with the rate limiting the SEC has.

deeleeramone:
Hi, thanks for the PR. The changes being introduced in #7525 use a global session with rate limit handling. These changes, therefore, will not be necessary.

PR #7657, opened 2026-09-05
fix(sec): persist company facts HTTP cache

### LINKED ISSUE #7655: openbb_sec company facts fetcher never actually caches: re-downloads full companyfacts.json for every statement type
## Summary

`openbb_sec.utils.company_facts.get_standardized_financials()` fetches SEC's full `companyfacts` JSON per call via a fresh `aiohttp_client_cache.CachedSession(expire_after=3600 * 6)`, but since no `cache=` argument is passed, `aiohttp_client_cache` defaults to a brand-new, empty, in-process `CacheBackend()`/`DictCache()` on every single call. The 6-hour cache therefore never actually caches anything across calls -- each call gets its own throwaway cache that is discarded when the function returns. Calling `obb.equity.fundamental.income()`, `.balance()`, and `.cash()` for the same symbol -- which all resolve to this same function and the same underlying SEC endpoint -- downloads the identical multi-megabyte payload three separate times instead of once.

## Environment

- `openbb` 4.7.2
- Python 3.12, Linux

## Root cause

`openbb_sec/utils/company_facts.py`, `get_standardized_financials()` -> `_fetch()`:

```python
async def _fetch(cik_str: str) -> dict:
    url = f"https://data.sec.gov/api/xbrl/companyfacts/CIK{cik_str}.json"
    if use_cache:
        from aiohttp_client_cache.session import CachedSession

        async with CachedSession(expire_after=3600 * 6) as session:
            try:
                resp = await amake_request(url, headers=HEADERS, session=session, timeout=300)
            finally:
                await session.close()
    else:
        resp = await amake_request(url, headers=HEADERS, timeout=300)
    ...
```

`CachedSession.__init__` (in `aiohttp_client_cache/session.py`) does:

```python
self.cache = cache or CacheBackend()
```

No `cache=` is ever passed at the call site above, so every invocation gets a fresh `CacheBackend()`, whose `responses` store is a plain in-process `DictCache()` (see `aiohttp_client_cache/backends/base.py`). That cache is scoped to the `CachedSession` instance, which is created and thrown away inside `_fetch()` on every call -- nothing persists between calls, so the 6-hour `expire_after` never gets a chance to produce a cache hit.

`income_statement.py`, `balance_sheet.py`, and `cash_flow.py` (same directory) each independently call `get_standardized_financials()` for their own statement type, with no shared cache or shared fetch between them, even though all three read the same underlying `companyfacts` JSON and simply project different fields from it.

## Reproduction / evidence

The endpoint itself, fetched directly, confirms the payload size involved:

```
$ curl -sS -H "User-Agent: test test@test.com" \
    "https://data.sec.gov/api/xbrl/companyfacts/CIK0000320193.json" | wc -c
3789099
$ curl -sS -H "User-Agent: test test@test.com" \
    "https://data.sec.gov/api/xbrl/companyfacts/CIK0000019617.json" | wc -c
7927352
```

~3.8MB and ~7.9MB respectively for these two example companies. Calling `obb.equity.fundamental.income(symbol, provider="sec")`, then `.balance(symbol, provider="sec")`, then `.cash(symbol, provider="sec")` for the same symbol in the same process triggers three independent full downloads of that same payload rather than one -- confirmed by timing each call and observing three separate multi-second-to-multi-minute fetches of comparable duration, rather than two fast cache hits following one real fetch. Larger/longer-filing-history companies (correspondingly larger `companyfacts.json` payloads) show proportionally worse cumulative cost, since the redundancy multiplies whatever the single-fetch cost already is.

## Suggested direction

Either:
1. Pass a shared, persistent `cache=` (e.g. a module-level or request-scoped `CacheBackend` instance, or a `SQLiteBackend` like the Form 4 fetcher's own on-disk cache) into `CachedSession(...)` in `_fetch()`, so the 6-hour TTL actually spans multiple calls, not just the lifetime of one `async with` block; or
2. Fetch `companyfacts` once per (symbol, session) and let `income_statement.py`/`balance_sheet.py`/`cash_flow.py` share that single result rather than each independently calling `get_standardized_financials()` end-to-end.

Either would cut real SEC bandwidth/time cost for a common usage pattern (fetching all three statements for one company) by roughly two-thirds.

Happy to provide more detail if useful.

sangtrx:
Fixes #7655.

get_standardized_financials previously created a fresh default in-memory CachedSession backend per invocation, so repeated income, balance, and cash calls re-downloaded the same SEC companyfacts payload.

This fix uses the existing OpenBB SEC SQLiteBackend pattern at <user-cache>/http/sec_company_facts, keeps the six-hour TTL and use_cache=False behavior, cleans expired responses, and closes sessions.

The focused network-free regression test openbb_platform/providers/sec/tests/test_company_facts_cache.py verifies repeated calls target the same persistent cache path, TTL, backend injection, expired-response cleanup, and session closure.

No new dependencies.

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/utils/company_facts.py
+++ b/openbb_platform/providers/sec/openbb_sec/utils/company_facts.py
@@ -532,7 +532,7 @@ async def get_standardized_financials(
     period : PeriodType
         Which periods to return.
     use_cache : bool
-        Whether to use in-memory HTTP caching (6-hour TTL).
+        Whether to use persistent HTTP caching (6-hour TTL).
     pit_mode : bool
         If True, skip the 10-K vintage override for quarterly data,
         preserving point-in-time fidelity for backtesting.
@@ -572,12 +572,22 @@ async def get_standardized_financials(
     async def _fetch(cik_str: str) -> dict:
         url = f"https://data.sec.gov/api/xbrl/companyfacts/CIK{cik_str}.json"
         if use_cache:
+            from aiohttp_client_cache import (
+                SQLiteBackend,
+            )  # pylint: disable=import-outside-toplevel
             from aiohttp_client_cache.session import (
                 CachedSession,
             )  # pylint: disable=import-outside-toplevel
+            from openbb_core.app.utils import (
+                get_user_cache_directory,
+            )  # pylint: disable=import-outside-toplevel
 
-            async with CachedSession(expire_after=3600 * 6) as session:
+            cache_dir = f"{get_user_cache_directory()}/http/sec_company_facts"
+            async with CachedSession(
+                cache=SQLiteBackend(cache_dir, expire_after=3600 * 6)
+            ) as session:
                 try:
+                    await session.delete_expired_responses()
                     resp = await amake_request(
                         url, headers=HEADERS, session=session, timeout=300
                     )
--- a/openbb_platform/providers/sec/tests/test_company_facts_cache.py
+++ b/openbb_platform/providers/sec/tests/test_company_facts_cache.py
@@ -1,0 +1,71 @@
+"""Tests for SEC company-facts HTTP caching."""
+
+import pytest
+from openbb_sec.utils import company_facts
+
+
+@pytest.mark.asyncio
+async def test_company_facts_uses_persistent_sqlite_cache(monkeypatch, tmp_path):
+    """Repeated calls should target the same persistent six-hour cache."""
+    import aiohttp_client_cache
+    import aiohttp_client_cache.session as cache_session
+    import openbb_core.app.utils as app_utils
+    import openbb_core.provider.utils.helpers as request_helpers
+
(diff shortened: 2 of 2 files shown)


piiq:
Thanks @sangtrx 

Please check if the behaviour you want to change is still available in the `feature/v5-sec` branch and if so re-target your PR to stack on top of https://github.com/OpenBB-finance/OpenBB/pull/7525

There is a v5 release in the works and it would be best to review it there.

sangtrx:
Checked feature/v5-sec as requested. The company-facts path now uses the unified disk-backed SEC cached_request from #7525 with a six-hour expiry, and the V5 cache tests cover miss-then-hit without a second transport call. That means the behavior from #7655 is already addressed there, so porting this patch would duplicate the new cache layer. I am closing this PR to avoid carrying redundant code. Thanks for pointing me to the V5 branch.

PR #7658, opened 2026-09-06
fix(sec): serialize Form 4 cache lifecycle

### LINKED ISSUE #7656: openbb_sec Form 4 fetcher's shared on-disk cache corrupts under concurrent calls
## Summary

`openbb_sec/utils/form4.py::download_data()` uses a single, shared on-disk cache file (`sec_form4.db` / `sec_form4.db.gz`, one file total, not keyed per symbol or per call) with a lifecycle that decompresses it at the start of every call and compresses-then-deletes the raw file at the end of every call. Under concurrent calls (e.g. fetching insider-trading data for several symbols at once from separate threads/tasks), one call's end-of-lifecycle `close_db()` (which sorts the table, closes the SQLite connection, gzips the file, and deletes the raw `.db`) can run while another call is still mid-flight against the same file, corrupting it. Once corrupted, every subsequent call fails.

## Environment

- `openbb` 4.7.2
- Python 3.12, Linux

## Root cause

`download_data()`'s cache lifecycle, every single call:

```python
if use_cache is True:
    db_dir = f"{get_user_cache_directory()}/sql"
    db_path = f"{db_dir}/sec_form4.db"
    if os.path.exists(f"{db_path}.gz"):
        decompress_db(db_path)   # reads .db.gz, (re)writes .db
    ...
    conn = sqlite3.connect(db_path)
    setup_database(conn)
    ...
...
if use_cache is True:
    close_db(conn, db_path)      # sorts, closes, gzips .db -> .db.gz, then os.remove(db_path)
```

`close_db()`:

```python
def close_db(conn, db_path):
    conn.execute("CREATE TABLE IF NOT EXISTS form4_data_sorted AS SELECT * FROM form4_data ORDER BY filing_date")
    conn.execute("DROP TABLE form4_data")
    conn.execute("ALTER TABLE form4_data_sorted RENAME TO form4_data")
    conn.commit()
    conn.close()
    compress_db(db_path)   # reads raw .db, writes .db.gz
    os.remove(db_path)     # deletes raw .db
```

This is a single shared file path with no locking around it at all. Two calls running concurrently -- e.g. `download_data()` invoked for two different symbols' insider-trading data at roughly the same time -- interleave in ways that corrupt the file: one call's `os.remove(db_path)` can delete the `.db` file another call still has an open `sqlite3.connect()` handle to; one call's `compress_db()` can read a `.db` file another call is mid-write on; a `decompress_db()` on one call's start can race a `compress_db()`/`os.remove()` on another call's end for the same `.gz` path. Once either file is left in a partial/inconsistent state, `decompress_db()`'s `gzip.open()` fails on read, or the SQLite file itself is left malformed -- and because `conn` is only assigned after the decompress step succeeds, a failure there hits the `except Exception as e:` handler's own `close_db(conn, db_path)` call with `conn` never having been assigned, raising `UnboundLocalError: cannot access local variable 'conn'` instead of the original error. Once the cache file is corrupted on disk, **every subsequent call fails the same way**, including fully sequential/non-concurrent ones, until the corrupted file is manually deleted.

## Reproduction / evidence

Running insider-trading fetches for several symbols concurrently (e.g. via a thread pool, one `obb.equity.ownership.insider_trading(symbol, provider="sec")` call per thread) reliably corrupts the cache within a few runs. Confirmed directly on the resulting files:

```
$ python3 -c "
import gzip
with gzip.open('sec_form4.db.gz', 'rb') as f:
    f.read()
"
BadGzipFile: Not a gzipped file (b'\xffb')

$ python3 -c "
import sqlite3
conn = sqlite3.connect('sec_form4.db')
conn.execute('select count(*) from form4_data').fetchone()
"
sqlite3.DatabaseError: database disk image is malformed
```

Both files existing simultaneously (`sec_form4.db` *and* `sec_form4.db.gz`) is itself a sign of an interrupted lifecycle -- a healthy end state should only ever have the `.gz`, since `close_db()` always deletes the raw file after compressing it.

## Suggested direction

Serialize the entire cache-file lifecycle (decompress-at-start through close/compress/delete-at-end) behind a lock shared across concurrent callers within a process, so only one caller ever has the file open/being rewritten at a time. A plain `threading.Lock` (not `asyncio.Lock`) is appropriate if concurrent callers may be running in separate threads/event loops rather than sharing one. The per-URL SEC fetches themselves (the `asyncio.gather` over chunks of the URL list) can stay concurrent as before; only the shared-file open/close boundary needs serializing. Separately, consider making `download_data()` resilient to an already-corrupted cache file (e.g. catching the decompress/connect failure and starting fresh rather than propagating an `UnboundLocalError`), since a corrupted cache currently blocks the fetcher entirely rather than degrading to an uncached fetch.

Happy to provide more detail if useful.

sangtrx:

## Description

- [x] Serialize the complete lifecycle of the shared `sec_form4.db` / `.gz` cache with a process-wide lock.
- [x] Acquire the lock through `asyncio.to_thread` so a waiting cached call does not block the event loop.
- [x] Guard cleanup with initialized connection state so decompression or connection failures preserve the original error.
- [x] Preserve the existing eight-URL concurrent batches and the `use_cache=False` path.
- [x] Fixes #7656.
- [x] Screenshot not applicable; this is a backend concurrency fix.
- [x] No new dependencies are required.

Open PRs #7653 and #7642 touch the same Form 4 utility and test path for separate fixes; neither addresses the shared cache lifecycle.

## How has this been tested?

- `.venv/bin/pytest openbb_platform/providers/sec/tests/test_form4.py -q` — 2 passed.
- `.venv/bin/ruff check --no-fix openbb_platform/providers/sec/openbb_sec/utils/form4.py openbb_platform/providers/sec/tests/test_form4.py` — passed.
- `git diff --check` — passed.
- The concurrency regression verifies cached lifecycles do not overlap while each call retains concurrent per-URL downloads.
- The failure regression verifies an exception before SQLite connection creation is reported without attempting database cleanup.

- [ ] Ensure all unit and integration tests pass.

## Checklist

- [x] I ensure that I am following the [CONTRIBUTING guidelines](https://github.com/OpenBB-finance/OpenBB/blob/develop/openbb_platform/CONTRIBUTING.md).
  - [x] I have added focused regression tests for the changed behavior.

### DIFF
--- a/openbb_platform/providers/sec/openbb_sec/utils/form4.py
+++ b/openbb_platform/providers/sec/openbb_sec/utils/form4.py
@@ -2,8 +2,11 @@
 
 import logging
 from datetime import date as dateType
+from threading import Lock
 
 from openbb_core.app.model.abstract.error import OpenBBError
+
+_FORM4_CACHE_LOCK = Lock()
 
 SEC_HEADERS: dict[str, str] = {
     "User-Agent": "Jesus Window Washing jesus@stainedglass.com",
@@ -504,9 +507,14 @@ async def download_data(urls, use_cache: bool = True):  # noqa: PLR0915
 
     results: list = []
     non_cached_urls: list = []
+    conn = None
+    db_path = None
+    cache_lock_acquired = False
 
     try:
         if use_cache is True:
+            await asyncio.to_thread(_FORM4_CACHE_LOCK.acquire)
+            cache_lock_acquired = True
             db_dir = f"{get_user_cache_directory()}/sql"
             db_path = f"{db_dir}/sec_form4.db"
             # Decompress the database file
@@ -598,19 +606,26 @@ async def download_data(urls, use_cache: bool = True):  # noqa: PLR0915
                     await asyncio.gather(*[get_one(url) for url in url_chunk])
                     await asyncio.sleep(1.125)
 
-        if use_cache is True:
-            close_db(conn, db_path)
+        if conn is not None and db_path is not None:
+            connection = conn
+            conn = None
+            close_db(connection, db_path)
 
         results = [entry for entry in results if entry.get("filing_date")]
 
         return sorted(results, key=lambda x: x["filing_date"], reverse=True)
 
     except Exception as e:  # pylint: disable=broad-except
-        if use_cache is True:
-            close_db(conn, db_path)
+        if conn is not None and db_path is not None:
+            connection = conn
+            conn = None
+            close_db(connection, db_path)
         raise OpenBBError(
             f"Unexpected error while downloading and processing data -> {e.__class__.__name__}: {e}"
         ) from e
+    finally:
+        if cache_lock_acquired:
+            _FORM4_CACHE_LOCK.release()
 
 
 def get_cached_data(urls, conn):
--- a/openbb_platform/providers/sec/tests/test_form4.py
+++ b/openbb_platform/providers/sec/tests/test_form4.py
@@ -1,0 +1,105 @@
+"""Tests for the SEC Form 4 utilities."""
+
+import asyncio
+import sqlite3
+from unittest.mock import Mock
+
+import pandas
+import pytest
+from openbb_core.app import utils as app_utils
+from openbb_core.app.model.abstract.error import OpenBBError
+from openbb_sec.utils import form4
+
+
+@pytest.mark.asyncio
+async def test_download_data_serializes_cached_calls(monkeypatch, tmp_path):
+    """Cached calls serialize their lifecycle but retain per-URL concurrency."""
+
+    class FakeDataFrame:
+        """Minimal DataFrame replacement for empty parsed filings."""
+
+        def __init__(self, *_args, **_kwargs):
+            pass
+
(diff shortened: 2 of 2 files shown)


deeleeramone:
Thanks for the PR. Take a look at this PR #7525, which introduces many changes to the package. As part of the refactor, all caching has been consolidated to use the `diskcache` library and a global rate limiter for file downloads.

Another consideration is that concurrent calls across threads for downloads from the SEC is not recommended. They strictly enforce rate limits by IP address and Form 4 produces many small downloads. The downloader is optimized to ensure the machine does not get blocked.

sangtrx:
Thanks for pointing me to #7525. I reviewed the V5 refactor and agree it supersedes this patch: Form 4 caching moves to the unified diskcache path and the downloader/rate-limiter design replaces the shared gzip/SQLite lifecycle this PR was fixing. I'll leave #7658 closed as redundant rather than forward-porting the lock-based approach. If a maintainer later wants a targeted 4.x/develop backport for the old cache path, I'm happy to adapt it on the existing branch.

PR #7663, opened 2026-09-14
[SEC] Preserve filing headers when comparative shares are missing

mowax3:
## Description

`SecBaseFiling._download_cover_page()` aborts NBIS's original FY2025 20-F with `cannot convert float NaN to integer`. Its real R1 table has shares in the 2025 column and missing values in the comparative 2024 column read by the existing parser. NaN passes the truthiness guard and reaches `int()`.

Check for missing shares and finite scaled values before integer conversion. Missing or non-finite shares remain unknown while the rest of the cover is parsed; a legitimate zero is preserved. This keeps the existing date/column selection and adds no dependencies.

Includes the original NBIS R1 and index-header HTML, source URLs and hashes, a full `SecFilingFetcher` replay, and focused missing/NaN/pandas.NA/infinity/zero/normal-value cases. The fixture HTML was obtained via a deployed OpenBB `regulators/sec/htm_file(provider=sec)` route; only the network download is substituted in the regression test.

Related downstream tracking: https://github.com/mowax3/tool-server/issues/44

## How has this been tested?

- Before the patch: 6 failures and 3 passes in the new tests, including the exact NBIS conversion error.
- After the patch: 9/9 new tests pass; 155 tests pass across `test_sec_filing_cover.py`, `test_sec_fetchers.py` and `test_company_facts.py`, with HTTP recordings in replay-only mode. Existing normal filing and HTML fetcher recordings pass.
- Python 3.13.5, pandas 3.0.5, local openbb-core 1.6.13 and openbb-sec 1.6.7 source at base `3e071fcc2cd9f891cac6040ae60296dba76dab46`.
- Black 25.1.0, Ruff 0.12.12 and Pylint pass for changed Python files; `git diff --check` passes.
- Downstream validation: repaired model output, wrapped in an OpenBB results envelope, returns `ok` through the actual local MCP `filings__get_headers` handler. The saved HTTP 400 still returns a typed provider failure. 95 related tool-server tests pass.

Minimal reproduction, with the core and SEC provider installed from this checkout:

```bash
PYTHONPATH=openbb_platform/providers/sec pytest -q openbb_platform/providers/sec/tests/test_sec_filing_cover.py
```

The 155-test run used a temporary OpenBB settings directory via `SystemService` constructor arguments because the runner's home is read-only. One pytest plugin rewrite warning remained. An initial wider taxonomy run hit that settings-directory restriction (9 failures, 16 errors); the full OpenBB suite and live taxonomy routes are not claimed as validated.

The deployed header endpoint was rechecked and still returned the original error. This PR has not been deployed; the deployed OpenBB package/image identity has not been obtained. Public-source replay, upstream acceptance and deployment verification remain separate evidence.

## Checklist

- [x] Self-reviewed the change and documented the missing-comparative-value guard.
- [x] Added regression tests using real source inputs and numerical boundary cases.
- [x] Used a `hotfix/` branch targeting `develop` and followed the contributor workflow.
- [ ] Full OpenBB unit/integration suite and deployed API/Python verification (limits above).


mowax3:
Withdrawn: submitted without the account owner’s approval. Please disregard this contribution.

